Issue #3594, opened 2021-12-03
Introduce PageStorage V3 to reduce CPU usage and write amplification


JaySon-Huang:
## Feature Request

**Teachability, Documentation, Adoption, Migration Strategy:**
To reduce CPU usage and write amplification, we file a proposal for new version of PageStorage: [Proposal: PageStorage V3 Design](https://docs.google.com/document/d/1l1GoIV6Rp0GEwuYtToJMKYACmZv6jf4kp1n8JdQidS8/edit#heading=h.h8ucduiiq45q)
This issue records the progress of PageStorage V3 development.

### Development progress
- [ ] some preparations and refactoration
  - [x] Rand IO benchmark on EBS: https://docs.google.com/document/d/1NoHeigyfNyB9Bp8cYzYFv-In0dQzjR7smLltMBXCmgs/edit#
  - [x] Make `PageStorage::Snapshot` an abstract class and make it compatible for V1, V2 and V3
  - [ ] Refactor out `PageStorage::getEntry` and make it an internal class inside PageStorageV2/V1
  - [x] Refactor out the default value in virtual methods of `PageStorage`
- [x] The framework and related data structure for PageStorage V3
- [ ] BlobStore
  - [x] SpaceMap
  - [x] Read, write logic for BlobStore
  - [x] Data compaction for BlobStore
  - [x] Multi-disks load balancing
  - [x] Read amplification on `PageMap BlobStore::read(FieldReadInfos & to_read, ...)` https://github.com/pingcap/tics/pull/3908/files#diff-1f8b8179cf4110ca801c7a156f936e6c7689f41e1bb967f2c9f46cdaffd9d1ebR345; fixed in https://github.com/pingcap/tiflash/pull/4181/files
- [ ] MVCC PageDirectory
  - [x] Basic apply edits, get PageEntry
  - [x] GC for cleaning up removed Page
  - [x] GC with BlobStore data movement
  - [x] GCApply needs to scan all living page ids with a read lock, which could cause high write blocking time. Find a better way for removing external pages; fixed in #4174
  - [ ] Stale snapshots detection
    - [x] Record the thread id, created time while creating a snapshot
    - [x] Log warnings when there exist long time snapshot
    - [ ] Optimization when stale snapshot(s) exist (MVCC GC, BlobStore GC)
- [x] WALLog
  - [x] LogFile format
  - [x] Meta format for persisting the PageEntry applied to MVCC PageDirectory
  - [x] Compaction of LogFiles
[...]

PR #4177, opened 2022-03-03
Fix a encryption problem in pagestorage v3.

jiaqizho:

Issue Number: ref #3594

Problem Summary:

- add some tests which can test encryption function in PageStorage V3

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/tests/gtest_page_storage.cpp
```diff
@@ -94,6 +97,75 @@ try
...
+    // Because WALStore can't get any record from it.
+    try
+    {
+        page_storage->restore();
+        page_storage->read(0);
+        ASSERT_FALSE(true); // Should not come here.
+    }
+    catch (DB::Exception & e)
+    {
+    }
```
JaySon-Huang, line 135: Use `ASSERT_ANY_THROW`?

flowbehappy (approved):

dbms/src/Storages/Page/V3/WAL/WALReader.cpp
```diff
@@ -209,7 +209,7 @@ bool WALStoreReader::openNextFile()
         auto read_buf = createReadBufferFromFileBaseByFileProvider(
             provider,
             fullname,
-            EncryptionPath{parent_path, filename},
```
jiaqizho, line 212: @JaySon-Huang different with `log reader`. so IV will changed.

dbms/src/Storages/Page/V3/WAL/WALReader.cpp
```diff
@@ -209,7 +209,7 @@ bool WALStoreReader::openNextFile()
...
+            EncryptionPath{parent_path + "/" + filename, ""},
```
JaySon-Huang, line 212: Make the L206 to be `const auto fullname = next_file.fullname(next_file.stage)` and make the L212 to be `EncryptionPath{fullname, ""},` may be helpful to elimate the differnece?
https://github.com/pingcap/tiflash/blob/f4ca432d741be05df8066f9cee9fe195c8956c5d/dbms/src/Storages/Page/V3/WALStore.cpp#L132-L140

dbms/src/Encryption/MockKeyManager.cpp
```diff
@@ -107,6 +107,10 @@ FileEncryptionInfo MockKeyManager::getFile(const String & fname)
     std::ignore = fname;
     if (encryption_enabled)
     {
+        if (!fileExist(fname))
+        {
+            files.emplace_back(fname);
+        }
```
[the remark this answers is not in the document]
jiaqizho: In fact, the creation of BlobFile does not pass this parameter. Because for BlobFile there is no way to know if it is a newly created or old one. I would think that this is a relatively safe approach so that the proxy can make its own judgment.

JaySon-Huang (approved):

Issue #4511, opened 2022-03-30
Fix some problem involve background gc thread


lidezhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1. load data into tiflash and set tiflash replica to 0 later
2. monitor the segment gc process

### 2. What did you expect to see? (Required)
The segment merged into an empty segment and stable_size_on_disk be 0.

### 3. What did you see instead (Required)
1. there are a lot of empty segments remained;
2. the stable_size_on_disk of the empty segments are not empty;

PR #4522, opened 2022-03-30
fix empty segment cannot merge after gc

lidezhu:
This is an cherry-pick of #4500

Issue Number: close #4511

Problem Summary:
1. empty segments cannot be merged after gc;
2. when a segment can be triggerred to gc because of too many delete range, we still wait for the gc_safe_point change, which is not needed.

1. when segment should be gced because of too many delete range, do it regardless of the gc_safe_point;
2. only update index cache for read thread;
3. fix unstable rough set filter test because of more fast gc speed.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
1. load data into tiflash and check segment status;
image
4. set tiflash replica to 0 and check segment status after a while
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the problem that empty segments cannot be merged after gc
```


lidezhu:
@JaySon-Huang PTAL

JaySon-Huang (approved):

Issue #4545, opened 2022-03-31
sanitizer: data race in PageStorage V3


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

build the tiflash ut with `-DCMAKE_BUILD_TYPE=TSan` and run it

### 2. What did you expect to see? (Required)
no warnings

### 3. What did you see instead (Required)
two warnings inside PageStorage V3
```
==================
WARNING: ThreadSanitizer: data race (pid=183648)
  Read of size 8 at 0x7b20000c54b0 by main thread (mutexes: write M1127161458996574888):
    #0 DB::PS::V3::BlobStore::BlobStats::chooseStat(unsigned long, std::__1::lock_guard<std::__1::mutex> const&) /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/Page/V3/BlobStore.cpp:934:26 (gtests_dbms+0xb21bb4b)
    #1 DB::PS::V3::BlobStore::getPosFromStats(unsigned long)::$_15::operator()() const /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/Page/V3/BlobStore.cpp:264:51 (gtests_dbms+0xb2133fb)
    #2 DB::PS::V3::BlobStore::getPosFromStats(unsigned long) /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/Page/V3/BlobStore.cpp:261:22 (gtests_dbms+0xb2133fb)
    #3 DB::PS::V3::BlobStore::write(DB::WriteBatch&, std::__1::shared_ptr<DB::WriteLimiter> const&) /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/Page/V3/BlobStore.cpp:123:38 (gtests_dbms+0xb2128c5)
    #4 DB::PS::V3::PageStorageImpl::write(DB::WriteBatch&&, std::__1::shared_ptr<DB::WriteLimiter> const&) /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/Page/V3/PageStorageImpl.cpp:93:28 (gtests_dbms+0xb2450f4)
    #5 DB::DM::WriteBatches::writeLogAndData() /DATA/disk1/jaysonhuang/tics-m/dbms/src/Storages/DeltaMerge/WriteBatches.h:114:29 (gtests_dbms+0x3630645)
[...]

PR #4548, opened 2022-03-31
PageStorage: Fix data race in BlobStore

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/4545

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -307,7 +303,7 @@ void BlobStore::removePosFromStats(BlobFileId blob_id, BlobFileOffset offset, si
 {
     const auto & stat = blob_stats.blobIdToStat(blob_id);
     auto lock = stat->lock();
-    stat->removePosFromStat(offset, size);
+    stat->removePosFromStat(offset, size, lock);
```
jiaqizho, line 306: if the current thread already holds a `lock_guand` from L305, Does it still need to pass the lock into the sub-methods?

jiaqizho: how about changing `stat->lock()` back to hold the `lock` from thread. then we won't pass the lock_guard many functions...

jiaqizho (approved):
LGTM with a suggestion: how about changing back `stat->lock()` to hold the lock from the thread. then we won't pass the lock_guard many functions.

SchrodingerZhu (approved):

JaySon-Huang:
> LGTM with a suggestion: how about changing back `stat->lock()` to hold the lock from the thread. then we won't pass the lock_guard many functions.

Adding the lock_guard to the signature of the function is mainly to remind the caller should take care of the concurrent problem, and easier for reviewing

PR #4570, opened 2022-04-02
Make performance of TPCH q15 stable 

### LINKED ISSUE #4451: TPCH q15 performance regression after introduce `local_delta` in MemoryTracker
## Bug Report

load tpch-100 data, using 1 TiFlash node
run tpch q15
The query time varies from 1.5 second to 3.0 second randomly
1. the query time should be stable
2. the query time should less than 2 second
master @ 225dabe6942966d1562af738c287b50872f26be8 
### 5. Root cause
The root cause is in TiFlash, the `ParallelAggregatingBlockInputStream` calculate the aggregation in 2 stage:
- stage 1: it use `ParallelInputsProcessor` to do a partial aggregation for each input pipeline
- stage 2: it merge the result of stage 1

Obviously, stage 1 is running using multiple threads, and depends on the size of result data set, stage 2 will use 1 threads or multiple threads: if the total agg key size exceeded `group_by_two_level_threshold` or the total result size exceeded `group_by_two_level_threshold_bytes`, stage 2 will use multiple threads othewise, it will use 1 thread. And if stage 2 is executed using 1 threads, all the result will be put into 1 single `block`.

The total result size is estimated using `memory_tracker`: before executed aggregation, the overall memory usage is saved as `memory_usage_before_aggregation` in [`Aggregator`](https://github.com/pingcap/tiflash/blob/225dabe6942966d1562af738c287b50872f26be8/dbms/src/Interpreters/Aggregator.h#L908), and during the executed of stage 1, it use [`current_memory_usage - memory_usage_before_aggregation`](https://github.com/pingcap/tiflash/blob/225dabe6942966d1562af738c287b50872f26be8/dbms/src/Interpreters/Aggregator.cpp#L607) to decide if need to convert  the aggregated hash table into two-level hash table. And if the hash table is converted into two-level hash table, it will use multiple threads to do the stage 2.

The problem is after introducing `local_delta`, if the memory usage is less than 8MB, it will not update the global memory tracker. By default the `group_by_two_level_threshold_bytes` is 100MB, so assuming that the stage 1 is executed using 20 threads, and each thread uses 7.9MB, then the actual memory usage will be ~158MB,  but since all of these memory usages is tracked in `local_delta`, the global memory tracker does not see these memory usage, so the hash table will not be converted to two-level hash table, thus the stage 2 is executed using 1 thread.

windtalker:

Issue Number: close #4451

Problem Summary:

Actually, there are 2 possible solutions
- Consider local delta memory usage when estimate the [result_size_bytes](https://github.com/pingcap/tiflash/blob/27452c63ad68362912ef4303199952eb9f78c93e/dbms/src/Interpreters/Aggregator.cpp#L605) in aggregator on each execute of `executeOnBlock`
- In [`Aggregator::mergeAndConvertToBlocks`](https://github.com/pingcap/tiflash/blob/27452c63ad68362912ef4303199952eb9f78c93e/dbms/src/Interpreters/Aggregator.cpp#L1662), check the result_size_bytes, and if it exceeds the threshold, convert all the hash table into two-level hash table

For the first solution, converting hash table to two-level hash table can be done by each threads in the first stage of ParallelAggregating, and for the second solution, this converting things are executed in 1 threads.

I've done some test for both solutions, and found first solution has ~20% perfomance gain compared to the second solution for TPCH q15. So I choose the first solution.

Why not enable two-level hash table by default for all the cases:
- not all the agg method support two-level hash table
- test results shows that always use two-level hash table is slower if the table data is small, and the size of agg results is small.

Testing query: 
`select count(*),id from test group by id;`
`test` has 65536 rows, all rows have the same `id`, that is to say the above query return only 1 row
Testing cluster has only 1 TiFlash, and the cpu has 36 core, each query is running 1000 times.
|          | currency = 1 | currency = 5 |
|----|---|---|
|two level hash table| 22s | 40s|
|normal hash table| 16s | 26s|

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


SchrodingerZhu (approved):

dbms/src/Common/MemoryTracker.cpp
```diff
@@ -226,6 +226,11 @@ void submitLocalDeltaMemory()
     local_delta = 0;
 }
 
+Int64 getLocalDeltaMemory()
+{
+    return local_delta;
```
fuzhe1989, line 231: `local_delta` is thread-local, so different threads will see different deltas, is that what you expect?

SchrodingerZhu: https://github.com/pingcap/tiflash/pull/4570/files#diff-773c77885e15377996db622e99646f024c4815893a7d44878db9bc05bbc4bf6fR612

windtalker: Yes

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -600,7 +607,13 @@ bool Aggregator::executeOnBlock(const Block & block, AggregatedDataVariants & re
     size_t result_size = result.sizeWithoutOverflowRow();
     Int64 current_memory_usage = 0;
     if (current_memory_tracker)
+    {
         current_memory_usage = current_memory_tracker->get();
+        auto updated_local_delta_memory = CurrentMemoryTracker::getLocalDeltaMemory();
+        auto local_delta_memory_diff = updated_local_delta_memory - local_delta_memory;
+        current_memory_usage += (local_memory_usage.fetch_add(local_delta_memory_diff) + local_delta_memory_diff);
```
fzhedu, line 614: local_delta_memory_diff is added twice?

windtalker: No, `fetch_add` returns the value before added, so need to add it again.

fzhedu (approved):
LGTM

PR #4630, opened 2022-04-12
Fix create table error

### LINKED ISSUE #4596: Fail to create table with some TiDB table info
## Bug Report

A simplified case from https://asktug.com/t/topic/662946

```mysql
> set sql_mode='';
> CREATE TABLE test ( id int(16) NOT NULL AUTO_INCREMENT, PRIMARY KEY (id));
> alter table test add column b longblob not null default '';

-- p.s. Note that `set sql_mode=""` is required or the SQL will get blocked like
> CREATE TABLE test ( id int(16) NOT NULL AUTO_INCREMENT, PRIMARY KEY (id));
> alter table test add column b longblob not null default '';
ERROR 1101 (42000): BLOB/TEXT/JSON column 'b' can't have a default value
```

Those SQLs will create a table with JSON info as below. Note that the column type of "b" is not null long blob, but its "default"/"origin_default" field is `""` and "type.FLen" is `-1`.

```
{
	"cols": [{
		"comment": "",
		"default": null,
		"default_bit": null,
		"id": 1,
		"name": {
			"L": "id",
			"O": "id"
		},
		"offset": 0,
		"origin_default": null,
		"state": 5,
		"type": {
			"Charset": "binary",
			"Collate": "binary",
			"Decimal": 0,
			"Elems": null,
			"Flag": 515,
			"Flen": 16,
			"Tp": 3
		}
	}, {
		"comment": "",
		"default": "",
		"default_bit": null,
		"id": 15,
		"name": {
			"L": "b",
			"O": "b"
		},
		"offset": 12,
		"origin_default": "",
		"state": 5,
		"type": {
			"Charset": "binary",
			"Collate": "binary",
			"Decimal": 0,
			"Elems": null,
			"Flag": 4225,
			"Flen": -1,
			"Tp": 251
		}
	}],
	"comment": "",
	"id": 330,
	"index_info": [],
	"is_common_handle": false,
	"name": {
		"L": "test",
		"O": "test"
	},
	"partition": null,
	"pk_is_handle": true,
	"schema_version": 465,
	"state": 5,
	"update_timestamp":999 
}
```

<del>However, if we run `CREATE TABLE test ( id int(16) NOT NULL AUTO_INCREMENT, b longblob not null, PRIMARY KEY (id));` in TiDB, we can only get a JSON as below, which "default"/"origin_default" is `null` instead of `""`. And this won't make any error.</del>

Table create successfully

TiFlash crashes with `basic_string::_M_replace_aux`

v5.2.1

Lloyd-Pottiger:

Issue Number: close #4596

Problem Summary:

make sure `Flen >= 0` before `append`.  

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Lloyd-Pottiger:
Two remaining problems:
1. do we need to support unsigned data structure?
2. for the TODO in `case TypeBit {}`, the current behavior is in line with expectations.

dbms/src/Databases/test/gtest_database.cpp
```diff
@@ -688,7 +737,7 @@ try
     DatabaseLoading::loadTable(ctx, *db, meta_path, db_name, db_data_path, "TiFlash", "t_45.sql", false);
 
     // Get storage from database
-    const auto tbl_name = "t_45";
+    const auto * const tbl_name = "t_45";
```
ywqzzy, line 740: The second const is not neccessary.

Lloyd-Pottiger: ok, I will fix.

Lloyd-Pottiger, line 740:
```suggestion
    const auto * tbl_name = "t_45";
```

dbms/src/Storages/Transaction/TiDB.cpp
```diff
@@ -155,10 +155,11 @@ Field ColumnInfo::defaultValueToField() const
         auto v = value.convert<String>();
         if (hasBinaryFlag())
         {
-            // For binary column, we have to pad trailing zeros according to the specified type length.
+            // For some binary column(like varchar(20)), we have to pad trailing zeros according to the specified type length.
```
JaySon-Huang, line 158: Add some comment to explain why we should check flen

dbms/src/Databases/test/gtest_database.cpp
```diff
@@ -652,6 +652,55 @@ try
...
+        interpreter.execute();
+    }
+
+    auto db = ctx.getDatabase(db_name);
+
+    const String tbl_name = "t_111";
+    {
+        /// Create table
+        ParserCreateQuery parser;
+        const String stmt = fmt::format("CREATE TABLE `{}`.`{}` ", db_name, tbl_name) + R"stmt( (`id` Int32,`b` String) Engine = DeltaMerge((`id`),'{"cols":[{"comment":"","default":null,"default_bit":null,"id":1,"name":{"L":"id","O":"id"},"offset":0,"origin_default":null,"state":5,"type":{"Charset":"binary","Collate":"binary","Decimal":0,"Elems":null,"Flag":515,"Flen":16,"Tp":3}},{"comment":"","default":"","default_bit":null,"id":15,"name":{"L":"b","O":"b"},"offset":12,"origin_default":"","state":5,"type":{"Charset":"binary","Collate":"binary","Decimal":0,"Elems":null,"Flag":4225,"Flen":-1,"Tp":251}}],"comment":"","id":330,"index_info":[],"is_common_handle":false,"name":{"L":"test","O":"test"},"partition":null,"pk_is_handle":true,"schema_version":465,"state":5,"update_timestamp":99999}'))stmt";
```
JaySon-Huang, line 677: Better break down the R"stmt(..." into multiple lines for better reading https://en.cppreference.com/w/cpp/language/string_literal

Lloyd-Pottiger: fix

dbms/src/Storages/Transaction/TiDB.cpp
Lloyd-Pottiger: fix

JaySon-Huang (approved):
LGTM

ywqzzy (approved):

PR #4668, opened 2022-04-14
clear old range before apply snapshot

### LINKED ISSUE #4414: TiFlash can't gc totally after set tiflash replica 0
## Bug Report

udmASBWGW8
PCj4V4N2Od

v6.0

lidezhu:

Issue Number: close #4414

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
1. load data into tidb cluster with multiple tiflash nodes by import csv file and set tiflash replica to 1;
2. set tiflash replica to 0 and see whether the data can be reclaimed;

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Avoid left data on tiflash node after set tiflash replica to 0
```

### DIFF
--- a/dbms/src/Storages/Transaction/ApplySnapshot.cpp
+++ b/dbms/src/Storages/Transaction/ApplySnapshot.cpp
@@ -153,20 +153,30 @@ void KVStore::onSnapshot(const RegionPtrWrap & new_region_wrap, RegionPtr old_re
                 // Acquire `drop_lock` so that no other threads can drop the storage. `alter_lock` is not required.
                 auto table_lock = storage->lockForShare(getThreadName());
                 auto dm_storage = std::dynamic_pointer_cast<StorageDeltaMerge>(storage);
-                auto key_range = DM::RowKeyRange::fromRegionRange(
+                auto old_key_range = DM::RowKeyRange::fromRegionRange(
+                    old_region->getRange(),
+                    table_id,
+                    storage->isCommonHandle(),
+                    storage->getRowKeyColumnSize());
+                auto new_key_range = DM::RowKeyRange::fromRegionRange(
                     new_region_wrap->getRange(),
                     table_id,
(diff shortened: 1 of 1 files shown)


CalvinNeo (approved):
LGTM

dbms/src/Storages/Transaction/ApplySnapshot.cpp
```diff
@@ -153,20 +153,30 @@ void KVStore::onSnapshot(const RegionPtrWrap & new_region_wrap, RegionPtr old_re
...
+                    storage->getRowKeyColumnSize());
+                auto new_key_range = DM::RowKeyRange::fromRegionRange(
                     new_region_wrap->getRange(),
                     table_id,
                     storage->isCommonHandle(),
                     storage->getRowKeyColumnSize());
+                if (old_key_range != new_key_range)
+                {
+                    LOG_FMT_INFO(log, "clear region {} old range {} before apply snapshot of new range {}", region_id, old_key_range.toDebugString(), new_key_range.toDebugString());
+                    dm_storage->deleteRange(old_key_range, context.getSettingsRef());
```
gengliqi, line 169: Can we just delete the non-intersect range? Do these two methods have the same performance impact?

lidezhu: The number of delete ranges is the main factor on the performance. And after the first delete range, the latter one will be much quicker because most data in its range should have been deleted.

And consider the following case, If the old range is [1, 6), and the new range is [2, 4), so the non-intersect ranges are [1, 2) and [4, 6). So in this case it will have three delete ranges instead which will cause more performance impact. 

What's more, trying to delete only the non-intersect ranges will make the code much complicated. So considering that this  a corner case,  I suggest we keep it simple enough.

gengliqi (approved):

PR #4673, opened 2022-04-15
clear storage data before drop its metadata

lidezhu:

Issue Number: ref https://github.com/pingcap/tiflash/issues/3594

Problem Summary: when tiflash drop table, it do it in the following steps:
1. drop the metadata file;
2. call `drop` method on the storage object;

And after drop the metadata file, the table is considered dropped and won't be restored after restart. This is ok for PageStorage V2 because all data of a table is located in it's owe folder, so it's easy to manully fix it even this case happens.

But for PageStorage V3, all tables data is stored in shared PageStorage instances, and it's hard to clear a table's data manually. So a table must clear it's data before being dropped.

What's more, because we need to delete multiple directories in `IStorage::drop`, so it's hard to make it an atomic operation. So we cannot simply exchange the order of the two steps in the drop process to fix this problem.

try to clear data of storage before drop it's metadata;

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lidezhu:
I will fix the logic `InterpreterDropQuery` in this pr.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -399,19 +401,44 @@ void DeltaMergeStore::drop()
...
         }
     }
+}
+
+void DeltaMergeStore::clearData()
+{
+    // Remove all background task first
+    shutdown();
+    LOG_FMT_INFO(log, "Clear DeltaMerge segments data [{}.{}]", db_name, table_name);
+    // We don't drop the last segment in clearData, because if we drop it and tiflash crashes before drop it's metadata,
```
lidezhu, line 427:
```suggestion
    // We don't drop the first segment in clearData, because if we drop it and tiflash crashes before drop it's metadata,
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -399,19 +401,44 @@ void DeltaMergeStore::drop()
...
         }
     }
+}
+
+void DeltaMergeStore::clearData()
+{
+    // Remove all background task first
+    shutdown();
+    LOG_FMT_INFO(log, "Clear DeltaMerge segments data [{}.{}]", db_name, table_name);
+    // We don't drop the first segment in clearData, because if we drop it and tiflash crashes before drop it's metadata,
```
lidezhu, line 427:
```suggestion
    // We don't drop the first segment in clearData, because if we drop it and tiflash crashes before drop the table's metadata,
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -399,19 +401,44 @@ void DeltaMergeStore::drop()
...
+    // when restart it will try to restore the first segment but failed to do it which cause tiflash crash again.
```
lidezhu, line 428:
```suggestion
    // when restart the table will try to restore the first segment but failed to do it which cause tiflash crash again.
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -399,19 +401,44 @@ void DeltaMergeStore::drop()
...
+    // The reason this happens is that even we delete all data in a PageStorage instance,
+    // the call to PageStorage::getMaxId is still not 0 and then tiflash will try to restore it's first segment.
```
lidezhu, line 430:
```suggestion
    // the call to PageStorage::getMaxId is still not 0 so tiflash treat it as an old table and will try to restore it's first segment.
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -369,12 +369,8 @@ void DeltaMergeStore::rename(String /*new_path*/, bool clean_rename, String new_
     db_name.swap(new_database_name);
 }
 
-void DeltaMergeStore::drop()
+void DeltaMergeStore::dropAllSegments(bool keep_first_segment)
```
JinheLin, line 372: I think we can add some failpoints to help testing whether the restoring and removing data perform as expected.

lidezhu: Added.

lidezhu:
Remove enable_ps_v3 config related code after this pr merge: https://github.com/pingcap/tiflash/pull/4831

JaySon-Huang (dismissed):
LGTM

dbms/src/Debug/dbgFuncMisc.cpp
```diff
@@ -97,4 +98,13 @@ void dbgFuncSearchLogForKey(Context & context, const ASTs & args, DBGInvoker::Pr
     else
         output("Invalid");
 }
+
+void dbgFuncTriggerGlobalPageStorageGC(Context & context, const ASTs & /*args*/, DBGInvoker::Printer /*output*/)
+{
+    auto global_storage_pool = context.getGlobalStoragePool();
+    if (global_storage_pool)
+    {
+        global_storage_pool->meta()->gc();
```
JaySon-Huang, line 107: what about also calling GC on the global data and log?

dbms/src/Debug/DBGInvoker.cpp
```diff
@@ -121,6 +121,7 @@ DBGInvoker::DBGInvoker()
 
     regSchemalessFunc("search_log_for_key", dbgFuncSearchLogForKey);
     regSchemalessFunc("tidb_dag", dbgFuncTiDBQueryFromNaturalDag);
+    regSchemalessFunc("trigger_global_storage_pool_gc", dbgFuncTriggerGlobalPageStorageGC);
```
JaySon-Huang, line 124:
```suggestion
    regSchemalessFunc("gc_global_storage_pool", dbgFuncTriggerGlobalPageStorageGC);
```

tests/delta-merge-test/raft/schema/partition_table_restart.test
```diff
@@ -35,6 +35,8 @@
 # schema syncer guarantees logical table creation at last, so there won't be cases that logical table exists whereas physical table not.
 => drop table default.test
 => DBGInvoke __reset_schemas()
+# remove obsolete entry left by previous dropped table
+=> DBGInvoke __trigger_global_storage_pool_gc()
```
JaySon-Huang, line 39:
```suggestion
=> DBGInvoke __gc_global_storage_pool()
```

dbms/src/Debug/dbgFuncMisc.h
```diff
@@ -26,4 +26,9 @@ class Context;
 //   ./storage-client.sh "DBGInvoke search_log_for_key(key)"
 void dbgFuncSearchLogForKey(Context & context, const ASTs & args, DBGInvoker::Printer output);
 
+// Trigger the gc process of global storage pool. Used to remove obsolete entry left by previous dropped table
```
JaySon-Huang, line 29:
```suggestion
// Trigger the gc process of global storage pool. Used to remove obsolete entries left by the previous dropped table
// Note that it is ok for now since we don't store external/ref on GlobalStoragePool.meta, or it may run into same
// problem as https://github.com/pingcap/tiflash/pull/4850
```

dbms/src/Debug/dbgFuncMisc.cpp
lidezhu: added.

JinheLin (approved):

lidezhu:
@JaySon-Huang PTAL

JaySon-Huang (approved):
LGTM

PR #4674, opened 2022-04-15
[DNM] refine handle table scan

SeaRise:

Issue Number: ref #4118

Problem Summary:

- remove the use of `DAGQueryBlock` in `DAGStorageInterpreter`
- move `StorageWithStructureLock`(table metadata and table lock) from `DAGStorageInterpreter` to `TiDBStorageTable`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -156,35 +181,27 @@ void DAGStorageInterpreter::execute(DAGPipeline & pipeline)
...
-
     analyzer = std::make_unique<DAGExpressionAnalyzer>(std::move(source_columns), context);
 
     FAIL_POINT_PAUSE(FailPoints::pause_after_learner_read);
 
     if (!mvcc_query_info->regions_query_info.empty())
         doLocalRead(pipeline, settings.max_block_size);
 
-    null_stream_if_empty = std::make_shared<NullBlockInputStream>(storage_for_logical_table->getSampleBlockForColumns(required_columns));
+    null_stream_if_empty = std::make_shared<NullBlockInputStream>(storage_table.getLogicalTableStorage()->getSampleBlockForColumns(required_columns));
```
JaySon-Huang, line 191:
```suggestion
    null_stream_if_empty = std::make_shared<NullBlockInputStream>(storage_table.getSampleBlockForColumns(required_columns));
```
Maybe adding `getSampleBlockForColumns` in stroage_table and making "logical table storage" hind from that class?

dbms/src/Flash/Coprocessor/TiDBTableScan.h
```diff
@@ -48,11 +48,12 @@ class TiDBTableScan
     }
     String getTableScanExecutorID() const
     {
-        return table_scan->executor_id();
+        return executor_id;
```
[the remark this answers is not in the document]
SeaRise: For coprocess read, table_scan->executor_id() may be empty.
query_block.source_name is guaranteed not to be empty(If executor_id is empty, an id will be assigned).
Would it be better to share the same executor_id for local read and remote read?

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
SeaRise: done

windtalker:
This pr move lock the table before do learner read, this is the same as our first implemention, but we met some deadlock related issue, and after, [this pr](https://github.com/pingcap/tiflash/pull/864) moved the `lock table` logical after leaner read is done. So maybe we should not lock the table so early. @JaySon-Huang could you please reconfirm this?

SeaRise:
> This pr locks the table before doing learner read, this is the same as our first implemention, but we met some deadlock related issue, and [this pr](https://github.com/pingcap/tiflash/pull/864) moved the `lock table` logical after leaner read is done. So maybe we should not lock the table so early. @JaySon-Huang could you please reconfirm this?

Maybe I need to move `learner read` to `TiDBStroageTable` too

JaySon-Huang:
@SeaRise I prefer to do it this way https://github.com/SeaRise/tiflash/pull/1

1. Only gets and locks storage(s) after learner read, and add comments about why
2. `DAGQueryBlockInterpreter` does not know about "TiDBStorageTable" except that inside `DAGQueryBlockInterpreter::handleTableScan`, receive a ptr to that to take over the locks
3. Make the ctor of `TiDBStorageTable` to be more simple, use a static method instead

PR #4724, opened 2022-04-20
Introduce FiberRWLock

### LINKED ISSUE #4725: We need a shared_lock under fiber env
## Enhancement

`boost::fibers::mutex` works well with `std::unique_lock` and is a perfect substitute of `std::mutex`. However there's no such a class as substitute of `std::shared_mutex`.

fuzhe1989:

Issue Number: close #4725 

Problem Summary:

Introduce `FiberRWLock` which could work with `std::shared_lock` under both fiber and thread env.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


LittleFall:
format check failed

fuzhe1989:
> format check failed

@LittleFall Sorry I'll fix it tonight. Anyway it's reviewable.

dbms/src/Common/FiberRWLock.h
```diff
@@ -0,0 +1,97 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <boost/fiber/all.hpp>
+
+namespace DB
+{
+/// FiberRWLock has the same interface with `std::shared_mutex` and is intended to
```
ywqzzy, line 21: Could you specific which type of the RWLock in the annotation, eg: Read first or Write first.

dbms/src/Common/FiberRWLock.h
```diff
@@ -0,0 +1,97 @@
...
+        std::unique_lock lock(mu);
+        if (--reader_cnt == 0 && writer_wait_cnt != 0)
+            writer_cv.notify_one();
+    }
+
+private:
+    boost::fibers::mutex mu;
+    boost::fibers::condition_variable reader_cv;
+    boost::fibers::condition_variable writer_cv;
+    int reader_cnt = 0;
```
ywqzzy, line 93: In class __shared_mutex_base, it only use one unsigned _state_ to track the reader or writer's state,  maybe it is more consice.  ref: https://zhuanlan.zhihu.com/p/463286822

dbms/src/Common/FiberRWLock.h
fuzhe1989: good idea.

dbms/src/Common/FiberRWLock.h
fuzhe1989: It combines two logical states into one field and messes them up. Since the memory usage isn't an issue, I'd like to use two states for better readability.

fuzhe1989:
@ywqzzy @LittleFall PTAL

dbms/src/Common/tests/gtest_fiber_rw_lock.cpp
```diff
@@ -0,0 +1,347 @@
...
+            std::shared_lock lock(rw_lock);
+            t1_entered.set_value();
+            future.get();
+        });
+
+        typename Traits::Thread t2([&] {
+            t1_entered_future.get();
+            std::unique_lock lock(rw_lock);
+            t2_entered.set_value();
+            ;
```
ywqzzy, line 238:
```suggestion
            
```

ywqzzy (approved):
LGTM

dbms/src/Common/tests/gtest_fiber_rw_lock.cpp
```diff
@@ -0,0 +1,347 @@
...
+            std::shared_lock lock(rw_lock);
+            t1_entered.set_value();
+            future.get();
+        });
+
+        typename Traits::Thread t2([&] {
+            t1_entered_future.get();
+            std::unique_lock lock(rw_lock);
+            t2_entered.set_value();
+
```
LittleFall, line 238:
```suggestion
```

dbms/src/Common/tests/gtest_fiber_rw_lock.cpp
```diff
@@ -0,0 +1,347 @@
...
+        typename Traits::Thread t2([&] {
+            t1_entered_future.get();
+            std::unique_lock lock(rw_lock);
+            t2_entered.set_value();
+            
+        });
+
+        std::atomic_bool t3_entered = false;
+        typename Traits::Thread t3([&] {
```
LittleFall, line 242: can we add a new test that makes t2 enter firstly, and then t3 enter, to ensure the assumption of write first?

LittleFall (approved):
LGTM

dbms/src/Common/tests/gtest_fiber_rw_lock.cpp
fuzhe1989: I tried but found it's hard in that we have to notify t3 after t2 starts waiting, but in this time t2 can't notify t3 (it's waiting!). Do you have any suggestion?

dbms/src/Common/FiberRWLock.h
```diff
@@ -0,0 +1,101 @@
...
+            writer_cv.notify_one();
+        else
+            reader_cv.notify_all();
+    }
+
+    void lock_shared()
+    {
+        std::unique_lock lock(mu);
+        reader_cv.wait(lock, [&] { return writer_cnt == 0 && writer_wait_cnt == 0; });
+        ++reader_cnt;
```
gengliqi, line 72: What if `reader_cnt` is the maximum of `int`?

dbms/src/Common/FiberRWLock.h
```diff
@@ -0,0 +1,101 @@
...
+        if (--reader_cnt == 0 && writer_wait_cnt != 0)
+            writer_cv.notify_one();
+    }
+
+private:
+    boost::fibers::mutex mu;
+    boost::fibers::condition_variable reader_cv;
+    boost::fibers::condition_variable writer_cv;
+    int reader_cnt = 0;
+    int writer_cnt = 0;
```
gengliqi, line 98: Seems it can be `bool`?

dbms/src/Common/FiberRWLock.h
```diff
@@ -0,0 +1,101 @@
...
+    FiberRWLock(const FiberRWLock &) = delete;
+    ~FiberRWLock() = default;
+
+    FiberRWLock & operator=(const FiberRWLock &) = delete;
+
+    void lock()
+    {
+        std::unique_lock lock(mu);
+        ++writer_wait_cnt;
+        writer_cv.wait(lock, [&] { return writer_cnt == 0 && reader_cnt == 0; });
```
gengliqi, line 42: I worry about the fairness of readers in this implementation.
For instance：
1. a writer holds a lock and some readers wait for this lock
2. another writer comes and adds `writer_wait_cnt`.
3. the first writer releases the lock and the second writer will get the lock no matter how long these readers wait.

I saw another implementation of rwlock in cmudb(cmu 15-445) that solves this problem.https://paste.ofcode.org/jKfGZ23H977wszmgmPMRdW
The writers and readers wait on `reader condition variable` first. So in above case, the second writer can not get the lock because it is the latter one.

gengliqi: Another reference, the implementation of `shared_mutex` https://code.woboq.org/llvm/libcxx/src/shared_mutex.cpp.html, it is basically the same as https://paste.ofcode.org/jKfGZ23H977wszmgmPMRdW.

JaySon-Huang:
/close

not introuding fiber so far

PR #4746, opened 2022-04-25
ignore delmark when add minmax for pk column

### LINKED ISSUE #4747: DeltaMerge return wrong result
## Bug Report

On a customer environment, select on a table return the following error:
DeltaMerge return wrong result, current handle[8070450532252511967]version[432677394077974556]@read[56]@pos[2022] is expected >= last_handle[8070450532252560947]last_version[432677778184470534]@read[56]@pos[2021] 

all relase branch influenced

lidezhu:

Issue Number: close #4747

Problem Summary:
If rows in a stable pack is all mark deleted, then it's minmax index for pk column is both 0.
And when `dt_enable_skippable_place` is enabled and try to place delta index on this segment, it will filter out all pack which is not in the pk range [start_key, +infinite). So the previous stable pack is ignored.
But for the place delta index algorithm, it's expected that all the rows in the pk range [start_key, +infinite) is placed no matter whether it's deleted. So when we try to read this segment using the placed delta index, it reports error like "DeltaMerge return wrong result"

Ignore delmark when add minmax for pk column.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix potential query error when select on a table with many delete operations
```


flowbehappy (approved):

JaySon-Huang (approved):
LGTM

PR #4765, opened 2022-04-26
BlobStore change the selection strategy.

jiaqizho:

Issue Number: ref #3594 

Problem Summary:

image
In multi-paths, BlobStore always chooses the lowest valid_rate to write. Then it will make the data imbalance.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/Page/V3/BlobStore.cpp
+++ b/dbms/src/Storages/Page/V3/BlobStore.cpp
@@ -1150,38 +1150,21 @@ std::pair<BlobStatPtr, BlobFileId> BlobStore::BlobStats::chooseStat(size_t buf_s
     auto stats_iter = stats_map.begin();
     std::advance(stats_iter, stats_map_path_index);
 
-    size_t path_iter_idx = 0;
-    for (path_iter_idx = 0; path_iter_idx < stats_map.size(); ++path_iter_idx)
+    // Try to find a suitable stat under current path (path=`stats_iter->first`)
+    for (const auto & stat : stats_iter->second)
     {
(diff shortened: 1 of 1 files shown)


jiaqizho:
@JaySon-Huang 
change back to the strategy
1. choose the path.
2. find the stat in the current path.

Is that make sense?

JaySon-Huang:
IMO, these changes don't make data to be more balance🤔

JaySon-Huang:
maybe we should iterate all path, and try to get "the smallest valid_rate inside each path" instead of "the smallest valid_rate among all paths"?
More specified, move line 1138 into the `for (path_iter_idx = 0; path_iter_idx < stats_map.size(); ++path_iter_idx)` loop

jiaqizho:
> maybe we should iterate all path, and try to get "the smallest valid_rate inside each path" instead of "the smallest valid_rate among all paths"? More specified, move line 1138 into the `for (path_iter_idx = 0; path_iter_idx < stats_map.size(); ++path_iter_idx)` loop

I am not sure whether I got your point. These changes' main logic is:

1. round-robin to choose the path
2. select a `BlobStat` by `valid_rate` from the current path. If can't find one, create a new `blobstat`

old logic is:

1. loop all `BlobStat` and find the lowest `valid_rate` one.
2. if can't find create a new one.

jiaqizho:
not root case I thought.

PR #4831, opened 2022-05-06
enable PageStorageV3 by default

lidezhu:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Server/StorageConfigParser.cpp
+++ b/dbms/src/Server/StorageConfigParser.cpp
@@ -197,24 +197,24 @@ void TiFlashStorageConfig::parseMisc(const String & storage_section, Poco::Logge
         format_version = *version;
     }
 
-    if (auto lazily_init = table->get_qualified_as<Int32>("lazily_init_store"); lazily_init)
-    {
-        lazily_init_store = (*lazily_init != 0);
-    }
-
-
-    if (table->contains("enable_ps_v3"))
-    {
-        if (auto enable_v3 = table->get_qualified_as<Int32>("enable_ps_v3"); enable_v3)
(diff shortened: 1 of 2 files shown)


JaySon-Huang (approved):
LGTM

jiaqizho (approved):

dbms/src/Storages/StorageDeltaMerge.cpp
```diff
@@ -95,6 +97,12 @@ StorageDeltaMerge::StorageDeltaMerge(
         is_common_handle = tidb_table_info.is_common_handle;
         pk_is_handle = tidb_table_info.pk_is_handle;
     }
+    else
+    {
+        const auto mock_table_id = MockTiDB::instance().newTableID();
```
lidezhu, line 102: It seems we cannot restore the table created through ch client after restart tiflash.

JaySon-Huang: Well
* We don't have any crash test cases in the mock test now, and no plan to add in the near future
* Even in mock test, `DBGInvoke mock_tidb_table` is the recommended way to create a table
* Those test cases using `create table ... ` through ch-client should be rewritten in other forms IMO, but I don't get time to do that. This is a workaround other than skipping those cases

lidezhu:
> A few of UT such as `gtest_dm_delta_merge_store.cpp` still use PageStorageV2

Can you create an issue for it and give some more explanation about why PageStorageV2 in these tests? So we can fix it later.

PR #4843, opened 2022-05-08
Add arg remove_invalid_files into restore

jiaqizho:

Issue Number: ref #3594 

Problem Summary:
- When we use mix mode to run `PageStorage`.Then V2 and V3 will use the same `StoragePool` in `KVStore`. So they will shared path `data/kvstore/`
- So when V2 restore, it will remove invalid files(detected not PageFile) by default. If we have V3 data in disk, it will be removed.

- add a arg to make sure won't remove invalid file when restore happened in KVStore

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/Page/PageStorage.h
+++ b/dbms/src/Storages/Page/PageStorage.h
@@ -211,7 +211,13 @@ public:
 
     virtual ~PageStorage() = default;
 
-    virtual void restore() = 0;
+    // Should not set `remove_invalid_file` to false except `KVStore`.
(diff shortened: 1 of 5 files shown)


jiaqizho:
Giveup this change. Will changed `delegate raft` into `globalmulti` in `Mix Mode`.

Issue #4473, opened 2022-03-28
execution time of simple agg query is not stable


windtalker:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. create a table and load some data into it.
2. run query select count(*), id from t group by id
### 2. What did you expect to see? (Required)
The running time should be stable
### 3. What did you see instead (Required)
The running time is not stable, for example, it could be varies from 0.2s - 0.6s
### 4. What is your TiFlash version? (Required)
master @ 225dabe6942966d1562af738c287b50872f26be8
### 5. Root cause
In TiFlash, aggregation will [recude current execution concurrency to 1](https://github.com/pingcap/tiflash/blob/5af453045984a4907b8dd09ebf6cc7534dd845ea/dbms/src/Flash/Coprocessor/DAGQueryBlockInterpreter.cpp#L820), in order to get better performance, we add a [restoreConcurrency](https://github.com/pingcap/tiflash/blob/5af453045984a4907b8dd09ebf6cc7534dd845ea/dbms/src/Flash/Coprocessor/InterpreterUtils.cpp#L27) function to restore the concurrency before aggregation:
The basic idea is put a `SharedQueryBlockInputStream`, and using n threads to read from `SharedQueryBlockInputStream`. 

In `SharedQueryBlockInputStream`, it use a [concurrent queue](https://github.com/pingcap/tiflash/blob/225dabe6942966d1562af738c287b50872f26be8/dbms/src/DataStreams/SharedQueryBlockInputStream.h#L152) to feed the block to it's readers. Each reader thread pull from that concurrent queue. And if `SharedQueryBlockInputStream` get EOF from its child input stream, it will return [`{}`](https://github.com/pingcap/tiflash/blob/225dabe6942966d1562af738c287b50872f26be8/dbms/src/DataStreams/SharedQueryBlockInputStream.h#L131) for the first reader thread it meet after got EOF. For other reading threads, it depends on the cancel flag in `SharedQueryBlockInputStream`: for each pull loop, the reading threads will check [if the current `SharedQueryBlockInputStream` is cancelled](https://github.com/pingcap/tiflash/blob/225dabe6942966d1562af738c287b50872f26be8/dbms/src/DataStreams/SharedQueryBlockInputStream.h#L106), if it is cancelled, the reader thread just return `{}`.

[...]

PR #4855, opened 2022-05-10
avoid premature exit of threads in `ParallelInputsProcessor`

### LINKED ISSUE #4856: `ParallelInputsProcessor` does not work well with extreme data skew
## Enhancement

https://github.com/pingcap/tiflash/blob/master/dbms/src/DataStreams/ParallelInputsProcessor.h#L287-L289
https://github.com/pingcap/tiflash/blob/master/dbms/src/DataStreams/ParallelInputsProcessor.h#L314-L315

If the number of threads is 10 and the number of streams is 10.
Only one stream `A` can be read 100 times and the other streams can only be read once.
In the end, only one thread will execute stream `A`, instead of 10 threads executing together.

SeaRise:

Issue Number: close #4856

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -353,6 +357,8 @@ class ParallelInputsProcessor
     using AvailableInputs = std::queue<InputData>;
     AvailableInputs available_inputs;
 
+    size_t may_available_inputs;
```
windtalker, line 360: Maybe use a better name like `active_inputs`?

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -106,6 +106,7 @@ class ParallelInputsProcessor
         , additional_input_at_end(additional_input_at_end_)
         , max_threads(std::min(inputs_.size(), max_threads_))
         , handler(handler_)
+        , may_available_inputs(inputs_.size())
```
windtalker, line 109: I think we should update this vairable in L232, it looks a little unsafe to me to set the value directly here.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -285,9 +286,12 @@ class ParallelInputsProcessor
                 std::lock_guard lock(available_inputs_mutex);
 
                 /// If there are no free sources, then this thread is no longer needed. (But other threads can work with their sources.)
-                if (available_inputs.empty())
+                if (0 == may_available_inputs)
```
windtalker, line 289: Sleep a while before next try?

windtalker: Or use condition variable since sleep will introduce extra random latency for queries, like `SharedQueryBlockInputStream`, refer to https://github.com/pingcap/tiflash/issues/4473 for details.

SeaRise: how about using `MPMCQueue`?

fuzhe1989:
How about use MPMCQueue to simplify the state management? 

Pseudo code example:

```
stream_cnt := atomic(n)
for each thread:
    pop one stream from queue
    if pop failed:
        return
    handle stream
    if stream is finished:
        if --stream_cnt == 0:
            queue.finish()
    else:
        push stream into queue
```

SeaRise:
will fixed by https://github.com/pingcap/tiflash/pull/5274

Issue #4728, opened 2022-04-21
Reduce the memory usage after TiFlash reboot.


flowbehappy:
## Enhancement
We found that after TiFlash reboot, the memory usage rises quickly. Especially the data size is large.

flowbehappy:
Confirmed that it is caused by huge transactions. Currently, TiFlash caches all uncommitted transaction data in memory. And looks like the space amplification is large. For example, ~50GB of uncommitted data causes TiFlash ~400GB to process.

lidezhu:
Reproduce step:
1. load tpch50 to tiflash;
2. run the sql `update lineitem set L_SUPPKEY=100000 limit 50000000;`

tidb metric:
![QQHNnMyDbQ](https://user-images.githubusercontent.com/47731263/164713615-d265c7f2-df53-46e7-931b-8c405105b9ab.jpg)
tiflash metric:
![CJasTIDVCR](https://user-images.githubusercontent.com/47731263/164713637-0aa9f3f5-0273-4df9-a01d-06d33298b617.jpg)
dashboard:
![oL5HpnnBu2](https://user-images.githubusercontent.com/47731263/164713731-2dd8aff3-5810-45fc-b259-affb464d1c0d.jpg)

lidezhu:
![image](https://user-images.githubusercontent.com/47731263/165215389-619e5bc2-e2f6-4ff6-91fa-280df20d9961.png)
About 10GB memory is consumed by raftstore entry cache.

flowbehappy:
So the memory space amplification is ~6.5x.

flowbehappy:
According to https://github.com/pingcap/tiflash/pull/4883, the memory amplification is ~1.5x starting from v6.3.

PR #4883, opened 2022-05-13
Optimize apply speed under heavy write pressure

lidezhu:

Issue Number: ref #4728 

Problem Summary: When tiflash is under heavy write pressure, it will consume a lot of write throughput and cause some latency on the process of apply raft log, which will make tiflash consume a log memory.

1. Only allow one delta flush process on a segment to process at any time to avoid unnecessary throughput consumption;
2. When processing `CompactLog` raft command, just try to flush the unsaved value in storage once and return `Persist` if the flush succed or `None` if the flush failed.(Originally we try to flush until it success)
3. Increase foreground flush threshold for write thread.
4. Avoid add minor compaction task to background if the segment is flushing to save some throughput and reduce lock contention on the segment.

### Check List

Tests 

Test step:
1. perform large transaction update through 5 tidb instance at the same time and the total transaction size of all tidb instance is about 24.5 GB;
Before optimize:
image
After optimize:
image

Detail result(unit: GB)
  | 1st run | 2nd run | 3rd run | 4th run | 5th run | 6th run) | 7th run | average value | amplification factor
-- | -- | -- | -- | -- | -- | --| -- | -- | --
before optimization | 40.8 | 37.2 | 37.3 | 36 | 34.7 | 33.6 | 34.5 | 36.3 | 1.48
after optimization | 38.1 | 29.4 | 29 | 34.3 | 29.6 | 32.6 | 37.7 | 32.9 | 1.34

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


flowbehappy (approved):
LGTM

flowbehappy:
Can you add some metrics to show the result of this optimization? @lidezhu

lidezhu:
> Can you add some metrics to show the result of this optimization? @lidezhu

Done

CalvinNeo (approved):
LGTM

PR #4885, opened 2022-05-13
feat: support the new compact RPC request

breezewish:

Issue Number: ref #4897

Problem Summary:

This PR adds the implementation of the Compact RPC. The Compact RPC is defined as:

- Compact one or more data partials (Segments in DeltaTree) according to the specified Table ID, and returns the key for compacting more data partials.

- See detailed RPC interface definition in https://github.com/pingcap/kvproto/pull/918

The TiDB PR to utilize this RPC endpoint: https://github.com/pingcap/tidb/pull/34741

New:

- The impl of Compact RPC: `Flash/Management/ManualCompact.cpp|h`.

- Public API to compact one segment in DeltaTree: `Storages/DeltaMerge/*`: Mainly `std::optional<DM::RowKeyRange> mergeDeltaBySegment(const Context & context, const DM::RowKeyValue & start_key);`

- Unit tests about the compact RPC and compact one segment: See `tests`.

Change:

- Moved `tiflashErrorCodeToGrpcStatusCode` from the `CoprocessorHandler` to a new shared utility called `ServiceUtils`, so that Compact RPC can use it as well.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [x] Affects user behaviors
- [x] Contains syntax changes

### Release note

None, as the user interface is in the TiDB side.

```release-note
None
```


dbms/src/Flash/Management/ManualCompact.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include <future>
+
+namespace DB
+{
+
+namespace Management
+{
+
+ManualCompactManager::ManualCompactManager(const Context & db_context_)
+    : db_context(db_context_)
```
breezewish, line 32: Not sure whether db_context is ok (or using a global_context better?)

breezewish:
@flowbehappy @lidezhu @JaySon-Huang The WIP in the title is only reserved for kvproto reference changes. There are no code base WIP remaining in this PR. Please take a review when you are free, thanks!

dbms/src/Flash/Management/ManualCompact.cpp
JaySon-Huang: This should be a reference to "global_context" as ManualCompactManager lives through the lifetime of the TiFlash process.
"db_context" usually indicates its lifetime is only for one query session

dbms/src/Storages/Page/PageUtil.h
```diff
@@ -77,6 +77,7 @@ extern const int FILE_SIZE_NOT_MATCH;
 
 namespace FailPoints
 {
+extern const char skip_check_segment_update[];
```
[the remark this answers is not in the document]
breezewish: This failpoint is enabled in https://github.com/pingcap/tiflash/blob/9407b18b3dd909a5ef3895ac01abb8d9ea163b83/dbms/src/Storages/DeltaMerge/tests/MultiSegmentTestUtil.h#L66 so that the splitted segment keeps stable and will not get merged or splitted again.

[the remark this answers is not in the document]
breezewish: Oh, good catch!

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1027,6 +1025,43 @@ void DeltaMergeStore::mergeDeltaAll(const Context & context)
     }
 }
 
+std::optional<DM::RowKeyRange> DeltaMergeStore::mergeDeltaBySegment(const Context & context, const RowKeyValue & start_key)
+{
+    auto dm_context = newDMContext(context, context.getSettingsRef(), /*tracing_id*/ "mergeDeltaBySegment");
+    if (start_key.is_common_handle != dm_context->is_common_handle)
+    {
+        return std::nullopt;
```
flowbehappy, line 1033: Simply throw an exception. It is unexpected status.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1027,6 +1025,43 @@ void DeltaMergeStore::mergeDeltaAll(const Context & context)
...
+        {
+            const auto segment_end = new_segment->getRowKeyRange().end;
+            if (unlikely(*segment_end.value <= *start_key.value))
+            {
+                // The next start key must be > current start key
+                LOG_FMT_ERROR(log, "Assert new_segment.end {} > start {} failed", segment_end.toDebugString(), start_key.toDebugString());
+                throw Exception("Assert segment range failed", ErrorCodes::LOGICAL_ERROR);
+            }
+            return new_segment->getRowKeyRange();
+        }
```
flowbehappy, line 1061: I think we should try again if a DeltaMerge failed. Because `full compact` is meant to do compactions for all segments. And we rely on it to clean up data in PS V2.

flowbehappy: And since DeltaMerge failures can be caused by Merge or Split, we need to obtain a new segment instance each time we failed, from the `segments` map by `start_key`.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
breezewish: The `start_key` comes from the RPC request. From the TiFlash perspective maybe better to think all RPC callers to be a bad guy in order to isolate component errors. Otherwise a mistake in TiDB may cause TiFlash to crash.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
breezewish: I implemented a retry mechanism from L1037 to L1048, retrieves a segment each time. Could you help confirm whether this is sufficient?

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
breezewish: Or may be better to leave this check in the ManualCompactManager? We could keep a property that all calls to `DeltaMergeStore` should be checked before calling.

dbms/src/Flash/Management/ManualCompact.cpp
```diff
@@ -0,0 +1,227 @@
...
+    {
+        try
+        {
+            ReadBufferFromString buf(request->start_key());
+
+            // TODO: This function seems to be not safe for accepting arbitrary client inputs.
+            // When client passes a string with "invalid length" encoded, e.g. a very large length, we will allocate memory first and then discover the input is invalid.
+            // This will cause OOM.
+            // Also it is not a good idea to use Try-Catch for this scenario.
+            start_key = DM::RowKeyValue::deserialize(buf);
```
JaySon-Huang, line 143: mark

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
flowbehappy: Sry. I miss the `while(true)`. LGTM.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
flowbehappy: IMO, it is better to rise some serious thing to notify the `unexpected status`, as it implies bugs usually. How about returning errors to TiDB when there are invalid params?

flowbehappy: BTW throw exception here should not crash TiFlash.

dbms/src/Interpreters/Settings.h
```diff
@@ -360,7 +360,11 @@ struct Settings
...
+    M(SettingUInt64, preallocated_request_count_per_poller, 20, "grpc preallocated_request_count_per_poller")                                                                                                                           \
+                                                                                                                                                                                                                                        \
+    M(SettingUInt64, manual_compact_pool_size, 1, "The number of worker threads to handle manual compact requests.")                                                                                                                    \
+    M(SettingUInt64, manual_compact_max_concurrency, 10, "Max concurrent tasks. It should be larger than pool size.")                                                                                                              \
+    M(SettingUInt64, manual_compact_more_until_ms, 60000, "Continuously compact more segments until reaching specified elapsed time. If 0 is specified, only one segment will be compacted each round.")
```
JaySon-Huang, line 367: Seems these two configs are not shown in the "config-template.toml" file?

breezewish: Yes, they are advanced config that normal users should not modify.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
breezewish: I have moved this check to the TiFlash service side. Please take a look again, thanks!

Also added a test in the DM to ensure that, when common handle is given for a int handle table, there will be an exception thrown.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1027,6 +1025,38 @@ void DeltaMergeStore::mergeDeltaAll(const Context & context)
...
+            std::shared_lock lock(read_write_mutex);
+            const auto segment_it = segments.upper_bound(start_key.toRowKeyValueRef());
+            if (segment_it == segments.end())
+            {
+                return std::nullopt;
+            }
+            segment = segment_it->second;
+        }
+
+        const auto new_segment = segmentMergeDelta(*dm_context, segment, TaskRunThread::Foreground);
```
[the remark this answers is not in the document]
breezewish: I'm not sure but, looks like `TaskRunThread` looks to serve similar purpose as `ThreadType`. How about merging them together later (in another PR)?

```
    enum ThreadType
    {
        Init,
        Write,
        Read,
        BG_Split,
        BG_Merge,
        BG_MergeDelta,
        BG_Compact,
        BG_Flush,
        BG_GC,
    };

    enum TaskRunThread
    {
        BackgroundThreadPool,
        BackgroundGCThread,
        Foreground,
    };
```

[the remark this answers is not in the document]
lidezhu: `TaskRunThread` is just for delta merge tasks to distinguish where the delta merge task is running. It seems that it's not a good name. We can change the name later.
`ThreadType` corresponds to the more generous concept of "thread type". And we can leave it as is.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1027,6 +1025,43 @@ void DeltaMergeStore::mergeDeltaAll(const Context & context)
     }
 }
 
+std::optional<DM::RowKeyRange> DeltaMergeStore::mergeDeltaBySegment(const Context & context, const RowKeyValue & start_key)
+{
+    auto dm_context = newDMContext(context, context.getSettingsRef(), /*tracing_id*/ "mergeDeltaBySegment");
```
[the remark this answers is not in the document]
breezewish: Updated.

Just curious, if we don't update the safepoint here, what will happen in worst case?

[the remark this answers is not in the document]
breezewish: BTW updating GC safepoint here means we will send 1 PD request for every segment. Is this fine?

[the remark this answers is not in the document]
breezewish: Another solution is to let `updateGCSafePoint` to be public, so that it will only get called once per compact request.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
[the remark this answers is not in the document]
breezewish: Got it. Now I added a TODO for renaming it. I will do this when 6.1 is switched out from master.

flowbehappy:
BTW, plz add this setting `manual_compact_pool_size`  to the document file. https://docs.pingcap.com/zh/tidb/v6.0/tiflash-configuration

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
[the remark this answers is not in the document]
JaySon-Huang: > Just curious, if we don't update the safepoint here, what will happen in worst case?

When the origin segment doesn't have any data in delta, and the stable contains rows that are older than gc safepoint. Even after we apply "merge delta" on the segment, we take some resources to rewrite the stable data (but it is effectively the same as before), but we don't get any performance improvement.

lidezhu (approved):
LGTM

Issue #4872, opened 2022-05-12
dtworkload run failed


JinheLin:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
./tiflash dtworkload --testing_type=daily_perf --verify_round=100
```

### 2. What did you expect to see? (Required)

Run successfully without any exception.

### 3. What did you see instead (Required)

Run failed.

### 4. What is your TiFlash version? (Required)

```
sh-4.2# /tiflash/tiflash --version
TiFlash
Release Version: v6.1.0-nightly
Edition:         Community
Git Commit Hash: c0459782af4499b98bef3e3edc523e103827e38b
Git Branch:      heads/refs/tags/v6.1.0-nightly
UTC Build Time:  2022-05-11 22:58:31
Enable Features: jemalloc avx avx512 unwind
Profile:         RELWITHDEBINFO

Raft Proxy
Git Commit Hash:   a03434d3758bee4fa335ce87da5f772eebe8f9cc
Git Commit Branch: HEAD
UTC Build Time:    2022-05-11 23:04:39
Rust Version:      rustc 1.60.0-nightly (1e12aef3f 2022-02-13)
Storage Engine:    tiflash
Prometheus Prefix: tiflash_proxy_
Profile:           release

```

Issue #4849, opened 2022-05-09
Exception when PageStorage V3 restore


hehechen:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
Enable PageStorage V3, run the UT:
https://github.com/pingcap/tiflash/blob/1e1bf1f1fd315edf88552b1092b606ee8686d58c/dbms/src/Storages/DeltaMerge/tests/gtest_dm_delta_merge_store.cpp#L1417

### 2. What did you expect to see? (Required)
Pass all the cases.
### 3. What did you see instead (Required)
Exception when PageStorage V3 restore.
[...]

PR #4845, opened 2022-05-09
Revert "Fix getMaxId won't get the right max id before GC"

jiaqizho:
### What problem does this PR solve?

Issue Number: ref #3594 

Problem Summary: If `getMaxId` ignore the deleted entry, we will reuse the PageId after a restart, which break the assumption that we will never reuse a PageID after deleted.

### What is changed and how it works?

Reverts pingcap/tiflash#4838

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #4889, opened 2022-05-13
Fix the bug for restoring data with duplicated page id

JaySon-Huang:

Issue Number: ref https://github.com/pingcap/tiflash/issues/3594

Problem Summary:
https://github.com/pingcap/tiflash/issues/4872
https://github.com/pingcap/tiflash/issues/4849
We fix those two bugs with previous PRs, but those changes introduce new misbehavior for TiFlash. Find a better way to fix those bugs.

Fix reusing same page_id for different kind by:
* Throw exception with error code == PS_DIR_APPLY_INVALID_STATUS when `VersionedPageEntries:: createNewEntry/createNewExternal/createNewRef/createNewExternal/createNewRef` run into invali state
* When restoring edit run into exception with error code == PS_DIR_APPLY_INVALID_STATUS, then call GC on mvcc_directory  entries and run apply edit again

Other changes:
* When restoring the storage pool for DeltaTree, get the max id for this ns_id by `PageStorage->getMaxId`
* `PageStorage->getMaxId` now ignore the invisible id
* Instead of `meta_max_id`, use `PageEntry.isValid()` to check whether page 1 is existed or not https://github.com/pingcap/tiflash/blob/3e46ac2a411560503abba977a37a5caa495e9bd2/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#L246-L247
* Enable some tests that are disabled in the previous changes

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/Context.cpp
+++ b/dbms/src/Interpreters/Context.cpp
@@ -1650,8 +1650,10 @@ bool Context::initializeGlobalStoragePoolIfNeed(const PathPool & path_pool)
     auto lock = getLock();
     if (shared->global_storage_pool)
     {
-        // GlobalStoragePool may be initialized many times in some test cases for restore.
-        LOG_WARNING(shared->log, "GlobalStoragePool has already been initialized.");
+        // Can't init GlobalStoragePool twice.
+        // Because we won't remove the gc task in BackGroundPool
+        // Also won't remove it from ~GlobalStoragePool()
+        throw Exception("GlobalStoragePool has already been initialized.", ErrorCodes::LOGICAL_ERROR);
     }
     CurrentMetrics::set(CurrentMetrics::GlobalStorageRunMode, static_cast<UInt8>(shared->storage_run_mode));
     if (shared->storage_run_mode == PageStorageRunMode::MIX_MODE || shared->storage_run_mode == PageStorageRunMode::ONLY_V3)
--- a/dbms/src/Interpreters/Context.cpp
+++ b/dbms/src/Interpreters/Context.cpp
@@ -1651,8 +1651,7 @@ bool Context::initializeGlobalStoragePoolIfNeed(const PathPool & path_pool)
     if (shared->global_storage_pool)
     {
         // Can't init GlobalStoragePool twice.
-        // Because we won't remove the gc task in BackGroundPool
-        // Also won't remove it from ~GlobalStoragePool()
+        // otherwise the pagestorage instances in `StoragePool` for each table won't be updated and cause unexpected problem.
         throw Exception("GlobalStoragePool has already been initialized.", ErrorCodes::LOGICAL_ERROR);
     }
     CurrentMetrics::set(CurrentMetrics::GlobalStorageRunMode, static_cast<UInt8>(shared->storage_run_mode));
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -243,7 +243,8 @@ DeltaMergeStore::DeltaMergeStore(Context & db_context,
     try
     {
         page_storage_run_mode = storage_pool->restore(); // restore from disk
-        if (!storage_pool->maxMetaPageId())
+        if (const auto entry1 = storage_pool->metaReader()->getPageEntry(1);
+            !entry1.isValid())
         {
             // Create the first segment.
             auto segment_id = storage_pool->newMetaPageId();
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -243,8 +243,8 @@ DeltaMergeStore::DeltaMergeStore(Context & db_context,
     try
     {
         page_storage_run_mode = storage_pool->restore(); // restore from disk
-        if (const auto entry1 = storage_pool->metaReader()->getPageEntry(1);
-            !entry1.isValid())
+        if (const auto first_segment_entry = storage_pool->metaReader()->getPageEntry(DELTA_MERGE_FIRST_SEGMENT_ID);
+            !first_segment_entry.isValid())
         {
             // Create the first segment.
             auto segment_id = storage_pool->newMetaPageId();
--- a/dbms/src/Storages/DeltaMerge/StoragePool.cpp
+++ b/dbms/src/Storages/DeltaMerge/StoragePool.cpp
@@ -82,26 +82,21 @@ PageStorage::Config extractConfig(const Settings & settings, StorageType subtype
 }
 
 GlobalStoragePool::GlobalStoragePool(const PathPool & path_pool, Context & global_ctx, const Settings & settings)
-    : // The iops and bandwidth in log_storage are relatively high, use multi-disks if possible
-    log_storage(PageStorage::create("__global__.log",
-                                    path_pool.getPSDiskDelegatorGlobalMulti("log"),
-                                    extractConfig(settings, StorageType::Log),
-                                    global_ctx.getFileProvider(),
-                                    true))
-    ,
-    // The iops in data_storage is low, only use the first disk for storing data
-    data_storage(PageStorage::create("__global__.data",
-                                     path_pool.getPSDiskDelegatorGlobalSingle("data"),
-                                     extractConfig(settings, StorageType::Data),
-                                     global_ctx.getFileProvider(),
-                                     true))
-    ,
-    // The iops in meta_storage is relatively high, use multi-disks if possible
-    meta_storage(PageStorage::create("__global__.meta",
-                                     path_pool.getPSDiskDelegatorGlobalMulti("meta"),
-                                     extractConfig(settings, StorageType::Meta),
-                                     global_ctx.getFileProvider(),
-                                     true))
+    : log_storage(PageStorage::create("__global__.log",
+                                      path_pool.getPSDiskDelegatorGlobalMulti("log"),
+                                      extractConfig(settings, StorageType::Log),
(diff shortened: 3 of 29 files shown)


dbms/src/Storages/DeltaMerge/tests/gtest_dm_storage_delta_merge.cpp
```diff
@@ -40,6 +40,7 @@
 
 #include <limits>
 
+#include "Storages/Transaction/TiDB.h"
```
flowbehappy, line 43:
```suggestion
#include <Storages/Transaction/TiDB.h>
```

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -101,80 +107,94 @@ PageDirectoryPtr PageDirectoryFactory::createFromEdit(String storage_name, FileP
...
         max_applied_page_id = std::max(r.page_id, max_applied_page_id);
 
-        auto [iter, created] = mvcc_table_directory.insert(std::make_pair(r.page_id, nullptr));
-        if (created)
+        // We can not avoid page id from being reused under some corner situation. Try to do gcInMemEntries
+        // and apply again to resolve the error.
+        if (bool ok = applyRecord(dir, r, /*throw_on_error*/ false); unlikely(!ok))
         {
-            iter->second = std::make_shared<VersionedPageEntries>();
+            dir->gcInMemEntries();
```
hehechen, line 120: Will it cause restore performance degradation if there are many edit entries?

tests/delta-merge-test/raft/schema/partition_table_restart.test
```diff
@@ -35,8 +35,6 @@
 # schema syncer guarantees logical table creation at last, so there won't be cases that logical table exists whereas physical table not.
 => drop table default.test
 => DBGInvoke __reset_schemas()
-# remove obsolete entry left by previous dropped table
-=> DBGInvoke __gc_global_storage_pool()
```
JaySon-Huang, line 39: Seems we still need this workaround, will investigate why

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -904,10 +939,31 @@ PageId PageDirectory::getMaxId(NamespaceId ns_id) const
...
+            }
+
+            // Current entry/ref/external is deleted and there are no entries before it.
+            if (iter == mvcc_table_directory.begin())
+            {
+                break;
+            }
+
+            iter--;
+        } while (true);
```
JaySon-Huang, line 964: These changes revert https://github.com/pingcap/tiflash/pull/4845 but use `isVisible` to check whether we can return the id in getMaxId /cc @jiaqizho @lidezhu

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -243,7 +243,8 @@ DeltaMergeStore::DeltaMergeStore(Context & db_context,
     try
     {
         page_storage_run_mode = storage_pool->restore(); // restore from disk
-        if (!storage_pool->maxMetaPageId())
+        if (const auto entry1 = storage_pool->metaReader()->getPageEntry(1);
```
flowbehappy, line 246: Let's give `1` a name.

lidezhu: It's already has a name `DELTA_MERGE_FIRST_SEGMENT_ID`.

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -375,6 +380,41 @@ std::optional<PageEntryV3> VersionedPageEntries::getLastEntry() const
     return std::nullopt;
 }
 
+// Returns true when **this id** is visible (not exist or already marked as deleted) by `seq`.
```
flowbehappy, line 383: "not exist or already marked as deleted" seems odd.

dbms/src/Interpreters/Context.cpp
```diff
@@ -1650,8 +1650,10 @@ bool Context::initializeGlobalStoragePoolIfNeed(const PathPool & path_pool)
     auto lock = getLock();
     if (shared->global_storage_pool)
     {
-        // GlobalStoragePool may be initialized many times in some test cases for restore.
-        LOG_WARNING(shared->log, "GlobalStoragePool has already been initialized.");
+        // Can't init GlobalStoragePool twice.
+        // Because we won't remove the gc task in BackGroundPool
```
jiaqizho, line 1654: Change back this comment pls. Also need change back to warning.

dbms/src/Storages/DeltaMerge/StoragePool.cpp
```diff
@@ -375,18 +367,18 @@ PageStorageRunMode StoragePool::restore()
             data_storage_writer = std::make_shared<PageWriter>(run_mode, /*storage_v2_*/ nullptr, data_storage_v3);
             meta_storage_writer = std::make_shared<PageWriter>(run_mode, /*storage_v2_*/ nullptr, meta_storage_v3);
 
-            max_log_page_id = global_storage_pool->getLogMaxId(ns_id);
-            max_data_page_id = global_storage_pool->getDataMaxId(ns_id);
-            max_meta_page_id = global_storage_pool->getMetaMaxId(ns_id);
+            max_log_page_id = global_storage_pool->log_storage->getMaxId(ns_id);
```
jiaqizho, line 370: Use `log_pagestorage_v3` to replace `global_storage_pool->log_storage` pls.
also dito meta/data and below.

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
jiaqizho, line 120: If there is too much multiplexing, will gcInMemEntries be triggered too many times ?
Maybe we can add an interface name `gcInMemEntries(page_id)`, or add a TODO here ?

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -101,80 +107,94 @@ PageDirectoryPtr PageDirectoryFactory::createFromEdit(String storage_name, FileP
...
+            applyRecord(dir, r, /*throw_on_error*/ true);
+            LOG_FMT_INFO(DB::Logger::get("PageDirectoryFactory"), "resolve from error status done, continue to restore");
```
jiaqizho, line 122: `DB::Logger::get("PageDirectoryFactory")` add a log in PageDirectoryFactory to replace it? cause it used many place

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -101,80 +107,94 @@ PageDirectoryPtr PageDirectoryFactory::createFromEdit(String storage_name, FileP
...
+    }
+    catch (DB::Exception & e)
+    {
+        e.addMessage(fmt::format(" [type={}] [page_id={}] [ver={}]", r.type, r.page_id, restored_version));
+        if (throw_on_error || e.code() != ErrorCodes::PS_DIR_APPLY_INVALID_STATUS)
         {
-            e.addMessage(fmt::format(" [type={}] [page_id={}] [ver={}]", r.type, r.page_id, restored_version));
             throw e;
         }
+        LOG_FMT_INFO(DB::Logger::get("PageDirectoryFactory"), "try to resolve error during restore: {}", e.message());
```
jiaqizho, line 194: Can we change this log into warning?

dbms/src/Interpreters/Context.cpp
JaySon-Huang: We do NOT re-init GlobalStoragePool even in test cases now.

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
JaySon-Huang:
```
PUT_EXT 100
REF 150 -> 100
DEL 100
REF 100 -> 40  <- exception with PS_DIR_APPLY_INVALID_STATUS thrown here
```

We can not only run GC on id==100 to get things right.

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
JaySon-Huang: Now we only create it when *failover* does happen.
Just a small optimization to reduce the size of global `_pLoggerMap` in Poco. If it does not occur, we won't call `Poco::Logger::get` and we won't add a new element to the global `_pLoggerMap`. 

https://github.com/pingcap/poco/blob/e411ea34492bf26e76b4e4a03ff5813a0c779240/Foundation/src/Logger.cpp#L475-L480

dbms/src/Storages/DeltaMerge/StoragePool.cpp
JaySon-Huang: Fixed, please check again

lidezhu (approved):
LGTM

dbms/src/Interpreters/Context.cpp
jiaqizho: okay, but the Notes, it's not right. need updated

jiaqizho (approved):
LGTM, But still need to change the notes(that reason is not right). https://github.com/pingcap/tiflash/blob/f2c47c9a98a40b2c832219f100040e418a7f1e74/dbms/src/Interpreters/Context.cpp#L1654-L1655

PR #4890, opened 2022-05-13
Add option to support clang LTO(Link Time Optimization) and time-traces

solotzg:

Issue Number: ref #4909

Problem Summary:

For example:

https://github.com/pingcap/tiflash/blob/e03dcc1e38e14906e7598c7e8775a2660e9ab58a/dbms/src/Storages/Transaction/TMTContext.cpp#L74-L82

Use `objdump -d ${original-tiflash-binary} | grep 'getKVStore'`, you can see at least 63 records like:

```
...
 7b18a17:       e8 34 27 e5 ff          callq  796b150 <_ZN2DB10TMTContext10getKVStoreEv>
 7b82ba7:       e8 b4 85 de ff          callq  796b160 <_ZNK2DB10TMTContext10getKVStoreEv>
...
```

Other modules only can invoke `TMTContext::getKVStore` by calling method through instruction `callq`.
Because other object files can not optimize invoking `TMTContext::getKVStore` unless implementations of `TMTContext::getKVStore` is declared in header file(`TMTContext.h`).

With `LTO`, `objdump -d ${new-tiflash-binary} | grep 'getKVStore'` will get nothing. Because linker can analyze the whole context and make it inline automatically. It's friendly for [forward declaration](https://en.wikipedia.org/wiki/Forward_declaration).

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


SchrodingerZhu:
For LTO, consider adding `-fvisibility=hidden`, `-fvisibility-inlines-hidden`, and enabling devirtualization flags if possible

SchrodingerZhu (approved):

solotzg:
> For LTO, consider adding `-fvisibility=hidden`, `-fvisibility-inlines-hidden`, and enabling devirtualization flags if possible

Add flags `-fvisibility=hidden -fvisibility-inlines-hidden -fwhole-program-vtables -fsplit-lto-unit `

CMakeLists.txt
```diff
@@ -102,6 +102,28 @@ if (COMPILER_CLANG)
     # Clang doesn't have int128 predefined macros, workaround by manually defining them
     # Reference: https://stackoverflow.com/questions/41198673/uint128-t-not-working-with-clang-and-libstdc
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__GLIBCXX_BITSIZE_INT_N_0=128 -D__GLIBCXX_TYPE_INT_N_0=__int128")
+
+    option(ENABLE_TIME_TRACES "Enable clang feature time traces" OFF)
```
zanmato1984, line 106: Are these options Clang-specific? If so, we should check if the compiler is Clang then set. Otherwise GCC will be broke - at least GCC 7.x is in the support list so far.

solotzg: Those options are clang specific. They are under `if (COMPILER_CLANG)`.

zanmato1984 (approved):
LGTM

PR #4955, opened 2022-05-22
flush cache before segment merge

### LINKED ISSUE #4956: Count result may be wrong after a lot of insert and delete
## Bug Report

Repeated do a lot of insert rows and delete alls.
Keep select count after after every operation.

Count result is right.

Count result is wrong.

lidezhu:

Issue Number: close #4956 

Problem Summary:
When do segment split, we try to copy the tail column files in the delta layer of the original segment to the new result
segments. So the new segments may contain data that doesn't belong to its segment range.

And this is ok for most cases, because the redundant data will be filtered out by the segment range when serve the read requests to the segment. So the redundant is invisible in almost all cases.

But when do segment merge later, if the previous redundant data is still not flushed to disk, it will be directly copied to the new merged segment again.

So the redundant data in each segment become visible again after segment merge which may cause potential data incorrectness.

Flush cache before every merge operation. So the potential unsaved data will be filtered out by the segment range when do merge.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix potential wrong result after a lot of insert and delete operations
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -1977,11 +1977,21 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
     /// So we flush here to make sure that all potential data left by previous split operation is saved.
     while (!left->flushCache(dm_context))
     {
-        // keep flush until success
+        // keep flush until success if not abandoned
+        if (left->hasAbandoned())
+        {
+            LOG_FMT_DEBUG(log, "Give up merge segments left [{}], right [{}]", left->segmentId(), right->segmentId());
+            return;
+        }
     }
     while (!right->flushCache(dm_context))
     {
-        // keep flush until success
+        // keep flush until success if not abandoned
+        if (right->hasAbandoned())
+        {
+            LOG_FMT_DEBUG(log, "Give up merge segments left [{}], right [{}]", left->segmentId(), right->segmentId());
+            return;
+        }
     }
 
     SegmentSnapshotPtr left_snap;
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -1975,8 +1975,14 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
     /// And only saved data in this segment will be filtered by the segment range in the merge process,
     /// unsaved data will be directly copied to the new segment.
     /// So we flush here to make sure that all potential data left by previous split operation is saved.
-    left->flushCache(dm_context);
-    right->flushCache(dm_context);
+    while (!left->flushCache(dm_context))
+    {
+        // keep flush until success
+    }
+    while (!right->flushCache(dm_context))
+    {
+        // keep flush until success
+    }
 
     SegmentSnapshotPtr left_snap;
     SegmentSnapshotPtr right_snap;
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -1971,6 +1971,13 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
         right->info(),
         dm_context.min_version);
 
+    /// This segment may contain some rows that not belong to this segment range which is left by previous split operation.
+    /// And only saved data in this segment will be filtered by the segment range in the merge process,
+    /// unsaved data will be directly copied to the new segment.
+    /// So we flush here to make sure that all potential data left by previous split operation is saved.
+    left->flushCache(dm_context);
+    right->flushCache(dm_context);
+
     SegmentSnapshotPtr left_snap;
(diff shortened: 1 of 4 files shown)


dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1971,6 +1971,13 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
         right->info(),
         dm_context.min_version);
 
+    /// This segment may contain some rows that not belong to this segment range which is left by previous split operation.
+    /// And only saved data in this segment will be filtered by the segment range in the merge process,
+    /// unsaved data will be directly copied to the new segment.
+    /// So we flush here to make sure that all potential data left by previous split operation is saved.
+    left->flushCache(dm_context);
+    right->flushCache(dm_context);
```
flowbehappy, line 1979: We need to check the result of flushCache, as it can be false.

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -1149,6 +1158,10 @@ SegmentPtr Segment::merge(DMContext & dm_context, const ColumnDefinesPtr & schem
     return merged;
 }
 
+/// Segments may contain some rows that not belong to its range which is left by previous split operation.
```
flowbehappy, line 1161: Do you think split needs flushCache?

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
lidezhu: Fixed.

dbms/src/Storages/DeltaMerge/Segment.cpp
lidezhu: We only need to flush segments either after split or before merge. And considering segment split mostly means it's a hot segment and segment merge means it's a cold segment. So flush before merge is a better choice which should do less flush work in most cases.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1971,6 +1971,19 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
...
+    /// unsaved data will be directly copied to the new segment.
+    /// So we flush here to make sure that all potential data left by previous split operation is saved.
+    while (!left->flushCache(dm_context))
+    {
+        // keep flush until success
+    }
+    while (!right->flushCache(dm_context))
+    {
+        // keep flush until success
+    }
```
JaySon-Huang, line 1985: If the `left`/`right` segment gets abandoned by another thread, will it run into an inf loop? Should we check `left->hasAbandoned`/`right->hasAbandoned` in the while loop?

https://github.com/pingcap/tiflash/blob/2e8d324e76a2c5a96c83fdab8e7e08f3a5424d61/dbms/src/Storages/DeltaMerge/Delta/DeltaValueSpace.cpp#L142-L207

dbms/src/Storages/DeltaMerge/tests/gtest_dm_segment.cpp
```diff
@@ -1009,9 +1015,13 @@ try
         ASSERT_EQ(num_rows_seg1 + num_rows_seg2, num_rows_write);
     }
 
+    // delete rows in the right segment
+    {
+        new_segment->write(dmContext(), new_segment->getRowKeyRange());
```
JaySon-Huang, line 1020:
```suggestion
        new_segment->write(dmContext(), /*delete_range*/ new_segment->getRowKeyRange());
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
flowbehappy: +1. Beet to check validation of segment instant. If is invalid then simply give up.

lidezhu: Fixed.

JaySon-Huang (approved):
LGTM

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -316,7 +316,10 @@ bool Segment::write(DMContext & dm_context, const Block & block)
 
     if (delta->appendColumnFile(dm_context, column_file))
     {
-        flushCache(dm_context);
+        if (flush_cache)
+        {
+            flushCache(dm_context);
```
flowbehappy, line 321: Does it need to check the flushCache result?

lidezhu: This is only used in tests. So the flushCache cannot fail here. And even the flushCache fail can help us cover more scenario.

flowbehappy: Hmmm, "even the flushCache fail can help us cover more scenario" sorry I don't get the sense here. Can you explain more?
Think that even if methods are only used for tests, we also need to guarantee correctness.

lidezhu: Got it and have fixed it.

flowbehappy (approved):

Issue #4957, opened 2022-05-23
Failed to restart with error `Restore postion from BlobStat failed, [offset=166292405] [buf_size=115365] [blob_id=19] is used or subspan is used`


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Run a TiDB cluster with TiFlash node of 16GiB memory. Run chbenchmark warehouse 1000 on that cluster. chbenchmark Q9 cause OOM frequently

### 2. What did you expect to see? (Required)
TiFlash run normally after restart caused by OOM

### 3. What did you see instead (Required)
After frequently restart caused by OOM run for over 15 hours, TiFlash fail to restart with error message like:
[fail_restart.log.zip](https://github.com/pingcap/tiflash/files/8751222/fail_restart.log.zip)

```
[2022/05/23 01:45:53.738 +08:00] [INFO] [BaseDaemon.cpp:315] ["BaseDaemon:Stop SignalListener thread"] [thread_id=2]
[2022/05/23 01:46:28.175 +08:00] [INFO] [<unknown>] ["Welcome to TiFlash"] [thread_id=1]
[2022/05/23 01:46:28.175 +08:00] [INFO] [<unknown>] ["Starting daemon with revision 54381"] [thread_id=1]
[2022/05/23 01:46:28.175 +08:00] [INFO] [<unknown>] ["TiFlash build info: TiFlash\nRelease Version: v6.1.0\nEdition:         Community\nGit Commit Hash: 71613fd8c059187a89d7858d3d723fea701869fc\nGit Branch:      heads/refs/tags/v6.1.0\nUTC Build Time:  2022-05-20 06:38:20\nEnable Features: jemalloc asimd unwind\nProfile:         RELWITHDEBINFO\n"] [thread_id=1]
[2022/05/23 01:46:28.175 +08:00] [INFO] [<unknown>] ["Application:starting up"] [thread_id=1]
[2022/05/23 01:46:28.175 +08:00] [INFO] [Server.cpp:363] ["Application:Got jemalloc version: 5.2.1-0-gea6b3e973b477b8061e0076bb257dbd7f3faa756"] [thread_id=1]
[2022/05/23 01:46:28.176 +08:00] [INFO] [Server.cpp:372] ["Application:Not found environment variable MALLOC_CONF"] [thread_id=1]
[...]

PR #4972, opened 2022-05-23
PageDirectory: need do gc twice in restore.

jiaqizho:

Issue Number: ref #4957 

Problem Summary:
- MVCC may not clean up deleted `entry` if that `entry` be refence by a deleted `ref`.
- For example: [p1, ref 2 -> 1, d 1, d 2]
- First time call `gcInMemEntries()`:
    - P1 entry 0 being_ref_count is not zero, won't be cleaned.
    - R2 entry 0 being_ref_count is zero, deref the P 1, make it being_ref_count is 0. Also R2 will be clean up.
- Second time call `gcInMemEntries()`:
    - P 1 being_ref_count is 0, it will be clean.
- If we can't clean up all of the deleted entries in `restore`. Then BlobStore may not restore success.

- call `gcInMemEntries()` twice to clean up all of deleted entries.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -41,6 +41,22 @@ PageDirectoryPtr PageDirectoryFactory::create(String storage_name, FileProviderP
...
+    //  P 1
+    //  R 2 -> 1
+    //  D 1
+    //  D 2
+    // First time call gcInMemEntries():
+    //  - P 1 entry 0 being_ref_count is not zero, won't be cleaned.
+    //  - R 2 entry 0 being_ref_count is zero, deref the P 1, make it being_ref_count is 0. Also R2 will be clean up.
+    //
+    // Second time call gcInMemEntries():
+    //  - P 1 being_ref_count is 0, it will be clean.
```
JaySon-Huang, line 57: Can you add a test case for it?

JaySon-Huang, line 57: Can you add a test case for it?

jiaqizho: done

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -42,6 +42,22 @@ PageDirectoryPtr PageDirectoryFactory::create(String storage_name, FileProviderP
...
+    dir->gcInMemEntries();
```
JaySon-Huang, line 59: Also add it to `PageDirectoryFactory::createFromEdit`

jiaqizho: done

JaySon-Huang: I mean in these lines, we should also call gcInMemEntries twice
https://github.com/pingcap/tiflash/blob/62e7c14141c0a3390ac5a4454708126511382838/dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp#L87-L100

JaySon-Huang:
Do not need this change because we have cleaned entries after decreasing the ref count in `VersionedPageEntries::derefAndClean`. 
https://github.com/pingcap/tiflash/blob/64a747a179d736475a66e02d926efbab5a133c32/dbms/src/Storages/Page/V3/PageDirectory.cpp#L599-L603

The test cases will be merged into https://github.com/pingcap/tiflash/pull/4987

Issue #4909, opened 2022-05-17
Add option to support clang LTO(Link Time Optimization)


solotzg:
## Enhancement

- [ ] Support LTO(Link Time Optimization) #4890

PR #4979, opened 2022-05-23
*: remove externs for failpoints

breezewish:

Issue Number: ref #4909

Problem Summary:

Use inline variables, so that we don't need to write `extern const` everywhere we want to use failpoints.

Note, there are downsides: When adding new failpoints, every translation unit that references one of the failpoint needs to be re-compiled (which is around 100 files, mostly TiFlash related files). But actually this is quite common when developing new features now.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


fuzhe1989 (approved):

JaySon-Huang:
But I do want to avoid almost all files being recompiled when I just add a failpoint in one file :(

lidezhu:
> But I do want to avoid almost all files being recompiled when I just add a failpoint in one file :(

+1

breezewish:
> But I do want to avoid almost all files being recompiled when I just add a failpoint in one file :(

Glad to know that. I will close this PR and try to discover whether we can have both benefits.

Issue #4922, opened 2022-05-18
Show the file usage of PageStorage V3 in Grafana


JaySon-Huang:
## Enhancement
Show the total file size on disk and valid data size for all PageStorage V3 instances for each TiFlash node

PR #4999, opened 2022-05-25
PageStorage: Report the file usage of BlobStore

JaySon-Huang:
Signed-off-by: JaySon-Huang <tshent@qq.com>

Issue Number: close https://github.com/pingcap/tiflash/issues/4922

Problem Summary: Report the file size, valid data size, file num of all PageStorage instances in a TiFlash node

- Report the file size, valid data size, file num of all PageStorage instances in `AsynchronousMetrics::update`
- Show the file size, valid data size and file num in Grafana
- Move "PageStorage Blob Status" into "StoragePool"

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code
  - Deploy and run sysbench oltp_update_index workload
  - http://172.16.5.81:21510/d/5xovuGr7k/verify-tiflash-summary?orgId=1&from=1653629400000&to=1653632700000
  - image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/AsynchronousMetrics.cpp
+++ b/dbms/src/Interpreters/AsynchronousMetrics.cpp
@@ -21,8 +21,12 @@
 #include <IO/UncompressedCache.h>
 #include <Interpreters/AsynchronousMetrics.h>
 #include <Storages/DeltaMerge/DeltaMergeStore.h>
+#include <Storages/DeltaMerge/StoragePool.h>
 #include <Storages/MarkCache.h>
+#include <Storages/Page/FileUsage.h>
 #include <Storages/StorageDeltaMerge.h>
+#include <Storages/Transaction/KVStore.h>
+#include <Storages/Transaction/TMTContext.h>
 #include <common/config_common.h>
 
 #include <chrono>
@@ -147,6 +151,7 @@ void AsynchronousMetrics::update()
     set("Uptime", context.getUptimeSeconds());
 
     {
+        // Get the snapshot status from all delta tree tables
         auto databases = context.getDatabases();
 
         double max_dt_stable_oldest_snapshot_lifetime = 0.0;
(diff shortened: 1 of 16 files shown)


dbms/src/Interpreters/AsynchronousMetrics.cpp
```diff
@@ -177,6 +182,32 @@ void AsynchronousMetrics::update()
         set("MaxDTBackgroundTasksLength", max_dt_background_tasks_length);
     }
 
+    do
```
jiaqizho, line 185: can we add a method to replace the sub logical?

flowbehappy (approved):

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -115,6 +116,38 @@ void BlobStore::registerPaths()
...
+            // We can access to these type without any locking.
+            if (stat->isReadOnly() || stat->isBigBlob())
+            {
+                usage.total_disk_size += stat->sm_total_size;
+                usage.total_valid_size += stat->sm_valid_size;
+            }
+            else
+            {
+                // Else the stat may being updated, acquire a lock to avoid data race.
+                auto lock = stat->lock();
```
jiaqizho, line 140: Can this lock be removed? we mat not need an exact value.

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -115,6 +116,38 @@ void BlobStore::registerPaths()
     }
 }
 
+FileUsageStatistics BlobStore::getFileUsageStatistics() const
+{
+    FileUsageStatistics usage;
+
+    // Get a copy of stats map to avoid the big lock on stats map
+    const auto stats_list = blob_stats.getStats();
```
jiaqizho, line 124: `const auto & stats_list`?

JaySon-Huang: * The return type of `getStats` itself is not reference, adding `&` here can not reduce copying
* And we do wang to copy the whole `stats` to avoid locking

dbms/src/Storages/Page/V3/BlobStore.cpp
JaySon-Huang: 1. Running under TSan may complain about that
2. We want a consistent value that total_size > valid size. Without lock, we may meet total_size < valid size, which is confusing
3. The frequency of calling getFileUsageStatistics is very low, call it only once every 30s 

So I would rather keep the lock here

dbms/src/Interpreters/AsynchronousMetrics.cpp
JaySon-Huang: done

dbms/src/Storages/Page/V3/BlobStore.cpp
jiaqizho: okay

jiaqizho (approved):
LGTM

PR #4861, opened 2022-05-11
Refine function test framework

windtalker:
### What problem does this PR solve?

Issue Number: close #4830 

Problem Summary:

We have a function test framework used by unit test of functions, however, the current implementation of the test framework use the following code to test the function:
```
    auto & factory = FunctionFactory::instance();

    Block block(columns);
    ColumnNumbers cns;
    for (size_t i = 0; i < columns.size(); ++i)
        cns.push_back(i);

    auto bp = factory.tryGet(func_name, context);
    if (!bp)
        throw TiFlashTestException(fmt::format("Function {} not found!", func_name));
    auto func = bp->build(columns, collator);
    block.insert({nullptr, func->getReturnType(), "res"});
    func->execute(block, cns, columns.size());
    return block.getByPosition(columns.size());
```
This is not the same in the real world when we executing functions, in the real world, we first build the function:
```
 const String & func_name = getFunctionName(expr);
 if (DAGExpressionAnalyzerHelper::function_builder_map.count(func_name) != 0)
 {
     ret = DAGExpressionAnalyzerHelper::function_builder_map[func_name](this, expr, actions);
 }
 else
 {
     ret = buildFunction(expr, actions);
 }
```
Then execute the function later. When building the function, we may rewrite functions, take `ifNull(col1, col2)` for example, we rewrite it to
`multiIf(isNull(col1), col2, assumeNotNull(col1))`

Current function test framework does not take this rewrite into account, it will either testing `ifNull(col1, col2)` by executing `ifNull(col1, col2)` directly, or rewrite it manually. Both solutions can't guarantee the testing logic is the same as the real world executing logic, which makes function unit test meaningless for functions that will be rewritten during `buildFunction`.(#4829 is an example that we already add ut for `ifNull`, but not catch the real world bug) 
### What is changed and how it works?
- add a new argument `raw_function_test` in `FunctionTest::executeFunction`, and if `raw_function_test == true`, will use the original way to test function, otherwise, will convert columns into TiPB expr first, then test the function after `DAGExpressionAnalyzerHelper::buildFunction`, `raw_function_test` is true by default

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
[...]

Issue #4830, opened 2022-05-06
Function unit test should follow the same logic in DAGExpressionAnalyzer when building function


windtalker:
## Enhancement
We already have a [function test framework](https://github.com/pingcap/tiflash/blob/e10c6eded7de50fee020be7b883b4e56fe5990a3/dbms/src/TestUtils/FunctionTestUtils.h#L653) for function unit test, however, it just use [executeFunction](https://github.com/pingcap/tiflash/blob/e10c6eded7de50fee020be7b883b4e56fe5990a3/dbms/src/TestUtils/FunctionTestUtils.h#L534) to build and execute the function, while when we build a function in `DAGExpressionAnalyzer`, we may have rewrite so the orignal function will be rewrite to some other functions, for exmaple:
`ifnull(col1, col2)` 
is rewriten to
`multiif(isnull(col1), col2, assumeNotNull(col1))`
in [buildIfNullFunction](https://github.com/pingcap/tiflash/blob/e10c6eded7de50fee020be7b883b4e56fe5990a3/dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp#L99)

We need to enhance our function test framework to ensure that function unit test follow the same logic in DAGExpressionAnalyzer when building function

PR #5029, opened 2022-05-30
enable new function test framework by default

windtalker:

Issue Number: ref #4830

Problem Summary:

Currently, function unit test will create function directly but DAGExpressionAnalyzer, it may add some rewrite when compiling function, #4861 add a new function unit test framework, which will convert columns to tipb expression first, and then use `DAGExpressionAnalyzer` to build the function. This pr enable this new function test framework by default.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Functions/tests/gtest_unix_timestamp.cpp
```diff
@@ -86,17 +86,17 @@ try
...
-        executeFunction(func_name_int, createConstColumn<Nullable<MyDateTime>>(std::make_tuple(0), 10, date_time_data[0])));
+        createConstColumn<UInt64>(10, date_time_int_result[0]),
+        executeFunction(func_name_int, createConstColumn<MyDateTime>(std::make_tuple(0), 10, date_time_data[0])));
     ASSERT_COLUMN_EQ(
-        createConstColumn<Nullable<Decimal64>>(std::make_tuple(12, 0), 10, date_time_decimal_result[0]),
-        executeFunction(func_name_dec, createConstColumn<Nullable<MyDateTime>>(std::make_tuple(0), 10, date_time_data[0])));
+        createConstColumn<Decimal64>(std::make_tuple(12, 0), 10, date_time_decimal_result[0]),
+        executeFunction(func_name_dec, createConstColumn<MyDateTime>(std::make_tuple(0), 10, date_time_data[0])));
     ASSERT_COLUMN_EQ(
-        createConstColumn<Nullable<Decimal64>>(std::make_tuple(18, 6), 10, date_time_with_fsp_result[0]),
```
SeaRise, line 98: Can we convert `Const<Nullable<Type>>` to `Const<Type>` in `FunctionTestUtils`, and then we don't need to modify the unit test.

dbms/src/Functions/tests/gtest_arithmetic_functions.cpp
```diff
@@ -128,6 +128,8 @@ try
     null_or_zero_field.push_back(Field(DecimalField<Decimal128>(0, 0)));
 
     std::vector<Int64> values{10, 2, 20, 8, 10, 0, 30, 8, 16, 4};
+    Int128 decimal_128_factor = 10000000000;
+    decimal_128_factor *= 1000000000;
```
SeaRise, line 132: add some comments?

dbms/src/Functions/tests/gtest_unix_timestamp.cpp
windtalker: This is an intentional behavior, because in TiDB, it will generate a constant expr with nullable type but not null value, we need to cover this case, if we convert `Const<Nullable<Type>>(not_null_value)` to `Const<Type>(not_null_value)` in `FunctionTestUtils`, we don't have case to cover this.

dbms/src/Functions/tests/gtest_arithmetic_functions.cpp
windtalker: Done

dbms/src/Functions/tests/gtest_unix_timestamp.cpp
SeaRise: Did we cover this case in the unit test?
what happens when `createConstColumn<Nullable<Decimal64>>(std::make_tuple(18, 6), 10, date_time_with_fsp_result[0])`?

dbms/src/Functions/tests/gtest_bitand.cpp
```diff
@@ -126,51 +126,40 @@ try
...
 CATCH
 
 TEST_F(TestFunctionBitAnd, Boundary)
 try
 {
-    ASSERT_BITAND(createColumn<Int8>({127, 127, -128, -128}), createColumn<UInt8>({0, 255, 0, 255}), createColumn<Int8>({0, 127, 0, -128}));
-    ASSERT_BITAND(createColumn<Int8>({127, 127, -128, -128}), createColumn<UInt16>({0, 65535, 0, 65535}), createColumn<Int16>({0, 127, 0, -128}));
-    ASSERT_BITAND(createColumn<Int16>({32767, 32767, -32768, -32768}), createColumn<UInt8>({0, 255, 0, 255}), createColumn<Int16>({0, 255, 0, 0}));
+    ASSERT_BITAND(createColumn<Int8>({127, 127, -128, -128}), createColumn<UInt8>({0, 255, 0, 255}), createColumn<UInt64>({0, 127, 0, 128}));
+    ASSERT_BITAND(createColumn<Int8>({127, 127, -128, -128}), createColumn<UInt16>({0, 65535, 0, 65535}), createColumn<UInt64>({0, 127, 0, 65408}));
```
SeaRise, line 157: Does this meet expectations?
What does `65408` mean?

dbms/src/Functions/tests/gtest_bitand.cpp
```diff
@@ -126,51 +126,40 @@ try
...
+    ASSERT_BITAND(createColumn<Int16>({32767, 32767, -32768, -32768}), createColumn<UInt8>({0, 255, 0, 255}), createColumn<UInt64>({0, 255, 0, 0}));
 
     ASSERT_BITAND(createColumn<Int64>({0, 0, 1, 1, -1, -1, INT64_MAX, INT64_MAX, INT64_MIN, INT64_MIN}),
                   createColumn<UInt64>({0, UINT64_MAX, 0, UINT64_MAX, 0, UINT64_MAX, 0, UINT64_MAX, 0, UINT64_MAX}),
-                  createColumn<Int64>({0, 0, 0, 1, 0, -1, 0, INT64_MAX, 0, INT64_MIN}));
+                  createColumn<UInt64>({0, 0, 0, 1, 0, UINT64_MAX, 0, INT64_MAX, 0, 9223372036854775808ull}));
```
SeaRise, line 162: ditto for `9223372036854775808ull`

dbms/src/Functions/tests/gtest_datetime_daymonthyear.cpp
```diff
@@ -76,15 +76,7 @@ class TestDateTimeDayMonthYear : public DB::tests::FunctionTest
...
-                column = ColumnWithTypeAndName(
-                    createConstColumn<Nullable<typename DT::FieldType>>(1,
-                                                                        {raw_input.toPackedUInt()})
-                        .column,
-                    makeNullable(std::make_shared<DT>()),
-                    "result");
-                result_day = createConstColumn<Nullable<UInt8>>(1, {day});
-                result_month = createConstColumn<Nullable<UInt8>>(1, {month});
-                result_year = createConstColumn<Nullable<UInt16>>(1, {year});
+                return;
```
SeaRise, line 79: or throw exception?

windtalker: I add these tests back.

dbms/src/Functions/tests/gtest_unix_timestamp.cpp
windtalker: Oh, my mistake, at the beginning I just remove all the tests withe `Const<Nullable<Type>>(not_null_value)`, then I realized that I need to keep these tests, because TiDB will geneate a constant expr with nullable type but not null value. So I decide to updated these tests' result instead of just remove them. Unfortunately I forgot to update the tests here.

SeaRise (approved):

PR #5044, opened 2022-06-02
Fix potential data inconsistency under heavy ddl operation

### LINKED ISSUE #5032: Data is inconsistent between TiFlash and TiKV after changing column types
## Bug Report

Run the full-stack test: https://github.com/pingcap/tiflash/blob/335b124cab124828aadc318b7355f8390950f032/tests/fullstack-test2/ddl/alter_column_when_pk_is_handle.test, it will fail probability: https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/6648/pipeline

Pass the test.
The test fail, 
飞书20220531-000239
Data is inconsistent between TiFlash and TiKV, the log shows that TiFlash use default value instead of `0`:
image
The `col_6` does not exist in stable:
image

7dff46338e179a047127abb151d2db586a5b9cce

lidezhu:

Issue Number: close #5032 

Problem Summary: Currently we cache decoding schema for decoding raft data if a table schema doesn't change. And we judge whether a table schema has changed based on the table schema version.

But the schema version is not strictly consistent with the actual table schema which can be seen at https://github.com/pingcap/tiflash/blob/master/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L362. That is when applying different schema changes in a diff, the table schema version will be set to the latest schema version after the first schema change is applied.

More concretely, when a lossy ddl change occurs, it will trigger `drop column` and `add column` schema changes and also rewrite some data at the same time.  The schema changes will be applied one by one, and because tiflash updates the table schema version ahead of time when applying schema diff, after applying the `drop column` schema change, it will update the schema version to the latest schema version.

And if the decode thread tries to obtain the current schema for decoding data before the `add column` is applied, the current schema and the latest schema version will be cached. Then after the subsequent `add column` operation is applied, the table schema version will not be updated, so the cache of the decode thread will not be invalidated. 

Therefore, the decode thread will decode the new data with an older schema, considering that the new added column is a dropped column and discarding it.

In addition, there is a lower chance of triggering this problem in the case of frequent add column and insert data.

Add an internal schema version for `DecodingStorageSchemaSnapshot`.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix potential data inconsistency under heavy ddl operation
```

### DIFF
--- a/dbms/src/Storages/StorageDeltaMerge.h
+++ b/dbms/src/Storages/StorageDeltaMerge.h
@@ -239,9 +239,9 @@ private:
     mutable std::mutex decode_schema_mutex;
     DecodingStorageSchemaSnapshotPtr decoding_schema_snapshot;
     // The following two members must be used under the protection of table structure lock
-    bool decoding_schema_changed;
+    bool decoding_schema_changed = false;
     // internal version for `decoding_schema_snapshot`
-    size_t decoding_schema_version = 1;
+    Int64 decoding_schema_version = 1;
 
     // avoid creating block every time when decoding row
     std::vector<BlockUPtr> cache_blocks;
--- a/dbms/src/Storages/StorageDeltaMerge.h
+++ b/dbms/src/Storages/StorageDeltaMerge.h
@@ -151,9 +151,9 @@ public:
 
     size_t getRowKeyColumnSize() const override { return rowkey_column_size; }
 
-    std::pair<DB::DecodingStorageSchemaSnapshotConstPtr, BlockUPtr> getSchemaSnapshotAndBlockForDecoding(bool /* need_block */) override;
+    std::pair<DB::DecodingStorageSchemaSnapshotConstPtr, BlockUPtr> getSchemaSnapshotAndBlockForDecoding(const TableStructureLockHolder & table_structure_lock, bool /* need_block */) override;
 
-    void releaseDecodingBlock(Int64 schema_version, BlockUPtr block) override;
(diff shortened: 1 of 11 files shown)


dbms/src/Storages/StorageDeltaMerge.h
```diff
@@ -238,6 +238,11 @@ class StorageDeltaMerge
 
     mutable std::mutex decode_schema_mutex;
     DecodingStorageSchemaSnapshotPtr decoding_schema_snapshot;
+    // The following two members must be used under the protection of table structure lock
+    bool decoding_schema_changed = false;
+    // internal version for `decoding_schema_snapshot`
+    size_t decoding_schema_version = 1;
```
JaySon-Huang, line 244: We use `Int64` as the param type in `void releaseDecodingBlock(Int64 block_decoding_schema_version, ...`, should also use Int64 here?

lidezhu: Fixed.

JaySon-Huang (approved):
LGTM

flowbehappy (approved):

lidezhu:
> In response to a cherrypick label: failed to apply #5044 on top of branch "release-6.1":
> 
> ```
> failed to git commit: exit status 1
> ```

release-6.1 has been manually cherry-picked in this pr: https://github.com/pingcap/tiflash/pull/5046

PR #5046, opened 2022-06-02
Fix potential data inconsistency under heavy ddl operation

lidezhu:

Issue Number: ref #5032

Problem Summary: Currently we cache decoding schema for decoding raft data if a table schema doesn't change. And we judge whether a table schema has changed based on the table schema version.

But the schema version is not strictly consistent with the actual table schema which can be seen at https://github.com/pingcap/tiflash/blob/master/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L362. That is when applying different schema changes in a diff, the table schema version will be set to the latest schema version after the first schema change is applied.

More concretely, when a lossy ddl change occurs, it will trigger `drop column` and `add column` schema changes and also rewrite some data at the same time.  The schema changes will be applied one by one, and because tiflash updates the table schema version ahead of time when applying schema diff, after applying the `drop column` schema change, it will update the schema version to the latest schema version.

And if the decode thread tries to obtain the current schema for decoding data before the `add column` is applied, the current schema and the latest schema version will be cached. Then after the subsequent `add column` operation is applied, the table schema version will not be updated, so the cache of the decode thread will not be invalidated. 

Therefore, the decode thread will decode the new data with an older schema, considering that the new added column is a dropped column and discarding it.

In addition, there is a lower chance of triggering this problem in the case of frequent add column and insert data.

cherry pick of https://github.com/pingcap/tiflash/pull/5044

Add an internal schema version for DecodingStorageSchemaSnapshot.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix potential data inconsistency under heavy ddl operation
```

### DIFF
--- a/dbms/src/Common/FailPoint.cpp
+++ b/dbms/src/Common/FailPoint.cpp
@@ -65,7 +65,8 @@ std::unordered_map<String, std::shared_ptr<FailPointChannel>> FailPointHelper::f
     M(exception_when_read_from_log)                               \
     M(exception_mpp_hash_build)                                   \
     M(exception_before_drop_segment)                              \
-    M(exception_after_drop_segment)
+    M(exception_after_drop_segment)                               \
+    M(exception_between_schema_change_in_the_same_diff)
 
 #define APPLY_FOR_FAILPOINTS(M)                              \
     M(skip_check_segment_update)                             \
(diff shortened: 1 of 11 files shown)


JaySon-Huang (approved):
LGTM

flowbehappy (approved):

Issue #4862, opened 2022-05-11
Supply comments


breezewish:
## Enhancement

We are now heavily relying on **internal** design docs. It is not friendly for the community to understand the code base. We should at least have comments for important functions, describing their usages.

PR #5139, opened 2022-06-13
enhencement: supplement the comment for SchemaActionType

hongyunyan:

Issue Number: ref #4862 

Problem Summary:

Supplement the comment for SchemaActionType.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


breezewish (approved):

JaySon-Huang:
we need more comments on these lines
https://github.com/pingcap/tiflash/blob/f4c2e012fc3067193412aca91c6b26e30173c722/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L373-L374

/cc @lidezhu for https://github.com/pingcap/tiflash/pull/5044

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -370,7 +370,8 @@ void SchemaBuilder<Getter, NameMapper>::applyAlterPhysicalTable(DBInfoPtr db_inf
         const auto & schema_change = schema_changes[i];
         /// Update column infos by applying schema change in this step.
         schema_change.second(orig_table_info);
-        /// Update schema version aggressively for the sake of correctness.
+        /// Update schema version aggressively for the sake of correctness（for read part).
+        /// In read action, we will use table_info.schema_version(storage_version) and TiDBSchemaSyncer.cur_version(global_version) to compare with query_version, to decide whether we can read under this query_version, or we need to make the schema newer. In our comparison logic, we must ensure storage_version <= gloabl_version. Thus, when the table need apply multi diffs here, we must update schema version aggressively to ensure storage_version <= gloabl_version. The more detail info you can refer the comments in DAGStorageInterpreter::getAndLockStorages.
```
lidezhu, line 374:
```suggestion
        /// In read action, we will use table_info.schema_version(storage_version) and TiDBSchemaSyncer.cur_version(global_version) to compare with query_version, to decide whether we can read under this query_version, or we need to make the schema newer. 
        /// In our comparison logic, we only serve the query when the query schema version meet the criterion: storage_version <= query_version <= global_version(The more detail info you can refer the comments in DAGStorageInterpreter::getAndLockStorages.)
        /// And when apply multi diffs here, we only update global_version when all diffs have been applied.
        /// So the global_version may be less than the actual "global_version" of the local schema in the process of applying schema changes.
        /// And if we don't update the storage_version ahead of time, we may meet the following scenario: storage_version <= global_version < actual "global_version".
        /// If we receive a query with the same version as global_version, we can have the following scenario: storage_version <= global_version == query_version < actual "global_version".
        /// And because storage_version <= global_version == query_version meet the criterion of serving the query, the query will be served. But query_version < actual "global_version" indicates that we use a newer schema to server an older query which may cause some inconsistency issue.
```

JaySon-Huang: LGTM for this change

JaySon-Huang (approved):
LGTM

PR #5202, opened 2022-06-23
add `MPPTask::handleError()`

### LINKED ISSUE #5095: Refine error handling/cancel in TiFlash MPP system
## Enhancement
Currently, error handling/cancel in TiFlash is error prone, it has caused many issues such as #4441, #4219, #4202 etc.
We want to refine the error handling/cancel logical in TiFlash MPP system to make it less error prone.

Some basic ideas:
- Refine `MPPTunnel`
  - `MPPTunnel` has 3 mode: `local`, `sync` and `async`, currently, the implementation of `MPPTunnel` is based on `is_local` and `is_async` flag, which makes the code complex and error prone.
- `MPPTunnel`/`BlockIO`/`ExchangeReceiver` should be treated as the top level components in `MPPTask`
- Each top level components in `MPPTask` should implement its own `cancel` and `handleError` method
- Like `MPPTask::cancel`, there should be a method like `MPPTask::handleError` method to handle errors based on task status
- Like `cancel`, there should be a query level error handling method, so once a `MPPTask` meet error, all the related tasks in the same TiFlash node can see the error and stop running 
- Local tunnel should not introduce direct dependency from send task to receive task 
- As #4208 avoid to print too much meaningless log when error/cancel happens

windtalker:

Issue Number: ref #5095

Problem Summary:

1. Add `MPPTask::abort()` function to abort task in case of being cancelled or encounter an error.
2. Both `MPPTask::cancel()` and `MPPTask::handleError()` will call `MPPTask::abort()` using different abort type.
3. In case of error handling, cancel the `DataStreams` silently to avoid useless error message.
 
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp
+++ b/dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp
@@ -198,8 +198,8 @@ void ParallelAggregatingBlockInputStream::Handler::onException(std::exception_pt
 
     /// can not cancel parent inputStream or the exception might be lost
     if (!parent.executed)
-        /// kill the processor so ExchangeReceiver will be closed
-        parent.processor.cancel(true);
+        /// use cancel instead of kill to avoid too many useless error message
+        parent.processor.cancel(false);
 }
 
 
--- a/dbms/src/Flash/Mpp/MPPTask.cpp
+++ b/dbms/src/Flash/Mpp/MPPTask.cpp
@@ -436,13 +436,14 @@ void MPPTask::runImpl()
     if (unregister)
         unregisterTask();
 
-    mpp_task_statistics.end(status.load(), err_msg);
+    mpp_task_statistics.end(status.load(), err_string);
     mpp_task_statistics.logTracingJson();
 }
 
 void MPPTask::handleError(const String & error_msg)
 {
-    abort(error_msg, AbortType::ONERROR);
+    if (manager == nullptr || !manager->isTaskToBeCancelled(id))
+        abort(error_msg, AbortType::ONERROR);
 }
 
 void MPPTask::abort(const String & message, AbortType abort_type)
@@ -471,6 +472,7 @@ void MPPTask::abort(const String & message, AbortType abort_type)
         }
         else if (previous_status == INITIALIZING && switchStatus(INITIALIZING, next_task_status))
         {
+            err_string = message;
             abortTunnels(message, abort_type);
             unregisterTask();
             LOG_WARNING(log, "Finish abort task from uninitialized");
@@ -481,6 +483,7 @@ void MPPTask::abort(const String & message, AbortType abort_type)
             /// abort the components from top to bottom because if bottom components are aborted
             /// first, the top components may see an error caused by the abort, which is not
             /// the original error
+            err_string = message;
             abortTunnels(message, abort_type);
             abortDataStreams(abort_type);
             abortReceivers();
--- a/dbms/src/Flash/Mpp/MPPTask.cpp
+++ b/dbms/src/Flash/Mpp/MPPTask.cpp
@@ -105,16 +105,18 @@ void MPPTask::abortReceivers()
 
 void MPPTask::abortDataStreams(AbortType abort_type)
 {
+    bool is_kill;
     switch (abort_type)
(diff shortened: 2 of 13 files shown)


dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -405,52 +436,68 @@ void MPPTask::runImpl()
     if (unregister)
         unregisterTask();
 
-    if (switchStatus(RUNNING, FINISHED))
-        LOG_INFO(log, "finish task");
-    else
-        LOG_WARNING(log, "finish task which was cancelled before");
-
     mpp_task_statistics.end(status.load(), err_msg);
```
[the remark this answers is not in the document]
windtalker: Oh, that's possible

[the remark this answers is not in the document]
SeaRise: maybe we can use MPPTask::err to hold the real exception...

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -81,6 +81,44 @@ MPPTask::~MPPTask()
...
+    bool is_kill;
+    switch (abort_type)
+    {
+    case AbortType::ONCANCELLATION:
+        is_kill = true;
+        break;
+    case AbortType::ONERROR:
+        /// When abort type is ONERROR, it means MPPTask already known it meet error, so let the remaining task stop silently to avoid too many useless error message
+        is_kill = false;
+        break;
```
yibin87, line 117: Better to add default case, or use if/else instead of switch will make it more clear.

dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp
```diff
@@ -198,8 +198,8 @@ void ParallelAggregatingBlockInputStream::Handler::onException(std::exception_pt
 
     /// can not cancel parent inputStream or the exception might be lost
     if (!parent.executed)
-        /// kill the processor so ExchangeReceiver will be closed
-        parent.processor.cancel(true);
+        /// use cancel instead of kill to avoid too many useless error message
+        parent.processor.cancel(false);
```
yibin87, line 202: Cancel(true) is introduced by fixing  ExchangeReceiver not abort issue, now, we change back because we abort receivers in MPPTask top level?

yibin87: Just curious: why does TiRemoteBlockInputStream only cancel when kill is true?
https://github.com/pingcap/tiflash/blob/69cbfdf8a6bfb1d98ac76dea6e70d87ab3a1ed84/dbms/src/DataStreams/TiRemoteBlockInputStream.h#L200

windtalker: > 

Because sometimes `cancel(false)` is used to stop the query silently, for example, if multiple data streams is reading from `SharedQueryBlockInputStream`, if any of the data stream read an empty Block, it will call `cancel(false)` to stop other data streams that is reading from the same `SharedQueryBlockInputStream`, in this case we don't expect to get an error because of `cancel(false)`. But if `RemoteReader::cancel()`, it always throws an exception. We can skip this check If we add `is_kill` argument to `RemoteReader::cancel()` to support silent stop.

dbms/src/Flash/Mpp/MPPTask.cpp
windtalker: Because there should be another AbortType `ONKILL`, I will use `if else` and change to `switch case` when `ONKILL` is added.

dbms/src/Flash/Mpp/MPPTask.cpp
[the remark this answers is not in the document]
windtalker: Yes, I store the error message in `MPPTask::err_string`

yibin87 (approved):
LGTM

SeaRise (approved):

PR #5245, opened 2022-06-28
Schema: allow loading empty schema diff when the version grows up.

### LINKED ISSUE #5244: Need allow empty schema diff when schema version increased.
## Enhancement
Before, `update schema version` and `set schema diff` in the same transaction. So after `schema version` grow up, we can get the `schema diff` at same time. 

**For now, We don't allow When `schema version` grow up, `schema diff` can not be read.**

But now, `update schema version` and `set schema diff` not in the same transaction. 

- TIDB can only make sure the `schema diff` from `lastest schema version - 1` is valuable.
   - If the `schema diff` from `lastest schema version - 1` is empty, Then it means may failed in TIDB, in TiFlash, we can skip it.
   - If the `schema diff` from `lastest schema version - 1` is empty, Then Tiflash should direct apply the `schema diff`.
- TIDB can't make sure the `schema diff` with ` lastest schema version` is not empty.

**So we need to allow empty schema diff when schema version increased. **

jiaqizho:

Issue Number: close #5244

Problem Summary:

- Make Tiflash `Schema Syncer` can deal with new `schema version/diff info`, Make it allow loading empty schema diff when the version grows up.
- Code change detail 
  - Changed `getSchemaDiff`, to make it can allow `empty schema diff` as the return value.
  - `tryLoadSchemaDiffs` will return the `applied version` or `error` with `-1`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,64 +145,82 @@ struct TiDBSchemaSyncer : public SchemaSyncer
         return it->second;
     }
 
-    bool tryLoadSchemaDiffs(Getter & getter, Int64 version, Context & context)
+    Int64 tryLoadSchemaDiffs(Getter & getter, Int64 lastest_version, Context & context)
     {
-        if (isTooOldSchema(cur_version, version))
+        if (isTooOldSchema(cur_version, lastest_version))
         {
             return false;
```
xiongjiwei, line 152: maybe also return -1 here?

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,64 +145,82 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
+                //  - `cur_version` is 1, `lastest_version` is 10
+                //  - The schema diff of schema version [2,4,6] is empty, Then we just skip it.
+                //  - The schema diff of schema version 10 is empty, Then we should just apply version into 9
+                if (used_version != lastest_version)
+                {
+                    LOG_FMT_WARNING(log, "Skip the schema diff from version {}. ", used_version);
+                    continue;
+                } // else (used_version == lastest_version)
+
+                LOG_FMT_DEBUG(log, "End load a part of schema diffs, current version is {} ", used_version);
```
xiongjiwei, line 181:
```suggestion
                LOG_FMT_DEBUG(log, "End load a part of schema diffs, current version is {} ", used_version - 1);
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,64 +145,82 @@ struct TiDBSchemaSyncer : public SchemaSyncer
         return it->second;
     }
 
-    bool tryLoadSchemaDiffs(Getter & getter, Int64 version, Context & context)
+    Int64 tryLoadSchemaDiffs(Getter & getter, Int64 lastest_version, Context & context)
```
JaySon-Huang, line 148: 1. Use `latest_version` instead of `lastest_version`
2. Add some comments about the meaning of the return value

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,64 +145,82 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
 
         LOG_FMT_DEBUG(log, "try load schema diffs.");
 
-        SchemaBuilder<Getter, NameMapper> builder(getter, context, databases, version);
+        SchemaBuilder<Getter, NameMapper> builder(getter, context, databases, lastest_version);
 
         Int64 used_version = cur_version;
-        std::vector<SchemaDiff> diffs;
-        while (used_version < version)
+        while (used_version < lastest_version)
```
JaySon-Huang, line 160: We should first fetch all schema diff and save them, and then apply them locally. Should not apply fetch and apply inside one while loop.
It is because that fetching schema diff from tikv is not exception-safe. If we do fetch and apply inside one while loop, we may get inconsistent schema.

For example:
The cur_version is 50 and we call tryLoadSchemaDiff with latest_version=70
Fetch and apply schema diff 51~60 successfully
When fetching schema diff 61, it throw an exception caused by network
Next time when we call tryLoadSchemaDiff, the cur_version is still 50 but the local schema copy is at schema_version=60

Fetch all schema diff and save them firstly, and then apply them locally in another loop won't cause this problem.

jiaqizho: How about add a try catch when call `getter.getSchemaDiff(used_version);`?
Then when 61 failed, it still can roll to 60.

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +147,58 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
         }
         LOG_FMT_DEBUG(log, "end load schema diffs with total {} entries.", diffs.size());
+
         try
         {
-            for (const auto & diff : diffs)
+            for (size_t diff_index = 0; diff_index < diffs.size(); diff_index++)
             {
-                builder.applyDiff(diff);
+                auto schema_diff = diffs[diff_index];
```
JaySon-Huang, line 178:
```suggestion
                const auto & schema_diff = diffs[diff_index];
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +147,58 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
+                    // If `schema diff` from [`cur_version`, `latest_version - 1`] got empty `schema diff`
+                    // Then we should just skip it.
+                    //
+                    // example:
+                    //  - `cur_version` is 1, `latest_version` is 10
+                    //  - The schema diff of schema version [2,4,6] is empty, Then we just skip it.
+                    //  - The schema diff of schema version 10 is empty, Then we should just apply version into 9
+                    if (diff_index != diffs.size() - 1)
+                    {
+                        LOG_FMT_WARNING(log, "Skip the schema diff from version {}. ", used_version);
```
JaySon-Huang, line 193:
```suggestion
                        LOG_FMT_WARNING(log, "Skip the schema diff from version {}. ", cur_version + diff_index + 1);
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +147,58 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
+                        continue;
+                    } // else (diff_index == diffs.size() - 1)
+
+                    LOG_FMT_DEBUG(log, "End load schema diffs without latest one, current version is {} ", used_version - 1);
```
JaySon-Huang, line 197: Also somehow duplicated with line 126, we can remove it

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +147,58 @@ struct TiDBSchemaSyncer : public SchemaSyncer
         return it->second;
     }
 
-    bool tryLoadSchemaDiffs(Getter & getter, Int64 version, Context & context)
+    // Return Values
+    // - if lastest schema diff is not empty, return the (latest_version)
+    // - if lastest schema diff is empty, return the (latest_version - 1)
+    // - if error happend, return (-1)
```
JaySon-Huang, line 153:
```suggestion
    // Return Values
    // - if latest schema diff is not empty, return the (latest_version)
    // - if latest schema diff is empty, return the (latest_version - 1)
    // - if error happend, return (-1)
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -113,12 +113,15 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
 
         GET_METRIC(tiflash_schema_apply_count, type_diff).Increment();
-        if (!tryLoadSchemaDiffs(getter, version, context))
+        Int64 version_after_load_diff = 0;
+        if (version_after_load_diff = tryLoadSchemaDiffs(getter, version, context); version_after_load_diff == -1)
         {
             GET_METRIC(tiflash_schema_apply_count, type_full).Increment();
             loadAllSchema(getter, version, context);
+            // After loadAllSchema, we need update `version_after_load_diff` by last diff value exist or not
+            version_after_load_diff = getter.checkSchemaDiffExists(version) ? version : version - 1;
```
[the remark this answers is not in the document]
xiongjiwei: yes, it determines which version we should use at that moment

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
jiaqizho: May not remove it? Cause we can notify that we load schema diff without the latest one.

JaySon-Huang: I think we can use line 126 to print the information, like `LOG_FMT_INFO(log, "end sync schema, version has been updated to {}{}", cur_version, cur_version == version_after_load_diff? "": "(latest diff is empty)");`

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -113,12 +113,15 @@ struct TiDBSchemaSyncer : public SchemaSyncer
         SCOPE_EXIT({ GET_METRIC(tiflash_schema_applying).Set(0.0); });
 
         GET_METRIC(tiflash_schema_apply_count, type_diff).Increment();
-        if (!tryLoadSchemaDiffs(getter, version, context))
+        Int64 version_after_load_diff = 0;
```
JaySon-Huang, line 116:
```suggestion
        // After the feature concurrent DDL, TiDB does `update schema version` before `set schema diff`, and they are done in separate transaction.
        // So TiFlash may see a schema version X but no schema diff X, meaning that the transaction of schema diff X has not been committed or has
        // been aborted.
        // However, TiDB make sure that if we get a schema version X, then the schema diff X-1 must exist. Otherwise the transaction of schema diff
        // X-1 is aborted and we can safely ignore it.
        // Since TiDB can not make sure the schema diff of latest scheam version X is not empty, under this situation we should set the `cur_version`
        // to X-1 and try to fetch the schema diff X next time.
        Int64 version_after_load_diff = 0;
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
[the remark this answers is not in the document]
JaySon-Huang: OK, I get it.

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -113,14 +113,24 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
             GET_METRIC(tiflash_schema_apply_count, type_full).Increment();
             loadAllSchema(getter, version, context);
+            // After loadAllSchema, we need update `version_after_load_diff` by last diff value exist or not
+            version_after_load_diff = getter.checkSchemaDiffExists(version) ? version : version - 1;
         }
-        cur_version = version;
+        cur_version = version_after_load_diff;
         GET_METRIC(tiflash_schema_version).Set(cur_version);
-        LOG_FMT_INFO(log, "end sync schema, version has been updated to {}", cur_version);
+        LOG_FMT_INFO(log, "End sync schema, version has been updated to {}{}", cur_version, cur_version == version ? "" : "(latest diff is empty)");
```
JaySon-Huang, line 133: If you turn "end" -> "End", then you should also change line 109 "start" -> "Start"

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +154,57 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
 
         LOG_FMT_DEBUG(log, "try load schema diffs.");
 
-        SchemaBuilder<Getter, NameMapper> builder(getter, context, databases, version);
+        SchemaBuilder<Getter, NameMapper> builder(getter, context, databases, latest_version);
 
         Int64 used_version = cur_version;
-        std::vector<SchemaDiff> diffs;
-        while (used_version < version)
+        std::vector<std::optional<SchemaDiff>> diffs;
```
JaySon-Huang, line 173:
```suggestion
        // First get all schema diff from `cur_version` to `latest_version`. Only apply the schema diff(s) if we fetch all
        // schema diff without any exception.
        std::vector<std::optional<SchemaDiff>> diffs;
```

JaySon-Huang (approved):
LGTM

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +154,59 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
             used_version++;
             diffs.push_back(getter.getSchemaDiff(used_version));
         }
-        LOG_FMT_DEBUG(log, "end load schema diffs with total {} entries.", diffs.size());
+        LOG_FMT_DEBUG(log, "End load schema diffs with total {} entries.", diffs.size());
+
         try
         {
-            for (const auto & diff : diffs)
+            for (size_t diff_index = 0; diff_index < diffs.size(); diff_index++)
```
Lloyd-Pottiger, line 185:
```suggestion
            for (size_t diff_index = 0; diff_index < diffs.size(); ++diff_index)
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
```diff
@@ -144,30 +154,59 @@ struct TiDBSchemaSyncer : public SchemaSyncer
...
+                    // example:
+                    //  - `cur_version` is 1, `latest_version` is 10
+                    //  - The schema diff of schema version [2,4,6] is empty, Then we just skip it.
+                    //  - The schema diff of schema version 10 is empty, Then we should just apply version into 9
+                    if (diff_index != diffs.size() - 1)
+                    {
+                        LOG_FMT_WARNING(log, "Skip the schema diff from version {}. ", cur_version + diff_index + 1);
+                        continue;
+                    } // else (diff_index == diffs.size() - 1)
+
```
Lloyd-Pottiger, line 205:
```suggestion
                    }
                    
                    // if diff_index == diffs.size() - 1, return used_version - 1;
```

dbms/src/TiDB/Schema/TiDBSchemaSyncer.h
jiaqizho: what for change `a++` to `++a`？

Lloyd-Pottiger: `++a` is more effective. https://stackoverflow.com/questions/24886/is-there-a-performance-difference-between-i-and-i-in-c

jiaqizho: just dump, example here.

```
a++
mov eax, double word ptr [a]
add eax, 1
mov double word ptr [a], eax

++a
mov eax, double word ptr [a]
add eax, 1
mov double word ptr [a], eax
```
so only do `a++` or `++a` here, won't be any effective difference.

Lloyd-Pottiger: yep, it will be optimize by compiler, there will not be any difference in binary, but give pressure to compiler. We can find difference by set `-O0` when compiling. Anyway, there are only tiny difference. The others LGTM.

Lloyd-Pottiger (approved):

Issue #5252, opened 2022-06-29
Support fast mode on TiFlash


hongyunyan:
## Enhancement

We provide a brand new mode called "fast mode" for TiFlash, to provide better query performance with a small loss of data consistency and timeliness.

hongyunyan:
TODO List:
1. fix min max index on del_mark
2. use min max index on del_mark to optimize read in fast mode.

hongyunyan:
todo:
the integration test about this case to check when do query slightly before 'set tiflash mode':
start query (set failpoint) --> set tiflash mode and refresh schema --> continue query, and due to query_version < local_schema --> query fail

PR #5253, opened 2022-06-29
feat: modify minor compaction to make each ColumnFile in persisted set is sorted

hongyunyan:

Issue Number: ref #5252 

Problem Summary:

Modify minor compaction to make each ColumnFile in persisted set is sorted, thus we can speed up reading under fast mode. Besides, a more ordered delta value space will also improve read performance in normal mode.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

test by dtworkload， environment: M1 Pro 32G
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


hongyunyan:
cc @lidezhu @flowbehappy

dbms/src/Storages/DeltaMerge/Delta/DeltaValueSpace.cpp
```diff
@@ -262,11 +252,26 @@ bool DeltaValueSpace::compact(DMContext & context)
...
+
         LOG_FMT_DEBUG(log, "{} {}", simpleInfo(), compaction_task->info());
     }
     wbs.writeRemoves();
 
     return true;
 }
+
+// updateDeltaTree should be used under lock.
+void DeltaValueSpace::updateDeltaTree(DeltaIndex::Updates delta_index_updates)
```
lidezhu, line 266:
```suggestion
void DeltaValueSpace::updateDeltaTree(const DeltaIndex::Updates & delta_index_updates)
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1608,6 +1608,7 @@ bool DeltaMergeStore::handleBackgroundTask(bool heavy)
         }
         case TaskType::Compact:
             task.segment->compactDelta(*task.dm_context);
+            task.segment->placeDeltaIndex(*task.dm_context);
```
lidezhu, line 1611: After `compactDelta`, it may be not a good point to `placeDeltaIndex`.  
Because after flush task complete, it indicates that there may some fresh data at the tail which just flush from memory to disk. So it's a good signal to try to `placeDeltaIndex`. 
But `compactDelta` is just a reorganization of the old data, so I suggest not adding this line here.

lidezhu (approved):
LGTM

flowbehappy:
/hold
There are some performance regressions. Should be figured out before merging.

PR #5261, opened 2022-06-30
Refine test framework code and tests

### LINKED ISSUE #5262: Test framework improvement
## Enhancement

Current executor test framework passes expected columns to `execute_streams()` and does equality checking within it. This approach has some drawbacks

1. hard to recognize which line of the unit test file causes failure. Note that, current implementation will always output the file name and line number of a single internal function.
2. hard to do different kinds of equality checking. To be specific, with higher concurrency, the final columns result might not be stable (the order might change). However, current implementation lacks the ability to handle this situation. And current workaround (appending a `topN` executor when results might be unstable) is not elegant. (we should keep each unit test as self-contained and less dependent as possible)

Willendless:

Issue Number: close #5262 

Problem Summary:

1. introduce two separate macros (ASSERT_COLUMNS_EQ_R/UR) to test executors. `ASSERT_COLUMNS_EQ_UR` allows equality checking in unstable situation. It achieves this by converting both expected and actual cols to rows then inserting them to separate `std::multiset` and do comparison. On the other hand, `ASSERT_COLUMNS_EQ_R` requires more restrict equality checking, namely each column's every element and their offset should be the same.
2. remove the `expected_cols` argument from execute_streams() and force to compare equality only in unit test file. This step allows more useful (file name:line number) output message when test fails.
3. update old tests according to above modification

One failure example output using ASSERT_COLUMNS_EQ_UR is like

```shell
/tiflash/dbms/src/Flash/tests/gtest_join_executor.cpp:167: Failure
Value of: DB::tests::columnsEqual((executeStreams(request, 5)), (ans), false)
  Actual: false (Columns row set mismatch
expected_row_set:
'1' '3' '1' '3' 
'3' '4' '2' '4' 
...
actual_row_set:
'1' '3' '1' '3' 
'2' '4' '2' '4' 
...
)
Expected: true
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Since c/c++ preprocessor recognizes `{col0, col1, col2}` as multiple input arguments for function-like macros, while using `ASSERT_COLUMNS_EQ_R/UR`, we need to wrap these columns with `createColumns()`, i.e. `createColumns({col0, col1, col2})`;

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/TestUtils/FunctionTestUtils.cpp
```diff
@@ -103,22 +111,114 @@ ::testing::AssertionResult columnEqual(
...
+        ++actual_it;
+
+        ret << "...\nactual_row_set:\n";
+        for (auto it = actual_row_set.begin(); it != actual_it; ++it)
+        {
+            ret << fmt_row(*it) << "\n";
+        }
+        ret << "...\n";
+
+        return ret;
```
SeaRise, line 218: can we use FmtBuffer?

dbms/src/Flash/tests/gtest_projection_executor.cpp
```diff
@@ -44,17 +44,17 @@ class ExecutorProjectionTestRunner : public DB::tests::ExecutorTest
     }
 
     template <typename T>
-    std::shared_ptr<tipb::DAGRequest> buildDAGRequest(T param, const String & sort_col)
+    std::shared_ptr<tipb::DAGRequest> buildDAGRequest(T param)
     {
         /// topN is introduced, so that we can get stable results in concurrency environment.
-        return context.scan(db_name, table_name).project(param).topN(sort_col, false, 100).build(context);
```
SeaRise, line 50: why remove `topN(sort_col, false, 100)`?

dbms/src/TestUtils/FunctionTestUtils.cpp
Willendless: Sure.

dbms/src/Flash/tests/gtest_projection_executor.cpp
Willendless: It seems `topN` only aims to get stable results instead of testing its correctness. Since now using `ASSERT_COLUMNS_EQ_UR` we can correctly test result even in unstable situation. It might be better to remove `topN` and make the ut more compact? I think I might forget to also remove the comment...

SeaRise: And `TopN` of https://github.com/pingcap/tiflash/blob/cbe6ab5593926ba82d15f8b2ac398fd98dff56bc/dbms/src/Flash/tests/gtest_executor.cpp#L94, https://github.com/pingcap/tiflash/blob/cbe6ab5593926ba82d15f8b2ac398fd98dff56bc/dbms/src/Flash/tests/gtest_executor.cpp#L126, https://github.com/pingcap/tiflash/blob/cbe6ab5593926ba82d15f8b2ac398fd98dff56bc/dbms/src/Flash/tests/gtest_executor.cpp#L144, https://github.com/pingcap/tiflash/blob/cbe6ab5593926ba82d15f8b2ac398fd98dff56bc/dbms/src/Flash/tests/gtest_executor.cpp#L174, https://github.com/pingcap/tiflash/blob/cbe6ab5593926ba82d15f8b2ac398fd98dff56bc/dbms/src/Flash/tests/gtest_executor.cpp#L211 can be removed. 
/cc @ywqzzy

Willendless: This might not be a problem.. I am working on moving these join tests to a new separate test file.

SeaRise (approved):
LGTM

dbms/src/TestUtils/FunctionTestUtils.cpp
```diff
@@ -24,6 +26,14 @@
 #include <TestUtils/FunctionTestUtils.h>
 #include <TestUtils/TiFlashTestBasic.h>
 #include <fmt/core.h>
+#include <gtest/gtest.h>
+
+#include <ext/enumerate.h>
+#include <iterator>
+#include <set>
+
+#include "Common/FmtUtils.h"
```
ywqzzy, line 35:
```suggestion
#include <Common/FmtUtils.h>
```

dbms/src/TestUtils/FunctionTestUtils.cpp
```diff
@@ -24,6 +26,14 @@
 #include <TestUtils/FunctionTestUtils.h>
 #include <TestUtils/TiFlashTestBasic.h>
 #include <fmt/core.h>
+#include <gtest/gtest.h>
```
ywqzzy, line 29: maybe not needed

dbms/src/TestUtils/FunctionTestUtils.cpp
```diff
@@ -103,22 +113,117 @@ ::testing::AssertionResult columnEqual(
...
-        ASSERT_EQ(actual_col.type->getName(), expected_col.type->getName());
-        ASSERT_COLUMN_EQ(expected_col.column, actual_col.column);
+        auto cmp_res = columnEqual(expected_col, actual_col);
+        if (!cmp_res)
+            return cmp_res;
     }
+    return ::testing::AssertionSuccess();
+}
+
+/// @pre size of each column should be the same
```
ywqzzy, line 136: pre?What does it mean

ywqzzy (commented):
Others LGTM

dbms/src/TestUtils/ExecutorTestUtils.cpp
```diff
@@ -21,6 +21,11 @@
 
 #include <functional>
 
+#include "Columns/IColumn.h"
+#include "Core/ColumnsWithTypeAndName.h"
+#include "TestUtils/FunctionTestUtils.h"
+#include "gtest/gtest.h"
```
SeaRise, line 27:
```suggestion
#include <Columns/IColumn.h>
#include <Core/ColumnsWithTypeAndName.h>
#include <TestUtils/FunctionTestUtils.h>
#include <gtest/gtest.h>
```

dbms/src/TestUtils/FunctionTestUtils.cpp
Willendless: Oh, it might not be a good idea to add "@pre". I will remove it. It actually means precondition.

ywqzzy (approved):

PR #5264, opened 2022-06-30
feat: init clang-15 support

### LINKED ISSUE #5265: Clang 15 support
## Enhancement

release notes: https://releases.llvm.org/download.html

### LIBCPP Highlists
[New Features](https://releases.llvm.org/14.0.0/projects/libcxx/docs/ReleaseNotes.html#id3)

    There’s support for the C++20 header <format>. Some parts are still missing, most notably the compile-time format string validation. Some functions are known to be inefficient, both in memory usage and performance. The implementation isn’t API- or ABI-stable and therefore considered experimental. (Some not-yet-implemented papers require an API-break.) As a result, it is disabled by default, however vendors can enable the header by using -DLIBCXX_ENABLE_INCOMPLETE_FEATURES=ON when configuring their build.

    More parts of <ranges> have been implemented. Since we still expect to make some API and ABI breaking changes, those are disabled by default. However, vendors that wish to enable <ranges> in their distribution may do so by defining -DLIBCXX_ENABLE_INCOMPLETE_FEATURES=ON when configuring their build.

    There’s a new CMake option LIBCXX_ENABLE_UNICODE to disable Unicode support in the <format> header. This only affects the estimation of the output width of the format functions.

    Support for building libc++ on top of a C Standard Library that does not support wchar_t was added. This is useful for building libc++ in an embedded setting, and it adds itself to the various freestanding-friendly options provided by libc++.

    Defining _LIBCPP_DEBUG to 1 enables the randomization of unspecified behavior in standard algorithms (e.g. the ordering of equal elements in std::sort, or the ordering of both sides of the partition in std::nth_element).

    Floating-point support for std::to_chars support has been added. Thanks to Stephan T. Lavavej and Microsoft for providing their implementation to libc++.

    The C++20 <coroutine> implementation has been completed.

    More C++20 features have been implemented. [libc++ C++20 Status](https://releases.llvm.org/14.0.0/projects/libcxx/docs/Status/Cxx20.html) has the full overview of libc++’s C++20 implementation status.

    More C++2b features have been implemented. [libc++ C++2b Status](https://releases.llvm.org/14.0.0/projects/libcxx/docs/Status/Cxx2b.html) has the full overview of libc++’s C++2b implementation status.

    16-bit wchar_t handling added for codecvt_utf8, codecvt_utf16 and codecvt_utf8_utf16.

### Clang Highlights
[clang-format](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#id29)

    Option AllowShortEnumsOnASingleLine: false has been improved, it now correctly places the opening brace according to BraceWrapping.AfterEnum.

    Option AlignAfterOpenBracket: BlockIndent has been added. If set, it will always break after an open bracket, if the parameters don’t fit on a single line. Closing brackets will be placed on a new line.

    Option QualifierAlignment has been added in order to auto-arrange the positioning of specifiers/qualifiers const volatile static inline constexpr restrict in variable and parameter declarations to be either Right aligned or Left aligned or Custom using QualifierOrder.

    Option QualifierOrder has been added to allow the order const volatile static inline constexpr restrict to be controlled relative to the type.

    Option RemoveBracesLLVM has been added to remove optional braces of control statements for the LLVM style.

    Option SeparateDefinitionBlocks has been added to insert or remove empty lines between definition blocks including functions, classes, structs, enums, and namespaces.

    Add a Custom style to SpaceBeforeParens, to better configure the space before parentheses. The custom options can be set using SpaceBeforeParensOptions.

    The command line argument -style=<string> has been extended so that a specific format file at location <format_file_path> can be selected. This is supported via the syntax: -style=file:<format_file_path>.

    Improved C++20 Modules and Coroutines support.

    Option AfterOverloadedOperator has been added in SpaceBeforeParensOptions to allow space between overloaded operator and opening parentheses.

### LLVM Highlights 

[Changes to the AArch64 Backend](https://releases.llvm.org/14.0.0/docs/ReleaseNotes.html#id6)

    Added support for the Armv9-A, Armv9.1-A and Armv9.2-A architectures.

    The compiler now recognises the “tune-cpu” function attribute to support the use of the -mtune frontend flag. This allows certain scheduling features and optimisations to be enabled independently of the architecture. If the “tune-cpu” attribute is absent it tunes according to the “target-cpu”.

    Fixed relocations against temporary symbols (e.g. in jump tables and constant pools) in large COFF object files.

    Auto-vectorization now targets SVE by default when available.

[Changes to the X86 Target](https://releases.llvm.org/14.0.0/docs/ReleaseNotes.html#id12)

    Support for AVX512-FP16 instructions has been added.

    Removed incomplete support for Intel MPX. ([D111517](https://reviews.llvm.org/D111517))

SchrodingerZhu:
Signed-off-by: SchrodingerZhu <i@zhuyi.fan>

Issue Number: close #5265

no extra clang, just suppress some warnings for all grpc sub targets (not causing errors actually)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


.toolchain.yml
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Hi, there!
+# This is passed via jenkins podTemplate argument.
+# Instead of trying injections, take a look at https://en.pingcap.com/careers/.
+image_tag_suffix: -llvmorg-14.0.6
```
JaySon-Huang, line 18: Seems that it didn't apply to the build process? https://ci.pingcap.net/blue/organizations/jenkins/tiflash-build-common/detail/tiflash-build-common/7534/pipeline/155
```
[2022-07-01T02:41:18.259Z] + cmake /home/jenkins/agent/workspace/tiflash-build-common/tiflash -DPREBUILT_LIBS_ROOT=/home/jenkins/agent/workspace/tiflash-build-common/tiflash/contrib/tiflash-proxy/ -G Ninja -DENABLE_FAILPOINTS=true -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/usr/local -DCMAKE_INSTALL_PREFIX=/home/jenkins/agent/workspace/tiflash-build-common/install/tiflash -DENABLE_TESTS=false -DUSE_CCACHE=true -DDEBUG_WITHOUT_DEBUG_INFO=true -DUSE_INTERNAL_TIFLASH_PROXY=false -DRUN_HAVE_STD_REGEX=0
[2022-07-01T02:41:18.259Z] -- The C compiler identification is Clang 13.0.0
[2022-07-01T02:41:18.517Z] -- The CXX compiler identification is Clang 13.0.0
```

solotzg:
How about enabling `bolt` for >= llvm 14.0.0

SchrodingerZhu:
@solotzg it still need profiling. It is worth trying but perhaps not in this PR?

SchrodingerZhu:
https://github.com/PingCAP-QE/ci/pull/1145/files

JaySon-Huang:
Is this pr ready to be merged?

breezewish (commented):
This PR changes the compiler used in **production release** from Clang 13 to Clang 14, which might be too aggressive and needs some careful QA work. How about let's separate into two:

PR 1: Mute warnings.

PR 2: Bump to use LLVM 14 in the release image. <-- We need to involve QA and at least 1 sprint of automated test time for it.

What do you think? @JaySon-Huang

JaySon-Huang:
Clickhouse has used clang-14 for a long time, so I think it is OK to switch to clang-14.

But now as clang-15 has been released on Sep 6th, https://llvm.org/ I think if we plan to upgrade the compiler, we can directly upgrade to clang-15. (BTW, Clickhouse switch to clang-15 as soon as it released 😂)

JaySon-Huang:
I think the performance gain (and new C++ features) would be the main reason for upgrading the compiler, but we lack getting a performance gain report and almost nobody treats it as an important thing now 😂...

breezewish:
Okay, then I would vote for using LLVM 15 directly, as it is always a big move for us to upgrade the toolchain, especially when upgrading the minimal required toolchain. Bumping to latest directly would make life easier.

release-centos7-llvm/env/Makefile
```diff
@@ -15,21 +15,21 @@
...
-tiflash-env/loader: tiflash-env
 	cp loader tiflash-env
-
-tiflash-env/tiflash-linker: tiflash-env
 	cp tiflash-linker tiflash-env
 
-tiflash-env-$(ARCH).tar.xz: tiflash-env/prepare-sysroot.sh tiflash-env/loader tiflash-env/tiflash-linker
-	docker run --rm -v $(MOUNT):/tiflash-env hub.pingcap.net/tiflash/tiflash-llvm-base:$(DOCKER_TAG) /tiflash-env/prepare-sysroot.sh
+tiflash-env-$(ARCH).tar.xz: scripts
+	sudo docker run --rm -v $(MOUNT):/tiflash-env hub.pingcap.net/tiflash/tiflash-llvm-base:$(DOCKER_TAG) /tiflash-env/prepare-sysroot.sh
```
JaySon-Huang, line 31:
```suggestion
	docker run --rm -v $(MOUNT):/tiflash-env hub.pingcap.net/tiflash/tiflash-llvm-base:$(DOCKER_TAG) /tiflash-env/prepare-sysroot.sh
```

breezewish:
Let's re-open this PR when appropiate.

Issue #5263, opened 2022-06-30
`ParallelInputsProcessor` should be able to parallelize `additional_input_at_end`


gengliqi:
## Enhancement

Consider the TPC-H sql 13:
``` sql
select
        c_custkey,
        count(o_orderkey) as c_count
from
        customer left outer join orders on
                c_custkey = o_custkey
group by
        c_custkey;
```
Execution plan:
``` SQL
+--------------------------------------------+------------+--------------+----------------+----------------------------------------------------------------------------------------------------------------------------------------------------+
| id                                         | estRows    | task         | access object  | operator info                                                                                                                                      |
+--------------------------------------------+------------+--------------+----------------+----------------------------------------------------------------------------------------------------------------------------------------------------+
| TableReader_61                             | 748800.00  | root         |                | data:ExchangeSender_60                                                                                                                             |
| └─ExchangeSender_60                        | 748800.00  | mpp[tiflash] |                | ExchangeType: PassThrough                                                                                                                          |
|   └─Projection_8                           | 748800.00  | mpp[tiflash] |                | sf10.customer.c_custkey, Column#18                                                                                                                 |
|     └─Projection_55                        | 748800.00  | mpp[tiflash] |                | Column#18, sf10.customer.c_custkey                                                                                                                 |
|       └─HashAgg_53                         | 748800.00  | mpp[tiflash] |                | group by:sf10.customer.c_custkey, funcs:count(sf10.orders.o_orderkey)->Column#18, funcs:firstrow(sf10.customer.c_custkey)->sf10.customer.c_custkey |
|         └─HashJoin_38                      | 7512019.23 | mpp[tiflash] |                | left outer join, equal:[eq(sf10.customer.c_custkey, sf10.orders.o_custkey)]                                                                        |
|           ├─ExchangeReceiver_24(Build)     | 750000.00  | mpp[tiflash] |                |                                                                                                                                                    |
|           │ └─ExchangeSender_23            | 750000.00  | mpp[tiflash] |                | ExchangeType: HashPartition, Hash Cols: [name: sf10.customer.c_custkey, collate: binary]                                                           |
|           │   └─TableFullScan_22           | 750000.00  | mpp[tiflash] | table:customer | keep order:false                                                                                                                                   |
|           └─ExchangeReceiver_27(Probe)     | 7500000.00 | mpp[tiflash] |                |                                                                                                                                                    |
[...]

Issue #4609, opened 2022-04-07
Interpreter & Executor & MPP unit test framework


ywqzzy:
## Enhancement
1. Since there is no unit test for the Interpreter module, Interpreter test framework aims to supplement the content of the unit test, improve the stability of the module, and find more hidden bugs at home.
2. TiFlash planner needs to rely on unit tests to ensure the quality and progress of development during the refactoring process.
3. User can mock input columns and feed them into the "Execution streams(BlockInputStreams in TiFlash)", then they can test executors(except MPP related logic) of TiFlash locally.
4. Further more, MPP related tests will be supported.
5. When developing more features in the future, good unit testing can be used as the basis for development to further improve the efficiency and quality of code development.

## Non MPP tests development progress:
- [x] #4610 
- [x] #4632 
- [x] #4706 
- [x] #4742 
- [x] #4788 
- [x] #4911 
- [x] #4858 
- [x] #5041 
- [x] #5021 
- [x] #5262 
- [x] #5295 
- [x] #5243 
- [x] #5347 
- [ ] #5345 
- [x] #5432 
- [ ] #5469 
- [x] #5510 
- [x] #5534 
- [x] #5419 
- [x] #5543
- [x] #5305  
- [x] #5889 add test log when failed.
- [x] #5957 
- [x] #5787
- [x] #5895 
- [x] #5915 
- [x] Test more types of executors[WIP].
- [ ] Support more expressions to test.
- [ ] More user friendly exception throw.
- [x] Check the legality and correctness of user input.
- [x] Add some easy to use template to write tests.
- [x] When result is wrong, print more user friendly message to help debug.
- [x] Add some built-in input columns for user to generate a bunch of tests.
- [x] Generate columns with different types and size. #5743 
- [ ] Generate columns following different data distribution.
- [x] Support regression test like this https://github.com/risinglightdb/sqlplannertest-rs. Here is the detail https://github.com/pingcap/tiflash/issues/6676
- [x] A document to demonstrate how to use the test framework.
- [ ] Bind test framework with google bench.
- [x] New executor should implement Serializer. New BlockInputStream should implement appendInfo().
- [ ] Improve the speed of tests.
- [ ] A test generator like sqlsmith.(Just for fun)
- [x] #6160 
- [x] #6561 
- [x] #7059 

## MPP tests development progress:
- [x] #5369 
- [x] #5450 
- [x] #5573
- [x] Orchestrate MPP Tasks.
[...]

PR #5274, opened 2022-07-01
Change one `additional_input_at_end` to many streams in `ParallelInputsProcessor` 

gengliqi:

Issue Number: close https://github.com/pingcap/tiflash/issues/5263 close https://github.com/pingcap/tiflash/issues/4856

Problem Summary:
This PR fixes two performance issues that are #5263 and #4856.

For #5263, when a query has a join then aggregation and its non-join data are enormous, this PR has a **huge speedup** in this scenario because multiple threads can be used to aggregate instead of a single thread in the previous implementation.

For #4856, this PR can optimize a scenario where some input streams' data are very huge and some threads may exit too early. For now, no thread can exit early. They work until all input steams are exhausted.

Refine the `ParallelInputsProcessor` code.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -205,27 +217,52 @@ class ParallelInputsProcessor
...
             }
+            else
+            {
+                wait_first_done.wait(lock, [this] {
+                    return running_first == 0;
+                });
+            }
+        }
 
+        try
```
fuzhe1989, line 264: L264-L276 is repeating L-218-L232, how about extract them into a function or lambda?

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -205,27 +217,52 @@ class ParallelInputsProcessor
...
+            {
+                handler.onFinish();
+            }
+        }
 
-                {
-                    std::lock_guard lock(available_inputs_mutex);
-                    available_inputs.push(unprepared_input);
-                }
+        {
```
fuzhe1989, line 242: Here contains two steps:
1. all threads wait for a barrier.
2. init streams into a pair of `unprepared_inputs` and `available_inputs`.

My suggestions:
1. explicitly express the barrier semantic.
2. how about encapsulate streams, `unprepared_inputs`, and `available_inputs` into one struct named as `WorkingStreams`? It will have two instances, one for inputs, another for additional_inputs_at_end. Then you can first call `loop(thread_num, inputsStreams)` then call `loop(thread_num, additionalStreams)` and needn't to do initialization between them.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -188,6 +182,24 @@ class ParallelInputsProcessor
...
+            catch (...)
+            {
+                /** If you can not ask one or more sources to stop.
+                      * (for example, the connection is broken for distributed query processing)
+                      * - then do not care.
+                      */
+                LOG_FMT_ERROR(log, "Exception while cancelling {}", child->getName());
+            }
+        }
+    }
```
SeaRise, line 201:
```suggestion
    void cancelStream(const BlockInputStreamPtr & stream, bool kill) {
        if (auto * p_stream = dynamic_cast<IProfilingBlockInputStream *>(&*stream))
        {
            try
            {
                p_stream->cancel(kill);
            }
            catch (...)
            {
                /** If you can not ask one or more sources to stop.
                      * (for example, the connection is broken for distributed query processing)
                      * - then do not care.
                      */
                LOG_FMT_ERROR(log, "Exception while cancelling {}", p_stream->getName());
            }
        }
    }
```

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -205,27 +217,53 @@ class ParallelInputsProcessor
...
-            while (!finish)
-            {
-                InputData unprepared_input;
-                {
-                    std::lock_guard lock(unprepared_inputs_mutex);
+            loop(thread_num);
+        }
+        catch (...)
+        {
+            exception = std::current_exception();
```
SeaRise, line 224: how about using `getCurrentExceptionMessage(true)`

gengliqi: The `exception` is a `std::exception_ptr` not string.

gengliqi:
@fuzhe1989 @SeaRise I have refined the code by using `MPMCQueue`. PTAL again, thanks!

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -181,13 +167,80 @@ class ParallelInputsProcessor
...
+
+        /// How many active input streams.
+        std::atomic<size_t> active_inputs;
+
+        /** For parallel preparing (readPrefix) child streams.
+          * First, streams are located here.
+          * After a stream was prepared, it is moved to "available_inputs" for reading.
+          */
+        using UnpreparedInputs = std::queue<InputData>;
+        UnpreparedInputs unprepared_inputs;
```
fuzhe1989, line 216: nit: How about let `unprepared_inputs` another `MPMCQueue`? Then we could get rid of `std::mutex` totally.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -237,127 +283,81 @@ class ParallelInputsProcessor
...
+    void loop(size_t thread_num, WorkingInputs & work)
+    {
+        if (work.active_inputs == 0)
+        {
+            return;
+        }
 
-        /// The last thread on the output indicates that there is no more data.
-        if (0 == --active_threads)
+        while (!finish)
```
fuzhe1989, line 298: After using `MPMCQueue` you needn't depend on `finish`, the queue itself contains enough information.

fuzhe1989:
great refactoring!

SeaRise:
It seems you can add `left_join --> agg` to the UT using https://github.com/pingcap/tiflash/issues/4609
like https://github.com/pingcap/tiflash/blob/57d001c1f91ef656d70eff13008e0fbff5588873/dbms/src/Flash/tests/gtest_interpreter.cpp and https://github.com/pingcap/tiflash/blob/57d001c1f91ef656d70eff13008e0fbff5588873/dbms/src/Flash/tests/gtest_executor.cpp

dbms/src/DataStreams/ParallelInputsProcessor.h
gengliqi: A no-waiting `tryPop` is needed for `MPMCQueue`. I will file a PR for this little fix soon.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -237,127 +283,72 @@ class ParallelInputsProcessor
...
+                    work.available_inputs.finish();
                     break;
-
-                if (block)
-                    publishPayload(input.in, block, thread_num);
+                }
             }
         }
+
+        // Should readSuffix be called here?
```
SeaRise, line 339: looks unnecessary.
https://github.com/pingcap/tiflash/blob/57d001c1f91ef656d70eff13008e0fbff5588873/dbms/src/DataStreams/UnionBlockInputStream.h#L247-L257
https://github.com/pingcap/tiflash/blob/57d001c1f91ef656d70eff13008e0fbff5588873/dbms/src/DataStreams/UnionBlockInputStream.h#L206-L209

SeaRise:
Basically LGTM, good job!

dbms/src/DataStreams/ParallelInputsProcessor.h
gengliqi: I think we can merge this PR https://github.com/pingcap/tiflash/pull/5311 first.

gengliqi: Done.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -237,132 +278,63 @@ class ParallelInputsProcessor
...
 
-                input = available_inputs.front();
+        while (work.unprepared_inputs.tryPop(input))
+        {
+            input.in->readPrefix();
 
-                /// We remove the source from the queue of available sources.
-                available_inputs.pop();
-            }
+            work.available_inputs.push(input);
```
SeaRise, line 299: nit: maybe we can check the result of push.

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -237,132 +278,63 @@ class ParallelInputsProcessor
...
             /// The main work.
             Block block = input.in->read();
 
+            if (block)
             {
-                if (finish)
-                    break;
-
-                /// If this source is not run out yet, then put the resulting block in the ready queue.
+                work.available_inputs.push(input);
```
SeaRise, line 311: ditto

SeaRise (approved):
Others LGTM

dbms/src/DataStreams/ParallelInputsProcessor.h
```diff
@@ -237,132 +278,63 @@ class ParallelInputsProcessor
...
-    std::mutex unprepared_inputs_mutex;
+    WorkingInputs working_inputs;
+    WorkingInputs working_additional_inputs;
 
     /// How many sources ran out.
     std::atomic<size_t> active_threads{0};
-    /// Finish the threads work (before the sources run out).
-    std::atomic<bool> finish{false};
     /// Wait for the completion of all threads.
     std::atomic<bool> joined_threads{false};
```
fuzhe1989, line 339: `joined_threads` seems needless now, we can reset `thread_manager` after join.

fuzhe1989 (approved):

dbms/src/DataStreams/ParallelInputsProcessor.h
gengliqi: Seems not necessary? Because the result may be true or false and if false, `tryPop` and `pop` will return false as well.

gengliqi:
Changed `InterpreterSelectQuery` to fix integration test. Make some code the same with `DAGQueryBlockInterpreter`.

PR #5277, opened 2022-07-02
Fix shell escaped character comparison bug in test scripts

### LINKED ISSUE #5275: TiFlash  doesn't support $ in Integration Tests
## Bug Report

cd tests
./run-test.sh fullstack-test/expr/upper.test
This test will success.
vi fullstack-test/expr/upper.test
1.add "insert into test.t values('pi$ng');" after origin insert sql
2.add "PI$NG" in excepted result at last
image

run test again, this test is failed
image

This may because the integration tests framework doesn't support $ .
It may delete $ and the letter followed by $

after add $ in upper.test ,this test case still success
after add $ in upper.test ,this test case will fail
TiFlash master branch

Willendless:

Issue Number: close #5275 

Problem Summary:

1. append `\` before "\\", "$", "`" characters to avoid shell escaping (order matters)
https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
2. ~use `--raw` option to ask mysql-client to not escape characters when return results. ref: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_raw~ (this might broke some tests that rely on output character escaping to do equality checking, ex. fullstack-test2/ddl/binary_default_value.test)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/run-test.py
```diff
@@ -41,12 +41,18 @@ def exec_func(cmd):
     err = p.close()
     return output, err
 
+# translate string to unescaped version in shell environment
+# we only need to consider '$', '`' and '\'
+# ref: https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
+def to_unescaped_str(cmd):
```
lidezhu, line 47: how about add tests for usage of '$' and '\' in the test?

Willendless: sure

JaySon-Huang (approved):
lgtm

tests/fullstack-test/expr/reverse.test
```diff
@@ -20,10 +20,16 @@ mysql> insert into test.t values('one week’s time test');
...
+mysql> insert into test.t values('+ѐ-ё*ђ/ѓ!є@ѕ#і$@ї%ј……љ&њ（ћ）ќ￥ѝ#ў@џ！^');
 mysql> insert into test.t values('αβγδεζηθικλμνξοπρστυφχψωσ');
 mysql> insert into test.t values('▲α▼βγ➨δε☎ζη✂θι€κλ♫μν✓ξο✚πρ℉στ♥υφ♖χψ♘ω★σ✕');
 mysql> insert into test.t values('թփձջրչճժծքոեռտըւիօպասդֆգհյկլխզղցվբնմշ');
+mysql> insert into test.t values('\$$\$$\$$');
+mysql> insert into test.t values('\$$\$$\$$\$');
+mysql> insert into test.t values('\`\$$\$$\$');
+mysql> insert into test.t values('\`$$`\$$\$$\$');
+mysql> insert into test.t values('\\$```');
+mysql> insert into test.t values('\\\```');
```
Willendless, line 32: explanation for the added tests:

1. `\$$\$$\$$`: note that `\` is also a escape character for mysql, therefore mysql will only insert `$$$$$$` into the table
2. similar to the first case, mysql will only insert `$$$$$$$` into the table
3. similar to the first case, mysql will only insert `$$$$$ into the table
4. similar to the first case, mysql will only insert \`$$ `$$$$$ into the table
5. mysql will insert \\$``` into the table, however since we don't use `--raw` option for mysql client, the output backslash will be escaped to be double backslash
6. mysql will insert \\``` into the table, however since we don't use `--raw` option for mysql client, the output backslash will be escaped to be double backslash

ref: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_raw

Willendless, line 32: explanation for the added tests:

1. `\$$\$$\$$`: note that `\` is also a escape character for mysql, therefore mysql will only insert `$$$$$$` into the table
2. similar to the first case, mysql will only insert `$$$$$$$` into the table
3. similar to the first case, mysql will only insert `$$$$$ into the table
4. similar to the first case, mysql will only insert \`$$ `$$$$$ into the table
5. mysql will insert \\$``` into the table, however since we don't use `--raw` option for mysql client, the output backslash will be escaped to be double backslash
6. mysql will insert \\``` into the table, however since we don't use `--raw` option for mysql client, the output backslash will be escaped to be double backslash

ref: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_raw

lidezhu (approved):

PR #5279, opened 2022-07-03
upgrade prometheus-cpp to v1.0.1

### LINKED ISSUE #5278: Upgrade prometheus-cpp
## Enhancement

The current prometheus-cpp version is too old, with some bugs in compilation:

```
/home/yangkeao/Project/github.com/YangKeao/tiflash/contrib/prometheus-cpp/core/src/histogram.cc:53:38: error: no member named 'numeric_limits' in namespace 'std'
                              ? std::numeric_limits<double>::infinity()
                                ~~~~~^
/home/yangkeao/Project/github.com/YangKeao/tiflash/contrib/prometheus-cpp/core/src/histogram.cc:53:59: error: expected '(' for function-style cast or type construction
                              ? std::numeric_limits<double>::infinity()
                                                    ~~~~~~^
/home/yangkeao/Project/github.com/YangKeao/tiflash/contrib/prometheus-cpp/core/src/histogram.cc:53:62: error: no member named 'infinity' in the global namespace
                              ? std::numeric_limits<double>::infinity()
```

This has been fixed in https://github.com/jupp0r/prometheus-cpp/pull/442,  Jan 5, 2021. The current version of prometheus-cpp in tiflash is [ca1f34](https://github.com/jupp0r/prometheus-cpp/tree/ca1f3463e74d957d1cccddd4a1a29e3e5d34bd83) in Oct 25, 2019. 

This problem only occurs with libstdc++ >= 11 (as described in [fix](https://github.com/jupp0r/prometheus-cpp/pull/442).

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

Issue Number: close #5278 #2103

Problem Summary:

The prometheus-cpp in tiflash is out of date, and will cause compile-error in new version of libstdc++

Upgrade the prometheus-cpp to v1.0.1. Modify the cmake files according to the repo.

### Check List

Tests 

- [x] No code

### Release Note

```release-note
none
```

### DIFF
--- a/contrib/prometheus-cpp-cmake/core/CMakeLists.txt
+++ b/contrib/prometheus-cpp-cmake/core/CMakeLists.txt
@@ -23,6 +23,12 @@ target_link_libraries(core
     $<$<AND:$<BOOL:UNIX>,$<NOT:$<BOOL:APPLE>>>:rt>
 )
 
+if(HAVE_CXX_LIBATOMIC)
+  # the exported library config must use libatomic unconditionally
+  # (the HAVE_CXX_LIBATOMIC variable should not leak into the target config)
+  target_link_libraries(core PUBLIC atomic)
+endif()
+
 target_include_directories(core
   PUBLIC
     $<BUILD_INTERFACE:${PROMETHEUS_SRC_DIR}/core/include>
--- a/contrib/prometheus-cpp-cmake/core/CMakeLists.txt
+++ b/contrib/prometheus-cpp-cmake/core/CMakeLists.txt
@@ -23,12 +23,6 @@ target_link_libraries(core
     $<$<AND:$<BOOL:UNIX>,$<NOT:$<BOOL:APPLE>>>:rt>
 )
 
-if(HAVE_CXX_LIBATOMIC)
-  # the exported library config must use libatomic unconditionally
-  # (the HAVE_CXX_LIBATOMIC variable should not leak into the target config)
-  target_link_libraries(core PUBLIC atomic)
-endif()
-
 target_include_directories(core
   PUBLIC
     $<BUILD_INTERFACE:${PROMETHEUS_SRC_DIR}/core/include>
--- a/contrib/prometheus-cpp-cmake/pull/CMakeLists.txt
+++ b/contrib/prometheus-cpp-cmake/pull/CMakeLists.txt
@@ -12,9 +12,18 @@ if(ENABLE_COMPRESSION)
 endif()
 
 add_library(pull
+  ${PROMETHEUS_SRC_DIR}/pull/src/basic_auth.cc
+  ${PROMETHEUS_SRC_DIR}/pull/src/basic_auth.h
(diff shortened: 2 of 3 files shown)


fuzhe1989 (approved):

contrib/prometheus-cpp-cmake/core/CMakeLists.txt
```diff
@@ -23,6 +23,12 @@ target_link_libraries(core
     $<$<AND:$<BOOL:UNIX>,$<NOT:$<BOOL:APPLE>>>:rt>
 )
 
+if(HAVE_CXX_LIBATOMIC)
```
SchrodingerZhu, line 26: may I ask which symbol is utilized here?

SchrodingerZhu: libatomic is a sublib of libgcc series to dispatch atomic operations. clang however should generate instrinction directly for most atomic operations as long as the hardware feature flags are set. We do not set `-mcx16` yet for tiflash, but that is something generally avaible for `x86-64-v2` target, which is within the assumption of tiflash. So, if that is the problem, I would prefer adding `-mcx16` on `x86-64` platform instead.

image

SchrodingerZhu:
```
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
  # First check if atomics work without the library.
  check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITHOUT_LIB)
  # If not, check if the library exists, and atomics work with it.
  if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
    check_library_exists(atomic __atomic_load_8 "" HAVE_CXX_LIBATOMIC)
    if(HAVE_CXX_LIBATOMIC)
      list(APPEND CMAKE_REQUIRED_LIBRARIES "atomic")
      check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITH_LIB)
      if(NOT HAVE_CXX_ATOMICS_WITH_LIB)
        message(FATAL_ERROR "Host compiler must support 64-bit std::atomic!")
      endif()
    else()
      message(FATAL_ERROR "Host compiler appears to require libatomic for 64-bit operations, but cannot find it.")
    endif()
  endif()
endif()
```

checking into the implementation, this seems to be a compatible flag for special architectures without even 8-byte atomic load. I think we can skip the check here?

YangKeao: I just (carelessly) copy these CMakeLists from the upstream. Nothing will set this variable so it's fine to remove this condition.

The upstream has a another CMakeList, which checked whether the compiler can compile without libatomic. If not, the `HAVE_CXX_LIBATOMIC` will be set. See https://github.com/jupp0r/prometheus-cpp/commit/d0e1056016d4ef1c91317f01e7cce1cb82eb1198

I've removed these lines in the latest commit

SchrodingerZhu (approved):

YangKeao:
:beers:

PR #5286, opened 2022-07-05
Refact MPPTunnel class to encapsulate different tunnel mode

### LINKED ISSUE #5095: Refine error handling/cancel in TiFlash MPP system

yibin87:

Issue Number: close #5095 

Problem Summary:

1. Separate 'Consumer' logic apart from MPPTunnel, which now named TunnelSender, and have three possible modes
2. Local Receiver, EstablishCall now owns TunnelSender instead of MPPTunnel
3. Refact MPPTunnel's status to an enum, and distinguish finished and consumerFinish

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -176,196 +167,229 @@ void MPPTunnelBase<Writer>::write(const mpp::MPPDataPacket & data, bool close_af
...
+        case TunnelSenderMode::SYNC_GRPC:
+            sync_tunnel_sender = std::make_shared<SyncTunnelSender>(mode, send_queue, writer, log, tunnel_id);
+            sync_tunnel_sender->startSendThread();
+            tunnel_sender = sync_tunnel_sender;
+            break;
+        case TunnelSenderMode::ASYNC_GRPC:
+            async_tunnel_sender = std::make_shared<AsyncTunnelSender>(mode, send_queue, writer, log, tunnel_id);
+            tunnel_sender = async_tunnel_sender;
+            RUNTIME_ASSERT(writer != nullptr, log, "Async writer shouldn't be null");
+            RUNTIME_ASSERT(tunnel_sender != nullptr, log, "Tunnel sender shouldn't be null");
```
windtalker, line 212: Why add this assert, line 210 already make sure that `tunnel_sender` is not nullptr

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -106,67 +91,73 @@ MPPTunnelBase<Writer>::~MPPTunnelBase()
...
         {
             if (!reason.empty())
             {
                 try
                 {
                     FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::exception_during_mpp_close_tunnel);
-                    send_queue.push(std::make_shared<mpp::MPPDataPacket>(getPacketWithError(reason)));
-                    if (!is_local && is_async)
-                        writer->tryFlushOne();
+                    send_queue->push(std::make_shared<mpp::MPPDataPacket>(getPacketWithError(reason)));
```
windtalker, line 122: When in `WaitingForSenderFinish` state, I think the `send_queue` is already `finished` or `cancelled`?

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -176,196 +167,229 @@ void MPPTunnelBase<Writer>::write(const mpp::MPPDataPacket & data, bool close_af
...
-    else
-        rest_work();
+    catch (...)
+    {
+        err_msg = getCurrentExceptionMessage(true);
+    }
+    if (!err_msg.empty())
+    {
+        err_msg = fmt::format("{} meet error: {}", tunnel_id, err_msg);
+        LOG_ERROR(log, err_msg);
```
windtalker, line 336: I suggest to trim the stack trace information in `err_msg` after it is logged. Otherwise the error message will be too long and hard to read. Please refer to #5304 for the details

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -176,196 +167,229 @@ void MPPTunnelBase<Writer>::write(const mpp::MPPDataPacket & data, bool close_af
...
+        }
+    }
+    catch (...)
+    {
+        err_msg = getCurrentExceptionMessage(true);
+    }
+    if (!err_msg.empty())
+    {
+        err_msg = fmt::format("{} meet error: {}", tunnel_id, err_msg);
+        LOG_ERROR(log, err_msg);
```
windtalker, line 378: Ditto

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -44,45 +44,164 @@ namespace DB
...
+ * Each MPPTunnel has a Sender to consume data. There're three kinds of senders: sync_remote, local and async_remote.
  * 
- * The protocol between MPPTunnel and consumer:
+ * The protocol between MPPTunnel and Sender:
  * - All data will be pushed into the `send_queue`, including errors.
- * - MPPTunnel may close `send_queue` to notify consumer normally finish.
- * - Consumer may close `send_queue` to notify MPPTunnel that an error occurs.
- * - After `connect` only the consumer can set `finished` to `true`.
- * - Consumer's state is saved in `consumer_state` and be available after consumer finished.
+ * - MPPTunnel may close `send_queue` to notify Sender normally finish.
```
windtalker, line 195: There are two ways to `close` send_queue: `finish` and `cancel`, is there any difference between the two `close`?

dbms/src/Flash/Mpp/MPPTunnel.cpp
yibin87: You're right. I'll remove the check.

dbms/src/Flash/EstablishCall.cpp
```diff
@@ -117,7 +117,7 @@ void EstablishCallData::writeDone(const ::grpc::Status & status)
     state = FINISH;
     if (stopwatch)
     {
-        LOG_FMT_INFO(mpp_tunnel->getLogger(), "connection for {} cost {} ms.", mpp_tunnel->id(), stopwatch->elapsedMilliseconds());
+        LOG_FMT_INFO(async_tunnel_sender->getLogger(), "connection cost {} ms.", stopwatch->elapsedMilliseconds());
```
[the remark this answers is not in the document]
yibin87: In previous version, tunnel_sender doesn't have tunnel_id field, and the Logger itself already has tunnel_id info as prefix, so just remove the tunnel_id. 
Since now add tunnel_id field in tunnel_sender, will add it back.

dbms/src/Flash/Mpp/MPPTunnel.cpp
yibin87: Seems in WaitingForSenderFinish state, MPPTunnel should just ignore the close event.

dbms/src/Flash/Mpp/MPPTunnel.cpp
yibin87: Got it!

yibin87: Got it, nice suggestions!

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Change 'close' to 'finish'.

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -101,122 +229,67 @@ class MPPTunnelBase : private boost::noncopyable
...
-        bool is_local_,
-        bool is_async_,
-        const String & req_id);
+    friend class tests::TestMPPTunnel;
+    enum class TunnelStatus
+    {
+        Unconnected, // Not connect to any writer, not able to accept new data
+        Connected, // Connected to some writer, accepting data
+        WaitingForSenderFinish, // Accepting all data already, wait for sender to finish
+        Finished // Final state, no more work to do
```
fuzhe1989, line 261: How about differentiate `Finished` (finish after connect) and `Cancelled` (close before connect)?

dbms/src/Common/MPMCQueue.h
```diff
@@ -76,7 +76,8 @@ class MPMCQueue
 
     /// Block until:
     /// 1. Pop succeeds with a valid T: return true.
-    /// 2. The queue is cancelled or finished: return false.
+    /// 2. The queue is cancelled: return false.
+    /// 3. The queue is finished: return true if the queue is not empty.
```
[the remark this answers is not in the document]
yibin87: How about make rule1 like: "The queue is normal, then blablabla" to make it more clear.

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Yeah, they are different states, I think they are just not so different, since cancel event may lead to Finished state too:) 
In the other side, it might be more meaningful to differentiate cancel and finished, I'll consider it later.

dbms/src/Flash/EstablishCall.h
```diff
@@ -115,7 +116,7 @@ class EstablishCallData : public PacketWriter
         FINISH
     };
     CallStatus state; // The current serving state.
-    std::shared_ptr<DB::MPPTunnel> mpp_tunnel = nullptr;
+    std::shared_ptr<DB::AsyncTunnelSender> async_tunnel_sender = nullptr;
```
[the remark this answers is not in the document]
yibin87: You're right, I'll remove it.

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -44,45 +44,164 @@ namespace DB
...
+    /// TunnelSender use consumer state to inform tunnel that whether sender has finished its work
+    class ConsumerState
+    {
+    public:
+        ConsumerState()
+            : future(promise.get_future())
+        {
+        }
+
+        // before finished, must be called without protection of mu
```
windtalker, line 104: Looks like this comment is out of date, there is no `mu` in `TunnelSender`

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -44,45 +44,164 @@ namespace DB
...
+        String getMsg()
+        {
+            future.wait();
+            return future.get();
+        }
+        void setMsg(const String & msg)
+        {
+            promise.set_value(msg);
```
windtalker, line 112: Not releated to this pr, but do we need some protections to make sure `promise.set_value` is not called multiple times? Since [here](https://en.cppreference.com/w/cpp/thread/promise/set_value) says "An exception is thrown if there is no shared state or the shared state already stores a value or exception."

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -101,122 +229,68 @@ class MPPTunnelBase : private boost::noncopyable
...
 
-    std::shared_ptr<mpp::MPPDataPacket> readForLocal();
-
     /// close() finishes the tunnel, if the tunnel is connected already, it will
     /// write the error message to the tunnel, otherwise it just close the tunnel
     void close(const String & reason);
 
     // a MPPConn request has arrived. it will build connection by this tunnel;
-    void connect(Writer * writer_);
+    virtual void connect(PacketWriter * writer);
```
windtalker, line 237: Why need to be a virtual function, seems I don't found override version of it

dbms/src/Common/MPMCQueue.h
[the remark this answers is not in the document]
fuzhe1989: It's kind of reverse to me in that I (the user) needn't to know the queue state until the `pop` failed. This means a normal queue and a finished queue seem no difference to me if they are both non-empty.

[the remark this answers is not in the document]
fuzhe1989: It's not a big issue though. It's ok if you think the new desc is more precise to you.

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Good catch, it is a mistake, I'll fix it.

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Currently, the only way to set consumer_state's msg is to call consumerFinish. In consumerFinish, there is the protection, check if msg already set, if so, just do nothing and return.

dbms/src/Common/MPMCQueue.h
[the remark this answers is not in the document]
yibin87: I see, previous comment is a little confusing to me, make me think pop finished queue will always return false.
To make it simple and clear, I'll add a table to describe the behavior for both pop/push.

dbms/src/Flash/Mpp/MPPTunnel.h
windtalker: How about we enhance `setMsg` as 
```
void setMsg(const String & msg)
{
    bool old_value = false;
    if (!msg_has_set.compare_exchange_strong(old_val, true, std::memory_order_seq_cst, std::memory_order_relaxed))
        return;
    promise.set_value(msg);
}
```
And `consumerFinish` don't need to involve lock.

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Yeah, I'll remove it.

dbms/src/Flash/Mpp/MPPTunnel.h
fuzhe1989: The tricky part of this is that, it's an obvious bug if `promise.set_value` is called more than once. Do we need to protect for potential bug and pretend nothing happened, or just leave it throw then we could quickly know there's a bug?

yibin87: Done

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -91,7 +210,16 @@ class MPPTunnelBase : private boost::noncopyable
...
+    // For gtest usage
+    MPPTunnel(
+        const String & tunnel_id_,
+        std::chrono::seconds timeout_,
+        int input_steams_num_,
+        bool is_local_,
+        bool is_async_,
+        const String & req_id);
+
+    virtual ~MPPTunnel();
```
windtalker, line 222: Not need to be virtual here.

windtalker (approved):
Others LGTM

dbms/src/Flash/Mpp/MPPTunnel.h
yibin87: Removed

fuzhe1989 (approved):

PR #5293, opened 2022-07-05
Implement Agg_BitOr function push down.

### LINKED ISSUE #5118: Call For Participation: implement `Agg_BitOr` function push down
## Feature Request

We'd like to support `Agg_BitOr` function push down.

At least one integration test and several unit tests.

RinChanNOWWW:
Signed-off-by: hezheyu <rinchannow@bupt.edu.cn>

Issue Number: close #5118

Problem Summary:

Apply `CAST` function to the bitwise aggregation functions' children exprs to make ordinary Clickhouse functions be able to work.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

### Release note

```release-note
None
```


RinChanNOWWW:
When the rows **are all `NULL`**, TiFlash will throw an exception: "cannot convert null value to non-null type". I don't know how to solve this problem. Please help me to reach the solution.

P.S. If there is at least one non-null row, it will work well.

RinChanNOWWW:
cc @xuhuaiyu

RinChanNOWWW:
> When the rows **are all `NULL`**, TiFlash will throw an exception: "cannot convert null value to non-null type". I don't know how to solve this problem. Please help me to reach the solution.
> 
> P.S. If there is at least one non-null row, it will work well.

Solved with `AggregateFunctionCombinatorNull::transformAggregateFunction`

dbms/src/AggregateFunctions/AggregateFunctionNull.h
```diff
@@ -429,12 +429,19 @@ class AggregateFunctionNullUnary final : public AggregateFunctionNullBase<result
...
         {
             const ColumnNullable * column = static_cast<const ColumnNullable *>(columns[0]);
+            const String & nexted_func_name = this->nested_function->getName();
             if (!column->isNullAt(row_num))
             {
                 this->setFlag(place);
                 const IColumn * nested_column = &column->getNestedColumn();
                 this->nested_function->add(this->nestedPlace(place), &nested_column, row_num, arena);
             }
+            else if (nexted_func_name == "groupBitOr" || nexted_func_name == "groupBitAnd" || nexted_func_name == "groupBitXor")
```
Lloyd-Pottiger, line 439:
```suggestion
            if (!column->isNullAt(row_num))
            {
                this->setFlag(place);
                const IColumn * nested_column = &column->getNestedColumn();
                this->nested_function->add(this->nestedPlace(place), &nested_column, row_num, arena);
            }
            else if (const String & nexted_func_name = this->nested_function->getName(); nexted_func_name == "groupBitOr" || nexted_func_name == "groupBitAnd" || nexted_func_name == "groupBitXor")
```
in this way looks better?

RinChanNOWWW: Fine.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -158,6 +158,34 @@ void DAGExpressionAnalyzer::fillArgumentDetail(
...
+        {
+            name = appendCast(uint64_type, actions, name);
+        }
+    }
+
+    arg_names.push_back(name);
+    arg_types.push_back(orig_type->isNullable() ? makeNullable(uint64_type) : uint64_type);
+    arg_collators.push_back(removeNullable(arg_types.back())->isString() ? getCollatorFromExpr(arg) : nullptr);
+}
+
```
Willendless, line 188: Can we separate the logic of appending cast and filling argument details?

RinChanNOWWW: You mean making two independent methods? But I think the logic is strongly related with current `fillArgumentDetailForAggFuncBitwise`, and cannot be used by others.

Lloyd-Pottiger (approved):
LGTM

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
Willendless: Sure! then maybe add some comments to indicate this function appends cast to expr actions and the corresponding reasons?

RinChanNOWWW: Ok.

tests/fullstack-test/expr/agg_bit_or_pushdown.test
```diff
@@ -0,0 +1,76 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+mysql> drop table if exists test.all_null
+mysql> create table test.all_null (a int, b int)
+mysql> alter table test.all_null set tiflash replica 1
+mysql> insert into test.all_null values(1, NULL), (1, NULL), (2, NULL), (2, NULL), (3, NULL);
+
+mysql> drop table if exists test.not_all_null
+mysql> create table test.not_all_null (a int, b int)
```
ywqzzy, line 21: Can the code support decimal type now？

RinChanNOWWW: Yes, and string types are also supported. It depends on whether `CAST` function can support the conversion or not.

dbms/src/AggregateFunctions/AggregateFunctionNull.h
```diff
@@ -467,6 +474,13 @@ class AggregateFunctionNullUnary final : public AggregateFunctionNullBase<result
                 arena,
                 if_argument_pos);
 
+            if (const String & nested_func_name = this->nested_function->getName(); nested_func_name == "groupBitOr" || nested_func_name == "groupBitAnd" || nested_func_name == "groupBitXor")
```
SeaRise, line 477: Can we use `if constexpr (result_is_nullable)`?

dbms/src/AggregateFunctions/AggregateFunctionNull.h
```diff
@@ -429,12 +429,19 @@ class AggregateFunctionNullUnary final : public AggregateFunctionNullBase<result
...
         {
             const ColumnNullable * column = static_cast<const ColumnNullable *>(columns[0]);
+
             if (!column->isNullAt(row_num))
             {
                 this->setFlag(place);
                 const IColumn * nested_column = &column->getNestedColumn();
                 this->nested_function->add(this->nestedPlace(place), &nested_column, row_num, arena);
             }
+            else if (const String & nested_func_name = this->nested_function->getName(); nested_func_name == "groupBitOr" || nested_func_name == "groupBitAnd" || nested_func_name == "groupBitXor")
```
SeaRise, line 439: Can we use `if constexpr (result_is_nullable)`?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -158,6 +158,36 @@ void DAGExpressionAnalyzer::fillArgumentDetail(
...
+    Names & arg_names,
+    DataTypes & arg_types,
+    TiDB::TiDBCollators & arg_collators)
+{
+    DataTypePtr uint64_type = std::make_shared<DataTypeUInt64>();
+    const Block & sample_block = actions->getSampleBlock();
+    String name = getActions(arg, actions);
+    DataTypePtr orig_type = sample_block.getByName(name).type;
+    // Bitwise operations can only be applied to integer types.
+    // So we need to cast the argument to UInt64.
```
SeaRise, line 173: why we need to add cast here?
Didn't tidb add a cast here?

RinChanNOWWW: No. We compute the result in TiFlash, so we should cast in TiFlash.

dbms/src/AggregateFunctions/AggregateFunctionNull.h
RinChanNOWWW: I don't know if it is right for non-bitwise functions.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
Willendless: It seems tidb does type cast during operator execution phase.
https://github.com/pingcap/tidb/blob/bce6901705cde161451c0bd707a07ecfac0c7578/expression/aggregation/bit_or.go#L37-L56. This behavior is kind of similar to non-agg bitwise function. I think we might have two choice

1. also do type cast during operator execution phase instead of appending cast beforehand
2. do the same as non-agg bitwise function, i.e. appending cast in advance

https://github.com/pingcap/tiflash/blob/bcb029eaa1e0bc616b279dfc37232e823f987d76/dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp#L326-L357

However, I think current comment is not accurate. TiDB always pushes integer types to tiflash. The reason why we append a cast, or do type cast in execution phase, is because mysql (prior to 8.0) handle only unsigned 64-bit integer argument and result values.

ref: https://dev.mysql.com/doc/refman/8.0/en/bit-functions.html

PR #5309, opened 2022-07-07
*: fix data race occurred in SharedQueryBlockInputStream.

ywqzzy:

Issue Number: ref #5302 

Problem Summary:
In restoreConcurrency(), it assigned multiple SharedQueryBlockInputStream to pipeline.
SharedQueryBlockInputStream enable multiple threads read from one stream, so the BlockStreamProfileInfo info may be accessed by multiple threads.

Change the read function in dbms/src/DataStreams/IProfilingBlockInputStream.h to identify streams which need to be shared with threads.
Add fine grained locks to protect the BlockStreamProfileInfo info.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ywqzzy:
The tsan result is here. https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-sanitizer/detail/tiflash-ghpr-sanitizer/19/pipeline/8/

ywqzzy:
Currently fix is not graceful, I will find a way to make the code clear.

dbms/src/DataStreams/IProfilingBlockInputStream.cpp
```diff
@@ -120,8 +131,7 @@ void IProfilingBlockInputStream::readPrefix()
     auto start_time = info.total_stopwatch.elapsed();
     readPrefixImpl();
 
-    forEachChild([&] (IBlockInputStream & child)
-    {
+    forEachChild([&](IBlockInputStream & child) {
```
SeaRise, line 134: should we add lock for `info.updateExecutionTime(info.total_stopwatch.elapsed() - start_time);`?

dbms/src/Flash/Coprocessor/DAGQueryBlockInterpreter.cpp
```diff
@@ -400,8 +400,6 @@ void DAGQueryBlockInterpreter::executeAggregation(
         pipeline.streams_with_non_joined_data.clear();
         pipeline.firstStream() = std::move(stream);
 
-        // should record for agg before restore concurrency. See #3804.
-        recordProfileStreams(pipeline, query_block.aggregation_name);
```
SeaRise, line 404: I think it should be kept here.
Because even in this pr, `ProfileInfo.ExecuteTime` of sharedQuery is still wrong.

SeaRise: May be we can modify the logic of `recordProfileStreams` to fix the https://github.com/pingcap/tiflash/issues/5314.
Such as `recordProfileStream(stream, executor_id, concurrency);`.

dbms/src/Common/UniqueLockGuard.h
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#pragma once
+
+#include <mutex>
+
+template <typename Mutex>
+struct UniqueLockGuard
```
gengliqi, line 19: Seems not necessary? Since there is a `get_lock` function to wrap the logic of acquiring the lock. `auto lock = get_lock()` is simple enough.

gengliqi (commented):
Although the cost of acquiring a lock in a single thread is very small due to `futex`, no overhead is always better than some. Most of `IProfilingBlockInputStream` don't need this lock so I think it's better to add a bool template parameter for `IProfilingBlockInputStream` to reduce the meaningless overhead.

dbms/src/Common/UniqueLockGuard.h
ywqzzy: Currently I use get_lock to try to reduce the overhead in single thread scene

ywqzzy: > Seems not necessary? Since there is a `get_lock` function to wrap the logic of acquiring the lock. `auto lock = get_lock()` is already simple enough.

I will remove the UniqueLockGuard

dbms/src/Flash/Coprocessor/DAGQueryBlockInterpreter.cpp
SeaRise: We can fix in https://github.com/pingcap/tiflash/pull/5367

dbms/src/Common/UniqueLockGuard.h
gengliqi: > Currently I use get_lock to try to reduce the overhead in single thread scene

Got it. LGTM.

SeaRise (dismissed):

dbms/src/DataStreams/IProfilingBlockInputStream.cpp
```diff
@@ -110,36 +116,52 @@ Block IProfilingBlockInputStream::read(FilterPtr & res_filter, bool return_filte
     }
 #endif
 
-    info.updateExecutionTime(info.total_stopwatch.elapsed() - start_time);
+    {
+        auto lock = get_lock();
+        info.updateExecutionTime(info.total_stopwatch.elapsed() - start_time);
+    }
```
gengliqi, line 122: In `progressImpl` function, `info` still may be used without lock protection.

gengliqi (commented):
It seems difficult to make `IProfilingBlockInputStream::read` thread-safe...

gengliqi:
Can we change the code here https://github.com/pingcap/tiflash/blob/fe2b539863e564acd0c0c74a556418a2cb85858f/dbms/src/Flash/Coprocessor/InterpreterUtils.cpp#L38 with different `SharedQueryBlockInputStream`s? And thes ' SharedQueryBlockInputStream`s share the data only in `SharedQueryBlockInputStream` like `queue`, `read_prefixed`, `read_suffixed`, `thread_manager`. The data from `IProfilingBlockInputStream` don't need to be shared.

dbms/src/DataStreams/IProfilingBlockInputStream.cpp
```diff
@@ -51,21 +50,25 @@ Block IProfilingBlockInputStream::read(FilterPtr & res_filter, bool return_filte
...
+        }
+        start_time = info.total_stopwatch.elapsed();
 
-    if (!checkTimeLimit())
-        limit_exceeded_need_break = true;
+        if (!checkTimeLimit())
+            limit_exceeded_need_break = true;
+    }
 
     if (!limit_exceeded_need_break)
```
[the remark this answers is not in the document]
ywqzzy: I have protected it, but IMO the code is hard to read.

dbms/src/DataStreams/IProfilingBlockInputStream.cpp
ywqzzy: ditto

ywqzzy:
> Can we change the code here
> 
> https://github.com/pingcap/tiflash/blob/fe2b539863e564acd0c0c74a556418a2cb85858f/dbms/src/Flash/Coprocessor/InterpreterUtils.cpp#L38
> 
> with different `SharedQueryBlockInputStream`s?
> And these `SharedQueryBlockInputStream`s share the data only from `SharedQueryBlockInputStream` like `queue`, `read_prefixed`, `read_suffixed`, `thread_manager`. The data from `IProfilingBlockInputStream` don't need to be shared.

Yes, I have discussed with searise before, but I think it will be too many code for this bug fix. Maybe it's a better way to solve it.

dbms/src/DataStreams/IProfilingBlockInputStream.cpp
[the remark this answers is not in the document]
gengliqi: Really? I see that `limit_exceeded_need_break` is not an atomic variable and is read without a lock. You can see https://stackoverflow.com/questions/14624776/can-a-bool-read-write-operation-be-not-atomic-on-x86 for more information.

ywqzzy:
> > Can we change the code here
> > https://github.com/pingcap/tiflash/blob/fe2b539863e564acd0c0c74a556418a2cb85858f/dbms/src/Flash/Coprocessor/InterpreterUtils.cpp#L38
> > 
> > with different `SharedQueryBlockInputStream`s?
> > And these `SharedQueryBlockInputStream`s share the data only from `SharedQueryBlockInputStream` like `queue`, `read_prefixed`, `read_suffixed`, `thread_manager`. The data from `IProfilingBlockInputStream` don't need to be shared.
> 
> Yes, I have discussed with searise before, but I think it will be too many code for this bug fix. Maybe it's a better way to solve it.

I will close it

PR #5358, opened 2022-07-12
fix: incorrect uptime in grafana panel

### LINKED ISSUE #5355: Incorrect uptime in grafana when tiflash meets supersize tables
## Bug Report

create a lot of tables in a TiDB cluster (maybe more than 100k).

uptime in grafana panel is correct.

uptime is incorrect, keep as 8 minutes as time goes by.
image

v6.1.0

Lloyd-Pottiger:

Issue Number: close #5355 

Problem Summary:

maybe due to the bug of linux kernel which cause monotonic clock may not be monotonic sometimes. It may cause some operations to fail with "Timeout exceeded: elapsed 18446744073.709553 seconds".Takes previously returned value and returns it again if time stepped back for some reason.
ref to:
https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/Stopwatch.h#L19-L21

result:
image
instance 172.16.5.81:51350 is the patch that add this pr. We can found that uptime grows correctly. However, it still maybe blocked by other thread.

### Release note

```release-note
None
```


dbms/src/Common/Stopwatch.h
```diff
@@ -23,21 +24,28 @@
...
-    return nanoseconds(clock_type) / 1000000000ULL;
-}
-} // namespace StopWatchDetail
+    UInt64 current_time = clock_gettime_ns(clock_type);
+    if (likely(prev_time <= current_time))
+        return current_time;
 
+    /// Something probably went completely wrong if time stepped back for more than 1 second.
+    assert(prev_time - current_time <= 1000000000ULL);
+    return prev_time;
```
fuzhe1989, line 47: `return std::max(prev_time, current_time)`?

dbms/src/Common/Stopwatch.h
```diff
@@ -23,21 +24,28 @@
 #include <common/apple_rt.h>
 #endif
 
-
-namespace StopWatchDetail
-{
-inline UInt64 nanoseconds(clockid_t clock_type)
+inline UInt64 clock_gettime_ns(clockid_t clock_type = CLOCK_MONOTONIC)
```
fuzhe1989, line 27: besides, for `StopWatch`, how about use `std::chrono::high_resolution_clock` which provides better compatibility?

dbms/src/Common/Stopwatch.h
Lloyd-Pottiger: good idea!

dbms/src/Common/Stopwatch.h
Lloyd-Pottiger: but seems `clock_gettime` has better performance?

fuzhe1989: sorry, `steady_clock` is what I thought. From [here](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc;h=5539d8cbedd110cd053950e86a1a4c18d4bfcf9f;hb=HEAD#l87) `steady_clock::now` also calls `clock_gettime`.

Lloyd-Pottiger: sometimes we need better performance, we can set `clock_type_ = CLOCK_MONOTONIC`, but `steady_clock` is only `CLOCK_MONOTONIC`?

dbms/src/Common/Stopwatch.h
Lloyd-Pottiger: sorry, I found that the previous version
```
if (likely(prev_time <= current_time))
    return current_time;

/// Something probably went completely wrong if time stepped back for more than 1 second.
assert(prev_time - current_time <= 1000000000ULL);
return prev_time;
```
seems to be more effective than `std::max`.

hongyunyan (approved):

hongyunyan (approved):

ywqzzy (approved):
LGTM

PR #5373, opened 2022-07-14
clear obsolete v2 data for tables that totally transform to ps v3

### LINKED ISSUE #5336: Remove the data of PageStorage V2 after all data has been migrated to V3
## Enhancement
1. Upgrade to V3
2. Run `alter table xxx compact tiflash replica`, all table's data has been migrated to V3
3. Restart TiFlash

There are some data files left on disk. They should be useless and removed after the 3rd step (Restart TiFlash).
image

image
image

lidezhu:

Issue Number: close #5336 

Problem Summary: After transform from all data from ps v2 to ps v3, there are still some v2 data left which may consume unnecessary disk space.

When the valid pages in ps v2 is 0, we write a mark file to indicate that ps v2 is deleted. And try to drop v2 data.
And at later restart, if we find that there is are ps v2 delete mark file, we will not restore ps v2 and may try to clear obsolete v2 data if previous drop process was interrupted.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
1. load data in ps v2 format;
2. edit config to use ps v3 format;
3. run sql `alter table ... compact tiflash replica;` to transform all data from ps v2 to ps v3;
4. restart again to check whether the ps v2 data in the table data path is deleted.

Before restart:
image
After restart and run compact:
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/Page/V2/PageStorage.cpp
+++ b/dbms/src/Storages/Page/V2/PageStorage.cpp
@@ -160,12 +160,14 @@ PageFileSet PageStorage::listAllPageFiles(const FileProviderPtr & file_provider,
 PageStorage::PageStorage(String name,
                          PSDiskDelegatorPtr delegator_, //
                          const Config & config_,
-                         const FileProviderPtr & file_provider_)
+                         const FileProviderPtr & file_provider_,
+                         bool no_more_insert_)
     : DB::PageStorage(name, delegator_, config_, file_provider_)
     , write_files(std::max(1UL, config_.num_write_slots.get()))
     , page_file_log(&Poco::Logger::get("PageFile"))
     , log(&Poco::Logger::get("PageStorage"))
     , versioned_page_entries(storage_name, config.version_set_config, log)
+    , no_more_insert(no_more_insert_)
 {
     // at least 1 write slots
     config.num_write_slots = std::max(1UL, config.num_write_slots.get());
@@ -405,18 +407,20 @@ DB::PageEntry PageStorage::getEntryImpl(NamespaceId /*ns_id*/, PageId page_id, S
 // - Writable, reuse `old_writer` if it is not a nullptr, otherwise, create a new writer from `page_file`
 // - Not writable, renew the `page_file` and its belonging writer.
 //   The <id,level> of the new `page_file` is <max_id + 1, 0> of all `write_files`
+// If `force` is true, we always roll writing_file unless it is empty.
 PageStorage::WriterPtr PageStorage::checkAndRenewWriter( //
     WritingPageFile & writing_file,
     PageFileIdAndLevel max_page_file_id_lvl_hint,
     const String & parent_path_hint,
     PageStorage::WriterPtr && old_writer,
-    const String & logging_msg)
+    const String & logging_msg,
+    bool force)
 {
     WriterPtr write_file_writer;
 
     PageFile & page_file = writing_file.file;
-    bool is_writable = page_file.isValid() && page_file.getType() == PageFile::Type::Formal //
-        && isPageFileSizeFitsWritable(page_file, config);
+    bool is_writable = (!force && page_file.isValid() && page_file.getType() == PageFile::Type::Formal //
+                        && isPageFileSizeFitsWritable(page_file, config));
     if (is_writable)
     {
         if (old_writer)
@@ -451,6 +455,7 @@ PageStorage::WriterPtr PageStorage::checkAndRenewWriter( //
         PageFileIdAndLevel max_writing_id_lvl{max_page_file_id_lvl_hint};
         for (const auto & wf : write_files)
             max_writing_id_lvl = std::max(max_writing_id_lvl, wf.file.fileIdLevel());
+
         delegator->addPageFileUsedSize( //
             PageFileIdAndLevel(max_writing_id_lvl.first + 1, 0),
             0,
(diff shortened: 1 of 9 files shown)


flowbehappy:
It looks like we use two strategies for tables' PS and kvstore's PS, can you clarify each of them? @lidezhu

dbms/src/Storages/Transaction/RegionPersister.cpp
```diff
@@ -276,6 +276,13 @@ RegionMap RegionPersister::restore(const TiFlashRaftProxyHelper * proxy_helper,
         }
         case PageStorageRunMode::MIX_MODE:
         {
+            // V2 data will be transformed to V3, so use config that more easy to gc
+            config.file_roll_size = PAGE_FILE_SMALL_SIZE;
```
flowbehappy, line 280: Since there could be no more pages written into PS v2, I wonder whether it will work to make the current writable file rotate. So that gc can work to free space.

flowbehappy: And, can we also use the same strategy in table's PS v2 in `gc()`?  i.e. If we find that a PS v2 already empty, and in MIX mode, then force it to run a writable file rotation, and then GC will delete old page files.

dbms/src/Storages/Transaction/RegionPersister.cpp
```diff
@@ -313,7 +320,9 @@ RegionMap RegionPersister::restore(const TiFlashRaftProxyHelper * proxy_helper,
             }
             else // no need do transform
             {
-                LOG_FMT_INFO(log, "Current kvstore translate already done before restored.");
+                LOG_FMT_INFO(log, "Current kvstore transform already done before restored.");
+                // running gc on v2 to decrease its disk space usage
+                page_storage_v2->gcImpl(/*not_skip=*/true, nullptr, nullptr);
```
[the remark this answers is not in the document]
lidezhu: Fixed.

lidezhu:
> It looks like we use two strategies for tables' PS and kvstore's PS, can you clarify each of them? @lidezhu

kvstore's ps v2 will just run once at restart, but table's ps v2 will always run if there is some data left in log storage at restart. So we can safely use some aggressive gc config for kvstore, but we must be carful with tables's ps.

dbms/src/Storages/Transaction/RegionPersister.cpp
lidezhu: Added.

flowbehappy:
> > It looks like we use two strategies for tables' PS and kvstore's PS, can you clarify each of them? @lidezhu
> 
> kvstore's ps v2 will just run once at restart, but table's ps v2 will always run if there is some data left in log storage at restart. So we can safely use some aggressive gc config for kvstore, but we must be carful with tables's ps.
> 
> Now ps v2 will rotate writing page file if no valid pages in non writeing page files.

Great! And I think it will be better to put this information in PR comment. @lidezhu

lidezhu:
> > > It looks like we use two strategies for tables' PS and kvstore's PS, can you clarify each of them? @lidezhu
> > 
> > 
> > kvstore's ps v2 will just run once at restart, but table's ps v2 will always run if there is some data left in log storage at restart. So we can safely use some aggressive gc config for kvstore, but we must be carful with tables's ps.
> > Now ps v2 will rotate writing page file if no valid pages in non writeing page files.
> 
> Great! And I think it will be better to put this information in PR comment. @lidezhu

Done.

dbms/src/Storages/Page/V2/PageStorage.cpp
```diff
@@ -1086,13 +1091,42 @@ bool PageStorage::gcImpl(bool not_skip, const WriteLimiterPtr & write_limiter, c
...
+            {
+                std::unique_lock lock(write_mutex);
+                LOG_FMT_DEBUG(log, "{} No valid pages in non writing page file. Try to roll all {} writing page files", storage_name, write_files.size());
+                idle_writers.clear();
+                for (auto & write_file : write_files)
+                {
+                    auto writer = checkAndRenewWriter(write_file, {0, 0}, /*parent_path_hint=*/"", nullptr, "", /*force*/ true);
+                    idle_writers.emplace_back(std::move(writer));
+                }
+            }
```
JaySon-Huang, line 1118: There is still one situation that we can not reclaim the disk usage. Assume that
```
write_files: [
pf_2_0 (disk size == 0),
pf_3_0 (disk_size == 64MB),
pf_4_0 (disk_size == 64MB),
pf_5_0 (disk_size == 64MB)]
```
even we roll pf_3_0, pf_4_0, pf_5_0 into new empty files, but the min_writing_file_id_level is still `<2,0>` and can not be GC-ed.

* * *

Maybe when **any page_file of writing_files is not empty**, run `checkAndRenewWriter` with force == true,
and change L422~423 to be 
```
bool is_writable = (force) || (!force && page_file.isValid() && page_file.getType() == PageFile::Type::Formal //
                                                                   && isPageFileSizeFitsWritable(page_file, config));
```

lidezhu: It's ok here becuase after roll pf_3_0, pf_4_0, pf_5_0, they will be set legacy regardless of the stats of pf_2_0 and reclaim most of their disk usage which should be good enough.
And if we always roll file when force is true, then we may have the risk to keep rolling many empty writing files.

JaySon-Huang: After roll, the `min_writing_file_id_level` is still `<2,0>`, and the pf_3_0, pf_4_0, pf_5_0 won't be set to legacy?
https://github.com/pingcap/tiflash/blob/eb26fc529a2b5ba91837218b142f86fe157c518e/dbms/src/Storages/Page/V2/PageStorage.cpp#L1085-L1087
https://github.com/pingcap/tiflash/blob/eb26fc529a2b5ba91837218b142f86fe157c518e/dbms/src/Storages/Page/V2/PageStorage.cpp#L1336-L1366

JaySon-Huang: > And if we always roll file when force is true, then we may have the risk to keep rolling many empty writing files.

I means force to roll when any of the writing file is empty

lidezhu: Got it. Fixed.

dbms/src/Storages/Page/V2/PageStorage.cpp
```diff
@@ -405,18 +407,20 @@ DB::PageEntry PageStorage::getEntryImpl(NamespaceId /*ns_id*/, PageId page_id, S
 // - Writable, reuse `old_writer` if it is not a nullptr, otherwise, create a new writer from `page_file`
 // - Not writable, renew the `page_file` and its belonging writer.
 //   The <id,level> of the new `page_file` is <max_id + 1, 0> of all `write_files`
+// If `force` is true, we always roll writing_file unless it is empty.
```
JaySon-Huang, line 410:
```suggestion
// If `force` is true, always create new page file for writing.
```

JaySon-Huang (approved):
LGTM

flowbehappy (approved):

Issue #5294, opened 2022-07-06
Improve the performance of new collation releated function/executor


windtalker:
## Enhancement
TiDB enable new collation by default since 6.0, however we found that performance of string column dropped a lot with new collation enabled in TiFlash including:
- collation related function: `strcmp`, `like`
- collation related executor: agg with string column as group by key, join with string column as join key
we need to improve the performance of new collation related function/executor.

- [x] #5299
- [x] #5375
- [x] #5489
- [x] #5545
- [x] #5569
- [x] #5658
- [x] #5834
- [x] #5908

### TPCH

- data tpch-100
- TiFlash x 1
- TIFLASH REPLICA 1 

Time(s) | Original: rollback all PR in pingcap/tiflash#5294 from commit a0f9865 | Optimized |   | Improvement: (Original) / (Optimized) - 1.0
-- | -- | -- | -- | --
Q1 | 11.31 | 10.37 |   | 9.06%
Q2 | 2.52 | 2.32 |   | 8.62%
Q3 | 5.6 | 5.47 |   | 2.38%
Q4 | 6.14 | 6.07 |   | 1.15%
Q6 | 1.98 | 1.98 |   | 0.00%
Q7 | 6.34 | 6.14 |   | 3.26%
Q8 | 9.03 | 8.36 |   | 8.01%
Q9 | 38.42 | 35.8 |   | 7.32%
Q10 | 7.82 | 7.15 |   | 9.37%
Q11 | 1.85 | 1.71 |   | 8.19%
Q12 | 4.4 | 4.19 |   | 5.01%
Q13 | 8.49 | 7.82 |   | 8.57%
Q14 | 2.11 | 2.11 |   | 0.00%
Q15 | 4.46 | 4.46 |   | 0.00%
Q16 | 2.25 | 2.11 |   | 6.64%
Q17 | 13.32 | 12.78 |   | 4.23%
Q18 | 18.09 | 17.41 |   | 3.91%
Q19 | 6.54 | 5.6 |   | 16.79%
Q20 | 2.99 | 2.92 |   | 2.40%
Q21 | 22.11 | 21.51 |   | 2.79%
Q22 | 1.85 | 1.78 |   | 3.93%

### ClickBench

- [ClickBench](https://github.com/ClickHouse/ClickBench/tree/d661b49dffda015415366f0e12aa1ebe3424c311/mysql)
- TiFlash x 1
- TIFLASH REPLICA 1 

[...]

PR #5375, opened 2022-07-15
Optimize string sort for default collation `UTF8MB4_BIN`

### LINKED ISSUE #5366: Wrong behavior when handle `strcmp` with default collation
## Bug Report

```
drop table if exists t;
create table t (a varchar(100)) CHARSET=utf8mb4 COLLATE=utf8mb4_bin;

insert into t values('1   '), ('1\n'), ('1');
select hex(min(a)) from t;
```

MySQL version

```
mysql> select version();
+-----------+
| version() |
+-----------+
| 8.0.29    |
+-----------+
1 row in set (0.00 sec)
```

```
MySQL [test]> select hex(min(a)) from t;
+-------------+
| hex(min(a)) |
+-------------+
| 31202020    |
+-------------+
```

```
MySQL [test]> select hex(min(a)) from t;
+-------------+
| hex(min(a)) |
+-------------+
| 31          |
+-------------+
```

master

### TiDB behavior

For mysql

```
mysql> SET NAMES utf8mb4 COLLATE utf8mb4_bin;
Query OK, 0 rows affected (0.00 sec)

mysql> select strcmp('1\0', '1');
+--------------------+
| strcmp('1\0', '1') |
+--------------------+
|                 -1 |
+--------------------+
1 row in set (0.00 sec)
```

For tidb

```
MySQL [test]> SET NAMES utf8mb4 COLLATE utf8mb4_bin;
Query OK, 0 rows affected (0.00 sec)

MySQL [test]> select strcmp('1\0', '1');
+--------------------+
| strcmp('1\0', '1') |
+--------------------+
|                  1 |
+--------------------+
1 row in set (0.00 sec)
```

MySQL cmp '1\0' and '1':
- get min length of both string is 1;
- '1' and '1' are equal;
- the remain str in '1\0' is '\0', cmp '\0' with whit space ' ';
- cmp '\0' and ' ', same like cmp 0x0 and 0x20, got -1;

---

According to https://docs.pingcap.com/tidb/dev/character-set-and-collation, the tidb choose to remove tail space.
 
### Bug in tiflash

https://github.com/pingcap/tiflash/blob/432c153b3d708f680cdcd28f4fc6e794564130d6/dbms/src/Columns/ColumnString.cpp#L317-L318

tiflash show remove tail '\0'.

solotzg:

Issue Number: close #5366 ref #5294

Problem Summary:

- #5366: the original string contains '\0' at the end. To compare it under tidb collation, it must not be contained.

- devirtualize function call when collation is `UTF8MB4_BIN`, which is the default one.

----

Tail '\0' doesn't be removed in those parts.

https://github.com/pingcap/tiflash/blob/3674439a1f1aeb09c70ad67f70cf5c2e961bd01a/dbms/src/Columns/ColumnString.h#L249
https://github.com/pingcap/tiflash/blob/3674439a1f1aeb09c70ad67f70cf5c2e961bd01a/dbms/src/Columns/ColumnString.h#L270
https://github.com/pingcap/tiflash/blob/432c153b3d708f680cdcd28f4fc6e794564130d6/dbms/src/Columns/ColumnString.cpp#L317-L318

### Benchmark

ENV:
- tpch100
- 1 tiflash
- limit cpu up to 200%
- original commit: 97342db2a1dd705b1bd98f3481d51f48a4f0300d

----

SQL
```
select min(L_SHIPMODE) from lineitem;
```

Time(s) | Original | Optimized | NoCollation |   |  
-- | -- | -- | -- | -- | --
  | 11.35 | 9.98 | 9.81 |   |  
  | 11.33 | 9.98 | 9.88 |   |  
  | 11.23 | 10 | 9.84 |   |  
  | 11.22 | 9.82 | 9.73 | NoCollation : Original | 15.22%
  | 11.58 | 9.95 | 9.96 | Optimized : Original | 14.04%
AVG | 11.342 | 9.946 | 9.844 | NoCollation : Optimized | 1.04%

----

SQL
```
select max(L_SHIPMODE) from lineitem;
```

Time(s) | Original | Optimized | NoCollation |   |  
-- | -- | -- | -- | -- | --
  | 13.56 | 12.62 | 12.77 |   |  
  | 13.74 | 12.51 | 12.27 |   |  
  | 13.35 | 12.61 | 12.32 |   |  
  | 13.63 | 12.63 | 12.45 | NoCollation : Original | 9.13%
  | 13.52 | 12.66 | 12.32 | Optimized : Original | 7.57%
AVG | 13.56 | 12.606 | 12.426 | NoCollation : Optimized | 1.45%

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Transaction/Collator.h
```diff
@@ -91,4 +119,119 @@ extern TiDBCollators dummy_collators;
...
+
+    explicit TiDBCollatorPtrImpl(class ITiDBCollator const * ptr_)
+        : TiDBCollatorID(ptr_ ? ptr_->getCollatorId() : 0)
+        , ptr(ptr_)
+    {
+    }
+
+    ALWAYS_INLINE inline int compare(const char * s1, size_t length1, const char * s2, size_t length2) const
+    {
+        if (likely(canUseFastPath()))
```
windtalker, line 140: Will this `likely` hurt the performance of non-utf8mb4_bin collator?

dbms/src/Storages/Transaction/Collator.h
```diff
@@ -91,4 +119,119 @@ extern TiDBCollators dummy_collators;
...
+        }
+        else
+        {
+            return compareIndirect(s1, length1, s2, length2);
+        }
+    }
+
+    ALWAYS_INLINE inline int compareIndirect(const char * s1, size_t length1, const char * s2, size_t length2) const
+    {
+        return ptr->compare(s1, length1, s2, length2);
```
windtalker, line 152: Don't need to check if `ptr` is nullptr?

solotzg: No need to check. Just like previous way, the caller must guarantee the pointer `class ITiDBCollator const *` is not null. `TiDBCollatorPtrImpl` is a wrap for raw poiner.

dbms/src/Storages/Transaction/Collator.h
solotzg: I think it will not hurt performance because the cost of local branch-jump is lower than virtual call.
`likely` is a **hint** for compiler to put branch closer to current instrument.

for example, with likely, maybe:
```
mov %eax,%esi // result = canUseFastPath()
test  %eax,%eax
je    ..... // jump to the branch DO NOT use fast path
.... // fast path branch
```

dbms/src/Storages/Transaction/Collator.h
solotzg: rollback those modifications.

dbms/src/Columns/ColumnString.cpp
```diff
@@ -321,54 +323,94 @@ int ColumnString::compareAtWithCollationImpl(size_t n, size_t m, const IColumn &
...
+    // optimize path for default collator `UTF8MB4_BIN`
+    if (TiDB::ITiDBCollator::getCollator(TiDB::ITiDBCollator::UTF8MB4_BIN) == &collator)
     {
-        if (reverse)
-            std::sort(res.begin(), res.end(), lessWithCollation<false>(*this, collator));
-        else
-            std::sort(res.begin(), res.end(), lessWithCollation<true>(*this, collator));
+        Utf8MB4BinCmp cmp_impl;
+        PermutationWithCollationUtils::getPermutationWithCollationImpl(*this, cmp_impl, reverse, limit, res);
+        ///
```
windtalker, line 408: useless comment here?

dbms/src/Columns/ColumnString.cpp
```diff
@@ -379,35 +421,73 @@ void ColumnString::updateWeakHash32(WeakHash32 & hash, const TiDB::TiDBCollatorP
...
+                auto sort_key = collator->sortKey(view.data(), view.size(), sort_key_container);
+                *hash_data = ::updateWeakHash32(reinterpret_cast<const UInt8 *>(sort_key.data), sort_key.size, *hash_data);
+                ++hash_data;
+            });
+        }
+    }
+    else
+    {
+        // Skip last zero byte.
+        LoopOneColumn(chars, offsets, offsets.size(), [&](const std::string_view & view, size_t) {
```
windtalker, line 450: Both `updateHashWithValues` and `updateHashWithValue` don't remove the last zero byte if `collator == nullptr`, I think we should also keep the last zero byte here, otherwise, it may make the others confused.

windtalker (approved):
LGTM

dbms/src/Columns/ColumnString.cpp
solotzg: No change here. https://github.com/pingcap/tiflash/blob/99d9eb0646c077cbb186e8b8b31ae96f705bff5f/dbms/src/Columns/ColumnString.cpp#L484-L491.

The web of github show mistake.

zanmato1984 (approved):
LGTM

PR #5377, opened 2022-07-15
Read ThreadPool and Data Sharing of DeltaTree

### LINKED ISSUE #5376: Read Thread Pool and DataSharing
## Feature Request

*TODO*

JinheLin:

Issue Number: close #5376

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
```diff
@@ -0,0 +1,69 @@
+#include <Poco/DirectoryIterator.h>
+#include <Poco/File.h>
+#include <Storages/DeltaMerge/ReadThread/CPU.h>
+
+#include <string>
+
+namespace DB::DM
+{
+// In Linux a numa node is represented by a device directory, such as '/sys/devices/system/node/node0', ''/sys/devices/system/node/node01'.
```
JaySon-Huang, line 9:
```suggestion
// In Linux a numa node is represented by a device directory, such as '/sys/devices/system/node/node0', '/sys/devices/system/node/node01'.
```

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
```diff
@@ -0,0 +1,69 @@
...
+static inline bool isNodeDir(const std::string & name)
+{
+    return name.substr(0, 4) == "node" && std::all_of(name.begin() + 4, name.end(), [](unsigned char c) { return std::isdigit(c); });
```
JaySon-Huang, line 12: do we need to ensure name.length() > 4?

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Under a numa node directory is CPU cores and memory, such as  '/sys/devices/system/node/node0/cpu0' and '/sys/devices/system/node/node0/memory0'.
+static inline bool isCPU(const std::string & name)
+{
+    return name.substr(0, 3) == "cpu" && std::all_of(name.begin() + 3, name.end(), [](unsigned char c) { return std::isdigit(c); });
```
JaySon-Huang, line 18: ensure name.length() > 3?

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1203,43 +1232,69 @@ BlockInputStreams DeltaMergeStore::read(const Context & db_context,
...
-    auto read_task_pool = std::make_shared<SegmentReadTaskPool>(std::move(tasks));
+    auto read_task_pool = std::make_shared<SegmentReadTaskPool>(
+        physical_table_id,
+        dm_context,
+        columns_to_read,
+        filter,
+        max_version,
+        expected_block_size,
+        false,
+        db_settings.dt_raw_filter_range,
```
flowbehappy, line 1257: Since that info are both used by  `SegmentReadTaskPool` and  `DMSegmentThreadInputStream`, how about grouping that info into a struct and passing around?

dbms/src/Storages/DeltaMerge/ReadThread/SegmentReader.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+void SegmentReaderPoolManager::addTask(MergedTaskPtr && task)
+{
+    static std::hash<uint64_t> hash_func;
+    auto idx = hash_func(task->getSegmentId()) % reader_pools.size();
+    reader_pools[idx]->addTask(std::move(task));
+}
+
+bool SegmentReaderPoolManager::isSegmentReader() const
```
flowbehappy, line 209: Please add some documents about this method.

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
flowbehappy: I think this needs better protection. How about using regex?

JinheLin: I think check `name.size() > 4` is simple and enought.
 
regex can slove this problem but I think it is not necessarily.

dbms/src/Storages/DeltaMerge/tests/gtest_circular_scan_list.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    for (uint64_t i = 0; i < 10; i++)
+    {
+        lst.add(std::make_shared<Node>(i));
+    }
+
+    {
+        auto [unvalid, valid] = lst.count(0);
+        ASSERT_EQ(unvalid, 10);
+        ASSERT_EQ(valid, 0);
```
JaySon-Huang, line 60: The return value name should be `auto [valid, unvalid] = lst.count(0)`? ditto for later vars in this test
https://github.com/pingcap/tiflash/blob/2760de874b46abe4e9701995097f7bfa353230f7/dbms/src/Storages/DeltaMerge/ReadThread/CircularScanList.h#L58

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
JinheLin: I think can use `SegmentReadTaskPool` to create inputsteam of segment for `DMSegmentThreadInputStream` and `DMSegmentThreadInputStream` is not necessarily to store these infomation.

I will refine the related code if I have time these two days or maybe later...

dbms/src/Storages/DeltaMerge/ReadThread/ColumnSharingCache.h
```diff
@@ -0,0 +1,199 @@
...
+            {
+                break;
+            }
+        }
+    }
+
+private:
+    std::mutex mtx;
+    // start_pack_id -> <pack_count, col_data>
+    std::map<size_t, ColumnData> packs;
```
flowbehappy, line 90:
```suggestion
    std::map<PackId, ColumnData> packs;
```

dbms/src/Storages/DeltaMerge/ReadThread/ColumnSharingCache.h
```diff
@@ -0,0 +1,199 @@
...
+};
+
+class ColumnSharingCacheMap
```
flowbehappy, line 93: Please add some documents to explain the strategy of this cache. Like why calling del by upper_start_pack_id after get.

dbms/src/Storages/DeltaMerge/ReadThread/ColumnSharingCache.h
```diff
@@ -0,0 +1,199 @@
...
+            addColumn(cd.id);
+        }
+    }
+
+    ~ColumnSharingCacheMap()
+    {
+        LOG_FMT_DEBUG(log, "dmfile {} stat {}", dmfile_name, statString());
+    }
+
+    void addStale()
```
flowbehappy, line 112: What does this method actually do?

flowbehappy: It seems to do nothing, only update a metric.

dbms/src/Storages/DeltaMerge/ReadThread/ColumnSharingCache.h
```diff
@@ -0,0 +1,199 @@
...
+        ColumnCacheStatus status;
+        std::lock_guard lock(mtx);
+        auto target = packs.find(start_pack_id);
+        if (target == packs.end())
+        {
+            status = ColumnCacheStatus::GET_MISS;
+        }
+        else if (target->second.pack_count < pack_count)
+        {
+            status = ColumnCacheStatus::GET_PART;
```
flowbehappy, line 54: I think there is space to optimize this part further, i.e. even if part of the cache that exists can still be helpful.
But we can do it later.

flowbehappy: And to reduce the chance of step mismatch of the split of packs, we can try to align the split point. For example, we always read packs by 4 steps.  

Thread 1:         2, 3 | 4, 5, 6, 7 | 8, 9
Thread 2: 0, 1, 2, 3 | 4, 5, 6, 7 | 8,

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
```diff
@@ -0,0 +1,112 @@
...
+    catch (std::exception & e)
+    {
+        LOG_FMT_WARNING(log, "{}", e.what());
+    }
+    catch (...)
+    {
+        LOG_FMT_WARNING(log, "Unknow Error");
+    }
+
+    std::vector<std::vector<int>> numa_nodes(1); // "One numa node"
```
flowbehappy, line 109: Maybe print a warning about choosing a fallback value?

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include <exception>
+#include <string>
+
+namespace DB::DM
+{
+// In Linux a numa node is represented by a device directory, such as '/sys/devices/system/node/node0', '/sys/devices/system/node/node01'.
+static inline bool isNodeDir(const std::string & name)
+{
+    return name.size() > 4 && name.substr(0, 4) == "node" && std::all_of(name.begin() + 4, name.end(), [](unsigned char c) { return std::isdigit(c); });
```
flowbehappy, line 16: I think using a regex is better. And we definitely need to do an exception protection caused by illegal `name`.

dbms/src/Storages/DeltaMerge/ReadThread/ColumnSharingCache.h
JinheLin: `addStale` just do some statistics.

dbms/src/Storages/DeltaMerge/ReadThread/CPU.cpp
JinheLin: These functions restrict the name to `node[digits]` and `cpu[digits]`. I think this is more clarify and strict. 
No illegal string can cause these functions throw a exception after we check the size of the string.

The advantage of regex is flexible but is not required here.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1203,43 +1232,69 @@ BlockInputStreams DeltaMergeStore::read(const Context & db_context,
 {
     // Use the id from MPP/Coprocessor level as tracing_id
     auto dm_context = newDMContext(db_context, db_settings, tracing_id);
-
-    SegmentReadTasks tasks = getReadTasksByRanges(*dm_context, sorted_ranges, num_streams, read_segments);
+    // TODO(jinhelin): If keep order is required, disable read thread.
+    auto enable_read_thread = db_context.getSettingsRef().dt_enable_read_thread;
+    SegmentReadTasks tasks = getReadTasksByRanges(*dm_context, sorted_ranges, num_streams, read_segments, !enable_read_thread);
```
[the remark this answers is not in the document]
JinheLin: Added.

dbms/src/Storages/DeltaMerge/ReadThread/MergedTask.h
```diff
@@ -0,0 +1,143 @@
...
+        if (!isStreamFinished(i))
+        {
+            streams[i] = nullptr;
+            pools[i] = nullptr;
+            finished_count++;
+        }
+    }
+
+    uint64_t seg_id;
+    SegmentReadTaskPools pools;
```
flowbehappy, line 113: `SegmentReadTaskPools` is really misleading here. Can we give it a better alias name?

flowbehappy: And `pools`.

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.cpp
```diff
@@ -95,4 +95,170 @@ SegmentReadTasks SegmentReadTask::trySplitReadTasks(const SegmentReadTasks & tas
     return result_tasks;
 }
 
+BlockInputStreamPtr SegmentReadTaskPool::getInputStream(uint64_t seg_id, SegmentReadTaskPtr & t)
```
flowbehappy, line 98: What is the relationship of `t` with `this->tasks`?

flowbehappy: This method looks very weird to me.

dbms/src/Storages/DeltaMerge/tests/gtest_circular_scan_list.cpp
```diff
@@ -0,0 +1,136 @@
...
+    l.next()->setInvalid();
+
+    ASSERT_EQ(l.next(), nullptr);
+    ASSERT_EQ(l.next(), nullptr);
+    l.add(std::make_shared<Node>(2));
+
+    ASSERT_EQ(l.next()->poolId(), 2);
+    ASSERT_EQ(l.next()->poolId(), 2);
+}
+} // namespace DB::DM::tests
```
flowbehappy, line 136: I think you will need to add more test cases.

flowbehappy: About other components of this PR.

dbms/src/Storages/DeltaMerge/ReadThread/MergedTask.h
```diff
@@ -0,0 +1,143 @@
...
+    std::vector<SegmentReadTaskPtr> tasks;
+    BlockInputStreams streams;
```
flowbehappy, line 115: Please document those vars, they are really confusing.

flowbehappy: And  use better names, like `tasks` can be `sharing_tasks`, `streams` -> "sharing_streams"

flowbehappy: And since pool, task, and stream are closely relative, how about putting them into a struct? Like 
```
class SharedUnit
{
SegmentReadTaskPtr pool; // Store the information of a read thread
SegmentReadTaskPtr task; // The specific task of the pool which is shared
BlockInputStreamPtr stream; // The stream created from the task
}

```

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.cpp
JinheLin: `t` is poped from `this-tasks` in `SegmentReadTaskScheduler::scheduleMergedTask`:
First, `scheduleSegmentUnlock` choose a segment to read.
Third, call `SegmentReadTaskPool::getTask` to pop task and build a `MergedTask`.

`MergedTask` includes `SegmentReadTaskPool`s and `SegmentReadTask`s.
Read threads use `SegmentReadTaskPool` and `SegmentReadTask` to build a segment inputstream before first read.

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
```diff
@@ -534,5 +540,52 @@ void DMFileReader::readFromDisk(
...
+                              size_t pack_count,
+                              size_t read_rows,
+                              size_t skip_packs,
+                              [[maybe_unused]] bool force_seek)
+{
+    if (!getCachedPacks(column_define.id, start_pack_id, pack_count, read_rows, column))
+    {
+        auto data_type = dmfile->getColumnStat(column_define.id).type;
+        auto col = data_type->createColumn();
+        readFromDisk(column_define, col, start_pack_id, read_rows, skip_packs, true /*force_seek*/);
```
JaySon-Huang, line 555: Why here always set `force_seek` to true?

JinheLin: If block is read from cache, we don't need to read from file and the file offset is not changed.

We can trace whether read from cache of the previous read to optimize this case.

dbms/src/Flash/Coprocessor/TiDBTableScan.cpp
```diff
@@ -24,6 +24,7 @@ TiDBTableScan::TiDBTableScan(
     , executor_id(executor_id_)
     , is_partition_table_scan(table_scan->tp() == tipb::TypePartitionTableScan)
     , columns(is_partition_table_scan ? table_scan->partition_table_scan().columns() : table_scan->tbl_scan().columns())
+    , keep_order(table_scan->tbl_scan().has_keep_order() ? table_scan->tbl_scan().keep_order() : true)
```
windtalker, line 27: If `is_partition_table_scan` is true, keep_order can be set to false since partition table scan will not require keep order.

dbms/src/Flash/Coprocessor/TiDBTableScan.cpp
```diff
@@ -24,6 +24,7 @@ TiDBTableScan::TiDBTableScan(
     , executor_id(executor_id_)
     , is_partition_table_scan(table_scan->tp() == tipb::TypePartitionTableScan)
     , columns(is_partition_table_scan ? table_scan->partition_table_scan().columns() : table_scan->tbl_scan().columns())
+    , keep_order(table_scan->tbl_scan().keep_order())
```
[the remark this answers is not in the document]
JinheLin: The default value of `bool` type of protobuf is false.
For safety, the default `keep_order` of TiFlash should be true, so I fix it as below:

`keep_order(table_scan->tbl_scan().has_keep_order() ? table_scan->tbl_scan().keep_order() : true)`

dbms/src/Flash/Coprocessor/TiDBTableScan.cpp
JinheLin: Does the table scan executor of partition table will set `keep_order` to be `false`?

windtalker: For patition table scan, you should get it by `table_scan->partition_table_scan()` and there is no `keep_order` field in `partition_table_scan`

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.cpp
flowbehappy: I think we should reconsider the code design. Maybe later PRs. Currently, it looks OK to me.

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
flowbehappy: I think we should not always set `force_seek = true`. It is a regression.

dbms/src/Storages/DeltaMerge/ReadThread/SegmentReadTaskScheduler.h
```diff
@@ -0,0 +1,86 @@
...
+    {
+        static SegmentReadTaskScheduler scheduler;
+        return scheduler;
+    }
+
+    ~SegmentReadTaskScheduler();
+    SegmentReadTaskScheduler(const SegmentReadTaskScheduler &) = delete;
+    SegmentReadTaskScheduler & operator=(const SegmentReadTaskScheduler &) = delete;
+    SegmentReadTaskScheduler(SegmentReadTaskScheduler &&) = delete;
+    SegmentReadTaskScheduler & operator=(SegmentReadTaskScheduler &&) = delete;
```
JaySon-Huang, line 47:
```suggestion
    DISALLOW_COPY_AND_MOVE(SegmentReadTaskScheduler);
```

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
JinheLin: Add `std::unordered_map<ColId, bool> last_read_from_cache;` to trace whether last read is from cache to avoid seek every time.

dbms/src/Flash/Coprocessor/TiDBTableScan.cpp
JinheLin: https://github.com/pingcap/tiflash/pull/5377/commits/93c92bf2a50df874beb6d66769e9db539bbda35c

@windtalker From better understanding, I have refined the code as below. PTAL.
```cpp
    if (is_partition_table_scan || (table_scan->tbl_scan().has_keep_order() && table_scan->tbl_scan().keep_order() == false))
    {
        keep_order = false;
    }
    else
    {
        keep_order = true;
    }
```

dbms/src/Interpreters/Settings.h
```diff
@@ -317,6 +317,9 @@ struct Settings
                                                                                                                                                                                                                                         \
     M(SettingDouble, dt_storage_blob_heavy_gc_valid_rate, 0.2, "Max valid rate of deciding a blob can be compact")                                                                                                                      \
     M(SettingDouble, dt_storage_blob_block_alignment_bytes, 0, "Blob IO alignment size")                                                                                                                                                \
+    M(SettingBool, dt_enable_read_thread, true, "Enable storage read thread or not")                                                                                                                                                          \
```
JaySon-Huang, line 320: Set it default to be false as it is experimental

flowbehappy:
LGTM

flowbehappy (approved):

PR #5413, opened 2022-07-19
fix read from ColumnFileBigReader

lidezhu:

Issue Number: ref #5252 

Problem Summary: Currently `ColumnFileBigReader::readNextBlock` doesn't work normally when just read two columns, because we never call `readNextBlock` when we just need read pk and version column.
But after support fast mode, there is a chance that the previous scenario happens. So we must fix it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang:
> But after support fast mode, there is a chance that the previous scenario happens

To be more specific, is it mean that under normal mode, at least we read 3 columns `<pk, tag, version>`. But for fast mode, we may only read `<pk, tag>`. And this bug will make the `select count(*) from t` return the wrong result?

JaySon-Huang (approved):
LGTM

hongyunyan:
LGTM

lidezhu:
> > But after support fast mode, there is a chance that the previous scenario happens
> 
> To be more specific, is it mean that under normal mode, at least we read 3 columns `<pk, tag, version>`. But for fast mode, we may only read `<pk, tag>`. And this bug will make the `select count(*) from t` return the wrong result?

Yes

hongyunyan (approved):

PR #5434, opened 2022-07-21
*: improve RUNTIME_CHECK and RUNTIME_ASSERT

### LINKED ISSUE #5444: Better `RUNTIME_CHECK` and `RUNTIME_ASSERT`
### `RUNTIME_CHECK`

Current `RUNTIME_CHECK` usage is:

```
RUNTIME_CHECK(a != b, ExceptionType, "{} != {}", a, b);
```

It's not clear for users to know which exception will be thrown. A better way is to explicit write code for generating an exception object like:

```
RUNTIME_CHECK(a != b, ExceptionType("{} != {}", a, b));
```

And the new style could enable better compilation checks.

### `RUNTIME_ASSERT`

Current `RUNTIME_ASSERT` always needs a `logger` and make it hard for replacing existent `assert`s. We should let it work without user provided logger (in which case a default logger is used).

fuzhe1989:
Signed-off-by: fuzhe1989 <fuzhe1989@gmail.com>

Issue Number: close #5444 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


fuzhe1989:
any update? @JaySon-Huang @SeaRise @gengliqi

dbms/src/Common/Exception.h
```diff
@@ -172,28 +162,123 @@ std::enable_if_t<std::is_pointer_v<T>, T> exception_cast(std::exception_ptr e)
...
+{
+    return Poco::Message(
+        logger_name,
+        generateFormattedMessage(condition, std::forward<Args>(args)...),
+        Poco::Message::PRIO_FATAL,
+        filename,
+        lineno);
+}
+
+Poco::Logger * getDefaultFatalLogger();
```
gengliqi, line 187: How about returning `LoggerPtr` then 
```cpp
inline void log(const char * filename, int lineno, const char * condition, Poco::Logger * logger)
```
and 
```cpp
inline void log(const char * filename, int lineno, const char * condition, Poco::Logger * logger, const char * fmt_str, Args &&... args)
```
can be removed.

fuzhe1989: hmmm ... looks like we don't need to support `Poco::Loger *` at the first revision?

gengliqi: I guess so...

fuzhe1989: done

dbms/src/Common/Exception.h
```diff
@@ -172,28 +162,98 @@ std::enable_if_t<std::is_pointer_v<T>, T> exception_cast(std::exception_ptr e)
...
+    }
+}
+
+inline void log(const char * filename, int lineno, const char * condition)
+{
+    log(filename, lineno, condition, getDefaultFatalLogger());
+}
+
+template <typename... Args>
+inline void log(const char * filename, int lineno, const char * condition, const LoggerPtr & logger, const char * fmt_str, Args &&... args)
```
gengliqi, line 204: How about removing `const char * fmt_str`? 
If so, `inline void log(const char * filename, int lineno, const char * condition, const LoggerPtr & logger)` can be removed too.

dbms/src/Common/Exception.h
```diff
@@ -172,28 +162,98 @@ std::enable_if_t<std::is_pointer_v<T>, T> exception_cast(std::exception_ptr e)
...
+            lineno,
+            condition,
+            fmt_str,
+            std::forward<Args>(args)...);
+        logger->log(message);
+    }
+}
+
+template <typename... Args>
+inline void log(const char * filename, int lineno, const char * condition, const char * fmt_str, Args &&... args)
```
gengliqi, line 220: ditto.

JaySon-Huang (approved):
LGTM

dbms/src/Common/Exception.h
fuzhe1989: done

dbms/src/Common/Exception.h
fuzhe1989: this `fmt_str` is used for overloadding from `const LoggerPtr &` or this version will always be chosen even when there's a `const LoggerPtr &`.

gengliqi (approved):

PR #5435, opened 2022-07-21
test: Simplify comparing the stream read result

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/5452

Problem Summary:
When comparing the result read from an inputstream, we write lots of redunt codes to locate the column, summing the num of rows, ensuring readPrefix/readSuffix, etc.
That is verbose and easy to forget.

Introduce some macro:
`ASSERT_INPUTSTREAM_COLS_UR` to compare the columns read from an inputstream (only compare a part of columns)
`ASSERT_INPUTSTREAM_BLOCK_UR` to compare the block read from an inputstream (exactly the same columns in the result block)
`ASSERT_INPUTSTREAM_NROWS` to compare the number of rows from an inputstream

Other change:
* Add `Block::dumpJsonStructure` for debugging output

Example:

Before this PR (27 lines)
https://github.com/pingcap/tiflash/blob/3ad743b58629fcb356701135b4dd2333632fc1da/dbms/src/Storages/DeltaMerge/tests/gtest_dm_delta_merge_store.cpp#L1903-L1930
After this PR (8 lines)
https://github.com/pingcap/tiflash/blob/0e77f859634d9442ad874acee259c8d802578162/dbms/src/Storages/DeltaMerge/tests/gtest_dm_delta_merge_store.cpp#L1552-L1559

Before this PR (22 lines):
https://github.com/pingcap/tiflash/blob/3ad743b58629fcb356701135b4dd2333632fc1da/dbms/src/Storages/DeltaMerge/tests/gtest_column_filter.cpp#L88-L112
After this PR (7 lines):
https://github.com/pingcap/tiflash/blob/0e77f859634d9442ad874acee259c8d802578162/dbms/src/Storages/DeltaMerge/tests/gtest_column_filter.cpp#L90-L96

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JinheLin (approved):

dbms/src/Storages/DeltaMerge/tests/DMTestEnv.h
```diff
@@ -96,6 +98,18 @@ inline std::vector<Int64> createSignedNumbers(size_t beg, size_t end)
     return values;
 }
 
+// Mock a common_pk_col that composed by number `rowkey_column_size` of int64 value
+inline String genMockCommonHandle(Int64 value, size_t rowkey_column_size)
+{
+    WriteBufferFromOwnString ss;
+    for (size_t index = 0; index < rowkey_column_size; index++)
```
Lloyd-Pottiger, line 105:
```suggestion
    for (size_t index = 0; index < rowkey_column_size; ++index)
```

dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
```diff
@@ -29,6 +29,8 @@
 #include <ext/scope_guard.h>
 #include <memory>
 
+#include "TestUtils/InputStreamTestUtils.h"
```
Lloyd-Pottiger, line 32:
```suggestion
#include <TestUtils/InputStreamTestUtils.h>
```

dbms/src/Storages/DeltaMerge/tests/DMTestEnv.h
```diff
@@ -285,13 +299,7 @@ class DMTestEnv
...
             {
                 Int64 value = reversed ? end - 1 - i : beg + i;
-                WriteBufferFromOwnString ss;
-                for (size_t index = 0; index < rowkey_column_size; index++)
-                {
-                    ::DB::EncodeUInt(static_cast<UInt8>(TiDB::CodecFlagInt), ss);
-                    ::DB::EncodeInt64(value, ss);
-                }
-                values.emplace_back(ss.releaseStr());
+                values.emplace_back(genMockCommonHandle(value, rowkey_column_size));
```
Lloyd-Pottiger, line 302: there is one question makes me confused:
in this case, any difference among `values.emplace_back(genMockCommonHandle(value, rowkey_column_size));`,  `values.emplace_back(std::move(genMockCommonHandle(value, rowkey_column_size)));` and `values.push_back(genMockCommonHandle(value, rowkey_column_size));`? Which one is the most effective?

JinheLin: There are two questions:
1. `push_back` vs `emplace_back`.
2. `genMockCommonHandle(value, rowkey_column_size)` vs `std::move(genMockCommonHandle(value, rowkey_column_size))`.

The first question, you can refer the cpp reference: [push_back](https://en.cppreference.com/w/cpp/container/vector/push_back) and [emplace_back](https://en.cppreference.com/w/cpp/container/vector/emplace_back). In short, `emplace_back` use placement-new to construct a object of the container. Strictly, `emplace_back` can be more effective than `push_back`.

The answer of the second question is it depends on the return value of the function. In the case of `genMockCommonHandle`, there is no different. 

Because `genMockCommonHandle` returns a unnamed temporary variable(`std::string`) and it is a rvalue and it is moveable (Don't need a `std::move`).

Lloyd-Pottiger: thank you, the answer make it clear~

dbms/src/Storages/DeltaMerge/tests/DMTestEnv.h
JaySon-Huang: done

dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
JaySon-Huang: done

Lloyd-Pottiger (approved):
LGTM

Issue #5368, opened 2022-07-14
Build fail on MacBook M1


zanmato1984:
## General Question

It's a brand new MacBook M1 I just obtained, with barely nothing installed.

Clang version:
```
Apple clang version 13.0.0 (clang-1300.0.27.3)
Target: arm64-apple-darwin21.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

CMake version:
```
cmake version 3.23.2
```

Build tiflash with:
```
cmake .. -DCMAKE_BUILD_TYPE=DEBUG -DENABLE_TESTS=ON
make -j
```
Two errors occurred:
```
Undefined symbols for architecture arm64:
  "_GetAarch64FeaturesEnumName", referenced from:
      _AddFlags in list_cpu_features.c.o
  "_GetAarch64FeaturesEnumValue", referenced from:
      _AddFlags in list_cpu_features.c.o
  "_GetAarch64Info", referenced from:
      _CreateTree in list_cpu_features.c.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [contrib/cpu_features/list_cpu_features] Error 1
make[1]: *** [contrib/cpu_features/CMakeFiles/list_cpu_features.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
```
error: failed to run custom build command for `kvproto v0.0.2 (https://github.com/pingcap/kvproto.git#0e2f26c0)`

Caused by:
  process didn't exit successfully: `/Users/zanmato/dev/tiflash/contrib/tiflash-proxy/target/debug/build/kvproto-53533e5785490f67/build-script-build` (exit status: 101)
  --- stdout
  `protoc` not in PATH, try using the bundled protoc

  --- stderr
  thread 'main' panicked at 'No suitable `protoc` (>= 3.1.0) found in PATH', /Users/zanmato/.cargo/registry/src/github.com-1ecc6299db9ec823/protobuf-build-0.13.0/src/protobuf_impl.rs:35:14
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
make[5]: *** [build] Error 101
make[4]: *** [debug] Error 2
make[3]: *** [release] Error 2
make[2]: *** [../contrib/tiflash-proxy/target/release/libtiflash_proxy.dylib] Error 2
[...]

PR #5451, opened 2022-07-22
*: try not to use lld as default linker in macos

solotzg:

Issue Number: ref #5368

Problem Summary:

- if there is llvm toolchain installed in macos and lld can be found in $PATH, then such lld will be used as defualt linker.
- the official clang from apple is NOT compatible with other llvm env so that build process will be broken.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


zanmato1984 (approved):
LGTM

Lloyd-Pottiger (approved):

PR #5470, opened 2022-07-26
feat: do del column clean read optimization for fast mode

hongyunyan:

Issue Number: ref #5252 

Problem Summary: For the query in fast mode, we always read the del_mark column, and then do filter with del_mark = 1. While for the packs that all rows' del_mark = 0, we don't need to read their del_mark column, and also don't need to do the filter with del_mark. Thus we add these optimization to cut down IO.

1. fix the min-max index calculation for del column.
2. add the is_delete in PackStat to store the delete rows in a pack.
3. use is_delete to make read optimization on del column in fast mode.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

TODO ：performance test 

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Functions/FunctionsString.cpp
+++ b/dbms/src/Functions/FunctionsString.cpp
@@ -4752,7 +4752,7 @@ public:
 
     DataTypePtr getReturnTypeImpl(const DataTypes & arguments) const override
     {
-        auto first_argument = arguments[0];
+        const auto& first_argument = arguments[0];
         if (!first_argument->isNumber() && !first_argument->isDecimal())
             throw Exception(
                 fmt::format("Illegal type {} of first argument of function {}", first_argument->getName(), getName()),
@@ -5285,7 +5285,7 @@ private:
         {
             UInt64 number = col->getUInt(i);
 
-            int print_size = sprintf(reinterpret_cast<char *>(&res_chars[prev_res_offset]), "%lX", number);
+            int print_size = sprintf(reinterpret_cast<char *>(&res_chars[prev_res_offset]), "%llX", number);
             res_chars[prev_res_offset + print_size] = 0;
             // Add the size of printed string and a tailing zero
             prev_res_offset += print_size + 1;
--- a/dbms/src/Server/DTTool/DTToolMigrate.cpp
+++ b/dbms/src/Server/DTTool/DTToolMigrate.cpp
@@ -235,7 +235,7 @@ int migrateServiceMain(DB::Context & context, const MigrateArgs & args)
             if (!args.dry_mode)
                 output_stream.write(
                     block,
-                    {stat_iter->not_clean, properties_iter->num_rows(), properties_iter->gc_hint_version()});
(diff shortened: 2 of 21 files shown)


breezewish:
This PR changes the PackStat. How is its compatibility when there are old data written in previous versions?

lidezhu:
> This PR changes the PackStat. How is its compatibility when there are old data written in previous versions?

Yes, it need to be added to `PackProperties` instead.

dbms/src/Storages/DeltaMerge/tests/MultiSegmentTestUtil.h
```diff
@@ -106,21 +106,21 @@ class MultiSegmentTestUtil : private boost::noncopyable
         }
         {
             // Check there is only one segment
-            ASSERT_EQ(store->segments.size(), 1);
+            ASSERT_EQ(store->segments.size(), static_cast<unsigned long>(1));
```
JaySon-Huang, line 109: try remove the `#include <gtest/gtest.h>` line (which is already included in `TiFlashTestBasic.h` with ignoring the sign-compare), and you can get rid of writing these casts

hongyunyan:
> > This PR changes the PackStat. How is its compatibility when there are old data written in previous versions?
> 
> Yes, it need to be added to `PackProperties` instead.

I change the field into PackProperties, and make the compatibility test with manually upgrade test.

lidezhu (commented):
LGTM. I will approve it after benchmark related code are removed.

dbms/src/Storages/DeltaMerge/tests/bench_del_optimization.cpp
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    return res;
+}
+
+Block createBlock(size_t rows, size_t columns, size_t delete_rows)
+{
+    Block block;
+    // 单独插入 handle/version/tag 列
```
JaySon-Huang, line 151:
```suggestion
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1042,7 +1042,12 @@ void DeltaMergeStore::mergeDeltaAll(const Context & context)
 
     for (auto & segment : all_segments)
     {
-        segmentMergeDelta(*dm_context, segment, TaskRunThread::Foreground);
+        auto res = segmentMergeDelta(*dm_context, segment, TaskRunThread::Foreground);
+        // just for the bench test
+        while (res == nullptr)
+        {
+            res = segmentMergeDelta(*dm_context, segment, TaskRunThread::Foreground);
+        }
```
JaySon-Huang, line 1050: just curious, there should be no concurrent task for the same segment in a bench test, why do we need this while loop?

JaySon-Huang (commented):
Rest LGTM

lidezhu (approved):
LGTM

JaySon-Huang (approved):
lgtm

PR #5520, opened 2022-08-02
fix wrong path for BlobStat under multi-disks env

### LINKED ISSUE #5519: The disk usage of blobfiles is not correct for PageStorageV3
## Bug Report

Deploy a cluster with 6.2

The valid rate is low in this cluster, much disk space are wasted

origin_img_v2_e777ef59-6d38-4706-80f1-ef8ffab9668g
origin_img_v2_6d3cd9f6-f78d-4f22-950c-2274db052f9g

https://github.com/pingcap/tiflash/commit/bbc5fb7fb551f2c7be115d54d9a95c2d2155fa05 @ release-6.2 branch

lidezhu:

Issue Number: close #5519 

Problem Summary: When creating BlobFile, it will call `delegator->choosePath()` two times when creating BlobStat and creating BlobFile. So the path of the BlobFile may not the same as BlobStat when tiflash is deployed with multiple disks. And when the BlobFile is removed, the associated BlobStat cannot be removed correctly which cause the related metric wrong.

Call `delegator->choosePath()` once when creating BlobStat and just use the path info in BlobStats when creating BlobFile later.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


flowbehappy (approved):

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -1353,10 +1352,9 @@ BlobStatPtr BlobStore::BlobStats::createBigPageStatNotChecking(BlobFileId blob_f
     return stat;
 }
 
-void BlobStore::BlobStats::eraseStat(const BlobStatPtr && stat, const std::lock_guard<std::mutex> &)
+void BlobStore::BlobStats::eraseStat(const BlobStatPtr && stat, const std::lock_guard<std::mutex> & lock)
 {
-    PageFileIdAndLevel id_lvl{stat->id, 0};
-    stats_map[delegator->getPageFilePath(id_lvl)].remove(stat);
```
flowbehappy, line 1359: I wonder why doesn't this method return correct path

lidezhu: This is also correct when run tiflash as a service. But for `BlobStoreStatsTest` in gtest_blob_store.cpp, it doesn's create any BlobFile, so it doesn't have any info in delegator, and the unit test cannot run.

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -1477,6 +1475,29 @@ BlobStatPtr BlobStore::BlobStats::blobIdToStat(BlobFileId file_id, bool ignore_n
...
+            if (stat->id == file_id)
+            {
+                return path;
+            }
+        }
+    }
+    throw Exception(fmt::format("Can't find BlobStat with [blob_id={}]",
+                                file_id),
+                    ErrorCodes::LOGICAL_ERROR);
+}
```
JaySon-Huang, line 1499: In BlobStore::removePosFromStats, we remove the stat from `blob_stats` first, after that call `getBlobFile` -> `getBlobFileParentPath` -> `blobIdToPath`. Exception will be thrown under this case?

https://github.com/pingcap/tiflash/blob/a2d7f16b24e0e5434c9c994be51d7542041512b0/dbms/src/Storages/Page/V3/BlobStore.cpp#L506-L515

lidezhu: Yes, we need remove BlobFile first.

JaySon-Huang (approved):
LGTM

Issue #5576, opened 2022-08-09
Logical split is incompatible with PageStorage V3


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Deploy tiflash with `profiles.default.dt_enable_logical_split = 1`.
The settings is default to be `0`, only happen when user explicitly set the settings (maybe upgrade from a lower version)

### 2. What did you expect to see? (Required)

TiFlash runs without exceptions

### 3. What did you see instead (Required)

As a simple example described in https://github.com/pingcap/tiflash/issues/5570. After several logical split, the third time logical split, PageStorage V3 will throw exception.

The exception stack is as below
```
Code: 0. DB::Exception: Trying to add ref to non-exist page [page_id=1000.998] [ori_id=1000.985] [ver=<6,0>] [resolve_id=1000.951] [resolve_ver=<4,0>]:  [type=2] [page_id=1000.998] [ver=<6,0>] [edit_size=2]
Stack trace:
     0x10db56d34	StackTrace::StackTrace() [gtests_dbms+4510706996]
                	dbms/src/Common/StackTrace.cpp:23
     0x10db56d70	StackTrace::StackTrace() [gtests_dbms+4510707056]
                	dbms/src/Common/StackTrace.cpp:22
     0x100da2530	DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) [gtests_dbms+4295009584]
                	dbms/src/Common/Exception.h:32
     0x100da1638	DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) [gtests_dbms+4295005752]
                	dbms/src/Common/Exception.h:34
     0x10c576bfc	DB::PS::V3::PageDirectory::applyRefEditRecord(std::__1::map<DB::UInt128, std::__1::shared_ptr<DB::PS::V3::VersionedPageEntries>, std::__1::less<DB::UInt128>, std::__1::allocator<std::__1::pair<DB::UInt128 const, std::__1::shared_ptr<DB::PS::V3::VersionedPageEntries> > > >&, std::__1::shared_ptr<DB::PS::V3::VersionedPageEntries> const&, DB::PS::V3::PageEntriesEdit::EditRecord const&, DB::PS::V3::PageVersion const&) [gtests_dbms+4487769084]
[...]

Issue #5570, opened 2022-08-09
PageStorage v3 reports Trying to add ref to non-exist page


breezewish:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Operation sequence:

```
// Flush
Put 951

// Segment Merge
Ref 954->951
Del 951

// Segment Logical Split
Ref 972->954
Ref 985->954
Del 954

// Segment Logical Split
Ref 998->985   <-- Crash
Ref ...
Del ...
```

Unit test code to reproduce:

```c++
TEST_F(PageStorageTest, MyTest)
try
{
    {
        PageStorage::Config config;
        config.blob_heavy_gc_valid_rate = 1.0; /// always run full gc
        page_storage = reopenWithConfig(config);
    }

    const size_t buf_sz = 1024;
    char c_buff[buf_sz];
    for (size_t i = 0; i < buf_sz; ++i)
    {
        c_buff[i] = i % 0xff;
    }

    const UInt64 tag = 12345;
    {
        // Page 951
        WriteBatch log;
        log.putPage(951, tag, std::make_shared<ReadBufferFromMemory>(c_buff, buf_sz), buf_sz, {});
        page_storage->write(std::move(log));
    }

    {
        // Page 954->951
        WriteBatch log;
        log.putRefPage(954, 951);
        page_storage->write(std::move(log));
    }
    {
        // DelPage 951
        WriteBatch removed_log;
        removed_log.delPage(951);
        page_storage->write(std::move(removed_log));
    }

    {
        // Page 972->954
        // Page 985->954
        WriteBatch log;
        log.putRefPage(972, 954);
        log.putRefPage(985, 954);
        page_storage->write(std::move(log));
    }
    {
        // DelPage 954
        WriteBatch removed_log;
        removed_log.delPage(954);
        page_storage->write(std::move(removed_log));
    }

    {
        // Page 998->985
        // Page 1011->985
        WriteBatch log;
        log.putRefPage(998, 985);
        log.putRefPage(1011, 985);
        page_storage->write(std::move(log));
    }

    {
        auto page = page_storage->read(998);
        for (size_t i = 0; i < buf_sz; ++i)
        {
            EXPECT_EQ(*(page.data.begin() + i), static_cast<char>(i % 0xff));
        }
    }
}
CATCH
```

### 2. What did you expect to see? (Required)

Test pass

### 3. What did you see instead (Required)

```
[...]

PR #5577, opened 2022-08-09
Force to disable logical split

JaySon-Huang:
Signed-off-by: JaySon-Huang <tshent@qq.com>

Issue Number: close https://github.com/pingcap/tiflash/issues/5576

Problem Summary:
If user deploys tiflash with explicit setting `profiles.default.dt_enable_logical_split = 1`.
After several logical split, PageStorage V3 will throw an exception for the third time logical split and make TiFlash crash.

Force to disable logical split in the code path

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  - ensure the logical split settings will be forced set to false by checking the log
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DMContext.h
+++ b/dbms/src/Storages/DeltaMerge/DMContext.h
@@ -114,7 +114,7 @@ public:
         , delta_small_column_file_rows(settings.dt_segment_delta_small_column_file_rows)
         , delta_small_column_file_bytes(settings.dt_segment_delta_small_column_file_size)
         , stable_pack_rows(settings.dt_segment_stable_pack_rows)
-        , enable_logical_split(settings.dt_enable_logical_split)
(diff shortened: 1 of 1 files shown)


breezewish:
Maybe we should use `ref issue` instead of `close issue`, so that the issue can be closed when the incompatibility is resolved?

JaySon-Huang:
> Maybe we should use `ref issue` instead of `close issue`, so that the issue can be closed when the incompatibility is resolved?

The issue https://github.com/pingcap/tiflash/issues/5576 is created for incompatibility and triage whether it needs to be resolved on release-6.2. The final incompatibility issue should be resolved as closing the original issue https://github.com/pingcap/tiflash/issues/5570

JaySon-Huang:
close as the issue is resolved by https://github.com/pingcap/tiflash/pull/5612

PR #5597, opened 2022-08-10
refine: shuffle column order before execute function ut

### LINKED ISSUE #5596: enhance ut framework to shuffle column order before execution
## Enhancement

Current `executeFunction` in function ut framwork manually constructs columns with the input order. Therefore, while executing ut, the logical offset and physical offset of each column is the same, which usually is not the case in real environment. 

```cpp
ColumnWithTypeAndName executeFunction(
    Context & context,
    const String & func_name,
    const ColumnsWithTypeAndName & columns,
    const TiDB::TiDBCollatorPtr & collator,
    bool raw_function_test)
{
    ColumnNumbers argument_column_numbers;
    for (size_t i = 0; i < columns.size(); ++i)
        argument_column_numbers.push_back(i);
    return executeFunction(context, func_name, argument_column_numbers, columns, collator, raw_function_test);
}
```

A relevant bug failed to be discovered by ut can be found at #5592.

Therefore, it would be better if we can introduce some randomness to the framework and change the order of input columns before doing execution.

Willendless:

Issue Number: close #5596 

Problem Summary:

Shuffle input columns before execution in function unit tests.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/TestUtils/FunctionTestUtils.cpp
+++ b/dbms/src/TestUtils/FunctionTestUtils.cpp
@@ -311,7 +311,7 @@ ColumnWithTypeAndName executeFunction(
         argument_column_numbers.push_back(i);
     std::shuffle(argument_column_numbers.begin(), argument_column_numbers.end(), g);
 
-    const ColumnsWithTypeAndName columns_reordered = toColumnsReordered(columns, argument_column_numbers);
+    const auto columns_reordered = toColumnsReordered(columns, argument_column_numbers);
 
     return executeFunction(context, func_name, argument_column_numbers, columns_reordered, collator, raw_function_test);
 }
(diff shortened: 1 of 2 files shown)


dbms/src/TestUtils/FunctionTestUtils.cpp
```diff
@@ -293,7 +305,15 @@ ColumnWithTypeAndName executeFunction(
     ColumnNumbers argument_column_numbers;
     for (size_t i = 0; i < columns.size(); ++i)
         argument_column_numbers.push_back(i);
-    return executeFunction(context, func_name, argument_column_numbers, columns, collator, raw_function_test);
+
+    /// shuffle input columns to assure function correctly use physical offsets instead of logical offsets
```
ywqzzy, line 309: explain why we need the shuffle process.

ywqzzy (approved):
LGTM

SeaRise (approved):

PR #5361, opened 2022-07-13
Refine cancelMPPQuery

windtalker:
### What problem does this PR solve?

Issue Number: ref #5095

Problem Summary:

### What is changed and how it works?
1. Do not hold `MPPTask` in `cancelMPPQuery` thread, so `MPPTask` can be destructed parallelly
2. Use ThreadManager instead of raw threads when cancel mpp task

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #5637, opened 2022-08-17
Fix bug that mpp_task is not move to cancel thread due to llvm SOO

### LINKED ISSUE #5638: mpptask is not moved to cancel thread as expected in `MPPTaskManager::cancelMPPQuery`
## Bug Report

ref #5637 for details

windtalker:

Issue Number: ref #5095, close #5638

Problem Summary:

After #5361, in `MPPTaskManager::cancelMPPQuery`, we try to move mpp task to the cancel thread so each mpp task can be destructed parallelly instead of be destructed one by one in `cancelMPPQuery` thread.
However, due to SOO in llvm(https://github.com/llvm/llvm-project/issues/32472), this move actaully failed, and there is a chance that mpp task is destructed inside the follow cancel loop:
https://github.com/pingcap/tiflash/blob/8404e6565e3ff2e47e50497550d838ea76e6e776/dbms/src/Flash/Mpp/MPPTaskManager.cpp#L102-L108

Destruct MPP inside the cancel loop means the task will be destructed before all mpp tasks are cancelled, this may cause some dead lock issues since there are some dependency between mpp tasks(especially in the case of local tunnel).

This pr fix this issue by 
- Use a Functor instead of lambda as a workaround of the SOO issues
- Save the moved Functor inside a vector to make sure mpp task never be destructed inside cancel loop, even if Functor move fails due to some unknown issues.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
run some random failpoint tests
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
+++ b/dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
@@ -125,23 +125,29 @@ void CreatingSetsBlockInputStream::createAll()
         }
         Stopwatch watch;
         auto thread_manager = newThreadManager();
-        for (auto & subqueries_for_sets : subqueries_for_sets_list)
+        try
         {
-            for (auto & elem : subqueries_for_sets)
+            for (auto & subqueries_for_sets : subqueries_for_sets_list)
             {
-                if (elem.second.source) /// There could be prepared in advance Set/Join - no source is specified for them.
+                for (auto & elem : subqueries_for_sets)
                 {
-                    if (is_cancelled || is_killed)
+                    if (elem.second.source) /// There could be prepared in advance Set/Join - no source is specified for them.
                     {
-                        thread_manager->wait();
-                        if (is_killed)
-                            throw Exception("Query was cancelled", ErrorCodes::QUERY_WAS_CANCELLED);
-                        return;
+                        if (isCancelledOrThrowIfKilled())
+                        {
+                            thread_manager->wait();
+                            return;
+                        }
+                        thread_manager->schedule(true, "CreatingSets", [this, &item = elem.second] { createOne(item); });
+                        FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::exception_in_creating_set_input_stream);
                     }
-                    thread_manager->schedule(true, "CreatingSets", [this, &item = elem.second] { createOne(item); });
-                    FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::exception_in_creating_set_input_stream);
                 }
             }
+        }
+        catch (...)
+        {
+            thread_manager->wait();
+            throw;
         }
 
         thread_manager->wait();
--- a/dbms/src/Flash/Mpp/MPPTaskManager.cpp
+++ b/dbms/src/Flash/Mpp/MPPTaskManager.cpp
@@ -128,21 +128,29 @@ void MPPTaskManager::cancelMPPQuery(UInt64 query_id, const String & reason)
     std::vector<std::function<void()>> cancel_functors;
     // TODO: cancel tasks in order rather than issuing so many threads to cancel tasks
     auto thread_manager = newThreadManager();
-    for (auto it = task_set->task_map.begin(); it != task_set->task_map.end();)
+    try
     {
-        fmt_buf.fmtAppend("{} ", it->first.toString());
-        auto current_task = it->second;
-        it = task_set->task_map.erase(it);
-        // Note it is not acceptable to destruct `current_task` inside the loop, because destruct a mpp task before all
-        // other mpp tasks are cancelled may cause some deadlock issues, so `current_task` has to be moved to cancel thread.
-        // At first, we use std::move to move `current_task` to lambda like this:
-        // thread_manager->schedule(false, "CancelMPPTask", [task = std::move(current_task), &reason] { task->cancel(reason); });
-        // However, due to SOO in llvm(https://github.com/llvm/llvm-project/issues/32472), there is still a copy of `current_task`
-        // remaining in the current scope, as a workaround we add a wrap(MPPTaskCancelFunctor) here to make sure `current_task`
-        // can be moved to cancel thread. Meanwhile, also save the moved wrap in a vector to guarantee that even if cancel functors
-        // fail to move due to some other issues, it still does not destruct inside the loop
-        cancel_functors.push_back(MPPTaskCancelFunctor(std::move(current_task), reason));
-        thread_manager->schedule(false, "CancelMPPTask", std::move(cancel_functors[cancel_functors.size() - 1]));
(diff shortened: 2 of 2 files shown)


gengliqi (approved):

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -70,6 +70,34 @@ MPPTaskPtr MPPTaskManager::findTaskWithTimeout(const mpp::TaskMeta & meta, std::
     return it->second;
 }
 
+static Poco::Logger * getLogger()
+{
+    static Poco::Logger * logger = &Poco::Logger::get("MPPTaskCancelFunctor");
+    return logger;
```
SeaRise, line 76: use `Logger::get("MPPTaskCancelFunctor", task->getId()->toString())`?

windtalker: I think we don't need to log task id here since it is not task specific, if move failed, all task's move will fail.

SeaRise: ok.
replace by `Logger::get("MPPTaskCancelFunctor)`?
Poco::Logger is deprecated in TiFlash.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -97,14 +125,24 @@ void MPPTaskManager::cancelMPPQuery(UInt64 query_id, const String & reason)
...
+        // Note it is not acceptable to destruct `current_task` inside the loop, because destruct a mpp task before all
+        // other mpp tasks are cancelled may cause some deadlock issues, so `current_task` has to be moved to cancel thread.
+        // At first, we use std::move to move `current_task` to lambda like this:
+        // thread_manager->schedule(false, "CancelMPPTask", [task = std::move(current_task), &reason] { task->cancel(reason); });
+        // However, due to SOO in llvm(https://github.com/llvm/llvm-project/issues/32472), there is still a copy of `current_task`
+        // remaining in the current scope, as a workaround we add a wrap(MPPTaskCancelFunctor) here to make sure `current_task`
+        // can be moved to cancel thread. Meanwhile, also save the moved wrap in a vector to guarantee that even if cancel functors
+        // fail to move due to some other issues, it still does not destruct inside the loop
+        cancel_functors.push_back(MPPTaskCancelFunctor(std::move(current_task), reason));
+        thread_manager->schedule(false, "CancelMPPTask", std::move(cancel_functors[cancel_functors.size() - 1]));
```
yibin87, line 145: Maybe, take a consideration if any exception thrown during the for loop.

yibin87 (approved):
Others LGTM.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
windtalker: OK

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -70,6 +70,34 @@ MPPTaskPtr MPPTaskManager::findTaskWithTimeout(const mpp::TaskMeta & meta, std::
...
+    static LoggerPtr logger = Logger::get("MPPTaskCancelFunctor");
+    return logger;
+}
+
+class MPPTaskCancelFunctor
+{
+public:
+    MPPTaskPtr task;
+    String reason;
+    MPPTaskCancelFunctor(const MPPTaskCancelFunctor & other)
```
fuzhe1989, line 84: why not declare it as `delete`?

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -97,14 +125,32 @@ void MPPTaskManager::cancelMPPQuery(UInt64 query_id, const String & reason)
     LOG_WARNING(log, fmt::format("Begin cancel query: {}", query_id));
     FmtBuffer fmt_buf;
     fmt_buf.fmtAppend("Remaining task in query {} are: ", query_id);
+    std::vector<std::function<void()>> cancel_functors;
```
fuzhe1989, line 128: guess we don't actually need this `vector`?

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -97,14 +125,32 @@ void MPPTaskManager::cancelMPPQuery(UInt64 query_id, const String & reason)
...
+            it = task_set->task_map.erase(it);
+            // Note it is not acceptable to destruct `current_task` inside the loop, because destruct a mpp task before all
+            // other mpp tasks are cancelled may cause some deadlock issues, so `current_task` has to be moved to cancel thread.
+            // At first, we use std::move to move `current_task` to lambda like this:
+            // thread_manager->schedule(false, "CancelMPPTask", [task = std::move(current_task), &reason] { task->cancel(reason); });
+            // However, due to SOO in llvm(https://github.com/llvm/llvm-project/issues/32472), there is still a copy of `current_task`
+            // remaining in the current scope, as a workaround we add a wrap(MPPTaskCancelFunctor) here to make sure `current_task`
+            // can be moved to cancel thread. Meanwhile, also save the moved wrap in a vector to guarantee that even if cancel functors
+            // fail to move due to some other issues, it still does not destruct inside the loop
+            cancel_functors.push_back(MPPTaskCancelFunctor(std::move(current_task), reason));
```
fuzhe1989, line 146: depend on the behavior of `std::function::move` is risky and hard to understand.

how about this?
```cpp
thread_manager->schedule(false, "CancelMPPTask", 
    [functor = new MPPTaskCancelFunctor(std::move(current_task), reason)] {
        std::unique_ptr(functor)->run(); // do not need `operator()`, just `run()`
});
```

IMO it's more stable than depending on some specific implementation of `std::function`.

dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
```diff
@@ -125,24 +125,30 @@ void CreatingSetsBlockInputStream::createAll()
...
                 }
             }
         }
+        catch (...)
+        {
+            thread_manager->wait();
+            throw;
+        }
 
         thread_manager->wait();
```
fuzhe1989, line 153: It's actually a finally:

```cpp
finally {
    thread_manager->wait();
}
```

dbms/src/Flash/Mpp/MPPTaskManager.cpp
windtalker: Good idea, since this pr is merged, I will use another pr to refine it.

PR #5132, opened 2022-06-13
Some refinements of `mpp_exchange_receiver_map` and `MPPTunnelSet`

windtalker:
Signed-off-by: xufei <xufeixw@mail.ustc.edu.cn>

### What problem does this PR solve?

Issue Number: ref #5095

Problem Summary:

### What is changed and how it works?
* Move `mpp_exchange_receiver_map` from `DAGContext` to `MPPTask`
* Rename `id_to_index_map` to `target_task_id_to_index_map` in `MPPTunnelSet` since the map key is the target task id. So the name is more self-explained.
* Add `registerTunnels` and `initExchangeReceivers` to setup MPPTunnel and ExchangeReceiver in MPPTask
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #5650, opened 2022-08-18
Fix data race in tunnel_set/receiver_set and establish call data

### LINKED ISSUE #5651: Data race in tunnel_set/receiver_set and establish call data
## Bug Report

ref #5650 for details info.

windtalker:

Issue Number: ref #5095, close #5651

Problem Summary:
- Data race in tunnel_set/receiver_set:
In #5132, it try to use the atomic variable `status` to protect tunnel_set/receiver_set in multiple thread read/write(MPPTask::runImpl thread to write and MPPTask::abort thread to read):
https://github.com/pingcap/tiflash/blob/b6536f6af56cc061eec1d65ff82ff79a1c6f81f0/dbms/src/Flash/Mpp/MPPTask.cpp#L145-L147
Unfortunately, it turns out not working becase the value of `status` can be changed after L145 and before L147
- Data race in establish call data:
https://github.com/pingcap/tiflash/blob/b6536f6af56cc061eec1d65ff82ff79a1c6f81f0/dbms/src/Flash/EstablishCall.cpp#L106-L113
After `write(packet)` in L109, the call data is put back to completion queue, and it can be deleted anytime, so there is a chance that before executing L110, the call data itself has been deleted.
- Add lock to protect tunnel_set and receiver_set
- Don't access the member variable after `write(packet)`, since `write(packet)` always return true, just set `err_status` to `OK` before `write(packet)`
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
run random failpoint tests
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Mpp/MPPTunnel.cpp
+++ b/dbms/src/Flash/Mpp/MPPTunnel.cpp
@@ -70,20 +70,7 @@ MPPTunnel::~MPPTunnel()
     });
     try
     {
-        {
-            std::unique_lock lock(*mu);
-            if (status == TunnelStatus::Finished)
-            {
-                LOG_DEBUG(log, "already finished!");
-                return;
-            }
-
-            /// make sure to finish the tunnel after it is connected
-            waitUntilConnectedOrFinished(lock);
-            finishSendQueue();
-        }
-        LOG_FMT_TRACE(log, "waiting consumer finish!");
-        waitForSenderFinish(/*allow_throw=*/false);
+        close("");
     }
     catch (...)
     {
--- a/dbms/src/Flash/Coprocessor/DAGContext.cpp
+++ b/dbms/src/Flash/Coprocessor/DAGContext.cpp
@@ -223,8 +223,12 @@ void DAGContext::addCoprocessorReader(const CoprocessorReaderPtr & coprocessor_r
 {
     if (!isMPPTask())
         return;
-    RUNTIME_ASSERT(mpp_receiver_set != nullptr, log, "MPPTask without receiver set");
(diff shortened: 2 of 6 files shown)


dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -70,20 +70,7 @@ MPPTunnel::~MPPTunnel()
...
-                return;
-            }
-
-            /// make sure to finish the tunnel after it is connected
-            waitUntilConnectedOrFinished(lock);
-            finishSendQueue();
-        }
-        LOG_FMT_TRACE(log, "waiting consumer finish!");
-        waitForSenderFinish(/*allow_throw=*/false);
+        close("");
```
windtalker, line 73: In MPPTask's destruct, it will close all tunnels, so it is safe to make this change.
https://github.com/pingcap/tiflash/blob/b6536f6af56cc061eec1d65ff82ff79a1c6f81f0/dbms/src/Flash/Mpp/MPPTask.cpp#L67-L74

bestwoody: What if one of tunnels throw error in tunnel::close() when closeAllTunnels is called? Is it safe?

windtalker: `MPPTunnel::close()` should not throw error, if it throws error, I think maybe the TiFlash server will be terminated.

bestwoody (approved):
lgtm

yibin87 (approved):
LGTM

PR #5695, opened 2022-08-24
storage: count more as write stalls

### LINKED ISSUE #5694: Count more write stalls
## Enhancement

Currently we may write stall when doing the flush or split the segment (foreground) in `checkSegmentUpdate`.

However we are not counting them as write stalls in the metrics.

breezewish:

Issue Number: close #5694

Problem Summary:

Count foreground split and foreground flush as write stalls. For example, foreground split may be heavily triggered during ingesting. This is how write stall metrics looks like in this PR, when adding TiFlash replica heavily:

image

Now we can clearly see that foreground split contributes to the slowness of adding replicas. We may have more interesting discovery about our performance bottlenecks -- For example, I will checkout whether segment split also affects the write sync performance under heavy write scenario.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -1500,6 +1501,17 @@ void DeltaMergeStore::checkSegmentUpdate(const DMContextPtr & dm_context, const
...
         if (should_foreground_flush)
         {
+            Stopwatch watch;
+            SCOPE_EXIT({
+                // We may be flushing in BG threads. Do not count them as write stalls.
+                if (thread_type == ThreadType::Write)
+                {
+                    // FIXME: We'd better count write stall duration for per-write, instead of per-call,
+                    //   in order to produce a meaningful value.
+                    GET_METRIC(tiflash_storage_write_stall_duration_seconds, type_flush).Observe(watch.elapsedSeconds());
```
flowbehappy, line 1511: How about name it more specifically, i.e. `type_fg_flush`.

breezewish: Operations in background will not be counted as write stalls (for now), so may be `fg` is redundant?

JaySon-Huang (approved):
LGTM

JaySon-Huang (approved):

Issue #5697, opened 2022-08-25
PageStorage gc for external pages my block creating tables


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

In `PageStorageImpl::gcImpl` `clean_external_page`, we hold the `callbacks_mutex` until all external page gc done.
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/Page/V3/PageStorageImpl.cpp#L287-L298

while the time cost in `clean_external_page` is not ignorable and may block new StorageDeltaMerge from being created 
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/Page/V3/PageStorageImpl.cpp#L400-L408
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#L342-L343

### 4. What is your TiFlash version? (Required)
master

JaySon-Huang:
And calling `getAliveExternalPageIds(ns_id)` inside the loop makes it cost more time when the number of tables is large

```
# create 1000 empty tables for testing
for ((i=0;i!=1000;++i)) do; mycli -h 172.16.5.85 -P 8010 --no-warn -D test -e "create table t${i} (c int);"; done
```

get alive external page for one namespace id inside loop
```
[2022/08/25 13:37:26.985 +08:00] [DEBUG] [PageStorageImpl.cpp:330] ["PageStorage:__global__.data GC finished without full gc. [total time=179ms] [dump snapshots=0ms] [gc in mem entries=1ms] [blobstore remove entries=0ms] [blobstore get status=0ms] [get gc entries=0ms] [blobstore full gc=0ms] [gc apply=0ms] [external callbacks=1052] [external gc=178ms] [get alive=92.77ms] [scanner=84.87ms] [remover=0.91ms]"] [thread_id=100]
[...]

PR #5699, opened 2022-08-25
PageStorage: Fine grained lock on external callbacks

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/5697

Problem Summary:
The GC total time in logging does not include the time costs in `clean_external_page`.

In `PageStorageImpl::gcImpl` `clean_external_page`, we hold the `callbacks_mutex` until all external page gc done.
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/Page/V3/PageStorageImpl.cpp#L287-L298

while the time cost in `clean_external_page` is not ignorable and may block new StorageDeltaMerge from being created 
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/Page/V3/PageStorageImpl.cpp#L400-L408
https://github.com/pingcap/tiflash/blob/7acf7f59f300d448d99b79619634b2751c01ff2c/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#L342-L343

* Make the external pages cleaning callbacks safe even after table is dropped
* Use shared_ptr to refine the lock granularity of running external cleaning callbacks, so that creating table won't be block by the whole processing time of `cleanExternalPage`.
* Remove the non-atomic renaming table logic

### Problem not yet resolve in this PR
* https://github.com/pingcap/tiflash/issues/5700

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the bug that page GC may block creating tables
```


dbms/src/Storages/Page/V3/PageStorageImpl.cpp
```diff
@@ -403,14 +495,37 @@ void PageStorageImpl::registerExternalPagesCallbacks(const ExternalPageCallbacks
     assert(callbacks.scanner != nullptr);
     assert(callbacks.remover != nullptr);
     assert(callbacks.ns_id != MAX_NAMESPACE_ID);
-    assert(callbacks_container.count(callbacks.ns_id) == 0);
+    {
+#ifdef NDEBUG
+        // In product env, we assume that NamespaceId(TableID) will not be reuse,
+        // so that creating table don't require to be blocked by `callbacks_remove_mutex`.
```
flowbehappy, line 501: Are there any cases that we might also create, drop, recreate a new DTStore in a real production env?

flowbehappy: I meant although TiDB won't do that, is there any case TiFlash could happen?

dbms/src/Storages/Page/V3/PageStorageImpl.cpp
```diff
@@ -269,13 +279,139 @@ void PageStorageImpl::traverseImpl(const std::function<void(const DB::Page & pag
...
+            std::scoped_lock remove_lock{callbacks_remove_mutex};
+            // If the ns_id is pushed into `remove_queue`, then the `callbacks`
+            // are invalid and may cause failure, break do..while to continue
+            // on next namespace
+            if (callbacks_remove_queue.count(ns_id) > 0)
+                break;
+
+            // `remove_lock` ensure that the `callbacks` won't be invalid
+            const auto & callbacks = callbacks_iter->second;
+            auto pending_external_pages = callbacks.scanner();
```
flowbehappy, line 389: Do we need the protection of `remove_lock` after already having `callbacks`?

JaySon-Huang: The `this` ptr (DeltaMergeStore instance) inside callback will be released and cause crash if we run the 
lambda `scanner`/`remover`.

https://github.com/pingcap/tiflash/pull/5699/files#diff-30b0bd8d7bcbd8a75bdfa77186768765e94209ab4c3dedcac9da2ba54720695aR1272-R1303

flowbehappy: Can optimize this part? e.g. we can cache some information in the callback object, which is not strongly linked to DeltaMergeStore instance. So that the lock scope can be narrowed down to only put and get `callbacks_container` operations, and does not need to protect the `callbacks.scanner()` and `callbacks.remover()`

flowbehappy: And the lock mechanism will be much simpler, without two lock scopes.

dbms/src/Storages/Page/V3/PageStorageImpl.cpp
JaySon-Huang: I think it will not happen unless we execute `drop table` by ch-client.

Or we need to acquire the lock on `callbacks_remove_mutex` that makes creating a table still being blocked by gc-ing DTFiles on "one table"

dbms/src/Storages/Page/V3/PageStorageImpl.cpp
JaySon-Huang: It is not that easy, in the callbacks we need the `StoragePathPool` instance, which the lifetime is the same as `DeltaMergeStore`, to remove useless DTFile 
https://github.com/pingcap/tiflash/blob/master/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#L316-L336

JaySon-Huang: If we want to decouple the callbacks with `DeltaMergeStore`, we may make the `StoragePathPool` to be a shared_ptr, and only save a weak_ptr to the `StoragePathPool` in the callbacks

JaySon-Huang: But even if we decouple the callbacks with `DeltaMergeStore`, we still need the lock `callbacks_remove_mutex`.
Because if thread A drop table that execute `callbacks.erase(ns_id)`, it invalid the iterator of ns_id. If thread B happen to hold the iterator of the same ns_id, thread B will crash.

We don't need that in `DeltaMergeStore.segments` and `PageDirctory.mvcc_table_directory` is that we won't execute `erase` in another thread

lidezhu (dismissed):
LGTM with minor comment

dbms/src/Storages/Page/V3/tests/gtest_page_storage.cpp
```diff
@@ -1223,6 +1227,81 @@ try
...
+        (*ptr) += 1; // mock access the storage inside callback
+        return {};
+    };
+    callbacks.remover = [&ptr](const ExternalPageCallbacks::PathAndIdsVec &, const std::set<PageId> &) -> void {
+        (*ptr) += 1; // mock access the storage inside callback
+    };
+    callbacks.ns_id = TEST_NAMESPACE_ID;
+    page_storage->registerExternalPagesCallbacks(callbacks);
+
+    // Start a segment merge and suspend it before applyMerge
```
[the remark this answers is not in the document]
JaySon-Huang: Fixed

dbms/src/Storages/Page/V3/PageStorageImpl.cpp
flowbehappy: Let's jump out of the details of the current code. The only reason we put a callback in PageStorage is that we want PS to clean up the useless DTFiles.
1. It does not matter that the callback is not put into `callbacks_container` in time, useless DTFiles are not required to be removed immediately.
2. Drop table will clean up the DTFiles of this table itself.

Back to the code, I don't understand why "thread A happen to hold the iterator of the same ns_id, thread A will crash".  As thread A should always hold a shared_ptr.

flowbehappy: The idea is trying to avoid blocking creating / deleting tables by GC, it will slow down the DDL performance.

JaySon-Huang:
Find a corner case: https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/8413/pipeline/300/

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -323,18 +344,17 @@ void DeltaMergeStore::setUpBackgroundTask(const DMContextPtr & dm_context)
                     continue;
 
                 // Note that page_id is useless here.
-                auto dmfile = DMFile::restore(global_context.getFileProvider(), id, /* page_id= */ 0, path, DMFile::ReadMetaMode::none());
+                auto dmfile = DMFile::restore(file_provider, id, /* page_id= */ 0, path, DMFile::ReadMetaMode::none());
                 if (dmfile->canGC())
                 {
                     delegate.removeDTFile(dmfile->fileId());
```
flowbehappy, line 350: If this method is called during the `DeltaMergeStore::drop()`, I think it will rise `Cannot find DMFile` exception.

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -1235,19 +1235,35 @@ void PageDirectory::gcApply(PageEntriesEdit && migrated_edit, const WriteLimiter
...
+                const auto ns_id = holder->high;
+                if (auto valid_ns_iter = valid_external_ids.find(ns_id); valid_ns_iter != valid_external_ids.end())
+                {
+                    // insert the valid external id for this ns_id
+                    valid_ns_iter->second.emplace(holder->low);
+                }
+                else
+                {
+                    // create a new set for this ns_id
+                    std::set<PageId> ns_valid_id;
```
flowbehappy, line 1263:
```suggestion
                    std::set<PageId> ns_valid_ids;
```

dbms/src/Storages/Page/V3/PageStorageImpl.cpp
```diff
@@ -269,13 +279,135 @@ void PageStorageImpl::traverseImpl(const std::function<void(const DB::Page & pag
...
+        //      pages (DTFile).
+        //   2. the `ns_id` is concurrently added after we get `alive_external_ids`, we
+        //      should not remove the external pages on disk.
+        // TODO: Is it ok?
+        if (auto iter = alive_external_ids.find(ns_callbacks->ns_id); iter != alive_external_ids.end())
+        {
+            // remove the external pages that is not alive now.
+            ns_callbacks->remover(pending_external_pages, iter->second);
+        }
+        // Else situation 2 mentioned above, just skip `remover`.
```
JaySon-Huang, line 391: /hold

There is a corner case that these lines lead to removing DTFile being used. We must ensure to execute `scanner`
before `getAliveExternalIds`
And that changes will make this PR too complicated. I will split the changes into another PR first.

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
JaySon-Huang: I'll try a better way to handle it

JaySon-Huang:
Pending for https://github.com/pingcap/tiflash/pull/5739

dbms/src/Storages/StorageDeltaMerge.cpp
```diff
@@ -1198,59 +1198,26 @@ void StorageDeltaMerge::rename(
...
-        global_context,
-        data_path_contains_database_name,
-        new_database_name,
-        new_table_name,
-        tidb_table_info.id,
-        std::move(table_column_defines),
-        std::move(handle_column_define),
-        is_common_handle,
-        rowkey_column_size,
-        settings);
```
JaySon-Huang, line 1253: Use RUNTIME_ASSERT to disable this logic. Cause we never expect running to this branch.
And after this PR, it will remove the callbacks, also cleaning the external ids ptrs from PageStorage. After restoring the DeltaMergeStore from disk, it will run into remove all dtfiles from disk.

JaySon-Huang: Likely, we should not call shutdown for a table but later restore the DeltaMergeStoer instance

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -323,18 +345,32 @@ void DeltaMergeStore::setUpBackgroundTask(const DMContextPtr & dm_context)
...
+                    }
+                    LOG_FMT_INFO(logger,
+                                 "GC try remove useless DM file, but file not found and may have been removed, dmfile={}",
+                                 DMFile::getPathByStatus(path, id, DMFile::Status::READABLE));
+                }
+                else if (dmfile->canGC())
+                {
+                    // scanner should only return dtfiles that can GC,
+                    // just another check here.
                     delegate.removeDTFile(dmfile->fileId());
```
flowbehappy, line 367: Should we also handle the exception here?

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
```diff
@@ -323,18 +345,32 @@ void DeltaMergeStore::setUpBackgroundTask(const DMContextPtr & dm_context)
...
+                auto dmfile = DMFile::restore(file_provider, id, /* page_id= */ 0, path, DMFile::ReadMetaMode::none());
+                if (unlikely(!dmfile))
                 {
+                    try
+                    {
+                        delegate.removeDTFile(id);
+                    }
+                    catch (DB::Exception & e)
+                    {
+                        // just ignore
```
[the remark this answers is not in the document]
JaySon-Huang: Added

dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
JaySon-Huang: Address

dbms/src/Storages/StorageDeltaMerge.cpp
JaySon-Huang: The deleted cases are rewrite as unit test in https://github.com/pingcap/tiflash/pull/5762

dbms/src/Storages/StorageDeltaMerge.cpp
```diff
@@ -1198,59 +1198,26 @@ void StorageDeltaMerge::rename(
...
-        if (storeInited())
-        {
-            _store->rename(new_path_to_db, clean_rename, new_database_name, new_table_name);
-        }
-        else
-        {
-            table_column_info->db_name = new_database_name;
-            table_column_info->table_name = new_table_name;
-        }
+        _store->rename(new_path_to_db, clean_rename, new_database_name, new_table_name);
```
lidezhu, line 1208: How about just remove `clean_rename` parameter from `DeltaMergeStore::rename`?

JaySon-Huang: removed

lidezhu (approved):
LGTM

flowbehappy (approved):

Issue #5710, opened 2022-08-26
Interaction between `EstablishCallData` and `MPPTunnel` should be more clear


gengliqi:
## Enhancement

In the current codebase, the interaction between `EstablishCallData` and `MPPTunnel` is complex and error-prone.
As we can see, these two structures can call the function of each other and also hold the pointer of each other.

Several bugs related to `EstablishCallData` and `MPPTunnel` are recently fixed. For example, https://github.com/pingcap/tiflash/pull/5553 and https://github.com/pingcap/tiflash/pull/5650.

We should make the logic clear in order to easily maintain it and maybe improve it in the future.

PR #5711, opened 2022-08-26
Decouple `EstablishCallData` and `MPPTunnel`

gengliqi:

Issue Number: close https://github.com/pingcap/tiflash/issues/5710

Problem Summary:
See https://github.com/pingcap/tiflash/issues/5710.

This PR does several changes to address the issue described in https://github.com/pingcap/tiflash/issues/5710.

This PR adds a `GRPCSendQueue` to decouple `EstablishCallData` and `MPPTunnel`. 
`GRPCSendQueue` is a multi-producer-single-consumer queue dedicated to async grpc streaming send work.
You can see more descriptions in the annotation of `GRPCSendQueue`.

So for now, several mutexes and variables are removed. Also, the grpc's `Write` is only called in grpc threads.
We don't need to care about the synchronization between grpc threads and other threads. 
The trouble is totally solved by `GRPCSendQueue`.

Another surprising thing is that this PR can improve some performance in high load.
Test with TPCH and 15 clients.
Each client run TPCH 2 times.

| Branch  | Time1 | Time2 |
| ------------- | ------------- | ------------- |
| Master  | 5min 25s | 5min 17s |
| This PR  | 4min 59s | 5min 2s |

However, in low load, the performance seems the same.
Test with TPCH and 1 client.
Each client run TPCH 2 times.

| Branch  | Time1 |
| ------------- | ------------- |
| Master  | 0min 27s |
| This PR  | 0min 26s |

The performance improvement may be because the grpc's `write` is totally moved to grpc threads. What I know so far is `write` has some serialization work. I will try to figure it out if I have time.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
Run failpoint tests for more than 10 hours.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -195,32 +189,63 @@ void MPPTunnel::writeDone()
...
+    }
+    LOG_DEBUG(log, "connected");
+}
+
+void MPPTunnel::connectAsync(IAsyncCallData * call_data)
+{
+    {
+        std::unique_lock lk(mu);
+        if (status != TunnelStatus::Unconnected)
+            throw Exception(fmt::format("MPPTunnel has connected or finished: {}", statusToString()));
```
gengliqi, line 228: Almost all codes of `MPPTunnel` are written for async grpc only. 
So we should totally separate the sync grpc code path from `MPPTunnel` to clarify the code.
Since this PR is too large, it will be done in future PR.

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -195,32 +189,63 @@ void MPPTunnel::writeDone()
...
-            RUNTIME_ASSERT(false, log, "Unsupported TunnelSenderMode: {}", mode);
+            RUNTIME_ASSERT(false, log, "Unsupported TunnelSenderMode in connect: {}", mode);
+        }
+        status = TunnelStatus::Connected;
+        cv_for_status_changed.notify_all();
+    }
+    LOG_DEBUG(log, "connected");
+}
+
+void MPPTunnel::connectAsync(IAsyncCallData * call_data)
```
gengliqi, line 223: Almost all codes of MPPTunnel are written for async grpc only.
So we should totally separate the sync grpc code path from MPPTunnel to clarify the code.
Since this PR is too large, it will be done in future PR.

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// Push the data from queue and kick the grpc completion queue.
+    ///
+    /// Return true if push succeed.
+    /// Else return false.
+    template <typename U>
+    bool push(U && u)
+    {
+        auto ret = send_queue.push(u) == MPMCQueueResult::OK;
```
fuzhe1989, line 105:
```suggestion
        auto ret = send_queue.push(std::forward<U>(u)) == MPMCQueueResult::OK;
```

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,233 @@
...
+        : send_queue(MPMCQueue<T>(queue_size))
+        , tag(nullptr)
+        , log(log_)
+    {
+        RUNTIME_ASSERT(call != nullptr, log, "call is null");
+        // If a call to `grpc_call_start_batch` with an empty batch returns
+        // `GRPC_CALL_OK`, the tag is pushed into the completion queue immediately.
+        // This behavior is well-defined. See https://github.com/grpc/grpc/issues/16357.
+        kick_func = [call](void * t) {
+            return grpc_call_start_batch(call, nullptr, 0, new KickTag(t), nullptr);
```
fuzhe1989, line 75: I had some concerns about here. Then I realize that when `tag` is not `nullptr` the EstablishCallData is stopped and only waiting on notification form here. So there won't be any race.

How about mentions this in comment?

dbms/src/Server/FlashGrpcServerHolder.cpp
```diff
@@ -102,9 +103,11 @@ FlashGrpcServerHolder::FlashGrpcServerHolder(Context & context, Poco::Util::Laye
     /// Init and register flash service.
     bool enable_async_server = context.getSettingsRef().enable_async_server;
     if (enable_async_server)
-        flash_service = std::make_unique<AsyncFlashService>(security_config, context);
+        flash_service = std::make_unique<AsyncFlashService>();
     else
-        flash_service = std::make_unique<FlashService>(security_config, context);
+        flash_service = std::make_unique<FlashService>();
+    flash_service->init(security_config, context);
```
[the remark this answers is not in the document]
gengliqi: `WithAsyncMethod_EstablishMPPConnection<FlashService>` can only use a default construction function of `FlashService`. I think this change is better than copying the `WithAsyncMethod_EstablishMPPConnection` code.

dbms/src/Flash/Mpp/EstablishCall.h
```diff
@@ -17,29 +17,30 @@
...
-    bool write(const mpp::MPPDataPacket & packet) override { return writer->Write(packet); }
+    /// Should return a non-null `grpc_call`.
+    virtual grpc_call * grpcCall() = 0;
 
-private:
-    ::grpc::ServerWriter<::mpp::MPPDataPacket> * writer;
+    /// Attach async sender in order to notify consumer finish msg directly.
+    virtual void attachAsyncTunnelSender(const std::shared_ptr<DB::AsyncTunnelSender> &) = 0;
+
+    // Return a `GRPCKickFunc` only used for test.
```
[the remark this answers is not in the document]
gengliqi: Done.

dbms/src/Flash/Mpp/EstablishCall.h
```diff
@@ -17,29 +17,30 @@
...
 {
 public:
-    explicit SyncPacketWriter(grpc::ServerWriter<mpp::MPPDataPacket> * writer)
-        : writer(writer)
-    {}
+    virtual ~IAsyncCallData() = default;
 
-    bool write(const mpp::MPPDataPacket & packet) override { return writer->Write(packet); }
+    /// Should return a non-null `grpc_call`.
+    virtual grpc_call * grpcCall() = 0;
```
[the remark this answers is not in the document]
gengliqi: But it's essential to construct `GRPCSendQueue`.

dbms/src/Flash/Mpp/EstablishCall.h
```diff
@@ -78,37 +73,37 @@ class EstablishCallData : public PacketWriter
...
-    void finishTunnelAndResponder();
+    void write(const mpp::MPPDataPacket & packet);
+
+    void writeErr(const mpp::MPPDataPacket & packet);
+    // Will try to call async_sender's consumerFinish if needed
+    void writeDone(String msg, const grpc::Status & status);
 
-    /// Will try to call async_sender's consumerFinish if needed
-    void setFinishState(const String & msg);
+    void unexpectedWriteDone();
```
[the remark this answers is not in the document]
gengliqi: Done.

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,233 @@
...
+        explicit KickTag(void * tag_)
+            : tag(tag_)
+        {}
+
+        bool FinalizeResult(void ** tag_, bool * /*status*/) override
+        {
+            *tag_ = tag;
+            // After calling this function, this `KickTag` will not be pointed
+            // to by any pointer. So it can be directly deleted here.
+            delete this;
```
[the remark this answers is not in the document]
gengliqi: Done.

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: I add more comments above `pop` because the `tag` is passed through it.

dbms/src/Flash/Mpp/EstablishCall.cpp
```diff
@@ -78,129 +110,95 @@ void EstablishCallData::initRpc()
...
-    }
-    responderFinish(status);
-}
-
-void EstablishCallData::notifyReady()
-{
-    std::unique_lock lk(mu);
-    ready = true;
-}
+        if (stopwatch)
```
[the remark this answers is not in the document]
gengliqi: Yes.

dbms/src/Flash/Mpp/EstablishCall.cpp
```diff
@@ -78,129 +110,95 @@ void EstablishCallData::initRpc()
...
+            LOG_FMT_INFO(async_tunnel_sender->getLogger(), "connection for {} cost {} ms.", async_tunnel_sender->getTunnelId(), stopwatch->elapsedMilliseconds());
+        }
 
-void EstablishCallData::cancel()
-{
-    if (state == NEW_REQUEST || state == FINISH) // state == NEW_REQUEST means the server is shutdown and no new rpc has come.
-    {
-        delete this;
-        return;
+        if (!async_tunnel_sender->isConsumerFinished())
```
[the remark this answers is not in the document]
gengliqi: This should not happen. If a connection is established, `MPPTunnel` must wait for `CallData` to finish before it can finish.

dbms/src/Flash/Mpp/EstablishCall.cpp
```diff
@@ -37,39 +39,69 @@ EstablishCallData::EstablishCallData(AsyncFlashService * service, grpc::ServerCo
...
+        assert(state == FINISH);
+        // Once in the FINISH state, deallocate ourselves (EstablishCallData).
+        // That's the way GRPC official examples do. link: https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/greeter_async_server.cc
+        delete this;
+        return;
+    }
 }
 
-void EstablishCallData::tryFlushOne()
+void EstablishCallData::cancel()
```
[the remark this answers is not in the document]
gengliqi: Done.

dbms/src/Flash/Mpp/EstablishCall.h
```diff
@@ -17,29 +17,30 @@
...
+    virtual ~IAsyncCallData() = default;
 
-    bool write(const mpp::MPPDataPacket & packet) override { return writer->Write(packet); }
+    /// Should return a non-null `grpc_call`.
+    virtual grpc_call * grpcCall() = 0;
 
-private:
-    ::grpc::ServerWriter<::mpp::MPPDataPacket> * writer;
+    /// Attach async sender in order to notify consumer finish msg directly.
+    virtual void attachAsyncTunnelSender(const std::shared_ptr<DB::AsyncTunnelSender> &) = 0;
```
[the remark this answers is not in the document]
gengliqi: `AsyncTunnelSender` must outlive `IAsyncCallData`.

dbms/src/Server/FlashGrpcServerHolder.cpp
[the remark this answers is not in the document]
fuzhe1989: got it.

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
fuzhe1989: Then I thought a raw pointer is enough? Or `IAsyncCallData` may mistakenly extend the lifetime of `AsyncTunnelSender`.

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
fuzhe1989: A trick is to let `grpc_call` be another template parameter.

dbms/src/Flash/Mpp/EstablishCall.cpp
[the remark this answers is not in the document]
fuzhe1989: Then you should assert it or do not have this if.

dbms/src/Flash/Mpp/EstablishCall.cpp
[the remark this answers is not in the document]
fuzhe1989: then here the check to `stopwatch` is needless, right?

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
fuzhe1989: My last comment is not quite correct, but here `shared_ptr` make the relationship too complicated (at least from the first glance).

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+#include <Common/MPMCQueue.h>
+#include <common/logger_useful.h>
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+#include <grpcpp/grpcpp.h>
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
```
fuzhe1989, line 28: guess other import points have to do the same thing. how about move all the macros to a new header file which wraps `#include <grpcpp/grpcpp.h>`?

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+    {
+        // No tag.
+        NONE,
+        // Waiting for kicking.
+        WAITING,
+        // Queuing in the grpc completion queue.
+        QUEUING,
+    };
+
+    KickTag(std::mutex * m, const LoggerPtr l)
```
fuzhe1989, line 55:
```suggestion
    KickTag(std::mutex * m, const LoggerPtr & l)
```

dbms/src/Flash/Mpp/EstablishCall.cpp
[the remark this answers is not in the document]
gengliqi: Done.

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+        }
+        RUNTIME_ASSERT(tag != nullptr, log, "status is waiting but tag is nullptr");
+        status = Status::QUEUING;
+        return true;
+    }
+
+private:
+    friend class tests::TestGRPCSendQueue;
+
+    std::mutex * mu;
```
fuzhe1989, line 115: Current usages to mu is error-prone in that:
1. not all methods lock mu, actually `toQueuing` implicitly requires the caller has locked mu.
2. It' not clear who holds mu.

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+        res = send_queue.tryPop(data);
+        switch (res)
+        {
+        case MPMCQueueResult::OK:
+            return GRPCSendQueueRes::OK;
+        case MPMCQueueResult::FINISHED:
+            return GRPCSendQueueRes::FINISHED;
+        case MPMCQueueResult::EMPTY:
+            // If empty, change status to WAITING.
+            RUNTIME_ASSERT(kick_tag.toWaiting(new_tag), log, "status {} is not none", kick_tag.getStatus());
```
fuzhe1989, line 224: It's not recommended to call non-pure functions in assert.

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
gengliqi: I am wrong. In the current implementation, the `consumerFinish` is called before calling `responder.Finish`. Then `EstablishCallData` will wait for the event for `responder.Finish` then it will destroy itself. For now, `AsyncTunnelSender` does not touched in `Finish` state.
I don't know if calling `consumerFinish` after `responder.Finish` is better than before.

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
gengliqi: Seems too verbose? The `IAsyncCallData` is just for the convenience of test.

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -136,38 +140,40 @@ class SyncTunnelSender : public TunnelSender
...
 private:
     friend class tests::TestMPPTunnel;
-    void sendJob();
+    void sendJob(PacketWriter * writer);
     std::shared_ptr<ThreadManager> thread_manager;
 };
 
 /// AsyncTunnelSender is mainly triggered by the Async PacketWriter which handles GRPC request/response in async mode, send one element one time
 class AsyncTunnelSender : public TunnelSender
+    , public GRPCSendQueue<MPPDataPacketPtr>
```
fuzhe1989, line 154: please either use private inheritance or use composition.

dbms/src/Flash/Mpp/EstablishCall.h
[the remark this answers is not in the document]
fuzhe1989: how about only return a `void *`?

dbms/src/Flash/Mpp/MPPTunnel.h
gengliqi: Change to use composition.

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: Added a `common/grpcpp.h`

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: Changed `toQueuing` and `toWaiting` to private functions.

dbms/src/Flash/Mpp/EstablishCall.cpp
```diff
@@ -78,129 +112,94 @@ void EstablishCallData::initRpc()
     try
     {
         FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::random_tunnel_init_rpc_failure_failpoint);
-        service->establishMPPConnectionSyncOrAsync(&ctx, &request, nullptr, this);
+
+        auto [status, err_msg] = service->establishMPPConnectionSyncOrAsync(&ctx, &request, nullptr, this);
```
fuzhe1989, line 116: nit: suggest a better name for `establishMPPConnectionSyncOrAsync`?

dbms/src/Flash/FlashService.cpp
```diff
@@ -181,62 +181,53 @@ ::grpc::Status FlashService::DispatchMPPTask(
...
-        calldata->writeDone(status);
-    }
-    return status;
-}
-
-::grpc::Status FlashService::establishMPPConnectionSyncOrAsync(::grpc::ServerContext * grpc_context,
-                                                               const ::mpp::EstablishMPPConnectionRequest * request,
-                                                               ::grpc::ServerWriter<::mpp::MPPDataPacket> * sync_writer,
-                                                               EstablishCallData * calldata)
+std::pair<::grpc::Status, std::string> FlashService::establishMPPConnectionSyncOrAsync(::grpc::ServerContext * grpc_context,
```
fuzhe1989, line 217: The caller may be confused by the two returned values, what's the meaning of each scenario:
1. {ok, empty}
2. {err, empty}
3. {ok, err}
4. {err, err}

dbms/src/Flash/Mpp/EstablishCall.cpp
```diff
@@ -78,129 +112,94 @@ void EstablishCallData::initRpc()
...
+        if (!status.ok())
```
fuzhe1989, line 117: What's the difference between L117 and L122?

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+
+    bool FinalizeResult(void ** tag_, bool * /*status*/) override
+    {
+        std::unique_lock lock(*mu);
+
+        RUNTIME_ASSERT(status == Status::QUEUING, log, "status {} is not queuing", status);
+        status = Status::NONE;
+
+        *tag_ = tag;
+        tag = nullptr;
```
fuzhe1989, line 68:
```suggestion
        *tag_ = std::exchange(tag, nullptr);
```

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,285 @@
...
+        return status;
+    }
+
+private:
+    friend class tests::TestGRPCSendQueue;
+    template <typename T>
+    friend class GRPCSendQueue;
+
+    /// Should acquire lock.
+    bool toWaiting(void * new_tag)
```
fuzhe1989, line 85: Still worry about the unclear ownership of mu.

A way to avoid depending on the caller locking mutex is to pass in the logic that should call in the critical area:

```cpp
template <typename F>
auto toWaiting(F && precheck, void * new_tag) {
    std::unique_lock lock(mu); // KickTag itself could own the mutex now.
    auto status = precheck();
    if (status == GRPCSendQueueRes::EMPTY) {
        ...
    }
    return status;
}
```

dbms/src/Flash/FlashService.cpp
gengliqi: Changed it to `std::variant`.

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: It's not very elegant.
I think a while and have changed it to a more reasonable manner(in my opinion).
The `status` and `tag` should belong to `GRPCSendQueue`. A `std::function` that contains the mutex logic is passed to `KickTag`. In this way, there is only one mutex.

dbms/src/Flash/Mpp/EstablishCall.cpp
gengliqi: I have no idea.

gengliqi:
@fuzhe1989 @bestwoody PTAL, thanks.

dbms/src/Common/MPMCQueue.h
```diff
@@ -427,6 +415,9 @@ class MPMCQueue
         std::unique_lock lock(mu);
         for (; read_pos < write_pos; ++read_pos)
             destruct(getObj(read_pos));
+
+        read_pos = 0;
```
bestwoody, line 419: what's the purpose of this?

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,264 @@
...
+        case MPMCQueueResult::FINISHED:
+            return GRPCSendQueueRes::FINISHED;
+        case MPMCQueueResult::EMPTY:
+            // Handle this case later.
+            break;
+        default:
+            RUNTIME_ASSERT(false, log, "Result {} is invalid", res);
+        }
+
+        std::unique_lock lock(mu);
```
bestwoody, line 152: can `pop(T & data, void * new_tag)`  be called concurrently? What will happen when multiple threads call `send_queue.tryPop` and  return `MPMCQueueResult::EMPTY` before and only one thread call the following `MPMCQueueResult::EMPTY` branch and other theads found `status` is `Status::WAITING` which cause `RUNTIME_ASSERT` fail

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,264 @@
...
+            std::unique_lock lock(mu);
+            if (status != Status::WAITING)
+            {
+                return;
+            }
+            RUNTIME_ASSERT(tag != nullptr, log, "status is waiting but tag is nullptr");
+            status = Status::QUEUING;
+        }
+
+        grpc_call_error error = kick_func(&kick_tag);
```
bestwoody, line 222: is it safe to call `kick_func(&kick_tag)` concurrently?

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -98,24 +97,26 @@ MPPTunnel::~MPPTunnel()
...
 }
 
 /// exit abnormally, such as being cancelled.
 void MPPTunnel::close(const String & reason)
 {
     SCOPE_EXIT({
-        // ensure the tracked memory is released and udpated before memotry tracker(in ProcListEntry) is released
-        send_queue->finishAndDrain(); // drain the send_queue when close
+        // ensure the tracked memory is released and updated before memory tracker(in ProcListEntry) is released
+        finishSendQueue(true); // drain the send_queue when close
```
bestwoody, line 116: The original one didn't triger a tryFlushOne()/kick, however, the new one will triger. Is it expected?

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: No, it can't. The comments above `GRPCSendQueue` says it's a `multi-producer-single-consumer` queue.

dbms/src/Flash/Mpp/GRPCSendQueue.h
gengliqi: Call `grpc_call_start_batch` with a empty batch is thread-safe. 
PS: I don't know why the comments here https://github.com/grpc/grpc/pull/16362/files says it's thread-compatible but https://github.com/tikv/grpc-rs/pull/217/files#diff-1b4f412f47df1edd4ad45800cfcc9134256c302b000d84646a75e0e7f022a944R765-R768 here says it's thread-safe.

dbms/src/Common/MPMCQueue.h
gengliqi: For safety, I worry about using `drain` without `finish` or `cancel` in future code.

dbms/src/Flash/Mpp/MPPTunnel.cpp
gengliqi: It's safe to call `kick` function multiple times. But calling `kick` function here is indeed useless in current implementation. I added a `kick` function in `GRPCSendQueue::finishAndDrain` just for semantic consistency.

dbms/src/Flash/Mpp/GRPCSendQueue.h
bestwoody: Do you mean kicking a NULL tag is safe? how about add a UT for it.

bestwoody (approved):
lgtm

dbms/src/Flash/Mpp/MPPTunnel.cpp
```diff
@@ -201,32 +198,63 @@ void MPPTunnel::writeDone()
...
+
+void MPPTunnel::connectAsync(IAsyncCallData * call_data)
+{
+    {
+        std::unique_lock lk(mu);
+        if (status != TunnelStatus::Unconnected)
+            throw Exception(fmt::format("MPPTunnel has connected or finished: {}", statusToString()));
+
+        LOG_FMT_TRACE(log, "ready to connect async");
+        if (mode == TunnelSenderMode::ASYNC_GRPC)
```
yibin87, line 240: Can use RUNTIME_ASSERT directly to reduce a branch.

yibin87 (approved):
LGTM

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -0,0 +1,264 @@
...
+class TestGRPCSendQueue;
+} // namespace tests
+
+/// In grpc cpp framework, the tag that is pushed into grpc completion
+/// queue must be inherited from `CompletionQueueTag`.
+class KickTag : public grpc::internal::CompletionQueueTag
+{
+public:
+    explicit KickTag(std::function<void *()> a)
+        : action(a)
```
fuzhe1989, line 39:
```suggestion
        : action(std::move(a))
```

dbms/src/Flash/EstablishCall.cpp
```diff
@@ -78,129 +113,107 @@ void EstablishCallData::initRpc()
...
+                                              writeErr(getPacketWithError(err_msg));
+                                              return false;
+                                          }
+                                          return true;
+                                      }},
+                                  res);
+        if (!success)
+        {
+            return;
+        }
```
fuzhe1989, line 140: bad readability

gengliqi: Reduced some code indentation.

fuzhe1989 (approved):

LittleFall (commented):
LGTM

PR #5731, opened 2022-08-30
PageStorage: avoid reallocate wal write buffer size

### LINKED ISSUE #5721: The page store size is abnormal
## Bug Report

case https://tcms.pingcap.net/dashboard/executions/plan/1171035 
origin_img_v2_932ef73e-bd96-4ef5-a918-0428a011dbeg
origin_img_v2_669f9a7d-46bd-4ec5-b078-36a880fa0f3g

master b109a12ec8ae99698110d44ad747d7b3a9bdd9fd

lidezhu:

Issue Number: close #5721

Problem Summary: The write buffer size of `LogWriter` is not right under some case.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the bug that TiFlash can not restart after page gc
```


dbms/src/Storages/Page/V3/LogFile/LogWriter.h
```diff
@@ -64,11 +64,11 @@ namespace PS::V3
  * Recyclable record format:
  *
  * +--------------+-----------+-----------+----------------+--- ... ---+
- * |CheckSum (8B) | Size (2B) | Type (1B) | Log number (4B)| Payload   |
+ * |CheckSum (8B) | Size (2B) | Type (1B) | Log number (8B)| Payload   |
  * +--------------+-----------+-----------+----------------+--- ... ---+
  *
  * Same as above, with the addition of
- * Log number = 32bit log file number, so that we can distinguish between
+ * Log number = 64bit log file number, so that we can distinguish between
```
flowbehappy, line 71: Is it compatible with old data?

lidezhu: Old data is 64 bit. This is just a change of code comment

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -92,23 +93,54 @@ void LogWriter::close()
...
+        // Flush the buffer if it has data before reset buffer
+        if (unlikely(write_buffer.offset() > 0))
+        {
+            flush(write_limiter, /* background */ false);
+        }
+        assert(write_buffer.offset() == 0);
+        // Accurate new_buff_size should be ((total_write_size - 1) / Format::BLOCK_SIZE + 1) * Format::BLOCK_SIZE
+        // Just be simple and conservative here.
+        size_t new_buff_size = (total_write_size / Format::BLOCK_SIZE + 1) * Format::BLOCK_SIZE;
         buffer = static_cast<char *>(realloc(buffer, buffer_size, new_buff_size));
```
JaySon-Huang, line 136: Do not realloc the buffer size, keep the buffer size stick to `Format::BLOCK_SIZE`. And call `flush` after we finish writing a block.
The places we need to call `flush` should be limited.

This should be the correct way to fix this misbehavior.

lidezhu: Fixed.

dbms/src/Storages/Page/V3/tests/gtest_wal_log.cpp
```diff
@@ -303,6 +315,15 @@ TEST_P(LogFileRWTest, Fragmentation)
...
 }
 
+// This test may take a lot of time
+TEST_P(LogFileRWTest, DifferentPayloadSize)
+{
+    for (size_t i = 0; i < 40000; i += 1)
+    {
+        write(repeatedString("a", i));
+    }
+}
```
JaySon-Huang, line 325: This slow down the ut running time in CI from about 7 min to 17 min...

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -56,8 +58,9 @@ void LogWriter::resetBuffer()
 
 LogWriter::~LogWriter()
 {
-    log_file->fsync();
-    log_file->close();
+    // In case of forgetting flush when `manual_flush` is true, should do nothing when `LogWriter` is used correctly
+    flush(/* write_limiter */ nullptr, /* background */ false);
```
JaySon-Huang, line 62: I think flush is too heavy for a deconstructor function. It is much likely throwing an I/O error than fsync in the deconstructor function. I think leaving it to the user with `manual_flush` calling `flush` explicitly is OK.

JaySon-Huang: For example, in WALStore::saveSnapshot:
https://github.com/pingcap/tiflash/blob/c85332e14abcd34953837a45bab7ba48898e4544/dbms/src/Storages/Page/V3/WALStore.cpp#L201-L206

If we do flush in `~LogWriter` and `compact_log->flush` throw an I/O exception, then when exiting `WALStore::saveSnapshot`, `~LogWriter` is called and another I/O exception is thrown. An I/O exception thrown in the deconstructor will make the process exit.

dbms/src/Storages/Page/V3/tests/gtest_wal_log.cpp
JaySon-Huang: Can we ease this test after we keep the buffer size unchanged?

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
lidezhu: Fixed.

dbms/src/Storages/Page/V3/tests/gtest_wal_log.cpp
lidezhu: Disabled test test.

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -92,40 +99,32 @@ void LogWriter::close()
...
     // we still want to iterate once to emit a single zero-length record.
     bool begin = true;
     size_t payload_left = payload_size;
-
-    size_t head_sizes = ((payload_size / Format::BLOCK_SIZE) + 1) * Format::RECYCLABLE_HEADER_SIZE;
-    if (payload_size + head_sizes >= buffer_size)
+    // Padding current block if needed
+    block_offset = block_offset % Format::BLOCK_SIZE; // 0 <= block_offset < Format::BLOCK_SIZE
+    size_t leftover = Format::BLOCK_SIZE - block_offset;
+    assert(leftover > 0);
```
flowbehappy, line 111: Better to use `RUNTIME_ASSERT`

lidezhu: This should be just a reminder for later maintenance. Because the condition`Format::BLOCK_SIZE - block_offset % Format::BLOCK_SIZE > 0` is really a simple mathematics and must be true.

lidezhu: It is more like a code comment and stronger than code comment. And `RUNTIME_ASSERT` is a little heavy and unnecessary.

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -92,40 +99,32 @@ void LogWriter::close()
 void LogWriter::addRecord(ReadBuffer & payload, const size_t payload_size, const WriteLimiterPtr & write_limiter)
```
JaySon-Huang, line 99: https://github.com/pingcap/tiflash/blob/3119fa401635635cfe7e7369f95d4250fcfeb63b/dbms/src/Storages/Page/V3/WALStore.cpp#L201-L206

We should add a bool param `background` and set `write_limiter`, `background=true` inside `WALStore::saveSnapshot`. Or it won't be limited and counted as the background write flow.

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -139,7 +138,21 @@ void LogWriter::addRecord(ReadBuffer & payload, const size_t payload_size, const
...
+        {
+            flush(write_limiter, /* background */ false);
+        }
+        try
+        {
+            emitPhysicalRecord(type, payload, fragment_length);
+        }
+        catch (...)
+        {
+            auto message = getCurrentExceptionMessage(false);
```
JaySon-Huang, line 152: should we call `getCurrentExceptionMessage` with `with_stacktrace=true` to get more information what happen here?

lidezhu: Change to `with_stacktrace=true`.

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
lidezhu: Fixed. By the way, it seems we may call `WALStore::apply` in a background thread but `addRecord` is always called with `background=true`. https://github.com/pingcap/tiflash/blob/79b36c3bf30dc2604b66971a54616512617e9676/dbms/src/Storages/Page/V3/WALStore.cpp#L105
Do we need to fix this? `BlobFile` also has similar problem.

lidezhu: May be we can check background like `RateLimiter` using `is_background_thread` in `PageUtil::writeFile` and avoid pass `background` around?

JaySon-Huang: `background` maybe a misleading name. I want to check the write flow that is caused by compaction (`saveSnapshot` as log_file compaction, `full gc` as blob_file compaction) and compare it to the total write flow.

JaySon-Huang: the thread local variable `is_background_thread` may be some write caused by compaction in the DeltaTree level, that is not what I want.

flowbehappy (approved):

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
JaySon-Huang: `WALStore::apply(const PageEntriesEdit & edit, const WriteLimiterPtr & write_limiter)` is a common method that will be used for both write from out of the `PageStorage` level and write from the compaction inside `PageStorage` level.

JaySon-Huang (approved):
LGTM

Issue #5689, opened 2022-08-24
Enable read thread of storage and deprecate configuation `dt_enable_read_thread`.


JinheLin:
1. Enable read thread of storage.
2. Deprecate configuation `dt_enable_read_thread` after https://github.com/pingcap/tidb/issues/35875 is fixed.

PR #5765, opened 2022-09-01
MemoryTracker*: Use std::shared_ptr to manage MemoryTracker.

JinheLin:

Issue Number: ref #5689

1. Computing layer uses `MemoryTracker` to track memory usage of `MPPTask` . And also uses `MemoryTracker` to estimate the size of the data and different execution strategies, such as multi-thread or single-thread, are adopted for different data sizes.
2. The reading threads of storage will set the `MemoryTracker` object of the corresponding `MPPTask` to `current_memory_tracker` when reading data and avoid inaccurate data size estimates.
3. If the `MPPTask` exists, the corresponding `MemoryTracker` will be destroyed. But the reading threads of storage can still use this object and cause the process crash.
4. For safety, reading threads should take a `shared_ptr` of `MemoryTracker` instead of a raw pointer.
5. In this PR, creating all the `MemoryTracker` objects by `MemoryTracker::create` which returns a `std::shared_ptr<MemoryTracker>` object.
6. In `DeltaMerge::read`, `SegmentReadTaskPool` will call `current_memory_tracker->shared_from_this()` to get the shared pointer.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/workload/PSBackground.cpp
```diff
@@ -51,9 +51,9 @@ void PSGc::doGcOnce()
     gc_stop_watch.start();
     try
     {
-        MemoryTracker tracker;
-        tracker.setDescription("(Stress Test GC)");
-        current_memory_tracker = &tracker;
+        auto tracker = MemoryTracker::create();
```
bestwoody, line 54: Is this a test code? since this memTracker not `setNext` a root memTracker.

dbms/src/Storages/Page/workload/PSRunnable.cpp
```diff
@@ -29,9 +29,9 @@ namespace DB::PS::tests
 void PSRunnable::run()
 try
 {
-    MemoryTracker tracker;
-    tracker.setDescription(nullptr);
-    current_memory_tracker = &tracker;
+    auto tracker = MemoryTracker::create();
```
bestwoody, line 32: ditto

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.h
```diff
@@ -139,7 +139,7 @@ class SegmentReadTaskPool : private boost::noncopyable
         , log(&Poco::Logger::get("SegmentReadTaskPool"))
         , unordered_input_stream_ref_count(0)
         , exception_happened(false)
-        , mem_tracker(current_memory_tracker)
+        , mem_tracker(current_memory_tracker == nullptr ? nullptr : current_memory_tracker->shared_from_this())
```
bestwoody, line 142: When will segReadTaskPool be destructed? Is it shared by multiple MPPTasks?

dbms/src/Storages/BackgroundProcessingPool.cpp
```diff
@@ -152,9 +152,9 @@ void BackgroundProcessingPool::threadFunction()
         addThreadId(getTid());
     }
 
-    MemoryTracker memory_tracker;
-    memory_tracker.setMetric(CurrentMetrics::MemoryTrackingInBackgroundProcessingPool);
-    current_memory_tracker = &memory_tracker;
+    auto memory_tracker = MemoryTracker::create();
```
bestwoody, line 155: Is this a test code? since this memTracker did not setNext a root memTracker.

bestwoody: This seems not a test code.

bestwoody: Root memtracker is total_mem_tracker in ProcessLsit.
if not `setNext` a root memtracker，the memory usage actually not be detected by process level of TiFlash.
The config `max_memory_usage_for_all_queries` will not work on it.

bestwoody: > These are the background threads that handle background tasks such as compaction and delta merge.
> 
> 
> 
> What is the use of `setNext` a root MemoryTracker?

ref：https://pingcap.feishu.cn/wiki/wikcnvA5vz6Mfk9VocSssEqPp3b

bestwoody: will it use a lot of memory？ for example：insert or query？

bestwoody (approved):
lgtm

dbms/src/Storages/BackgroundProcessingPool.cpp
JaySon-Huang: > Root memtracker is total_mem_tracker in ProcessLsit.
> if not setNext a root memtracker，the memory usage actually not be detected by process level of TiFlash.
> The config max_memory_usage_for_all_queries will not work on it.

The memory used in here is not caused by queries, so I think it should not be a child of the `ProcessList::total_mem_tracker`(limited by `max_memory_usage_for_all_queries`).
But we do need another root MemoryTracker for all background tasks and maybe a new config item for controlling the memory usage for it.

JaySon-Huang (approved):
LGTM

dbms/src/Storages/BackgroundProcessingPool.cpp
bestwoody: if there are two root memTracker，none of them can track all the memory of the tiflash process. In that case，we can track and throttle the process memory

JaySon-Huang: 1. I agree that a root memory tracker that track all memory of a TiFlash instance can avoid OOM under some case. But it is weird that config with the name `max_memory_usage_for_all_queries` limit the memory usage for non-queries task. Maybe we need to refactor the config ;)
2. Some background tasks are lower priority than queries, so those background tasks can be delayed or canceled if the total memory usage reaches the high water mark. We need another root memory tracker besides queires_memory_tracker.

PR #5812, opened 2022-09-07
Auto Load Certificate

### LINKED ISSUE #5503: Support certificates reloading
## Feature Request

We use [cert-manager](https://github.com/cert-manager/cert-manager) to manage the TLS cert for the TiDB Cluster. When the cert expired, cert-manager renewed the cert, but TiFlash didn't reload the new cert online.

TiFlash reloads the TLS cert online, both for client connection and mTLS.

for TiKV, I found https://github.com/tikv/tikv/pull/7150, but it seems still does not work well. for other components of TiDB, I haven't tested them yet.

birdstorm:
Signed-off-by: birdstorm <samuelwyf@hotmail.com>

Issue Number: close #5503

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Context.cpp
```diff
@@ -1399,6 +1420,12 @@ void Context::createTMTContext(const TiFlashRaftConfig & raft_config, pingcap::C
     shared->tmt_context = std::make_shared<TMTContext>(*this, raft_config, cluster_config);
 }
 
+void Context::updateTMTContext(const TiFlashRaftConfig & raft_config, pingcap::ClusterConfig && cluster_config)
```
lidezhu, line 1423: I don't think we should create a new `TMTContext` just for reload config. Because most of our service objects is managed in `TMTContext` which has nothing to with TLS. And create `TMTContext` is also a heavy operation and have a high possibility to introduce other race problems at run time.

birdstorm: you are right, is there a better way?

lidezhu: Just reload the service related to TLS.

birdstorm: > Just reload the service related to TLS.

@lidezhu 
The services related to TLS include the pd_client, rpc_client, region_cache, and lock_resolver.
I believe it is almost as heavy as reloading the TMTContext.

If it is true that other services depending on the mentioned services do not need to reload again, I can fix it by just reloading these services. For now, I am not certain about the dependencies of other services.

lidezhu: For example, `KVStore` and `ManagedStorages` are services that should not be recreated when reload TLS certificate. And there are lots of other things that need to be done after create `TMTContext`, for example restore data for KVStore, register all tables in ManagedStorages and a lot more.

PR #5833, opened 2022-09-09
Storage: (Refactor) Separate segment impl out 

breezewish:
### What problem does this PR solve?

Issue Number: ref #5831

Problem Summary:

### What is changed and how it works?

Move internal segment operations impl into a standalone file.

- [x] This PR is based on https://github.com/pingcap/tiflash/pull/5832, so that merge it first to get a nicer diff.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

Issue #5831, opened 2022-09-09
Separate DeltaMergeStore into multiple files


breezewish:
## Enhancement

DeltaMergeStore provides most of public interfaces of the storage engine. Currently `DeltaMergeStore.cpp/h` contains too many things together, which is hard for new developers to understand who are related together.

We can separate it into multiples via the following steps:

- [ ] 1. Simply move implementations in DeltaMergeStore.cpp into multiple files: DeltaMergeStore_Segment.cpp, DeltaMergeStore_Ingest.cpp, DeltaMergeStore_Background.cpp

- [ ] 2. Isolate private functions from `DeltaMergeStore` and place them in separated files, close to their impl places.

PR #5835, opened 2022-09-09
Storage: (Refactor) Avoid referencing DeltaMergeStore.h

breezewish:

Issue Number: ref #5831

1. Currently the `DeltaMerge/DeltaMergeStore.h` is imported in `Storages/StorageDeltaMerge.h`, which makes the pimpl useless -- When changing private member functions in `DeltaMerge/DeltaMergeStore.h`, a lot of external functions need to be recompiled.
2. It does not make sense to leak the "ThreadType" in the Delta Merge Public API, as background threads are purely internal.

1. Changed `TaskRunThread` to `MergeDeltaReason`, which makes more sense.
2. No need to specify `TaskRunThread` or `MergeDeltaReason` in DeltaMerge public API `mergeDeltaBySegment`, as run thread or merge reason are purely internal.
3. Forward declare `RowKeyValue` so that the import of internal DeltaMerge headers can be avoided.

- [x] This PR is based on https://github.com/pingcap/tiflash/pull/5833, so merge that PR first to get a nicer diff.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/StorageDeltaMerge.h
```diff
@@ -14,29 +14,24 @@
 
 #pragma once
 
+#include <Common/Logger.h>
```
hongyunyan, line 17: Can we remove this header file？

breezewish: It should be a better practice to include everything you used in this header file, even if it may be included transitively by other included header files. In this way, this header is less likely to break when other headers are refactored.

dbms/src/Common/TiFlashMetrics.h
```diff
@@ -107,20 +107,20 @@ namespace DB
...
     M(tiflash_storage_subtask_count, "Total number of storage's sub task", Counter,                                                       \
-        F(type_delta_merge, {"type", "delta_merge"}),                                                                                     \
-        F(type_delta_merge_fg, {"type", "delta_merge_fg"}),                                                                               \
-        F(type_delta_merge_fg_rpc, {"type", "delta_merge_fg_rpc"}),                                                                       \
+        F(type_delta_merge_bg, {"type", "delta_merge_bg"}),                                                                               \
         F(type_delta_merge_bg_gc, {"type", "delta_merge_bg_gc"}),                                                                         \
+        F(type_delta_merge_fg, {"type", "delta_merge_fg"}),                                                                               \
+        F(type_delta_merge_manual, {"type", "delta_merge_manual"}),                                                                       \
```
flowbehappy, line 113: Does grafana need to adjust?

breezewish: In Grafana expressions are using "delta_merge|seg" as matchers, so there is no need to adjust.

flowbehappy (approved):
LGTM with a minor comment

dbms/src/Storages/StorageDeltaMerge.h
JaySon-Huang: We can use forward declaration here, using
```
namespace DB
{
class Logger;
using LoggerPtr = std::shared_ptr<Logger>;
}
```
instead of including the `<Common/Logger.h>`.

Here I think both are OK since `Common/Logger.h` itself is a clean header file that doesn't bring many other header files.

JaySon-Huang (approved):
lgtm

Issue #5237, opened 2022-06-27
Big region snapshot adaption and applying snapshot improvement


flowbehappy:
## Enhancement
Applying a snapshot requires writing a delete range and a DTFile to the Delta layer of 1 or more segments. Because the snapshot is relatively large, it will 

- Very likely trigger a delta merge, which causes write amplification.
- A big snapshot will directly cause write stall.

Note that, most of the time, the snapshot is not overlapped with the data in the relative segments. I think we can find a way to avoid writing snapshots to the Delta layer, as LSM-Tree does.

breezewish:
Related TiKV change: https://github.com/tikv/tikv/issues/11515

PR #5848, opened 2022-09-12
storage: Allow merging multiple segments in the segment API

breezewish:

Issue Number: ref #5237

Problem Summary:

This is one of a series of PRs for #5237, where you can see the overall roadmap.

1. Segment - Adjusted interface to accept merging multiple segments.

   **Note:** The path of merging multiple segments is covered in tests. **It is not used in production in this PR**, as `checkSegmentUpdate` will always merge two segments.

2. Segment - Added `[[nodiscard]]` to the new segments returned in segment APIs to avoid forgetting using them.

3. DeltaMergeStore - Added sleep to the flush inside segment merge, in case of a flush is already in progress.

4. Segment - Added utility function `Segment::simpleInfo` and `Segment::info` for pretty-printing a vector of `SegmentPtr`s. They are also used in merging segments.

5. SegmentTest - Added new unit tests:
   - TestSegmentMemTableDataAfterSplit
   - TestSegmentMergeTwo
   - TestSegmentMergeThree
   - TestSegmentMergeInvalid

7. SegmentTest - Updated the segment test framework to test with merging multiple segments. The randomized test will randomly pick 2 ~ 6 continuous segments.

8. SegmentTest - Added a probability config for different segment operations in the randomized segment test, so that there is more chance to test with merging multiple segments.

9. SegmentTest - Changed `write_rows` from 8192 to default 100 when writing rows with deleted pack, because when writing rows with data we also write 100 rows by default.

10. SegmentTest - Write, Ingest and WriteWithDeletedPack now write continuous rows **randomly** in the segment range, instead of always write at the beginning of the segment.

11. SegmentTest - Randomized write now write 20 ~ 100 rows randomly, instead of always write 100 rows. 

12. SegmentTest - Changed to use the `random` in c++11, so that we can use different distribution utilities.

13. SegmentTest - Fix: Now the segment range generated by the test conforms with the check `range->isStartInfinite` or `range->isEndInfinite` when `is_common_handle = true`. This prevents possible bug that only happens in the test due to a wrong segment range. For example, the check whether this is the last segment will not work correctly in previous test framework, but will work in production.

14. SegmentTest - Fix: Prevent writing 0-sized block when int segment range cannot contain any valid rows. This may happen with large number of splits in the test.

15. SegmentTest - Improved some loggings to better track operations and understand what happened.

- [ ] This PR is based on https://github.com/pingcap/tiflash/pull/5835. Merge it first to get a nicer diff.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -163,71 +163,85 @@ SegmentPair DeltaMergeStore::segmentSplit(DMContext & dm_context, const SegmentP
...
+            }
         }
-        if (!isSegmentValid(lock, right))
+
+        for (const auto & seg : ordered_segments)
         {
-            LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because right not valid, left={} right={}", left->simpleInfo(), right->simpleInfo());
-            return;
-        }
+            // TODO: Should we ensure the ordering of "segments" first?
```
JaySon-Huang, line 222: The segments ordering is checked inside `Segment::prepareMerge`. Seems it is good enough?
https://github.com/pingcap/tiflash/blob/9a975379a502fead7e6c148de2597c0faef519dd/dbms/src/Storages/DeltaMerge/Segment.cpp#L1273-L1277

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -163,71 +163,85 @@ SegmentPair DeltaMergeStore::segmentSplit(DMContext & dm_context, const SegmentP
...
         std::shared_lock lock(read_write_mutex);
 
-        if (!isSegmentValid(lock, left))
+        for (const auto & seg : ordered_segments)
         {
-            LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because left not valid, left={} right={}", left->simpleInfo(), right->simpleInfo());
-            return;
+            if (!isSegmentValid(lock, seg))
+            {
+                LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because not valid, segment={}", seg->info());
```
JaySon-Huang, line 215: I think using `seg->simpleInfo` here is good enough.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -163,71 +163,85 @@ SegmentPair DeltaMergeStore::segmentSplit(DMContext & dm_context, const SegmentP
...
-            return;
-        }
+            // TODO: Should we ensure the ordering of "segments" first?
 
-        left_snap = left->createSnapshot(dm_context, /* for_update */ true, CurrentMetrics::DT_SnapshotOfSegmentMerge);
-        right_snap = right->createSnapshot(dm_context, /* for_update */ true, CurrentMetrics::DT_SnapshotOfSegmentMerge);
+            auto snap = seg->createSnapshot(dm_context, /* for_update */ true, CurrentMetrics::DT_SnapshotOfSegmentMerge);
+            if (!snap)
+            {
+                LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because snapshot failed, segment={}", seg->info());
```
JaySon-Huang, line 227: ditto

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -242,37 +256,40 @@ void DeltaMergeStore::segmentMerge(DMContext & dm_context, const SegmentPtr & le
...
 
-        if (!isSegmentValid(lock, left) || !isSegmentValid(lock, right))
+        for (const auto & seg : ordered_segments)
         {
-            LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because left or right not valid, left={} right={}", left->simpleInfo(), right->simpleInfo());
-            wbs.setRollback();
-            return;
+            if (!isSegmentValid(lock, seg))
+            {
+                LOG_FMT_DEBUG(log, "Merge - Give up segmentMerge because not valid, segment={}", seg->info());
```
JaySon-Huang, line 272: ditto

JaySon-Huang (approved):
LGTM with minor comments

flowbehappy (approved):
LGTM

lidezhu (approved):

PR #5860, opened 2022-09-13
clean up segments with large space amplification ratio generated by logical split

### LINKED ISSUE #5817: Clean up segments referencing a small subset of DTFile pack to reduce space amplification
## Enhancement

When doing logical split, we may experience at most 2x space amplification. It happens when a segment is referencing a very small subset of a DTFile stable layer while the rest subset of DTFile is already useless and not referenced any more in other segments.

In this case, we could discover such segments in the GC thread and rewrite the stable layer, so that this DTFile can be removed and clean up the space.

lidezhu:

Issue Number: close #5817 

Problem Summary: After logical split, the result segments will ref a `DTFile` whose range is almost twice as larger than the corresponding segments and these cause space amplification problem which cannot be ignored.

1. calculate the ratio of the valid_size/valid_bytes of `StableValueSpace` to the size/bytes of underlying `DTFile` and trigger a bg gc if the ratio is too low.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
1. enable logical split and load ch-benchmark;
2. restart to check whether the store size change;
3. patch new binary to check whether the store size change;
After patch, the bg gc start to clean the segments generated by logical split and the store size begin to decrease:
image
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -249,9 +249,32 @@ bool DeltaMergeStore::handleBackgroundTask(bool heavy)
 
 namespace GC
 {
+enum Type
+{
+    Unknown,
+    TooManyDeleteRange,
+    TooLargeDTFileRange,
```
flowbehappy, line 256: I think `TooMuchOutOfRange` is better. The thing here is not too large range, but small range.

dbms/src/Storages/DeltaMerge/Segment.h
```diff
@@ -424,6 +427,7 @@ class Segment : private boost::noncopyable
     const StableValueSpacePtr stable;
 
     bool split_forbidden = false;
+    std::atomic<bool> check_valid_data_ratio = false;
```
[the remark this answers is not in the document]
lidezhu: Added.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
lidezhu: Fixed.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -291,8 +314,26 @@ bool shouldCompactDeltaWithStable(const DMContext & context, const SegmentSnapsh
     //   because before write apply snapshot file, it will write a delete range first, and will meet the following gc criteria.
     //   But the cost should be really minor because merge delta on an empty segment should be very fast.
     //   What's more, we can ignore this kind of delete range in future to avoid this extra gc.
-    bool should_compact = (delete_rows >= stable_rows * ratio_threshold) || (delete_bytes >= stable_bytes * ratio_threshold);
-    return should_compact;
+    return (delete_rows >= stable_rows * invalid_data_ratio_threshold) || (delete_bytes >= stable_bytes * invalid_data_ratio_threshold);
+}
+
+bool shouldCompactStableWithTooMuchDataOutSegmentRange(const SegmentSnapshotPtr & snap, double invalid_data_ratio_threshold, const LoggerPtr & log)
```
flowbehappy, line 320:
```suggestion
bool shouldCompactStableWithTooMuchDataOutOfSegmentRange(const SegmentSnapshotPtr & snap, double invalid_data_ratio_threshold, const LoggerPtr & log)
```

flowbehappy (approved):
LGTM

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -291,8 +314,26 @@ bool shouldCompactDeltaWithStable(const DMContext & context, const SegmentSnapsh
...
+    size_t total_bytes = 0;
+    for (const auto & file : dm_files)
+    {
+        total_rows += file->getRows();
+        total_bytes += file->getBytes();
+    }
+
+    LOG_FMT_TRACE(log, "valid_rows [{}], valid_bytes [{}] total_rows [{}] total_bytes [{}]", valid_rows, valid_bytes, total_rows, total_bytes);
+
+    return (valid_rows < total_rows * (1 - invalid_data_ratio_threshold)) || (valid_bytes < total_bytes * (1 - invalid_data_ratio_threshold));
```
breezewish, line 336: Does this mean, if we have logical split enabled, then logical split will be nearly useless (and becomes very similar to physical split), because all segments split-out by logical split usually only contains 50% data (which < the 70% threshold)?

lidezhu: Yes, but the heavy work is moved to background gc thread. And I thought this is exactly the solution that we talked about previously(

breezewish: I'm not sure, but may be a better way would be counting by DTFiles, for example, check whether a DTFile is < 70% utilized. When a DTFile is < 70% utilized, it means delta-merging all segments who use this DTFile can result in reclaiming 30% space. Otherwise delta-merging these segments will not have notable benefits.

However, the newly proposed check could be more complex and I'm not sure whether covering this case is useful. @JaySon-Huang @flowbehappy What do you think?

breezewish (commented):
Except for the comment in https://github.com/pingcap/tiflash/pull/5860/files#r969439210, the rest LGTM.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
lidezhu: Got it. This seems really a complex solution.

lidezhu: Add a check for whether this segment share any DTFile with neighbor segments before check the invalid data ratio.
So for segments generated by logical split, it will only do clean work after its neighbor has been updated by merge delta/merge/physical split.

For DTFile which is shared by two segments, this is exactly the correct behavior we want.

For DTFile which is shared by more than two segments, 
1. this is really a rare case, because it is hard to do yet another logical split on a segment which is generated by logical split.(Check `Segment::getSplitPointFast`);
2. even it happens, if one of the multiple segments is updated and on longer ref the shared DTFile, it is beneficial to do the clean work on the remaining segments;

So I think the current solution is good enough for our purpose.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -317,22 +315,77 @@ bool shouldCompactDeltaWithStable(const DMContext & context, const SegmentSnapsh
...
-    auto valid_bytes = snap->stable->getBytes();
+    std::unordered_set<UInt64> file_ids;
+    // We get the file ids in the segment no matter it is abandoned or not,
+    // because even it is abandoned, we don't know whether the new segment will ref the old dtfiles.
+    // So we just be conservative here to return the old file ids.
+    // This is ok because the next check will get the right file ids in such case.
+    if (seg)
+    {
+        const auto & dm_files = seg->getStable()->getDMFiles();
+        for (const auto & file : dm_files)
```
breezewish, line 328: BTW I'm curious, when will we have multi DTfiles in the stable currently?

breezewish (approved):

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
lidezhu: Actually there is just one DTFile now. But we keep the interface like this to avoid the assumption that there is only one DTFile for future extenstion.

PR #5878, opened 2022-09-14
feat: Enable minmax index on the pack with nullable value

### LINKED ISSUE #5850: Enhance null value support for minmax index
## Enhancement
Currently, we not enable the minmax index when the pack has null values. Considering we now support nullable datatype in minmax index(ref #4787)  , we can support minmax index for the packs with null values, to enhance the query performance.

Besides, we also want support  'is null' , 'is not null' predicates with our minmax index.

hongyunyan:

Issue Number: close #5850

Problem Summary:

1. exclude the null values in the calculation of the minmax index.
2. enable to use minmax index on the pack with nuallable value.
3. support the function 'is null' and 'is not null' to use our minmax index to make filter. 
4. make more tests on the nullable column and is null filter.
5. make some tests to guarantee the requirements of the correctness on null values.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
```diff
@@ -310,6 +313,48 @@ RSOperatorPtr parseTiExpr(const tipb::Expr & expr,
...
+            {
+                op = createUnsupported(
+                    expr.ShortDebugString(),
+                    "filter IsNull with " + DB::toString(expr.children_size()) + " children",
+                    false);
+            }
+            else
+            {
+                const auto & child = expr.children(0);
+                if (likely(isColumnExpr(child)))
```
hehechen, line 335: Maybe we can't assume that users don't write sql like `isnull(function(xx))`

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -510,6 +586,19 @@ RSResult MinMaxIndex::checkGreaterEqual(size_t pack_index, const Field & value,
     return RSResult::Some;
 }
 
+RSResult MinMaxIndex::checkIsNull(size_t pack_index)
+{
+    if (!(*has_value_marks)[pack_index])
+    {
+        return RSResult::All;
```
hehechen, line 593: Maybe there is no null value, but all value are deleted? I think here should return `RSResult::Some`.

dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
```diff
@@ -1246,6 +1329,57 @@ try
...
+            }
+            auto type_value_pair = generateTypeValue(static_cast<MinMaxTestDatatype>(datatype), true);
+            ASSERT_EQ(true, checkMatch(case_name, *context, type_value_pair.first, type_value_pair.second, generateRSOperator(static_cast<MinMaxTestDatatype>(datatype), static_cast<MinMaxTestOperator>(operater_type), true)));
+            ASSERT_EQ(true, checkMatch(case_name, *context, type_value_pair.first, type_value_pair.second, generateRSOperator(static_cast<MinMaxTestDatatype>(datatype), static_cast<MinMaxTestOperator>(operater_type), false)));
+        }
+    }
+}
+CATCH
+
+
```
hehechen, line 1382: Can add `isnull` UT for scenarios of whole pack deleted, like `DMMinMaxIndexTest.DelMark`.

dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
hongyunyan: Do you means the statement like where isnull(a>1)?
But now we only assume the filter IsNull can handle the case for column only, just like isnull(a), but not isnull(a>1) or a > 1 is null, because it may cause some correctness problem. So for the statement like where isnull(a>1), I think we should make the filter unsupported.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
hongyunyan: Yes, I forget the deleted case. I fix it now.

dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
hongyunyan: Yes, I add a new case in DMMinMaxIndexTest.DelMark.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
hongyunyan: I find another point. For our minmax index, whatever the RSResult is Some or ALL, the use_pack[i] is always true. In this case, whether we return All or return Some in this case is no difference.
But if the case is not(isnull), and then we have a pack contains deleted values and null. If we return All here, the final result of not(isnull) is None, then we don't need to read this pack. But if we return Some here, the final result of not(isnull) is Some, then we still need to read this pack.(Although the case I think may be very rare)

dbms/src/Storages/tests/gtest_filter_parser.cpp
```diff
@@ -367,6 +368,39 @@ try
...
+        // And with IsNULL
+        auto rs_operator = generateRsOperator(table_info_json, "select * from default.t_111 where col_2 = 789 and col_3 is null");
+        EXPECT_EQ(rs_operator->name(), "and");
+        EXPECT_EQ(rs_operator->getAttrs().size(), 2);
+        EXPECT_EQ(rs_operator->getAttrs()[0].col_name, "col_2");
+        EXPECT_EQ(rs_operator->getAttrs()[0].col_id, 2);
+        EXPECT_EQ(rs_operator->getAttrs()[1].col_name, "col_3");
+        EXPECT_EQ(rs_operator->getAttrs()[1].col_id, 3);
+        EXPECT_EQ(rs_operator->toDebugString(), "{\"op\":\"and\",\"children\":[{\"op\":\"equal\",\"col\":\"col_2\",\"value\":\"789\"},{\"op\":\"isnull\",\"col\":\"col_3\"}]}");
+    }
```
JaySon-Huang, line 382: do we support "col_3 is not null" parsed as RSOperator?

hongyunyan: yes, it will be parse like not(isnull(xxx))

hongyunyan: And TiDB will transfer `is not null` to `not(isnull(xx)`.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -79,9 +110,22 @@ void MinMaxIndex::addPack(const IColumn & column, const ColumnVector<UInt8> * de
                 break;
             }
         }
+
+        if (has_null)
+        {
+            auto min_max_pair = details::minmax(column, del_mark, null_mark_data, 0, column.size());
+            min_index = min_max_pair.first;
+            max_index = min_max_pair.second;
```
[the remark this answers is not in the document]
hongyunyan: Yes, I change the code now.

JaySon-Huang (approved):

hehechen (approved):

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -58,12 +61,40 @@ inline std::pair<size_t, size_t> minmax(const IColumn & column, const ColumnVect
 
     return {batch_min_idx, batch_max_idx};
 }
+
+// minmax calculation for the nullable column
+inline std::pair<size_t, size_t> minmax(const IColumn & column, const ColumnVector<UInt8> * del_mark, const PaddedPODArray<UInt8> & null_mark, size_t offset, size_t limit)
```
breezewish, line 66: Better to add comment that this function actually bypasses the NULL (does not treat NULL as what its semantic is supposed to be) and then calculate a MinMax only for not NULL values.

It is definitely not a function calculating MinMax for a nullable column.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -159,6 +199,15 @@ std::pair<UInt64, UInt64> MinMaxIndex::getUInt64MinMax(size_t pack_index)
 RSResult MinMaxIndex::checkNullableEqual(size_t pack_index, const Field & value, const DataTypePtr & type)
 {
     const auto & column_nullable = static_cast<const ColumnNullable &>(*minmaxes);
+    const auto & null_map = column_nullable.getNullMapColumn();
+    bool min_is_null = null_map.getElement(pack_index * 2);
+
+    // if min value is null, then it must be the old version of minmax index.
```
breezewish, line 205: May be you could describe it as "prior to v6.x, the min max index may ...." instead of using "old", because "old" and "new" is not a good description after this PR merged into master. Consider the case, when you look at the code, you see some code describe "old" and "new" -- you don't even know how old the old is.. and this piece of code may exist even years ago.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -397,7 +457,18 @@ RSResult MinMaxIndex::checkNullableGreaterEqual(size_t pack_index, const Field &
 {
     const auto & column_nullable = static_cast<const ColumnNullable &>(*minmaxes);
 
+    const auto & null_map = column_nullable.getNullMapColumn();
+    bool min_is_null = null_map.getElement(pack_index * 2);
+
+    // if min value is null, then it must be the old version of minmax index.
+    // In this case, we directly return RSResult::Some, because we can not know about the real min value.
+    if (min_is_null)
```
breezewish, line 465: Seems that this logic is not checked by the unit test, could you add some tests for it?

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -29,6 +29,9 @@
 #include <Storages/DeltaMerge/Index/MinMaxIndex.h>
 #include <Storages/DeltaMerge/Index/RoughCheck.h>
 
+#include "Storages/DeltaMerge/Index/RSResult.h"
```
breezewish, line 32: We prefer `#include <...>`

breezewish (approved):
The rest LGTM.

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -29,6 +29,8 @@
 #include <Storages/DeltaMerge/Index/MinMaxIndex.h>
 #include <Storages/DeltaMerge/Index/RoughCheck.h>
 
+#include "common/types.h"
```
breezewish, line 32: Also this one.

dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
```diff
@@ -30,6 +30,8 @@
 #include <ext/scope_guard.h>
 #include <memory>
 
+#include "Storages/DeltaMerge/Index/RSResult.h"
```
breezewish, line 33: And this one

Issue #5859, opened 2022-09-13
Decode error when "NULL" value in the column with "primary key" flag


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

TiFlash assumes all values in the column with "primary key" flag are non-null value. https://github.com/pingcap/tiflash/blob/6d0cbc8cc78b4048fb88e0845d6c9c2881e01515/dbms/src/Storages/Transaction/RowCodec.cpp#L323-L330

However, users could create such "NULL" value in the "primary key" column by following steps

```sql
-- create table with `source` is nullable
create table test.t_case (
`case_no` varchar(32) not null,
`source` varchar(20) default null,
`p` varchar(12) DEFAULT NULL,
primary key (`case_no`)
);

-- insert some data and left `source` to be empty
insert into test.t_case(case_no)
values ("1"), ("2"), ("3"), ("4");

-- drop the primary key, fill the `source` to be non-empty
alter table t_case drop primary key;
update test.t_case set `source` = '' where `source` is NULL;
-- add new primary key with case_no and source
alter table t_case add primary key (case_no, source);

-- send the snapshot data to tiflash
alter table t_case set tiflash replica 1;
```

### 2. What did you expect to see? (Required)

No exception thrown

### 3. What did you see instead (Required)

Exception throw with "Parameter out of bound in IColumnString::insertRangeFrom method" in v4.0.x/v5.0.x/v5.1.x/v5.2.x/v5.3.x/v5.4.0~v5.4.1/v6.1.0

```
[2022/09/13 15:30:02.786 +08:00] [ERROR] [<unknown>] ["bool DB::GCManager::work():Code: 12, e.displayText() = DB::Exception: Parameter out of bound in IColumnString::insertRangeFrom method, start=0, length=8, src_size=4:  while reading DTFile /data2/jaysonhuang/tiup-deploy/data/tiflash-5018/data/t_77/stable/dmf_2: while apply gc Segment [1] [range=[-9223372036854775808,9223372036854775807)] [table=t_77], e.what() = DB::Exception, Stack trace:
0. bin/tiflash/tiflash(StackTrace::StackTrace()+0x16) [0x36d42e6]
1. bin/tiflash/tiflash(DB::Exception::Exception(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int)+0x26) [0x36c8116]
[...]

PR #5879, opened 2022-09-14
Fix decode error when "NULL" value in the column with "primary key" flag

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/5859

Problem Summary:
TiFlash assumes all values in the column with "primary key" flag are non-null value.

However, users could create such "NULL" value in the "primary key" column by following steps

```sql
-- create table with `source` is nullable
create table test.t_case (
`case_no` varchar(32) not null,
`source` varchar(20) default null,
`p` varchar(12) DEFAULT NULL,
primary key (`case_no`)
);

-- insert some data and left `source` to be empty
insert into test.t_case(case_no)
values ("1"), ("2"), ("3"), ("4");

-- drop the primary key, fill the `source` to be non-empty
alter table t_case drop primary key;
update test.t_case set `source` = '' where `source` is NULL;
-- add new primary key with case_no and source
alter table t_case add primary key (case_no, source);

-- send the snapshot data to tiflash
alter table t_case set tiflash replica 1;
```

* mainly change the logic in `addDefaultValueToColumnIfPossible`
  * For non-clustered index table(is_common_handle=false && pk_is_handle=false), fill default value even if the column contains primary key flag when the column is not appear in the encoded value(those rows are not supposed to be return to users)
  * For clustered index/pk_is_handle table, ignore the primary key column as they are supposed to be decoded from the encoded key
* Refine the exception message in `insertRangeFrom`
* Add more region info when exception thrown in `preHandleSnapshot`/`ingestSST`

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the panic issue after creating the primary index with a column containing `NULL` value
```


tests/fullstack-test2/ddl/alter_pk.test
```diff
@@ -57,3 +57,56 @@ mysql> alter table test.t drop primary key;
...
+## drop the primary key, fill the `source` to be non-empty
+## add new primary key with case_no and source
+mysql> alter table test.t_case drop primary key;
+mysql> update test.t_case set `source` = '' where `source` is NULL;
+mysql> alter table test.t_case add primary key (`case_no`, `source`);
+
+## send the snapshot data to tiflash
+mysql> alter table test.t_case set tiflash replica 1;
+func> wait_table test t_case
+mysql> select * from test.t_case;
```
JaySon-Huang, line 77:
```suggestion
mysql> select case_no,p,source from test.t_case;
```

tests/fullstack-test2/ddl/alter_pk.test
```diff
@@ -57,3 +57,56 @@ mysql> alter table test.t drop primary key;
...
+
+mysql> alter table test.t_case add column `source` varchar(20) not null;
+## drop the primary key, add new primary key with case_no and source
+mysql> alter table test.t_case drop primary key;
+mysql> alter table test.t_case add primary key (`case_no`, `source`);
+
+## send the snapshot data to tiflash
+mysql> alter table test.t_case set tiflash replica 1;
+func> wait_table test t_case
+mysql> select * from test.t_case;
```
JaySon-Huang, line 102:
```suggestion
mysql> select case_no,p,source from test.t_case;
```

JaySon-Huang:
The helper functions in `RegionBlockReaderTestFixture` is buggy when testing with primary key columns. I will split it into another PR to refine the test suite first.

dbms/src/Storages/Transaction/RowCodec.cpp
```diff
@@ -285,53 +290,96 @@ void encodeRowV2(const TiDB::TableInfo & table_info, const std::vector<Field> &
...
+        return is_big ? appendRowV2ToBlockImpl<true>(raw_value, column_ids_iter, column_ids_iter_end, block, block_column_pos, column_infos, pk_handle_id, fill_missing_column_with_pk_flag, force_decode)
+                      : appendRowV2ToBlockImpl<false>(raw_value, column_ids_iter, column_ids_iter_end, block, block_column_pos, column_infos, pk_handle_id, fill_missing_column_with_pk_flag, force_decode);
+    }
+    default:
+        return appendRowV1ToBlock(raw_value, column_ids_iter, column_ids_iter_end, block, block_column_pos, column_infos, pk_handle_id, fill_missing_column_with_pk_flag, force_decode);
+    }
 }
 
-inline bool addDefaultValueToColumnIfPossible(const ColumnInfo & column_info, Block & block, size_t block_column_pos, bool force_decode)
+inline bool addDefaultValueToColumnIfPossible(const ColumnInfo & column_info, Block & block, size_t block_column_pos, bool fill_column_with_pk_flag, bool force_decode)
```
hongyunyan, line 356: I think the variable name "fill_column_with_pk_flag" may by a little difficult to understand.

JaySon-Huang: what about `table_is_non_clustered`? The name means that the keys for row data consist of internal `_tidb_rowid` implicitly assigned by TiDB, and we should fill the missing value for those columns with primary key flags under this situation

https://docs.pingcap.com/tidb/dev/clustered-indexes#clustered-indexes

lidezhu: How about `ignore_pk_if_absent`?

hongyunyan: I think `ignore_pk_if_absent` is more straight forward to understand.

lidezhu (approved):
LGTM

dbms/src/Storages/Transaction/RowCodec.cpp
JaySon-Huang: updated

hongyunyan (approved):
LGTM

dbms/src/Storages/Transaction/RowCodec.cpp
```diff
@@ -290,53 +290,96 @@ void encodeRowV2(const TiDB::TableInfo & table_info, const std::vector<Field> &
...
+        if (ignore_pk_if_absent)
+            return true;
+
+        assert(!ignore_pk_if_absent);
+        if (!force_decode)
+            return false;
+        // Else non-clustered index, and not pk_is_handle, it could be a row encoded by older schema,
+        // we need to fill the column wich has primary key flag with default value.
+        // fallthrough to fill default value when force_decode
+        LOG_FMT_WARNING(Logger::get("RowCodec"), "fill default value for missing primary key column, column_id={} column_name={}", column_info.id, column_info.name);
```
flowbehappy, line 373: I'm afraid this warning too chatty....

flowbehappy: I suggest using trace

flowbehappy (approved):
LGTM with a comment

dbms/src/Storages/Transaction/RowCodec.cpp
JaySon-Huang, line 373:
```suggestion
```

JaySon-Huang: removed

breezewish:
This PR's description really refreshed my mind LOL

PR #5890, opened 2022-09-14
storage: Allow specifying split point and split method in the segment API

breezewish:

Issue Number: ref #5237

1. API: Added `opt_split_point` to segment split APIs, so that the caller can split at an arbitrary position. 

   When `opt_split_point` is not specified, a calculated mid split point will be used (the same as previous).

   Note: Now it is allowed to specify a split point and split the segment even when segment is empty.

2. API: Added `opt_split_mode` to segment split APIs, so that the caller can force to use LogicalSplit or PhysicalSplit.

   - Mode = Auto: The same as previous behavior
   - Mode = Logical: Always do a logical split. Failed if mid logical split cannot be figured.
   - Mode = Physical: Always do a physical split.

   Fail point `try_segment_logical_split` and `force_segment_logical_split` is deprecated in favor of the split mode parameter.

3. Test: Added a lot of tests to cover split at middle and split at a point.

4. Test: Rearranged randomized segment test. It is taken out to be a standalone test file. The name of the randomized test is changed, which it is fine, because the DISABLED test is actually not enabled in the CI.

6. Test: Cover split mode and customized split point in the randomized segment test.

7. Test: Limit the randomized test to run over a small ranged segment, so that data related operations are more likely to be covered.

8. Test: Enable randomized test for Int handle.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMergeStore.h
```diff
@@ -483,11 +458,81 @@ class DeltaMergeStore : private boost::noncopyable
...
     /**
      * Split the segment into two.
      * After splitting, the segment will be abandoned (with `segment->hasAbandoned() == true`) and the new two segments will be returned.
+     *
+     * When `opt_split_at` is not specified, this function will try to find a mid point for splitting, and may lead to failures.
      */
-    SegmentPair segmentSplit(DMContext & dm_context, const SegmentPtr & segment, bool is_foreground);
+    SegmentPair segmentSplit(DMContext & dm_context, const SegmentPtr & segment, SegmentSplitReason reason, std::optional<RowKeyValue> opt_split_at = std::nullopt, SegmentSplitMode opt_split_mode = SegmentSplitMode::Auto);
+
+#pragma endregion Segment Split
```
flowbehappy, line 535: I'm sorry what is it used for?

breezewish: It triggers a collapseable block in IDEs like VS, Intellij Idea and VSCode via extensions, like this, so that we could better organize large files:

image

image


However I just figured out that GCC will generate a warning for this instruction (although I don't have an environment to verify it):
https://gcc.gnu.org/bugzilla//show_bug.cgi?id=85487

So maybe I should drop it when we are still considering GCC compatibility (in the README)?

flowbehappy (approved):
LGTM

JaySon-Huang (approved):
LGTM

PR #5904, opened 2022-09-15
storage: Introduce replaceData in the segment API

breezewish:

Issue Number: ref #5237

1. In addition to current segment operations Split, Merge, MergeDeta, a new operation called ReplaceData is introduced.

   ReplaceData unconditionally replace everything in the segment to be the specified data.

   There are many restrictions to correctly use the ReplaceData API, as documented. This may be improved in future:

```c++
   /**
     * Discard all data in the current delta and stable layer, and use the specified DMFile as the stable instead.
     * This API does not have a prepare & apply pair, as it should be quick enough. The specified DMFile is safe
     * to be shared for multiple segments.
     *
     * Note 1: Should be protected behind the Segment update lock to ensure no new data will be appended to this
     *         segment during the function call. Otherwise these new data will be lost in the new segment.
     *
     * Note 2: This function will not enable GC for the new_stable_file for you, in case of you may want to share the same
     *         stable file for multiple segments. It is your own duty to enable GC later.
     *
     * Note 3: You must ensure the specified new_stable_file has been managed by the storage pool, and has been written
     *         to the PageStorage's data. Otherwise there will be exceptions.
     *
     * Note 4: This API is subjected to be changed in future, as it relies on the knowledge that all current data
     *         in this segment is useless, which is a pretty tough requirement.
     */
    [[nodiscard]] SegmentPtr dangerouslyReplaceData(const Lock &, DMContext & dm_context, const DMFilePtr & data_file, WriteBatches & wbs) const;
```

   The ReplaceData API is currently not used in the production execution path. It is only covered by unit tests.

2. Segment Test Framework: Share same implementation for Write / Ingest / WriteDeletedPack.

3. Test: Cover replaceData in segment randomized tests (iterated for 50000 times).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -706,6 +706,65 @@ SegmentPtr Segment::applyMergeDelta(const Segment::Lock &, //
...
+                                            new_stable);
+    new_me->serialize(wbs.meta);
+
+    delta->recordRemoveColumnFilesPages(wbs);
+    stable->recordRemovePacksPages(wbs);
+
+    LOG_FMT_DEBUG(log, "ReplaceData - Finish, old_me={} new_me={}", info(), new_me->info());
+
+    return new_me;
+}
```
breezewish, line 766: For reviewers: Please carefully review the logic in this function. Thank you!

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -487,6 +487,46 @@ SegmentPtr DeltaMergeStore::segmentMergeDelta(
...
+    {
+        std::unique_lock lock(read_write_mutex);
+        if (!isSegmentValid(lock, segment))
+        {
+            LOG_FMT_DEBUG(log, "ReplaceData - Give up segment replace data because segment not valid, segment={} data_file={}", segment->simpleInfo(), data_file->path());
+            return {};
+        }
+
+        auto segment_lock = segment->mustGetUpdateLock();
+        new_segment = segment->dangerouslyReplaceData(segment_lock, dm_context, data_file, wbs);
```
flowbehappy, line 509: If the usage scenario is to first split an empty segment, then ingest DTFile to stable layer, then how about we make sure the segment is empty before calling `segment->dangerouslyReplaceData`?

breezewish: Awesome! This could be a possible solution to support ingestBySplit for IngestSST scenario. For ApplySnapshot scenario, as we always discard existing data, so there is no need to check.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
```diff
@@ -487,6 +487,46 @@ SegmentPtr DeltaMergeStore::segmentMergeDelta(
...
+        wbs.writeLogAndData();
+        wbs.writeMeta();
+
+        segment->abandon(dm_context);
+        segments[segment->getRowKeyRange().getEnd()] = new_segment;
+        id_to_segment[segment->segmentId()] = new_segment;
+        RUNTIME_CHECK(compare(segment->getRowKeyRange().getEnd(), new_segment->getRowKeyRange().getEnd()) == 0, segment->info(), new_segment->info());
+        RUNTIME_CHECK(segment->segmentId() == new_segment->segmentId(), segment->info(), new_segment->info());
```
JaySon-Huang, line 517: These two checks should be done before `wbs.writeLogAndData`. Checking it after we commit the new segment to disk and memory is useless

JaySon-Huang (commented):
Rest LGTM

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
flowbehappy: So `dangerouslyReplaceData` is designed to be used by apply snapshot?

flowbehappy (approved):

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
breezewish: For this sprint, yes. For next sprint we could change its semantic to something like: tryIngestData (no longer dangerous), which:
1. Checks if the target is empty --- do segment replace
2. If not empty --- do ingest

lidezhu (approved):
LGTM

JaySon-Huang (approved):
LGTM

PR #5927, opened 2022-09-16
storage: Ingest snapshot by split

breezewish:

Issue Number: ref #5237

Use LogicalSplitAt + ReplaceData to ingest snapshots when `clear_data_in_range == true`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test

**Workload: Ingest 100GB data using Full-Speed, then query a few times**

this PR vs master:

image
image
image
image
image
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Improve ApplySnapshot performance
```


dbms/src/Storages/DeltaMerge/ExternalDTFileInfo.h
```diff
@@ -0,0 +1,45 @@
...
+
+struct ExternalDTFileInfo
+{
+    /**
+     * The allocated PageId of the file.
+     */
+    PageId id;
+
+    /**
+     * The handle range of contained data.
```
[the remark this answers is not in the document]
breezewish: Actually according to the unit test I see that the handle range would be never nullopt. I'm discovering how to remove this optional wrapper safely. Thanks for pointing it out!

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -210,6 +140,412 @@ void DeltaMergeStore::ingestFiles(
...
+        else
+        {
+            RUNTIME_CHECK(!external_file.range);
+        }
+
+        files.emplace_back(std::move(file));
+    }
+
+    bool ingest_using_split = false;
+    if (clear_data_in_range && bytes >= 4 * 1024 * 1024)
```
JaySon-Huang, line 494: maybe reuse `dm_context.delta_small_column_file_bytes` for the threshold is better

breezewish:
Not sure how integration test failed, but looks like the storage meta data is corrupted, which seems to be not related with ingestion:

DB::Exception: Cannot open file /tmp/tiflash/data/db/metadata/db_401/t_398.sql, errno: 2, strerror: No such file or directory:  table name: t_398, table id: 398 

Cannot reproduce the problem locally. Try rerun again.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -210,6 +140,412 @@ void DeltaMergeStore::ingestFiles(
...
+        /*
+         * The segment and the ingest range is perfectly matched. We can
+         * simply replace all of the data from this segment.
+         *
+         * Example:
+         *    │----------- Segment ----------│
+         *    │-------- Ingest Range --------│
+         */
+        const auto new_segment_or_null = segmentDangerouslyReplaceData(dm_context, segment, file);
+        const bool succeeded = !!new_segment_or_null;
```
flowbehappy, line 391: Why `!!`?

breezewish: `new_segment_or_null` is a `std::optional<...>`, so `!!` casts it to boolean unconditionally

JaySon-Huang: maybe use `new_segment_or_null != nullptr`

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -210,6 +140,412 @@ void DeltaMergeStore::ingestFiles(
...
+         *
+         * We will logical split the segment to form a perfect matching segment:
+         *    │--------------- Segment ------│--------│
+         *    │-------- Ingest Range --------│
+         */
+        const auto [left, right] = segmentSplit(dm_context, segment, SegmentSplitReason::IngestBySplit, ingest_range.end, SegmentSplitMode::Logical);
+        // We don't care whether it is succeeded or not, because the caller always need to retry
+        // from the beginning in this case.
+        UNUSED(left);
+        UNUSED(right);
```
JaySon-Huang, line 409:
```suggestion
        UNUSED(left, right);
```
`UNUSED` can accept multiple prarms

JaySon-Huang:
We can add some tests that applySnapshot and read/applySnapshot and background segmentMergeDelta run concurrently (maybe in later PR)

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
breezewish: Updated. Thanks for the suggestion!

dbms/src/Storages/DeltaMerge/ExternalDTFileInfo.h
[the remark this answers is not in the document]
breezewish: The optional wrapper has been replaced by some RUNTIME checks.

dbms/src/Storages/DeltaMerge/SSTFilesToDTFilesOutputStream.cpp
```diff
@@ -289,6 +307,42 @@ void SSTFilesToDTFilesOutputStream<ChildStream>::cancel()
...
+    {
+        current_file_range = RowKeyRange( //
+            block_start.toRowKeyValue(),
+            block_end,
+            schema_snap->is_common_handle,
+            schema_snap->rowkey_column_size);
+    }
+    else
+    {
+        RUNTIME_CHECK(compare(block_start, current_file_range->getStart()) > 0);
```
breezewish, line 340: @JaySon-Huang Will we split a key into multiple blocks (due to multiple versions)? If so, then this check may fail I think.

dbms/src/Storages/DeltaMerge/SSTFilesToDTFilesOutputStream.cpp
```diff
@@ -289,6 +307,42 @@ void SSTFilesToDTFilesOutputStream<ChildStream>::cancel()
...
+    const auto rowkey_column = RowKeyColumnContainer(pk_col.column, schema_snap->is_common_handle);
+    auto & current_file_range = ingest_files_range.back();
+
+    auto const block_start = rowkey_column.getRowKeyValue(0);
+    auto const block_end = rowkey_column.getRowKeyValue(pk_col.column->size() - 1) //
+                               .toRowKeyValue()
+                               .toPrefixNext(); // because range is right-open.
+
+    // TODO: what if there are a lof of versions, exceeding a block size? Will our checks fail?
+    RUNTIME_CHECK(compare(block_start, block_end.toRowKeyValueRef()) < 0);
```
JaySon-Huang, line 328: For rough set filter working correctly, `BoundedSSTFilesToBlockInputStream` ensures that if there are multiple rows with the same rowkey, they must be in the same block.

But we do need to handle that a block that only contains one rowkey. Calling `toPrefixNext` for getting the block_end may be wrong under this situation.

https://github.com/pingcap/tiflash/blob/b232bcd66ccc2684f9b0688daf941832077affd2/dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.h#L115-L117

breezewish: Interesting, when will we have a block containing only one key? The block size is small, or the versions of key is too much?

Seems that the boundary will be fine, as long as we don't meet one key in multiple blocks.

Consider:

```
[1, 1, 1, 1], [2, ....]
```

The range would be:

```
[1, 2) [2, ...)     (int handle)
[0x1, 0x2) [0x2, ...)  (common handle)
```

JaySon-Huang: When the versions of key is too much(8192 rows per block by default), we will get a block containing only one key.

JaySon-Huang: I get it, `toPrefixNext` solve the problem

dbms/src/Storages/DeltaMerge/SSTFilesToDTFilesOutputStream.cpp
JaySon-Huang: Never have overlapped ranges, the children `BoundedSSTFilesToBlockInputStreamPtr` ensure that for the correctness of rough set filter in DTFile

JaySon-Huang (approved):
LGTM

but I think it would be better to let 5879 get merged first. Or it will introduce some conflict for apply snapshot codes when cherry-picking 5879 to previous versions

breezewish:
> LGTM
> 
> but I think it would be better to let 5879 get merged first. Or it will introduce some conflict for apply snapshot codes when cherry-picking 5879 to previous versions

Makes sense. Let's merge #5879 first.

dbms/src/Storages/DeltaMerge/SSTFilesToDTFilesOutputStream.cpp
breezewish: Good news. I will add some checks in the next PR along with the test.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -210,6 +140,401 @@ void DeltaMergeStore::ingestFiles(
...
+     *  │               │-- DMFile --│- DMFile --│---- DMFile ----│- DMFile --│       │
+     *  │- Segment --│-- Seg --│------- Segment -----│- Seg -│------- Segment --------│
+     *
+     * This is what we will ingest:
+     * Iterate 0:
+     * -Inf                                                                         +Inf
+     *  │        │--------------- Ingest Range ---------------│                       │
+     *  │               │-- DMFile --│                                                │
+     *  │            │-- Seg --│------- Segment -----│                                │
+     *                   ↑              ↑ The segment we ingest DMFile into
```
flowbehappy, line 246:
```suggestion
     *                   ↑                The segment we ingest DMFile into
```

flowbehappy (approved):
LGTM

breezewish:
@JaySon-Huang @flowbehappy Great thanks for the review! Let me use the latest branch to perform some e2e tests as the final verifications. I will trigger the merge after that.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
breezewish: Oh, may be I should made comments more clear. I would like to illustrate what we will iterate for the outmost loop `for (size_t file_idx = 0; file_idx < files.size(); file_idx++)`.

PR #5942, opened 2022-09-17
Add boost macro BOOST_NO_CXX98_FUNCTION_BASE.

### LINKED ISSUE #5717: Build boost fail when using clang15 and libc++15
## Bug Report

Build tiflash using clang15 and libc++15

Build pass

tiflash/contrib/boost/boost/container_hash/hash.hpp:131:33: error: no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
        struct hash_base : std::unary_function<T, std::size_t> {};
                           ~~~~~^~~~~~~~~~~~~~
                                __unary_function
/usr/lib/llvm-15/bin/../include/c++/v1/__functional/unary_function.h:46:1: note: '__unary_function' declared here
using __unary_function = __unary_function_keep_layout_base<_Arg, _Result>;
^

master 61e436be45a9df7daa3869fd4794cd155ff9896b

`std::unary_function` is deprecated in C++11 and removed in C++17. From [Boost Macro Reference](https://www.boost.org/doc/libs/1_76_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.macros_that_describe_features_that_have_been_removed_from_the_standard_) 
 set `BOOST_NO_CXX98_FUNCTION_BASE` could solve this issue.

cakebytheoceanLuo:
Signed-off-by: Jigao Luo <luojigao@outlook.com>

Issue Number: ref #5717

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


contrib/boost-cmake/CMakeLists.txt
```diff
@@ -30,5 +30,7 @@ target_include_directories (boost_internal BEFORE PUBLIC ${Boost_INCLUDE_DIRS})
 
 target_compile_definitions (boost_internal PUBLIC BOOST_SYSTEM_NO_DEPRECATED)
 
+target_compile_definitions (boost_internal PUBLIC BOOST_NO_CXX98_FUNCTION_BASE)
```
fuzhe1989, line 33: This only applies to `boost_internal`, however can't solve errors from where boost hpp files are included.

fuzhe1989 (commented):
Please mark this pr as "ref #5717" since `BOOST_NO_CXX98_FUNCTION_BASE` is only one build error raised when clang15 and libc++15 are used.

Maybe we could wrap boost hpp files into some new header files like
https://github.com/pingcap/tiflash/blob/3124b24e3dea2b9783ba15c99ad1539ba6aae9a2/dbms/src/Common/grpcpp.h#L17-L24

fuzhe1989:
And thanks for your contribution!

cakebytheoceanLuo:
> only one build error raised when clang15 and libc++15 are used.

I am not sure. I have a GCC with 12.2.0 and Clang with 14.0.6, and I have this build failed due to boost.

contrib/boost-cmake/CMakeLists.txt
cakebytheoceanLuo: Yes. You are right. The reason is that `boost_internal` includes only cpp files: https://github.com/pingcap/tiflash/pull/5942/files/e0239ffac27e0f8c28c5e2d140422a28a58eac1a#diff-37a87c321b131d55ec9843b8701fdddc8720a0f0b22687e840fbafdd055dc8a8R22-R25

cakebytheoceanLuo: Is this reason that you think a hpp wrapper is much better than Cmake with Boost Marco?

fuzhe1989: I'm not a cmake expert. You can check compile_commands.json if other cpps also have this macro defined.

cakebytheoceanLuo: Sorry, my question is not clear enough and I was not asking about cmake. I want to ask, why the header wrapper is a better way than using Boost Macro? In my eyes, the Boost Macro solves the issue. 

---

I have read `tiflash/dbms/src/Common/grpcpp.h`, which is a wrapper of `<grpcpp/grpcpp.h>` and is only use inside the tiflash, not used in the external `contri` folder.

But we have this boost issue when compiling the boost inside the external `contri` folder: https://github.com/pingcap/tiflash/issues/5717#issuecomment-1250135583 
Adding a new `tiflash/dbms/src/Common/hash.h` is not helpful to the issue because `contri/boost` doesn't know the existence of `tiflash/dbms/src/Common/hash.h`. The new wrapper should be created under this folder https://github.com/pingcap/boost-extra , but I don't think it is a good idea.

fuzhe1989 (approved):
I've verified that by applying this patch, `BOOST_NO_CXX98_FUNCTION_BASE` is correctly set for all compile targets that are depending on `boost_internal`.

However new errors appear when using clang15 so there's still work to do.

fuzhe1989 (approved):
I've verified that by applying this patch, `BOOST_NO_CXX98_FUNCTION_BASE` is correctly set for all compile targets that are depending on `boost_internal`.

However new errors appear when using clang15 so there's still work to do.

JaySon-Huang (approved):
LGTM

Thanks for your contribution!

PR #5945, opened 2022-09-18
storage: Always split the pre-ingested DTFile

breezewish:

Issue Number: ref #5237

Consider that we received a 2GB-sized snapshot (which may happen when dynamic region is enabled).

- Improved: In ApplySnapshot, we use ingestBySplit. In this case, 2GB sized snapshot becomes a 2GB sized segment immediately, which causes segment split again. If we pre-split the DTFile, then the segment will be stable and don't need further operations.

- No change: In IngestSST, we use ingestByColumnFile. In this case, 2GB sized snapshots are all written to the segment delta layer. Even if we split into multiple DTFiles, they are still all written to the segment delta layer (for now). So there will be no changes.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test - There are e2e tests to cover the big region scenario.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/Transaction/ApplySnapshot.cpp
+++ b/dbms/src/Storages/Transaction/ApplySnapshot.cpp
@@ -328,6 +328,8 @@ std::vector<DM::ExternalDTFileInfo> KVStore::preHandleSSTsToDTFiles(
             }
             physical_table_id = storage->getTableInfo().id;
 
+            auto & global_settings = context.getGlobalContext().getSettingsRef();
+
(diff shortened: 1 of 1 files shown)


flowbehappy (approved):

hehechen (approved):

CalvinNeo (approved):
lgtm

JaySon-Huang (approved):

PR #6010, opened 2022-09-22
storage: GC segments contained by the pack

### LINKED ISSUE #5975: tiflash can't gc completely after set tiflash replica 0
## Bug Report

image
image

master

breezewish:

Issue Number: close #5975

Previously, we count segment rows by using the sum rows of overlapped packs. When the segment rows are < 0.7 rows of DTFiles, the DTFile will be regarded as "reclaimable" and the stable will be re-generated.

In this PR, we count segment rows by using the sum rows of **fully contained packs**. This means, the calculated segment rows will be smaller than before, and segments will more likely being GCed.

Here is an illustration of the new contained rule:

image

Consider the case where there is only 1 pack in the DTFile. When segment is fully or partially contained by this pack, the segment will be GCed, **no matter how many actual rows are there in the segment**. This leads to extra GCs in some cases, but the negative impact is small, as the cost of rewriting 1 pack should be small.

- [x] This PR is based on https://github.com/pingcap/tiflash/pull/5982. So merge it to get a nicer diff.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -451,29 +464,32 @@ bool shouldCompactStableWithTooMuchDataOutOfSegmentRange(const DMContext & conte
...
-        total_rows += file->getRows();
-        total_bytes += file->getBytes();
+        dt_file_rows += file->getRows();
+        dt_file_bytes += file->getBytes();
     }
-    auto valid_rows = snap->stable->getRows();
-    auto valid_bytes = snap->stable->getBytes();
 
-    auto check_result = (valid_rows < total_rows * (1 - invalid_data_ratio_threshold)) || (valid_bytes < total_bytes * (1 - invalid_data_ratio_threshold));
+    auto check_result = (at_least_result.rows < dt_file_rows * (1 - invalid_data_ratio_threshold)) //
```
flowbehappy, line 479: Better to add some documents to this method to explain the idea behind this.

flowbehappy (approved):
LGTM

lidezhu (approved):
LGTM

JaySon-Huang (approved):
lgtm

Issue #6016, opened 2022-09-23
Enhance minmax index to support more data types.


hongyunyan:
## Enhancement

Our minmax index now is only support for the handle column, and the columns whose data type satisfies isInteger() or isDateorDateTime()).

For lots of other data types, such as float, double, decimal, string, etc, minmax index is unsupported currently.

In order to enhance our index ability, we first can enhance our minmax index for more data types, to utilize our minmax index better.

JaySon-Huang:
For float, double - Comparing floating-point values is not an accurate operation, can we ensure it won't filter the wrong result? How does other database systems handle that?

For string
* The "min" "max" of random long string is usually useless for filtering. Unless we can ensure the strings are low-cardinality
* Generating the min max value of string column could slow down the write speed a lot
* We need to pay attention for the collation of string content when using it, which makes it difficult for comparing

There are more kinds of rough set indexes (or some systems call them zone-maps) other than min-max that could help us better filter the irrelevant data when executing queries.

PR #6017, opened 2022-09-23
Support minmax index for float and double datatype

hongyunyan:

Issue Number: ref #6016 

Problem Summary:

1. Support establish minmax index for columns whose data type is float or double.
2. Add the UT for float and double data types.

### Check List

Tests 

- [ x ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/File/DMFileWriter.cpp
+++ b/dbms/src/Storages/DeltaMerge/File/DMFileWriter.cpp
@@ -74,7 +74,7 @@ DMFileWriter::DMFileWriter(const DMFilePtr & dmfile_,
         // TODO: currently we only generate index for Integers, Date, DateTime types, and this should be configurable by user.
         /// for handle column always generate index
         auto type = removeNullable(cd.type);
-        bool do_index = cd.id == EXTRA_HANDLE_COLUMN_ID || type->isInteger() || type->isDateOrDateTime();
+        bool do_index = cd.id == EXTRA_HANDLE_COLUMN_ID || type->isNumber() || type->isDateOrDateTime();
         if (options.flags.isSingleFile())
         {
             if (do_index)
--- a/dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
+++ b/dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
@@ -47,6 +47,8 @@ inline bool isRoughSetFilterSupportType(const Int32 field_type)
     case TiDB::TypeLongLong:
     case TiDB::TypeInt24:
     case TiDB::TypeYear:
+    case TiDB::TypeFloat:
+    case TiDB::TypeDouble:
         return true;
     // For these date-like types, they store UTC time and ignore time_zone
     case TiDB::TypeNewDate:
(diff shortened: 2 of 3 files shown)


Lloyd-Pottiger (commented):
The code LGTM, but maybe a perf test is needed. And can we use statistics about the data distribution to judge whether we should use index?

hongyunyan:
Good point ! I will add a case to reflect the performance improvement of the new corresponding type

hongyunyan:
Good Point!
I will add a case to reflect the performance improvement for columns with float/double data types.

dbms/src/Storages/DeltaMerge/File/DMFileWriter.cpp
```diff
@@ -74,7 +74,7 @@ DMFileWriter::DMFileWriter(const DMFilePtr & dmfile_,
         // TODO: currently we only generate index for Integers, Date, DateTime types, and this should be configurable by user.
         /// for handle column always generate index
         auto type = removeNullable(cd.type);
-        bool do_index = cd.id == EXTRA_HANDLE_COLUMN_ID || type->isInteger() || type->isDateOrDateTime();
+        bool do_index = cd.id == EXTRA_HANDLE_COLUMN_ID || type->isNumber() || type->isDateOrDateTime();
```
breezewish, line 77: Not sure, but decimal may be also covered by `isNumber()` check?

breezewish (commented):
So we have already supported the MinMax for float and double and they were just not enabled?

hongyunyan:
> So we have already supported the MinMax for float and double and they were just not enabled?

I thought like that before, but @JaySon-Huang  reminder me that we need to consider that comparing floating-point values is not an accurate operation, we have to ensure it won't filter the wrong result. So I need to recheck the comparing logic here, to know whether it can work.

breezewish:
> > So we have already supported the MinMax for float and double and they were just not enabled?
> 
> I thought like that before, but @JaySon-Huang reminder me that we need to consider that comparing floating-point values is not an accurate operation, we have to ensure it won't filter the wrong result. So I need to recheck the comparing logic here, to know whether it can work.

Okay. This reminds me that there is another tricky thing you need to consider. The native IEEE 754 float is not total order, which may bring you some trouble when involving NaN, ±Inf, ±0.

Ref: 
https://stackoverflow.com/questions/20097380/iee-754-total-order-in-standard-c11
https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#how-can-i-implement-partialord
https://docs.rs/ordered-float/latest/ordered_float/

dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
```diff
@@ -47,6 +47,8 @@ inline bool isRoughSetFilterSupportType(const Int32 field_type)
     case TiDB::TypeLongLong:
     case TiDB::TypeInt24:
     case TiDB::TypeYear:
+    case TiDB::TypeFloat:
```
hehechen, line 50: I think we [can't compare float directly](https://stackoverflow.com/questions/4548004/how-to-correctly-and-standardly-compare-floats). May need to modify code to support float comparison in [https://github.com/pingcap/tiflash/blob/master/dbms/src/Storages/DeltaMerge/Index/ValueComparison.h](url).

PR #5610, opened 2022-08-11
Improve memtrack

bestwoody:
### What problem does this PR solve?

Issue Number: close #5609

Problem Summary:

### What is changed and how it works?
Solve the following problems which may cause memory quota not working, in order to reduce OOM cases.

1.Track some kind of object was not tracked before, such as object generated by Protobuf.
2.Fix a bug in memtracker, which causes memory quota not working. 
3.Fix bugs in our MPPTask construct/destruct logic, which cause memtracker can‘t be propagated into MPPTunnel & ExchangeReceiver
4. Use a background thread to collect RSS metric periodicaly and check how much it exceed the user setting limit in mem tracker. And throttle when it's too large, to control those un-trackable memory such as: cache used by Allocator 、stack allocated by threads.

How RSS check works：
Besides a simply condition check RSS > limit + 5GB.  A dynamic linear check logic is also defined. 
An example for how this logic works: if limit is 30GB, and real_rss is 40GB, since then mem alloc will fail when tracked amount > 25GB(30-(10-5)=25), when memory usage turn down below 25GB, all query tasks can continue work, where there is no need to wait the real_rss metric updated in next collect 100ms period. The larger real_rss over than limit+5GB, the lower ceiling of memory tracked amount where we can continue allocating memory.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

Issue #5609, opened 2022-08-11
Improve memtrack to reduce OOM cases


bestwoody:
## Enhancement
There are some problems which may cause memory quota not working, so that more OOM cases.
1. Some kind of object is not tracked, such as object generated by Protobuf.
2. A bug in memtracker, which causes memory quota not working. ref: [https://pingcap.feishu.cn/wiki/wikcnvA5vz6Mfk9VocSssEqPp3b](https://pingcap.feishu.cn/wiki/wikcnvA5vz6Mfk9VocSssEqPp3b)
3. Some bugs in our `MPPTask` construct/destruct logic,  which causememtracker can‘t be propagated into `MPPTunnel` & `ExchangeReceiver`
4. Some un-trackable memory such as: cache used by Allocator 、stack allocated by threads.

PR #6020, opened 2022-09-26
Refine the lifecycle of memory_tracker for MPP query

windtalker:

Issue Number: ref #5609

Problem Summary:
This is a refine pr for #5610
1. create `processlist/memory_tracker` in `MPPTask::prepare`, so when creating `MPPTunnel/ExchangeReceiver`, there is already a memory_tracker to use.
3. save memory tracker in `TunnelSender`, so each `TrackedMPPPacket` does not need to hold the shared ptr of memory tracker
4. remove some workaround code
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  - run random failpoint tests for more than 12 hours
  - run mpp fail tests for more than 12 hours
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/MemoryTracker.h
+++ b/dbms/src/Common/MemoryTracker.h
@@ -112,7 +112,9 @@ public:
     /// next should be changed only once: from nullptr to some value.
     void setNext(MemoryTracker * elem)
     {
-        next.store(elem, std::memory_order_relaxed);
+        MemoryTracker * old_val = nullptr;
+        if (!next.compare_exchange_strong(old_val, elem, std::memory_order_seq_cst, std::memory_order_relaxed))
+            return;
         next_holder = elem ? elem->shared_from_this() : nullptr;
     }
 
--- a/dbms/src/Flash/EstablishCall.cpp
+++ b/dbms/src/Flash/EstablishCall.cpp
@@ -202,6 +202,11 @@ void EstablishCallData::trySendOneMsg()
     switch (async_tunnel_sender->pop(res, this))
     {
     case GRPCSendQueueRes::OK:
+        /// Note: has to switch the memory tracker before `write`
+        /// because after `write`, `async_tunnel_sender` can be destroyed at any time
+        /// so there is a risk that `res` is destructed after `aysnc_tunnel_sender`
+        /// is destructed which may cause the memory tracker in `res` become invalid
+        res->switchMemTracker(nullptr);
         write(res->packet);
         return;
     case GRPCSendQueueRes::FINISHED:
--- a/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
+++ b/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
@@ -246,7 +246,7 @@ public:
             if (auto error_message = getErrorFromPackets(); !error_message.empty())
                 setDone(fmt::format("Exchange receiver meet error : {}", error_message));
             else if (!sendPackets())
-                setDone("Exchange receiver meet error : push packets fail, " + err_info);
+                setDone("Exchange receiver meet error : push packets fail");
             else if (read_packet_index < batch_packet_count)
             {
                 stage = AsyncRequestStage::WAIT_FINISH;
--- a/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
+++ b/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
@@ -204,7 +204,10 @@ public:
         }
         case AsyncRequestStage::WAIT_BATCH_READ:
             if (ok)
+            {
                 ++read_packet_index;
+                packets[read_packet_index - 1]->recomputeTrackedMem();
+            }
             if (!ok || read_packet_index == batch_packet_count || packets[read_packet_index - 1]->hasError())
                 notifyReactor();
             else
@@ -244,7 +247,7 @@ public:
                 has_data = true;
 
             if (auto packet = getErrorPacket())
-                setDone("Exchange receiver meet error : " + packet->error().msg());
+                setDone("Exchange receiver meet error : " + packet->error());
             else if (!sendPackets(err_info))
                 setDone("Exchange receiver meet error : push packets fail, " + err_info);
             else if (read_packet_index < batch_packet_count)
@@ -360,7 +363,6 @@ private:
             // We shouldn't throw error directly, since the caller works in a standalone thread.
             try
             {
-                packet->recomputeTrackedMem();
                 if (!pushPacket<enable_fine_grained_shuffle, false>(
                         request->source_index,
                         req_info,
@@ -415,8 +417,7 @@ ExchangeReceiverBase<RPCContext>::ExchangeReceiverBase(
     size_t max_streams_,
     const String & req_id,
     const String & executor_id,
-    uint64_t fine_grained_shuffle_stream_count_,
-    bool setup_conn_manually)
+    uint64_t fine_grained_shuffle_stream_count_)
     : rpc_context(std::move(rpc_context_))
     , source_num(source_num_)
     , max_streams(max_streams_)
@@ -442,11 +443,7 @@ ExchangeReceiverBase<RPCContext>::ExchangeReceiverBase(
             msg_channels.push_back(std::make_unique<MPMCQueue<std::shared_ptr<ReceivedMessage>>>(max_buffer_size));
         }
         rpc_context->fillSchema(schema);
-        if (!setup_conn_manually)
-        {
-            // In CH client case, we need setUpConn right now. However, MPPTask will setUpConnection manually after ProcEntry is created.
-            setUpConnection();
-        }
+        setUpConnection();
     }
     catch (...)
     {
@@ -494,8 +491,6 @@ void ExchangeReceiverBase<RPCContext>::close()
 template <typename RPCContext>
 void ExchangeReceiverBase<RPCContext>::setUpConnection()
 {
-    if (thread_count)
-        return;
     mem_tracker = current_memory_tracker ? current_memory_tracker->shared_from_this() : nullptr;
     std::vector<Request> async_requests;
 
@@ -627,7 +622,7 @@ void ExchangeReceiverBase<RPCContext>::readLoop(const Request & req)
                 if (packet->hasError())
                 {
                     meet_error = true;
-                    local_err_msg = fmt::format("Read error message from mpp packet: {}", packet->getPacket().error().msg());
+                    local_err_msg = fmt::format("Read error message from mpp packet: {}", packet->error());
                     break;
                 }
 
--- a/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
+++ b/dbms/src/Flash/Mpp/ExchangeReceiver.cpp
@@ -204,10 +204,7 @@ public:
         }
         case AsyncRequestStage::WAIT_BATCH_READ:
             if (ok)
-            {
                 ++read_packet_index;
-                packets[read_packet_index - 1]->recomputeTrackedMem();
-            }
             if (!ok || read_packet_index == batch_packet_count || packets[read_packet_index - 1]->hasError())
                 notifyReactor();
             else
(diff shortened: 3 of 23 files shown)


dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -204,7 +204,10 @@ class AsyncRequestHandler : public UnaryCallback<bool>
         }
         case AsyncRequestStage::WAIT_BATCH_READ:
             if (ok)
+            {
                 ++read_packet_index;
+                packets[read_packet_index - 1]->recomputeTrackedMem();
```
bestwoody, line 209: `recomputeTrackedMem` may throw error, we need handle the exception properly.

windtalker: Yes, I move the `try/catch` part into `recomputeTrackedMem `

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -281,6 +284,21 @@ void MPPTask::prepare(const mpp::DispatchTaskRequest & task_request)
 
     context->setDAGContext(dag_context.get());
 
+    if (context->getSettingsRef().enable_planner)
+        query_source = std::make_unique<PlanQuerySource>(*context);
+    else
+        query_source = std::make_unique<DAGQuerySource>(*context);
+
+    auto [query, ast] = query_source->parse(context->getSettingsRef().max_query_size);
```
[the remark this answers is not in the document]
windtalker: Refined.

dbms/src/Common/MemoryTracker.h
```diff
@@ -112,7 +112,9 @@ class MemoryTracker : public std::enable_shared_from_this<MemoryTracker>
     /// next should be changed only once: from nullptr to some value.
     void setNext(MemoryTracker * elem)
     {
-        next.store(elem, std::memory_order_relaxed);
+        MemoryTracker * old_val = nullptr;
+        if (!next.compare_exchange_strong(old_val, elem, std::memory_order_seq_cst, std::memory_order_relaxed))
```
bestwoody, line 116: in what cases that `next` will be called multiple times？

windtalker: For example:
https://github.com/pingcap/tiflash/blob/5609e68c0fa213e9740ff52ec60eac97c884e45a/dbms/src/Interpreters/ProcessList.cpp#L168

Different query may belong to the same user, the the user's memory tracker will be call multiple times.

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -59,14 +61,16 @@ MPPTask::MPPTask(const mpp::TaskMeta & meta_, const ContextPtr & context_)
...
+    current_memory_tracker = nullptr;
+}
 
 MPPTask::~MPPTask()
 {
     /// MPPTask maybe destructed by different thread, set the query memory_tracker
     /// to current_memory_tracker in the destructor
-    if (current_memory_tracker != memory_tracker)
-        current_memory_tracker = memory_tracker;
+    if (current_memory_tracker != process_list_entry->get().getMemoryTrackerPtr().get())
```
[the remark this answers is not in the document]
windtalker: Yes, we need to check null there.

dbms/src/Interpreters/executeQuery.cpp
```diff
@@ -232,16 +232,8 @@ std::tuple<ASTPtr, BlockIO> executeQueryImpl(
 
             context.setProcessListElement(&process_list_entry->get());
         }
```
SeaRise, line 234: maybe
```
if (!internal && nullptr == typeid_cast<const ASTShowProcesslistQuery *>(&*ast))
{
    process_list_entry = xxx
}
else if (context.getDAGContext())
{
    process_list_entry = context.getDAGContext()->getProcessListEntry();
}
```
and set the `internal` must be true in https://github.com/pingcap/tiflash/blob/5609e68c0fa213e9740ff52ec60eac97c884e45a/dbms/src/Flash/executeQuery.cpp#L29-L35
is better;

dbms/src/Flash/Mpp/TrackedMppDataPacket.h
```diff
@@ -93,7 +88,7 @@ struct MemTrackerWrapper
     {
         int bak_size = size;
         freeAll();
-        memory_tracker = getSharedPtrOfMemTracker(new_memory_tracker);
+        memory_tracker = new_memory_tracker;
```
SeaRise, line 91: maybe we can add `likely` in `MemTrackerWrapper::alloc` and `MemTrackerWrapper::free`

dbms/src/Interpreters/executeQuery.cpp
SeaRise: or
```
if (context.getProcessListElement())
    process_list_entry = context.getProcessListElement()
else if (!internal && nullptr == typeid_cast<const ASTShowProcesslistQuery *>(&*ast))
{
    xxxxx
}
```

windtalker: I think it is a little hack to set `internal` to be true, and `context.getDAGContext() != nullptr` does not means process_list_entry is not nullptr, only mpp task will construct process_list_entry earlier, for cop/batch cop, it still need to be constructed here.

SeaRise: Then maybe we just need to check `if (context.getProcessListElement())` here?

windtalker: But if we just check `if (context.getProcessListElement())`, the reader may be confused because they don't know in which case context.getProcessListElement() will be nullptr?

windtalker: I prefer the orignal implementation because it shows clearly that if current query is a mpp task, it does not need to create process_list_entry here. I can add some comments to explain that mpp task already create process_list_entry so we don't need to create it here.

SeaRise: ok, add `assert(context.getProcessListElement());` and some comments is ok to me

dbms/src/Flash/EstablishCall.cpp
```diff
@@ -202,6 +202,11 @@ void EstablishCallData::trySendOneMsg()
     switch (async_tunnel_sender->pop(res, this))
     {
     case GRPCSendQueueRes::OK:
+        /// Note: has to switch the memory tracker before `write`
+        /// because after `write`, `async_tunnel_sender` can be destroyed at any time
+        /// so there is a risk that `res` is destructed after `aysnc_tunnel_sender`
+        /// is destructed which may cause the memory tracker in `res` become invalid
+        res->switchMemTracker(nullptr);
```
bestwoody, line 209: It may cause MemTracker incorrect... How about switching to RootMemTracker?

dbms/src/Common/MemoryTracker.h
windtalker: Yes, total_memory_tracker is the same. But even if total_memory_tracker is the same, there is a write condition race when concurrent write to `next_holder`, that is why I made this change.

dbms/src/Flash/EstablishCall.cpp
windtalker: Why? It just release the memory before this change, the memory is also released after `write()`

dbms/src/Flash/Mpp/MPPHandler.cpp
```diff
@@ -44,7 +46,9 @@ void MPPHandler::handleError(const MPPTaskPtr & task, String error)
 grpc::Status MPPHandler::execute(const ContextPtr & context, mpp::DispatchTaskResponse * response)
 {
     MPPTaskPtr task = nullptr;
-    current_memory_tracker = nullptr; /// to avoid reusing threads in gRPC
+    SCOPE_EXIT({
+        current_memory_tracker = nullptr; /// to avoid reusing threads in gRPC
```
bestwoody, line 50: why a `SCOPE_EXIT` is added? since the meanings are quiet different

windtalker: Because after `MPPTask::prepare()`, `current_memory_tracker` will be set to the memory tracker of the `MPPTask`, and if we don't set it back to nullptr, then this thread may have illegal memory tracker.

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -336,6 +348,7 @@ void MPPTask::preprocess()
 void MPPTask::runImpl()
 {
     CPUAffinityManager::getInstance().bindSelfQueryThread();
+    current_memory_tracker = process_list_entry->get().getMemoryTrackerPtr().get();
```
bestwoody, line 351: is it necessary? since `newThreadManager()->scheduleThenDetach` will propgate memTracker into `runImpl`

windtalker: Oh, yes, it is not necessary.

windtalker: Maybe change it to RUNTIME_ASSERT

dbms/src/Flash/EstablishCall.cpp
bestwoody: I'm worried about an extreme case that a lot of threads are doing `EstablishCallData::trySendOneMsg()` and early freeing the MemTracker cause real memory usage too large than tracked

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -351,29 +359,19 @@ class AsyncRequestHandler : public UnaryCallback<bool>
...
 
-    bool sendPackets(std::string & err_info)
+    bool sendPackets()
     {
         // note: no exception should be thrown rudely, since it's called by a GRPC poller.
         for (size_t i = 0; i < read_packet_index; ++i)
         {
             auto & packet = packets[i];
-            // We shouldn't throw error directly, since the caller works in a standalone thread.
-            try
```
bestwoody, line 361: I seems `pushPacket` can throw exception except recomputeTrackedMem such as `fiu_do_on(FailPoints::random_receiver_async_msg_push_failure_failpoint, push_succeed = false;);`

windtalker: This failpoint just set the return value(`push_succeed`) to false, it does not throw exception.

dbms/src/Flash/EstablishCall.cpp
windtalker: But in before this pr, it also has the early free issue since async tunnel's write actually does not wait until write finish.

dbms/src/Flash/Mpp/MPPTask.cpp
windtalker: Done

dbms/src/Flash/Mpp/TrackedMppDataPacket.h
windtalker: Done

bestwoody (approved):
lgtm

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -20,12 +20,14 @@
...
 #include <DataStreams/SquashingBlockOutputStream.h>
 #include <Flash/Coprocessor/DAGCodec.h>
+#include <Flash/Coprocessor/DAGQuerySource.h>
 #include <Flash/Coprocessor/DAGUtils.h>
 #include <Flash/Mpp/ExchangeReceiver.h>
 #include <Flash/Mpp/GRPCReceiverContext.h>
 #include <Flash/Mpp/MPPTask.h>
 #include <Flash/Mpp/MPPTunnelSet.h>
 #include <Flash/Mpp/Utils.h>
+#include <Flash/Planner/PlanQuerySource.h>
```
SeaRise, line 30: useless #include

SeaRise (approved):
The rest LGTM

Issue #5813, opened 2022-09-07
Raw TiDB PK: 80000000007DACAC, Prewrite ts: 435819990079504388 can not found in default cf for key


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```sql
 CREATE TABLE `other_handle4` (
  `a` bigint(20) DEFAULT NULL,
  `b` int(11) DEFAULT NULL,
  `c` varchar(100) NOT NULL,
  `d` char(100) DEFAULT NULL,
  `e` float DEFAULT NULL,
  `g` double DEFAULT NULL,
  `h` decimal(8,4) DEFAULT NULL,
  `i` date DEFAULT NULL,
  `j` datetime DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;

select count(*) from other_handle4;
```
sometimes the query can get result ,but result if wrong(result inconsistent tso 435820060019523587, tikv: [7100000], tiflash [3484384]) ; sometime it raise error
ERROR 1105 (HY000): other error for mpp stream: From MPP<query:435824304366551041,task:3>: Code: 9008, e.displayText() = DB::Exception: Raw TiDB PK: 800000000081E6F6, Prewrite ts: 435819992582717448 can not found in default cf for key: 7480000000000000FFB45F728000000000FF81E6F60000000000FAF9F3A8069B4BFFEF: (while doing learner read for table, logical table_id: 180), e.what() = DB::Exception,
### 4. What is your TiFlash version? (Required)
master

flowbehappy:
Similar to https://github.com/pingcap/tiflash/issues/5648

CalvinNeo:
We observe 26606 comes from 20687 in its first split
```
[...]

PR #6041, opened 2022-09-27
KVStore: Support multi file snapshot

### LINKED ISSUE #6086: Support multi sst file snapshot for TiFlash
## Enhancement

TiFlash can support applying big snapshot for now, so we can support ingesting multi sst file snapshot, by combining them into a stream of a imaginary big sst file for every cf.

CalvinNeo:

Issue Number: close #6086

Problem Summary:

see issue

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Integration test:

Whether and to what extent this case is covered by integration test is totally up to TiKV.

Currently, integration test can seldomly cover this case, which is caused by bugs on TiKV's side. See https://github.com/pingcap/tiflash/issues/5813.

Once TiKV loose its constraint, integration test will cover this case.

Manual test:
1. disable TiKV's limitation
2. create a playground with new TiKV and TiFlash
3. create a table only on TiKV
4. write many rows into this table
5. alter table set tiflash replica 1
6. wait this table until it becomes available

Found we can handle snapshot:
on Proxy
```
[2022/10/08 16:10:11.423 +08:00] [INFO] [observer.rs:700] ["pre apply snapshot"] [pending=0] [snap_key="SnapKey { region_id: 1005, term: 6, idx: 7 }"] [region_id=1005] [peer_id=1007]
[2022/10/08 16:10:11.424 +08:00] [INFO] [observer.rs:654] ["observe multi-file snapshot"] [total=2] [cf="\"default\""] [snap="Snapshot { key: SnapKey { region_id: 1005, term: 6, idx: 7 }, display_path: \"/Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_(default|lock|write).sst\" }"]
[2022/10/08 16:10:11.616 +08:00] [INFO] [region.rs:439] ["begin apply snap data"] [peer_id=1007] [region_id=1005]
[2022/10/08 16:10:11.653 +08:00] [INFO] [snap.rs:670] ["set_snapshot_meta total cf files count: 4"]
[2022/10/08 16:10:11.792 +08:00] [INFO] [io.rs:259] ["apply_sst_cf_file starts on cf default. All files [\"/Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_default.sst.clone\", \"/Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_default_0001.sst.clone\"]"]
[2022/10/08 16:10:11.793 +08:00] [INFO] [observer.rs:767] ["post apply snapshot"] [region="id: 1005 start_key: 7480000000000000FF4B5F720000000000FA end_key: 7480000000000000FF4C00000000000000F8 region_epoch { conf_ver: 2 version: 37 } peers { id: 1006 store_id: 1 } peers { id: 1007 store_id: 70 role: Learner }"] [snap_key="SnapKey { region_id: 1005, term: 6, idx: 7 }"] [peer_id=1007]
[2022/10/08 16:10:11.922 +08:00] [INFO] [observer.rs:785] ["get prehandled snapshot success"] [pending=1] [region="id: 1005 start_key: 7480000000000000FF4B5F720000000000FA end_key: 7480000000000000FF4C00000000000000F8 region_epoch { conf_ver: 2 version: 37 } peers { id: 1006 store_id: 1 } peers { id: 1007 store_id: 70 role: Learner }"] [peer_id="SnapKey { region_id: 1005, term: 6, idx: 7 }"]
```

... and on TiFlash
```
[2022/10/08 16:10:11.440 +08:00] [INFO] [SSTFilesToBlockInputStream.cpp:84] ["SSTFilesToBlockInputStream:Read SST Files with MultiSSTReader"] [thread_id=12]
[2022/10/08 16:10:11.440 +08:00] [INFO] [SSTReader.h:93] ["MultiSSTReader:Init with /Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_default.sstþ\u0007\u0005"] [thread_id=12]
[2022/10/08 16:10:11.440 +08:00] [INFO] [TiDBSchemaSyncer.h:113] ["SchemaSyncer:Start to sync schemas. current version is: 40 and try to sync schema version to: 41"] [thread_id=13]
[2022/10/08 16:10:11.440 +08:00] [DEBUG] [TiDBSchemaSyncer.h:170] ["SchemaSyncer:Try load schema diffs."] [thread_id=13]
[2022/10/08 16:10:11.441 +08:00] [INFO] [SSTReader.h:93] ["MultiSSTReader:Init with /Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_write.sst\tâ\u0005"] [thread_id=12]
[2022/10/08 16:10:11.441 +08:00] [INFO] [SSTFilesToBlockInputStream.cpp:131] ["SSTFilesToBlockInputStream:Finish Construct MultiSSTReader, write 1 lock 0 default 2"] [thread_id=12]
[2022/10/08 16:10:11.441 +08:00] [DEBUG] [TiDBSchemaSyncer.h:181] ["SchemaSyncer:End load schema diffs with total 1 entries."] [thread_id=13]
[2022/10/08 16:10:11.442 +08:00] [DEBUG] [SchemaGetter.cpp:234] ["SchemaGetter:Get DB Info from TiKV : {\"id\":73,\"db_name\":{\"O\":\"test99\",\"L\":\"test99\"},\"charset\":\"utf8mb4\",\"collate\":\"utf8mb4_bin\",\"state\":5,\"policy_ref_info\":null}"] [thread_id=13]

...

[2022/10/08 16:10:11.622 +08:00] [INFO] [SSTReader.h:78] ["MultiSSTReader:Swithch to /Users/calvin/.tiup/data/multi/tiflash-0/flash/snap/rev_1005_6_7_default_0001.sst).sst\" }\u0004"] [thread_id=12]
[2022/10/08 16:10:11.686 +08:00] [DEBUG] [SSTFilesToBlockInputStream.cpp:251] ["SSTFilesToBlockInputStream:Done loading all kvpairs from [CF=default] [offset=1000] [write_cf_offset=1000] "] [thread_id=12]
[2022/10/08 16:10:11.694 +08:00] [DEBUG] [SSTFilesToBlockInputStream.cpp:251] ["SSTFilesToBlockInputStream:Done loading all kvpairs from [CF=lock] [offset=0] [write_cf_offset=1000] "] [thread_id=12]
[2022/10/08 16:10:11.806 +08:00] [DEBUG] [SSTFilesToBlockInputStream.cpp:251] ["SSTFilesToBlockInputStream:Done loading all kvpairs from [CF=default] [offset=1000] [write_cf_offset=1000] "] [thread_id=12]
[2022/10/08 16:10:11.806 +08:00] [DEBUG] [SSTFilesToBlockInputStream.cpp:251] ["SSTFilesToBlockInputStream:Done loading all kvpairs from [CF=lock] [offset=0] [write_cf_offset=1000] "] [thread_id=12]
[2022/10/08 16:10:11.810 +08:00] [DEBUG] [SSTFilesToDTFilesOutputStream.cpp:152] ["SSTFilesToDTFilesOutputStream:Create new DTFile for snapshot data, region=[region 1005, applied: term 6 index 7] file_idx=0 file=/Users/calvin/.tiup/data/multi/tiflash-0/data/data/t_75/stable/.tmp.dmf_2"] [thread_id=12]
```

We found default has 2 files, and MultiSSTReader can switch to the second file when finish reading the first one. The whole table can be available finally.

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Transaction/tests/gtest_new_kvstore.cpp
```diff
@@ -0,0 +1,255 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "kvstore_helper.h"
+
+namespace DB
+{
+namespace tests
+{
+TEST_F(RegionKVStoreTest, KVStoreFailRecovery)
```
JaySon-Huang, line 21: What is the difference between gtest_kvstore and gtest_new_kvstore?

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,110 @@
...
 
+/// MultiSSTReader helps when there are multiple sst files in a column family.
+/// It is derived from virtual class SSTReader, so it can be holded in a SSTReaderPtr.
+/// It also maintains instance of `R` which is normaly SSTReader(and MockSSTReader in tests),
+/// to read a single sst file.
+/// An `Initer` function is need to create a instance of `R` from a instance of `E`,
+/// which is usually path of the SST file.
+/// We introduce `R` and `E` to make this class testable.
+template <typename R, typename E>
+struct MultiSSTReader : SSTReader
```
JaySon-Huang, line 51:
```suggestion
struct MultiSSTReader : public SSTReader
```

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,110 @@
...
+        return mono->remained();
+    }
+    BaseBuffView key() const override
+    {
+        return mono->key();
+    }
+    BaseBuffView value() const override
+    {
+        return mono->value();
+    }
```
JaySon-Huang, line 69: Suggest adding "View" suffix to these two methods as they return "BaseBuffView". The same for class `SSTReader`.
```suggestion
    BaseBuffView keyView() const override
    {
        return mono->key();
    }
    BaseBuffView valueView() const override
    {
        return mono->value();
    }
```

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,110 @@
...
+        // The last sst reader will be dropped with inner.
+    }
+
+private:
+    Poco::Logger * log;
+    mutable std::unique_ptr<R> mono;
+    // Overwrite
+    const TiFlashRaftProxyHelper * proxy_helper;
+    // Overwrite
+    ColumnFamilyType type;
```
JaySon-Huang, line 117: make `SSTReader.proxy_helper` and `SSTReader.type` to be protected?

dbms/src/Storages/Transaction/tests/gtest_new_kvstore.cpp
CalvinNeo: gtest_kvstore is old tests of KVStore without a complete mock of proxy.
gtest_new_kvstore is new tests of KVStore which involves pipeline of proxy, like commit/apply etc..

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: Why public here?

JaySon-Huang: It is `Private Inheritance` if you ignore "public" here. Why do you need to make it a private inheritance?

CalvinNeo: Inheritance is public by default for structs, the following compiles
```C++
struct B {
    int z = 1;
};

struct D : B {

};

int main() {
    B* p = new D();
    printf("%d\n", p->z);
}
```

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: Maybe it is better the base class exposed no its fields to either outside or its derived classes. Since `MultiSSTReader` inherits `SSTReader` only to implement its interfaces. So I think it is safer that we hide all fields of these classes by setting them private.

dbms/src/Storages/Transaction/SSTReader.h
JaySon-Huang: Well... I didn't know that struct is public inheritance by default before. No matter what, I think explicit is better than implicit.
And as SSTReader now has virtual methods, all its members are private, I think it is more likely to be a `class` instead of `struct`, though they are almost the same in C++. So does `MultiSSTReader`

dbms/src/Storages/Transaction/SSTReader.h
JaySon-Huang: I think defining members with duplicated names in the derived classes is a worse idea. If you want to make the fields not exposed its private method, then you can define some protect methods for read-only access to these fields.

CalvinNeo: What I want is the derived class `MultiSSTReader` to have no access to all fields of its base class `SSTReader`, which is the reason why declaring them as private. Meanwhile, I find it hard to find another name for `type` and `proxy_helper` in derived class. So if `MultiSSTReader` already has no access to `SSTReader::tupe/proxy_helper`, we can define our own.

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: Change from struct to class makes sense.

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: OK, I know what the point is now. My mainly concern it make `MultiSSTReader` aware of nothing in `SSTReader` other than interfaces. This surely is a waste of memory, but only two integers in a limited used class. 
However, I think by totally deprivation of `SSTReader`, `MultiSSTReader` can no longer care about how `SSTReader` is implemented.

dbms/src/Debug/MockSSTReader.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "MockSSTReader.h"
+
+#include <Common/Exception.h>
+
+namespace DB
+{
+SSTReaderPtr fn_get_sst_reader(SSTView v, RaftStoreProxyPtr)
+{
+    std::string s(v.path.data, v.path.len);
+    LOG_FMT_INFO(&Poco::Logger::get("!!!!! fn_get_sst_reader"), "AAAAA {}", s);
```
CalvinNeo, line 24: will be removed

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,110 @@
...
+    {
+        return mono->keyView();
+    }
+    BaseBuffView valueView() const override
+    {
+        return mono->valueView();
+    }
+    void next() override
+    {
+        this->maybe_next_reader();
```
breezewish, line 74: What about next after the read instead of before the read? In this way may be we can eliminate the case where current mono is drained.

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,110 @@
...
+/// We introduce `R` and `E` to make this class testable.
+template <typename R, typename E>
+class MultiSSTReader : public SSTReader
+{
+public:
+    using Initer = std::function<std::unique_ptr<R>(const TiFlashRaftProxyHelper *, E)>;
+
+    DISALLOW_COPY_AND_MOVE(MultiSSTReader);
+
+    bool remained() const override
```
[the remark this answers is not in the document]
CalvinNeo: It has the side-effect that switch to the next SSTReader(if any), and ask whether it remains.
I think repeatedly calling this function will not cause bad side effects? I will add test for this.

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: Seems it will fail if a reader is empty, not knowing whether this will happen.

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -64,20 +64,97 @@ SSTFilesToBlockInputStream::~SSTFilesToBlockInputStream() = default;
...
+
+        auto make_inner_func = [&](const TiFlashRaftProxyHelper * proxy_helper, SSTView snap) {
+            return std::make_unique<SSTReader>(proxy_helper, snap);
+        };
+        auto generate_cf_reader = [&]() {
+            // Generate old cf reader.
+            switch (prev_type)
+            {
+            case ColumnFamilyType::Default:
+                size_default = ssts.size();
```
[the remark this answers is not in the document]
CalvinNeo: The order is ensured by using exactly the interface `CfFile::file_paths()` in proxy.

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -64,20 +64,97 @@ SSTFilesToBlockInputStream::~SSTFilesToBlockInputStream() = default;
...
 void SSTFilesToBlockInputStream::readPrefix()
 {
-    for (UInt64 i = 0; i < snaps.len; ++i)
+    bool use_multi = false;
+    if (snaps.len > 3)
     {
-        const auto & snapshot = snaps.views[i];
-        switch (snapshot.type)
+        use_multi = true;
+    }
```
JaySon-Huang, line 71: If TiKV send 2 default cf file and 1 write cf file (can it?) and make this assumption break? Maybe always use `MultiSSTReader` for reading?

CalvinNeo: This can actually happen, so we will check again if `snaps.len <= 3`.
MultiSSTReader can be always used though.

JaySon-Huang: If we always use MultiSSTReader, then I think
* We can eliminate some unnecessary branches, which make the code more clear and testable
* Make `SSTReader` a pure interface

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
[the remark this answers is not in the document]
JaySon-Huang: I vote for map+vector or 1 vector for each cf, then we don't rely on the assumption of TiKV. Cause it's hard to constrain tikv behavior change by tiflash.
And suggest adding an exception when tiflash meets unknown cf.

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
CalvinNeo: I think we'd better not move all logic of `SSTReader` into `MultiSSTReader` and make it pure virtual, since SSTReader reads an sst file, which is quite cohesive. Meanwhile, there are already some tests either based on or directly test `SSTReader`.

CalvinNeo: I create a new `MonoSSTReader` for `SSTReader`, and then make `SSTReader` an abstract class, how about this? @JaySon-Huang

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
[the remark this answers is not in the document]
CalvinNeo: I am confused what the problem is? @breezewish @JaySon-Huang Is the case
```
default 1
lock 1
default 2
lock 2
```
rather than the case
```
default 2
default 1
```
you all concern?
Then proxy will guarantee an order. However three vectors is also acceptable.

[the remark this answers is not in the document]
JaySon-Huang: The point is can we be aware of TiKV behavior changes in the early stage (like internal testing). Or automatically adapt to the behavior changes.

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,121 @@
...
+    BaseBuffView valueView() const override;
+    void next() override;
 
-    DISALLOW_COPY_AND_MOVE(SSTReader);
-    SSTReader(const TiFlashRaftProxyHelper * proxy_helper_, SSTView view);
-    ~SSTReader();
+    DISALLOW_COPY_AND_MOVE(MonoSSTReader);
+    MonoSSTReader(const TiFlashRaftProxyHelper * proxy_helper_, SSTView view);
+    // Should only be used when MonoSSTReader is a base class.
+    MonoSSTReader();
```
JaySon-Huang, line 45: Seems we don't need this default ctor now?

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -64,22 +64,43 @@ SSTFilesToBlockInputStream::~SSTFilesToBlockInputStream() = default;
...
-            write_cf_reader = std::make_unique<SSTReader>(proxy_helper, snapshot);
+            ssts_write.push_back(snapshot);
             break;
         case ColumnFamilyType::Lock:
-            lock_cf_reader = std::make_unique<SSTReader>(proxy_helper, snapshot);
+            ssts_lock.push_back(snapshot);
             break;
         }
     }
+    if (ssts_default.size())
```
JaySon-Huang, line 91: boolean in `if(...)` statement is better
```suggestion
    if (!ssts_default.empty())
```

dbms/src/Debug/MockRaftStoreProxy.h
```diff
@@ -193,12 +238,19 @@ struct MockRaftStoreProxy : MutexLockWrap
...
+        table_id = 1;
+    }
 
     std::unordered_set<uint64_t> region_id_to_drop;
     std::unordered_set<uint64_t> region_id_to_error;
     std::map<uint64_t, MockProxyRegionPtr> regions;
     std::list<std::shared_ptr<RawMockReadIndexTask>> tasks;
     AsyncWaker::Notifier notifier;
+    TableID table_id;
+    Poco::Logger * log;
```
JaySon-Huang, line 253: DB::LoggerPtr is preferred than Poco::Logger
```suggestion
    LoggerPtr log;
```

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,121 @@
...
+        mono = initer(proxy_helper, args[current]);
+    }
+
+    ~MultiSSTReader() override
+    {
+        // The last sst reader will be dropped with inner.
+    }
+
+private:
+    Poco::Logger * log;
```
JaySon-Huang, line 125:
```suggestion
    LoggerPtr log;
```

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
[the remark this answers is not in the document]
breezewish: @CalvinNeo  Yeah, I'm indeed concerning about some day this case may happen:

```
default 1
lock 1
default 2
lock 2
```

It would be better if we could be (even) compatible with this kind of order, or we could have some checks to detect it.

In the current impl, when this kind of order happens, seems that we will silently lose some data. Then it may be a hard work for us to troubleshoot.

dbms/src/Storages/Transaction/SSTReader.h
breezewish: You are right. A usual convention (as a cursor like thing) could be:

- `keyView` and `valueView` is valid after an initial `next()`
- The copied value of `keyView` and `valueView` may be invalidated if `next()` is called.

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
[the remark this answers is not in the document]
CalvinNeo: For the first question, though proxy prevent this situation, I have adopt the map+vector(actually 3 vectors) idea in my later commits in TiFlash, so there is a double ensurance.
For the second question, we have no means to validate if a list of file is in order. The order is totally a blackbox for us. We use what TiKV gives us.

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: Maybe your suggestion is correct, `reader->remain()` can be interpreted into a "valid" function rather than a "hasNext" function.

dbms/src/Storages/Transaction/ApplySnapshot.cpp
```diff
@@ -276,6 +276,7 @@ std::vector<DM::ExternalDTFileInfo> KVStore::preHandleSnapshotToFiles(
     }
     catch (DB::Exception & e)
     {
+        LOG_INFO(log, "!!!! [error={}]", e.displayText());
```
JaySon-Huang, line 279: do we need this log? I think the log is print by `tryLogCurrentException` in ProxyFFI.cpp

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,119 @@
...
+    MultiSSTReader(const TiFlashRaftProxyHelper * proxy_helper_, ColumnFamilyType type_, Initer initer_, std::vector<E> args_)
+        : log(Logger::get("MultiSSTReader"))
+        , proxy_helper(proxy_helper_)
+        , type(type_)
+        , initer(initer_)
+        , args(args_)
+        , current(0)
+    {
+        assert(args.size() > 0);
+        LOG_INFO(log, "Init with {}", buffToStrView(args[current].path));
```
JaySon-Huang, line 113:
```suggestion
        LOG_INFO(log, "Open sst file {}", buffToStrView(args[current].path));
```

JaySon-Huang: This could helpful for grepping log

dbms/src/Debug/MockRaftStoreProxy.cpp
```diff
@@ -505,6 +512,127 @@ void MockRaftStoreProxy::replay(
...
+    ColumnsDescription columns;
+    auto & data_type_factory = DataTypeFactory::instance();
+    columns.ordinary = NamesAndTypesList({NameAndTypePair{"a", data_type_factory.get("Int64")}});
+    auto tso = tmt.getPDClient()->getTS();
+    MockTiDB::instance().newDataBase("d");
+    UInt64 table_id = MockTiDB::instance().newTable("d", "t", columns, tso, "", "dt");
+    try
+    {
+        auto schema_syncer = tmt.getSchemaSyncer();
+        schema_syncer->syncSchemas(ctx);
```
JaySon-Huang, line 624: can we drop the table after the mock proxy instance is not used?

dbms/src/Storages/Transaction/tests/kvstore_helper.h
```diff
@@ -0,0 +1,154 @@
...
+        kvstore->restore(*path_pool, proxy_helper.get());
+        return *kvstore;
+    }
+    void createDefaultRegions()
+    {
+        proxy_instance->init(100);
+    }
+    void initStorages()
+    {
+        registerStorages();
```
JaySon-Huang, line 115: we need to try ... catch and ignore the exception thrown by this method. Because other unit test may already call `registerStorages` when we run without any gtest-filter

JaySon-Huang (commented):
Rest LGTM

dbms/src/Storages/Transaction/ApplySnapshot.cpp
CalvinNeo: thats quite embarrassing

JaySon-Huang (approved):

dbms/src/Storages/Transaction/SSTReader.h
```diff
@@ -14,28 +14,119 @@
...
+    // and we have a next sst file to read.
+    void maybe_next_reader() const
+    {
+        if (!mono->remained())
+        {
+            current++;
+            if (current < args.size())
+            {
+                // We don't drop if mono is the last instance for safety,
+                // and it will be dropped as MultiSSTReader is dropped.
```
breezewish, line 97: What kind of safety do you want to have? For example, not crashing the TiFlash due to nullptr when there are `valueView()` after `next()` and drained or something else?

breezewish (approved):
Looks good, just some questions.

dbms/src/Storages/Transaction/SSTReader.h
CalvinNeo: We the mono hold is the last sst file, `MultiSSTReader` acts like a `MonoSSTReader`

Issue #5726, opened 2022-08-29
Log format unification for easier grep


breezewish:
## Enhancement

As there is no fields support in the current logger, when meeting fields like `segment_id`, or `table_name`, we are now outputting them in different ways in the log. For example, it may be like `segment [1]`, or may be like `[id:1]`. This makes the `grep` process hard to catch all occurrences, because we need to grep for different keywords. Additionally, notice that `[id:1]` is not friendly for grep because it may be conflicted with other "id".

We should arrange these fields in a unified way, so that searching a specific keyword will lead and only lead to related log lines.

The following are tasks so that we can pinpoint logs related with different things:

- [ ] Related with a segment: grep by `segment_id=XX `
- [ ] Related with a DMFile: TBD
- [ ] Related with a stable layer: TBD
- [ ] Related with a delta layer: TBD
- [ ] Related with a dm store: TBD
- [ ] Related with a region: TBD

PR #6055, opened 2022-09-28
storage: include segment and table info in more logs

breezewish:

Issue Number: ref #5726

Attach the table_id and segment_id in more storage modules for easier debugging.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below): WIP

A. All segment logs now contains table_id, in case of multiple tables contains same segment id:

image

The recommended grep command would be:

```
grep 'segment_id=X\>' | grep 'table_id=X\>'
```

B. DeltaValueSpace and StableValueSpace now contains segment_id and table_id. A useful scenario is that we can see flush operations when grepping by the segment_id:

image

C. SSTFilesToDTFilesOutputStream now contains table_id. When grepping by the table we can know more information:

image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/StableValueSpace.h
```diff
@@ -141,14 +157,14 @@ class StableValueSpace : public std::enable_shared_from_this<StableValueSpace>
 
         ColumnCachePtrs column_caches;
 
-        Snapshot()
-            : log(&Poco::Logger::get("StableValueSpace::Snapshot"))
+        Snapshot(StableValueSpacePtr stable_)
+            : stable(stable_)
+            , log(Logger::get("StableValueSpace::Snapshot", fmt::format("<{}>", stable->log_prefix)))
```
JaySon-Huang, line 162: Maybe use `stable->log->identifier()` instead of saving `log_prefix` in `StableValueSpace`?

breezewish: Updated. Now there are still some `log_prefix` variables remain, when we need to concat multiple properties together. We may resolve it in https://github.com/pingcap/tiflash/issues/6068

JaySon-Huang (approved):
LGTM

dbms/src/Storages/DeltaMerge/Delta/ColumnFilePersistedSet.h
```diff
@@ -69,19 +69,19 @@ class ColumnFilePersistedSet : public std::enable_shared_from_this<ColumnFilePer
...
+    LoggerPtr log;
 
 private:
     inline void updateColumnFileStats();
 
     void checkColumnFiles(const ColumnFilePersistedLevels & new_column_file_levels);
 
 public:
-    explicit ColumnFilePersistedSet(PageId metadata_id_, const ColumnFilePersisteds & persisted_column_files = {});
+    explicit ColumnFilePersistedSet(const std::string & log_prefix_, PageId metadata_id_, const ColumnFilePersisteds & persisted_column_files = {});
```
Lloyd-Pottiger, line 80: use `String`?

Lloyd-Pottiger (approved):

dbms/src/Storages/DeltaMerge/Delta/ColumnFilePersistedSet.h
breezewish: Not sure about the idiom, but seems that in our code base both `std::string` and `String` are heavily used :/

Issue #6061, opened 2022-09-29
Support onlyNull in test framework


xzhangxian1008:
## Enhancement

We jude if this is a const null column by `onlyNull` function in `getNullPresense`(located in IFunction.cpp).  However, test framework doesn't support `onlyNull` because some settings is ignored. In `gtest_regexp.cpp`, many const null columns have been created and input into the function. However, regexp functions can't judge if this is a const null column in gtest because of the above reason and it does the same judgement by another workaround to pass the gtest.

PR #6082, opened 2022-10-07
Refactor regexp and regexp_like function

### LINKED ISSUE #5984: Refactor regexp

xzhangxian1008:

Issue Number: close #5984 , ref #6115

Problem Summary:

changes:
- empty pattern is invalid.
- we re-implement them in FunctionStringRegexp class. (regexp and regexp_like functions are implemented in FunctionsStringSearch and MatchImpl classes at first)
- RegexpReplace functions related codes has been moved to FunctionsRegexp.h and FunctionsRegexp.cpp files.

main new codes:
- class FunctionStringRegexpBase
- class FunctionStringRegexp
- class Param related
- macros which convert columns to Params

main moved codes:
- class ReplaceStringImpl
- class ReplaceRegexpImpl
- class FunctionStringReplace

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+        bool has_nullable_col = false;
+        bool has_data_type_nothing = false;
+
+        for (size_t i = 0; i < REGEXP_XXX_MIN_PARAM_NUM; ++i)
+            checkInputArg(arguments[i], &has_nullable_col, &has_data_type_nothing);
+
+        // check match_type arg for regexp_like
+        if constexpr (class_name_sv == regexp_like_name_sv)
+            if (arg_num == args_max_num && !arguments[args_max_num - 1]->isString())
+                checkInputArg(arguments[args_max_num - 1], &has_nullable_col, &has_data_type_nothing);
```
windtalker, line 401: why not check the input arg like this:
```
for (auto & arg : arguments)
    checkInputArg(arg, &has_nullable_col, &has_data_type_nothing)
```

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+
+        if (has_nullable_col)
+        {
+            if (has_data_type_nothing)
+                return std::make_shared<DataTypeNullable>(std::make_shared<DataTypeNothing>());
+            return std::make_shared<DataTypeNullable>(std::make_shared<DataTypeNumber<ResultType>>());
+        }
+        else
+        {
+            if (has_data_type_nothing)
```
windtalker, line 411: if has_data_type_nothing is true, has_nullable_col must be true

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+    {
+        // Do something related with nullable columns
+        NullPresence null_presence = getNullPresense(block, arguments);
+
+        const ColumnPtr & col_expr = block.getByPosition(arguments[0]).column;
+
+        if (null_presence.has_const_null_col || null_presence.has_data_type_nothing)
+        {
+            // There is a const null column in the input
+            block.getByPosition(result).column = block.getByPosition(result).type->createColumnConst(block.rows(), Null());
```
windtalker, line 427: I'm not sure if this line can work if the return type if `DataTypeNullable(DataTypeNothing)`

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+            return;
+        }
+
+        const ColumnPtr & col_pat = block.getByPosition(arguments[1]).column;
+
+        if (col_expr->empty())
```
windtalker, line 433: why need to handle the empty case explicitly?

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+        {
+            auto null_col_res = ColumnNullable::create(ColumnString::create(), ColumnUInt8::create());
```
windtalker, line 435: why it has to be nullable column if the col_expr is empty?

dbms/src/Functions/FunctionsRegexp.cpp
```diff
@@ -0,0 +1,1116 @@
...
+
+    // generate match type flag
+    String flags;
+    for (auto flag : applied_flags)
+        flags += flag;
+
+    return flags;
+}
+
+NullPresence getNullPresense(const Block & block, const ColumnNumbers & args)
```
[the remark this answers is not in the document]
xzhangxian1008: > Why can not reuse the `getNullPresense` defined in `IFunction.cpp`?

see issue: https://github.com/pingcap/tiflash/issues/6061

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+            // This is a const column
+            auto col_const_data = col_const->getDataColumnPtr();
+            if (col_const_data->isColumnNullable())
+            {
+                // This is a const nullable column
+                // const null can't be here as we should have handle it in the previous
+                Field field;
+                auto p = static_cast<const ColumnNullable &>(*col_const_data).getNestedColumnPtr();
+                col_const->get(0, field);
+                data_string = field.safeGet<String>();
```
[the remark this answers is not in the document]
xzhangxian1008: > why not use `getDataAt` here?

nullable column does not support this function and it will throw exception.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+
+            is_const = true;
+            return;
+        }
+
+        if (col_ptr->isColumnNullable())
+        {
+            // Handle nullable column
+            auto nested_ptr = static_cast<const ColumnNullable &>(*col_ptr).getNestedColumnPtr();
+            col_int64 = checkAndGetColumn<ColumnInt64>(&(*nested_ptr));
```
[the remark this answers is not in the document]
xzhangxian1008: > I think all integer type is possible here, `checkAndGetColumn<ColumnInt64>` will return nullptr if the input type is not Int64

I will resolve it

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+    // Max parameter number the regexp_xxx function could receive
+    static constexpr size_t REGEXP_MAX_PARAM_NUM = 2;
+    static constexpr size_t REGEXP_LIKE_MAX_PARAM_NUM = 3;
+    static constexpr size_t REGEXP_INSTR_MAX_PARAM_NUM = 6;
+    static constexpr size_t REGEXP_REPLACE_MAX_PARAM_NUM = 6;
+    static constexpr size_t REGEXP_SUBSTR_MAX_PARAM_NUM = 5;
+
+    void memorize(const Param & pat_param, const std::unique_ptr<const Param> & match_type_param, TiDB::TiDBCollatorPtr collator) const
+    {
+        String && final_pattern = pat_param.getString(0);
```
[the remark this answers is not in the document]
xzhangxian1008: > why use `String &&`?

typo, getString() returns String&& at the beginning..

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+        if (!is_const && col_str != nullptr)
+            dst = col_str->getDataAt(idx);
+        else
+        {
+            dst.data = data_string.c_str();
+            dst.size = data_string.size();
+        }
+    }
+
+    String getString(size_t idx) const
```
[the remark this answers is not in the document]
xzhangxian1008: > Can we use `StringRef` instead of `String`?

We also provide a `StringRef` function.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+            // This is a pure vector column
+            col_int64 = checkAndGetColumn<ColumnInt64>(&(*col_ptr));
+        }
+    }
+
+    Int64 getInt64(size_t idx) const
+    {
+        // Use default value when arg is const or not provided.
+        // For safety, nullptr should be checked
+        return !is_const && col_int64 != nullptr ? col_int64->getInt(idx) : data_int64;
```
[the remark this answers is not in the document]
xzhangxian1008: > And you can make `Param` a template class, so don't have to check `!is_const && col_int64 != nullptr` for column vector.

done

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,944 @@
...
+            return ret_str;
+        }
+    }
+
+    bool isNullAt(size_t idx) const
+    {
+        if (null_map == nullptr)
+            return false;
+
+        return (*null_map)[idx];
```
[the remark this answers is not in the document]
xzhangxian1008: > The size of null_map for constant column is only 1

okk, I have considered it in the latest codes

dbms/src/Functions/FunctionsRegexp.cpp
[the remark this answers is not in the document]
xzhangxian1008: > Why can not reuse the `getNullPresense` defined in `IFunction.cpp`?

resolved

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+}
+
+// Use this type when param is not provided
+class ParamDefault
+{
+public:
+    explicit ParamDefault(Int64 val)
+        : default_int(val)
+        , default_string("")
+    {}
```
windtalker, line 132: what happens if `val` is not 0?

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+
+    // Max parameter number the regexp_xxx function could receive
+    static constexpr size_t REGEXP_MAX_PARAM_NUM = 2;
+    static constexpr size_t REGEXP_LIKE_MAX_PARAM_NUM = 3;
+    static constexpr size_t REGEXP_INSTR_MAX_PARAM_NUM = 6;
+    static constexpr size_t REGEXP_REPLACE_MAX_PARAM_NUM = 6;
+    static constexpr size_t REGEXP_SUBSTR_MAX_PARAM_NUM = 5;
+
+    template <typename ExprT, typename MatchTypeT>
+    void memorize(const ExprT & pat_param, const MatchTypeT & match_type_param, TiDB::TiDBCollatorPtr collator) const
```
windtalker, line 482: Don't need to check null value here?

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+            }
+        }
+        else
+        {
+            if (!arg->isString())
+            {
+                if (arg->getTypeId() != TypeIndex::Nothing)
+                    throw Exception(fmt::format("Illegal type {} of argument of function {}", arg->getName(), getName()), ErrorCodes::ILLEGAL_TYPE_OF_ARGUMENT);
+                else
+                    *has_data_type_nothing = true;
```
windtalker, line 790: TypeNothing should always be inside TypeNullable, so just throw exception there is ok.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+private:
+    Int64 const_int_val;
+
+    // for vector int
+    const Container * int_container;
+};
+
+// Columns may be const, nullable or plain vector, we can conveniently handle
+// these different type columns with Param.
+template <typename ParamImplType, bool is_null>
```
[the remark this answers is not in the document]
xzhangxian1008: > rename `is_null` to `is_nullable`?

get

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+        , data(int_container_)
+    {}
+
+    Int64 getInt(size_t idx) const { return data.getInt(idx); }
+    void getStringRef(size_t idx, StringRef & dst) const { return data.getStringRef(idx, dst); }
+    String getString(size_t idx) const { return data.getString(idx); }
+
+    bool isNullAt(size_t idx) const
+    {
+        if constexpr (is_null)
```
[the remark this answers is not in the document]
xzhangxian1008: > what abount nullable constant? (*null_map)[idx] maybe out of bound

Resolved

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+            }                                                                                                       \
+            else                                                                                                    \
+            {                                                                                                       \
+                Param<ParamString<true>, false>(param_name)(col_size, col_const->getDataAt(0));                     \
+                next_convertion;                                                                                       \
+            }                                                                                                       \
+        }                                                                                                           \
+        else                                                                                                        \
+        {                                                                                                           \
+            CONVERT_NULL_STR_COL_TO_PARAM((param_name), (converted_col), next_convertion)                              \
```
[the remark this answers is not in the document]
xzhangxian1008: > 

I have renamed these common methods.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1078 @@
...
+
+#define SELF_CLASS_NAME (name)
+#define ARG_NUM_VAR_NAME arg_num
+
+// Unify the name of functions that actually execute regexp
+#define REGEXP_CLASS_MEM_FUNC_IMPL_NAME process
+
+// Common method to convert nullable string column
+// converted_col is impossible to be const here
+#define CONVERT_NULL_STR_COL_TO_PARAM(param_name, converted_col, next_convertion)                                                                                               \
```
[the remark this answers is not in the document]
xzhangxian1008: > maybe rename is to `CONVERT_STR_COL_TO_PARAM` since it also handles non-nullable string col.

ditto

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > what happens if `val` is not 0?

All ParamDefault has been replaced by const ParamString or const ParamInt

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > Don't need to check null value here?

const null case has been handled at the beginning of `executeImpl`, it's impossible to appear const null column

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > TypeNothing should always be inside TypeNullable, so just throw exception there is ok.

okk

dbms/src/Functions/re2Util.cpp
```diff
@@ -0,0 +1,83 @@
...
+    if (!match_type.empty())
+    {
+        for (const auto & c : match_type)
+        {
+            switch (c)
+            {
+            case 'i':
+                /// according to MySQL doc: if either argument is a binary string, the arguments are handled in
+                /// case-sensitive fashion as binary strings, even if match_type contains the i character.
+                /// However, test in MySQL 8.0.25 shows that i flag still take affect even if the collation is binary,
```
windtalker, line 49: make sure this behavior is the same in TiDB/TiKV/TiFlash

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,795 @@
...
+    static constexpr size_t REGEXP_REPLACE_MAX_PARAM_NUM = 6;
+    static constexpr size_t REGEXP_SUBSTR_MAX_PARAM_NUM = 5;
+
+    // We should pre compile the regular expression when:
+    //  - only pattern column is provided and it's a constant column
+    //  - pattern and match type columns are provided and they are both constant columns
+    template <typename ExprT, typename MatchTypeT>
+    std::unique_ptr<Regexps::Regexp> memorize(const ExprT & pat_param, const MatchTypeT & match_type_param, TiDB::TiDBCollatorPtr collator) const
+    {
+        String final_pattern = pat_param.getString(0);
```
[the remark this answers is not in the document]
xzhangxian1008: > Why don't check if `pat` or `match_type` is null?

regexp can be memorized only when pat and match type are all const, and const null situation has been handled at the beginning of the `executeImple`. So, `pat` and `match_type` is impossible to be null here.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,795 @@
...
+            {
+                // expr column must be a nullable column here, so we need to check null for each elems
+                auto nullmap_col = ColumnUInt8::create();
+                typename ColumnUInt8::Container & nullmap = nullmap_col->getData();
+                nullmap.resize(col_size);
+
+                StringRef expr_ref;
+                for (size_t i = 0; i < col_size; ++i)
+                {
+                    if (expr_param.isNullAt(i))
```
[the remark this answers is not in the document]
xzhangxian1008: > why don't check if `pat` or `match_type` is null?

const null situation has been handled at the beginning of `executeImpl`

dbms/src/Functions/re2Util.cpp
xzhangxian1008: > make sure this behavior is the same in TiDB/TiKV/TiFlash

okk

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > Don't need to check null value here?

resolved

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,807 @@
...
+
+    // for vector int
+    const Container * int_container;
+};
+
+// Columns may be const, nullable or plain vector, we can conveniently handle
+// these different type columns with Param.
+template <typename ParamImplType, bool is_nullable>
+class Param
+{
```
gengliqi, line 261: `is_nullable == (null_map == nullptr)` is always true. 
So how about adding some exceptions in constructor functions.

gengliqi (commented):
Lots of micros are hard to read and maintain in the future.
How about using template functions?

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > `is_nullable == (null_map == nullptr)` is always true. So how about adding some exceptions in constructor functions.

okk, I will ensure it

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,807 @@
...
+                regexp = memorize(pat_param, match_type_param, collator);
+                if (regexp == nullptr)
+                {
+                    auto nullmap_col = ColumnUInt8::create();
+                    typename ColumnUInt8::Container & nullmap = nullmap_col->getData();
+                    nullmap.resize(col_size, 1);
+                    res_arg.column = ColumnNullable::create(std::move(col_res), std::move(nullmap_col));
+                    return;
+                }
+            }
```
windtalker, line 662: how about adding an `else` branch to return directly if `col_size == 0`?

tests/fullstack-test/expr/regexp.test
```diff
@@ -71,13 +71,12 @@ mysql> set @@tidb_isolation_read_engines='tiflash'; set @@tidb_enforce_mpp=1; se
...
 mysql> alter table test.t set tiflash replica 1
 func> wait_table test t
-mysql> set @@tidb_isolation_read_engines='tiflash'; set @@tidb_enforce_mpp=1; select data regexp pattern, data regexp '', '' regexp pattern from test.t;
-+---------------------+----------------+-------------------+
-| data regexp pattern | data regexp '' | '' regexp pattern |
-+---------------------+----------------+-------------------+
-|                   1 |              1 |                 1 |
-|                   1 |              1 |                 0 |
-+---------------------+----------------+-------------------+
+mysql> set @@tidb_isolation_read_engines='tiflash'; set @@tidb_enforce_mpp=1; select data regexp pattern, '' regexp pattern from test.t;
```
windtalker, line 77: Need add some tests to test the match type

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > how about adding an `else` branch to return directly if `col_size == 0`?

I think it's unnecessary, as we will add more duplicated codes.

tests/fullstack-test/expr/regexp.test
xzhangxian1008: > Need add some tests to test the match type

I will do this

gengliqi (commented):
Much code of different param type can be reduced by mircors.
https://github.com/pingcap/tiflash/blob/f97e30fb804d43d6c19e2c912b0f765973832968/dbms/src/Interpreters/Aggregator.cpp#L62-L67
https://github.com/pingcap/tiflash/blob/f97e30fb804d43d6c19e2c912b0f765973832968/dbms/src/Interpreters/Aggregator.h#L675-L683
You can see above codes for more details.

xzhangxian1008:
> 

Done

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1009 @@
...
+// Unify the name of functions that actually execute regexp
+#define REGEXP_CLASS_MEM_FUNC_IMPL_NAME executeRegexpFunc
+
+// Do not merge GET_ACTUAL_STRING_PARAM and GET_ACTUAL_INT_PARAM together,
+// as this will generate more useless codes and templates.
+
+#define ENUMERATE_PARAM_VARIANT_CASES(NAME, pv_name, param_name, next_process) \
+    case ParamVariant::ParamType::NAME:                                        \
+    {                                                                          \
+        ParamVariant::Param##NAME *(param_name) = (pv_name).getParam##NAME();  \
```
gengliqi, line 604: How about using a micro for this conversion? If so, `getParamXXX` function family can be removed.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1009 @@
...
+    {                                                                                                  \
+        GET_ACTUAL_STRING_PARAM(EXPR_PV_VAR_NAME, EXPR_PARAM_PTR_VAR_NAME, ({GET_PAT_ACTUAL_PARAM()})) \
+    } while (0);
+
+// The entry to get actual params and execute regexp functions
+#define GET_ACTUAL_PARAMS_AND_EXECUTE() \
+    do                                  \
+    {                                   \
+        GET_EXPR_ACTUAL_PARAM()         \
+    } while (0);
```
gengliqi, line 757: Maybe we can combine these micros into one？

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > How about using a micro for this conversion? If so, `getParamXXX` function family can be removed.

Converted pointer is contained in ParamVariant, it will add complexity when we want to convert it as we have to get it from ParamVariant. Moreover, debugging with macro is hard, I think we should avoid macro as much as possible. At last, these simple functions must be inlined by compiler and there is no performance difference between macro and function.

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > Maybe we can combine these micros into one？

I believe that readability is very important and it's more clear to seperate them.

dbms/src/Functions/FunctionsRegexp.h
gengliqi: How about
```suggestion
        #define ParamName(NAME) ParamVariant::Param##NAME
        #define ToParamPtr(NAME, ptr) (reinterpret_cast<ParamName(NAME) *>(ptr))
        auto *(param_name) = ToParamPtr(NAME, (pv_name).param);  \
```

xzhangxian1008: > How about this?

okk, I will resolve it.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -0,0 +1,1007 @@
...
+#define M(NAME, pv_name, param_name, next_process)     \
+    case ParamType::NAME:                              \
+    {                                                  \
+        delete reinterpret_cast<Param##NAME *>(param); \
+        break;                                         \
+    }
+
+                // Expand the macro to enumerate int param cases
+                APPLY_FOR_PARAM_INT_VARIANTS(M, placeholder1, placeholder2, placeholder3)
+#undef M
```
gengliqi, line 495: Seems these two macros are the same?  
BTW, `ENUMERATE_PARAM_VARIANT_CASES` can be used here.

gengliqi (commented):
Rest LGTM

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > Seems these two macros are the same? BTW, `ENUMERATE_PARAM_VARIANT_CASES` can be used here.

Done

gengliqi (approved):
LGTM

windtalker (approved):
LGTM

PR #6098, opened 2022-10-11
Fix I/O limiter auto-tuner.

### LINKED ISSUE #6099: IOLimitTuner may not work well in a special case
If both `max_bg_read_bytes_per_sec` and `max_fg_read_bytes_per_sec` are less than `io_config.min_bytes_per_sec` and `bg_read` runs out of the bandwidth quota, but `fg_read` has not been used.

The usage rate of read is `(max_bg_read_bytes_per_sec + 0 ) / (max_bg_read_bytes_per_sec + max_fg_read_bytes_per_sec)`, about 50%. It is less than `io_config.medium_pct` (60% by default), so watermark is `LOW`.

The `LOW` watermark means that bandwidth quota of read is sufficient since the usage rate is less than 60%, so it is unnessary to increase its bandwidth quota by decreasing the bandwidth quota of write.

In this case, `bg_read` will increase its bandwidth quota by decreasing the bandwidth quota of `fg_read`. However, `fg_read` is less than `io_config.min_bytes_per_sec`, it is too small to decrease.

JinheLin:

Issue Number: close #6099

Try to tune bandwidth quota between read-bandwidth-quota and write-bandwidth-quota if bandwidth quota of bg_read/fg_read (or bg_write/fg_write) is small.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Encryption/RateLimiter.h
```diff
@@ -412,13 +412,14 @@ class IOLimitTuner
...
     Watermark writeWatermark() const
     {
-        return getWatermark(writePct());
+        return getWatermark(fg_write_stat, bg_write_stat, writePct());
     }
     Watermark readWatermark() const
     {
-        return getWatermark(readPct());
+        return getWatermark(fg_read_stat, bg_read_stat, readPct());
     }
```
Lloyd-Pottiger, line 420: we need to tune bg and fg, why revise those two functions? Should we revise https://github.com/pingcap/tiflash/blob/187a46c19267746d1f0a89234bece30ea1ff7eee/dbms/src/Encryption/RateLimiter.cpp#L819-L820

Lloyd-Pottiger: thank you, I got it.

dbms/src/Encryption/RateLimiter.cpp
```diff
@@ -897,4 +897,32 @@ IOLimitTuner::Watermark IOLimitTuner::getWatermark(int pct) const
...
 }
+
+IOLimitTuner::Watermark IOLimitTuner::getWatermark(const LimiterStatUPtr & fg, const LimiterStatUPtr & bg, int pct) const
+{
+    // There is a cornar case:
+    // 1. Both `fg_mbps` and `bg_mbps` are less than `min_bytes_per_sec`.
+    // 2. `bg` runs out of the quota, but `fg` has not been used.
+    // 3. In this case, `pct` can be less than `medium_pct` and watermark is LOW.
+    // 4. Low watermark means it is unnecessary to increase the quota by descrease the quota of others.
+    // 5. However, because `fg_mbps` is less than `min_bytes_per_sec`, `bg_mbps` cannot be increased by decreasing `fg_mbps`.
```
Lloyd-Pottiger, line 908: Actually there are four similar corner case, it can hard to summary them,  maybe take one as an example is better?

Lloyd-Pottiger (approved):
LGTM

flowbehappy (approved):
LGTM

PR #6128, opened 2022-10-13
Set linker name when using AppleClang

Lloyd-Pottiger:
Signed-off-by: Lloyd-Pottiger <yan1579196623@gamil.com>

Issue Number: ref #6129

Problem Summary:

set linker name equal to `ld` when using AppleClang.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,7 +182,7 @@ if (ENABLE_LLVM_PGO)
     endif ()
 endif ()
 
-if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     # clang: warning: argument unused during compilation: '-stdlib=libc++'
     # clang: warning: argument unused during compilation: '-specs=/usr/share/dpkg/no-pie-compile.specs' [-Wunused-command-line-argument]
     set (COMMON_WARNING_FLAGS "${COMMON_WARNING_FLAGS} -Wno-unused-command-line-argument")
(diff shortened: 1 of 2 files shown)


JaySon-Huang:
Can you elaborate more why the warnings appears and why this can suppress the warnings?

ywqzzy (approved):

Lloyd-Pottiger:
> Can you elaborate more why the warnings appears and why this can suppress the warnings?

I found an explanation: https://discourse.itk.org/t/visibility-settings/596/4

breezewish:
Is it possible to specify this flag more precisely (by setting it as a flag to some targets), instead of globally?

Lloyd-Pottiger:
> Is it possible to specify this flag more precisely (by setting it as a flag to some targets that set visibility incorrectly), instead of globally? The major reason drives me think about this direction is setting this flag causes everything to be rebuilt .

In my device, `ninja tiflash` output 164106 lines ld warnings each time..... almost every subdirectories......

> It would be also better to discover why only MacOS platforms are affected, while we using LLVM compiler for both Linux and MacOS.

I am working on it, but have not find the answer yet.

Lloyd-Pottiger:
@JaySon-Huang @breezewish it is caused by the default linker `ld` in apple. If we use llvm linker `lld`, this problem can fixed. But why the `ld` report those warnings, I can not find the answer.

CMakeLists.txt
```diff
@@ -248,17 +248,22 @@ endif()
...
 find_program (GOLD_PATH NAMES "gold")
+if (APPLE)
+    find_program (LD_PATH NAMES "ld")
+endif ()
 
-if (NOT LINKER_NAME AND NOT APPLE)
-    if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND LLD_PATH)
+if (NOT LINKER_NAME)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND LD_PATH)
+        set (LINKER_NAME "ld")
```
breezewish, line 257: Why do we specifically use ld instead of lld for AppleClang?

Lloyd-Pottiger: use lld may cause https://github.com/pingcap/tiflash/issues/5368

breezewish (approved):

PR #6146, opened 2022-10-15
storage: include memtable in for_update snapshots

### LINKED ISSUE #6045: Include memtable in the snapshot for update
## Enhancement

Currently, when snapshot is generated for update, memtable is not included. This means, we need to ensure data is flushed in many scenarios. This characteristic has caused several issues previously:

- In https://github.com/pingcap/tiflash/pull/4959, as there is no memtable in the snapshot, we copy memtable directly during merge, which causes the issue.
- In https://github.com/pingcap/tiflash/issues/6039, the GC ignores the delete range in the memtable.

The reason that memtable is not included for update is to ensure that no data will be written to the shared memtable, because there may be appendable column files. As an improvement, we could mark column files to be not appendable, when generating a snapshot for update. In this way, we could also ensure that new writes will not change the result of the snapshot.

After this improvement takes effect, may be we could revert PRs like https://github.com/pingcap/tiflash/pull/4959 to reduce unnecessary flushes. Leaving unflushed data for long time is still a problem, because it means data will be lost after a power lost (when persisted index has been pushed forward). Instead of trying to flush the data in more cases, we could also use some background threads to better discover them.

breezewish:

Issue Number: close #6045

Problem Summary:

1. Include memtable in the ForUpdate snapshot. The appendable ColumnFileTiny will be changed to not appendable when ForUpdate snapshot is created.

   Consequently, the DeltaMerge, Merge and Split process is changed. Take DeltaMerge as an example, the new process looks like:

image

   In the apply stage, we only include new memtable CFs and new persisted CFs in the new segment. There may be CFs flushed from memtable into the persisted. These CFs will not be included. This is handled by the new `DeltaVS::cloneNewlyAppendedColumnFiles`. Here is a visualization of its step:
   
   image

2. Add `merge(start, end)` in the DeltaMergeStore test framework, to merge segments in `[start, end)`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/Delta/DeltaValueSpace.cpp
```diff
@@ -65,15 +66,135 @@ void DeltaValueSpace::saveMeta(WriteBatches & wbs) const
...
+}
+
+std::pair<ColumnFiles, ColumnFilePersisteds> DeltaValueSpace::cloneAllColumnFiles(
+    const DeltaValueSpace::Lock &,
+    DMContext & context,
+    const RowKeyRange & target_range,
+    WriteBatches & wbs) const
+{
+    auto [new_mem_files, flushed_mem_files] = mem_table_set->diffColumnFiles({});
+    RUNTIME_CHECK(flushed_mem_files.empty());
```
flowbehappy, line 187: Why `flushed_mem_files` should be empty? Can you add some comments?

flowbehappy (approved):
LGTM with a minor comment

JaySon-Huang (approved):
lgtm

PR #6173, opened 2022-10-20
Fine grained partition writer optimization

### LINKED ISSUE #6157: Join & Aggregation Fine Grained Partition Optimization
## Enhancement
This issue tracks Join&Aggregation operator utilizing fine grained partition tech to optimize its implementation.
For Join operator:
The time saved by utilizing fine grained partition tech in Prob side is much fewer than the cost brought by fine grained partition itself, so just consider optimize Build side. Besides, the optimization requires:

1. Prob side uses the same (or compatible) hash function to choose build side stream, and uses original hash function to do the match work.
2. The prehash key is the same as the Join hash table's hash key

For Agg operator:
Nothing special.

Related issue: https://github.com/pingcap/tiflash/issues/4631
## TiDB

- [x] Support fine grained parition Join & Aggregation in physical plan https://github.com/pingcap/tidb/pull/40121
- [x] Add column prune rule for exchange + join in physical plan phase https://github.com/pingcap/tidb/pull/38536

## TiFlash

- [x] Optimize FineGrainedShuffleWriter to reuse block memory https://github.com/pingcap/tiflash/pull/6173
- [x] Optimize Receiver side decode implementation to include squashing functionality https://github.com/pingcap/tiflash/pull/6202
- [x] Utilize Fine Grained Partition information to optimize Join & Aggregation https://github.com/pingcap/tiflash/pull/6279
- [x] Gtests for utilizing Fine Grained Partition information to optimize Join & Aggregation https://github.com/pingcap/tiflash/pull/6445

After apply all these optimizations, TPCH_100 3 nodes, 2-tiflash replica, total execution time will reduce about 10%, 80.6 => 72.3:
acbfac11-b14f-4347-b4a6-7cb3fc371e7f

yibin87:

Issue Number: close #6157 

Problem Summary:
1. Current FineGrainedShuffleWriter allocates new memory for scattered columns each time when batchWrite is triggered.
2. "Long tail problem", cached blocks under thres won't write until readSuffix, which's executed in a single thread. 

This PR aims to reuse memory for scattered columns and "Long tail problem".
Referenced PR: [https://github.com/pingcap/tiflash/pull/3787](https://github.com/pingcap/tiflash/pull/3787)
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Coprocessor/StreamingDAGResponseWriter.cpp
```diff
@@ -78,8 +78,15 @@ void StreamingDAGResponseWriter<StreamWriterPtr>::finishWrite()
...
-void StreamingDAGResponseWriter<StreamWriterPtr>::write(const Block & block)
+void StreamingDAGResponseWriter<StreamWriterPtr>::write(const Block & block, bool last_null_block)
 {
+    if (last_null_block)
+    {
+        if (rows_in_blocks > 0)
+            encodeThenWriteBlocks<false>();
+        return;
+    }
+
```
SeaRise, line 89:
```suggestion
template <bool last_null_block>
void StreamingDAGResponseWriter<StreamWriterPtr>::write(const Block & block)
{
    if constexpr (last_null_block)
    {
        if (rows_in_blocks > 0)
            encodeThenWriteBlocks<false>();
        return;
    }

```

yibin87: Nice suggestion, extracted a new interface 'flush' instead of reusing the write interface to make it more clear.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,105 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
-                    final_dest_tbl_columns[bucket_idx][col_id]->insertRangeFrom(*src_col, 0, src_col->size());
-                }
-            }
+            header = blocks[0].cloneEmpty();
+            num_columns = header.columns();
+            // fine_grained_shuffle_stream_count is in (0, 1024], and partition_num is uint16_t, so will not overflow.
+            num_bucket = partition_num * fine_grained_shuffle_stream_count;;
+            partition_key_containers_for_reuse.resize(collators.size());
+            resetScatterColumns();
+            inited = true;
```
SeaRise, line 123: Can we move to `init(const Block & header)` and Called by `readPrefixImpl`?

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,105 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
         }
-        assert(blocks.empty());
-        rows_in_blocks = 0;
 
-        // For i-th stream_count buckets, send to i-th tiflash node.
-        for (size_t bucket_idx = 0; bucket_idx < bucket_num; bucket_idx += fine_grained_shuffle_stream_count)
+        for (const auto & block : blocks)
+            HashBaseWriterHelper::scatterColumnsInplace(block, num_bucket, collators, partition_key_containers_for_reuse, partition_col_ids, hash, selector, scattered);
```
SeaRise, line 127: It looks like this optimization can also be used in `HashPartitionWriter`

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,105 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
+            partition_key_containers_for_reuse.resize(collators.size());
+            resetScatterColumns();
+            inited = true;
         }
-        assert(blocks.empty());
-        rows_in_blocks = 0;
 
-        // For i-th stream_count buckets, send to i-th tiflash node.
-        for (size_t bucket_idx = 0; bucket_idx < bucket_num; bucket_idx += fine_grained_shuffle_stream_count)
+        for (const auto & block : blocks)
```
SeaRise, line 126: Maybe we should inherit the original way 
```
while (!blocks.empty())
{
    const auto & block = blocks.back();
    xxx
    blocks.pop_back();
}
```
to release Block's memory as soon as possible?

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,105 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
     size_t rows = block.rows();
     rows_in_blocks += rows;
     if (rows > 0)
     {
         blocks.push_back(block);
+        cached_block_count++;
     }
 
-    if (static_cast<UInt64>(rows_in_blocks) >= fine_grained_shuffle_batch_size)
+    if (cached_block_count == fine_grained_shuffle_stream_count || static_cast<UInt64>(rows_in_blocks) >= fine_grained_shuffle_batch_size * fine_grained_shuffle_stream_count)
```
SeaRise, line 83: Why not just use `blocks.size()`?

dbms/src/DataStreams/ExchangeSenderBlockInputStream.cpp
```diff
@@ -42,6 +42,10 @@ Block ExchangeSenderBlockInputStream::readImpl()
         total_rows += block.rows();
         writer->write(block);
     }
+    else 
+    {
+        writer->flush();
```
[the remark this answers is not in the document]
yibin87: Not proper here, I think. finishWrite interface will gather summary info among streams, and should be invoked in a single thread

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Nice suggestion, I'll try.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Yes, it actually should work for all batchWriter, we can refact it later.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: You're right, it's a reasonable way, I'll change it.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Yeah, I'll remove this field.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
SeaRise: ok

dbms/src/DataStreams/ExchangeSenderBlockInputStream.cpp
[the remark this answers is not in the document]
SeaRise: ok

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Done

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Done

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Done.

dbms/src/Flash/Mpp/tests/gtest_mpp_exchange_writer.cpp
```diff
@@ -408,6 +410,7 @@ try
         *dag_context_ptr);
     for (const auto & block : blocks)
         dag_writer->write(block);
+    dag_writer->flush();
     dag_writer->finishWrite();
```
windtalker, line 414: maybe we can also call `flush` inside `finishWrite`, so these test code don't need to be changed.

yibin87: For example, "testBatchWriteFineGrainedShuffle" test, expect the write and finishWrite only produce one packet.

windtalker: It looks like that these tests are meaningless since we now will always flush?

yibin87: Yeah, just invoke it as a test.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,107 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
+
     size_t rows = block.rows();
     rows_in_blocks += rows;
     if (rows > 0)
     {
         blocks.push_back(block);
     }
 
-    if (static_cast<UInt64>(rows_in_blocks) >= fine_grained_shuffle_batch_size)
+    if (blocks.size() == fine_grained_shuffle_stream_count || static_cast<UInt64>(rows_in_blocks) >= fine_grained_shuffle_batch_size * fine_grained_shuffle_stream_count)
```
SeaRise, line 94: It seems that `fine_grained_shuffle_batch_size * fine_grained_shuffle_stream_count` can be pre-calculated

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,107 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
-
     if (!blocks.empty())
     {
         assert(rows_in_blocks > 0);
         assert(fine_grained_shuffle_stream_count <= 1024);
 
-        // fine_grained_shuffle_stream_count is in (0, 1024], and partition_num is uint16_t, so will not overflow.
-        uint32_t bucket_num = partition_num * fine_grained_shuffle_stream_count;
-
         HashBaseWriterHelper::materializeBlocks(blocks);
```
SeaRise, line 125: from https://github.com/pingcap/tiflash/pull/6173/files#diff-ad533224c30a2d203c877d1c8eaa03cc0bcacef358a8169cf5f8915fe7970265R100-R113,
it seems that we don't need to call `HashBaseWriterHelper::materializeBlocks`.

yibin87: Materialize constant columns, seems work for ser/deser columns, not for scatter/insertRangeFrom:
https://github.com/pingcap/tiflash/blob/40ad20c9439b18221297ed2b87d9d7421d17b174/dbms/src/Flash/Coprocessor/CHBlockChunkCodec.cpp#L70

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Yes, will change it.

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +58,107 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
 }
 
+template <class StreamWriterPtr>
+void FineGrainedShuffleWriter<StreamWriterPtr>::prepare(const Block & sample_block)
+{
+    header = sample_block.cloneEmpty();
+    num_columns = header.columns();
+    // fine_grained_shuffle_stream_count is in (0, 1024], and partition_num is uint16_t, so will not overflow.
+    num_bucket = partition_num * fine_grained_shuffle_stream_count;
+    ;
```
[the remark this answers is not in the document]
yibin87: Done

dbms/src/Columns/tests/gtest_column_scatterTo.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+#include <Columns/ColumnArray.h>
+#include <Columns/ColumnFixedString.h>
+#include <Columns/ColumnNothing.h>
+#include <Columns/ColumnSet.h>
+#include <Columns/ColumnTuple.h>
+#include <TestUtils/FunctionTestUtils.h>
+#include <TestUtils/TiFlashTestBasic.h>
+
+#include "Common/COWPtr.h"
```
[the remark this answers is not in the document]
yibin87: Done

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
SeaRise: yes, you are right.

dbms/src/Columns/tests/gtest_column_scatterTo.cpp
```diff
@@ -0,0 +1,233 @@
...
+        }
+        ASSERT_COLUMN_EQ(expected, actual);
+    }
+
+    void doTestWork(ColumnWithTypeAndName & col_with_type_and_name)
+    {
+        auto column_ptr = col_with_type_and_name.column;
+        ASSERT_TRUE(rows == column_ptr->size());
+        MutableColumns scatterTo_res(scattered_num_column);
+        for (size_t i = 0; i < scattered_num_column; i++)
```
SeaRise, line 53:
```suggestion
        for (size_t i = 0; i < scattered_num_column; ++i)
```
from https://clickhouse.com/docs/en/development/style/

SeaRise (approved):
LGTM

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +59,113 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
+
     size_t rows = block.rows();
     rows_in_blocks += rows;
     if (rows > 0)
     {
         blocks.push_back(block);
     }
 
-    if (static_cast<UInt64>(rows_in_blocks) >= fine_grained_shuffle_batch_size)
+    if (blocks.size() == fine_grained_shuffle_stream_count || static_cast<UInt64>(rows_in_blocks) >= batch_send_row_limit)
```
guo-shaoge, line 101: 1. So we changed the meaning of `fine_grained_shuffle_batch_size`, maybe need to talk to PM and also change the [doc](https://docs.pingcap.com/zh/tidb/dev/system-variables#tiflash_fine_grained_shuffle_batch_size-%E4%BB%8E-v620-%E7%89%88%E6%9C%AC%E5%BC%80%E5%A7%8B%E5%BC%95%E5%85%A5)
2. Just a question: why trigger sending data when blocks.size() equals to fine_grained_shuffle_stream_count

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +59,113 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
 
+template <class StreamWriterPtr>
+void FineGrainedShuffleWriter<StreamWriterPtr>::prepare(const Block & sample_block)
+{
+    /// Materialize sample_block so that header and reserved scatterColumns are full columns
+    /// Because ser/der don't support constant columns now
+    std::vector<Block> sample_blocks;
+    sample_blocks.push_back(sample_block);
+    HashBaseWriterHelper::materializeBlocks(sample_blocks);
+    header = sample_blocks.back().cloneEmpty();
```
SeaRise, line 70:
```suggestion
    header = materializeBlock(sample_block);
```
from `#include <DataStreams/materializeBlock.h>`

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Good suggestion and question. For 1, I'll follow. For 2, just think like this:
Previously, no fine_grained_shuffle_stream, N partitions, then one block => N blocks
With fine_grained_shuffle_stream, N partitions, M fine_grained_shuffle_stream_count,  then one block => N * M blocks, thus here use blocks.size() == stream_count, can make final block close to previous block size.

guo-shaoge: Got it. So compared to the original situation, will trigger sending more frequent when blocks are small. Guess this is the reason why `long tail problem` can be fix?

yibin87: Long tail problem is actually fixed by the "flush" interface, which will be invoked concurrently in different threads. Previously, the remain cached blocks that are below thres, will be flushed in "finishWrite" interface, which is invoked sequentially in a single thread.

guo-shaoge (approved):

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +59,111 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
     }
 }
 
+template <class StreamWriterPtr>
+void FineGrainedShuffleWriter<StreamWriterPtr>::prepare(const Block & sample_block)
+{
+    /// Materialize sample_block so that header and reserved scatterColumns are full columns
+    /// Because ser/der don't support constant columns now
+    header = sample_block.cloneEmpty();
```
windtalker, line 67: The `column` in `ColumnWithTypeAndName` can be nullptr sometimes, so I'm a little warried that `HashBaseWriterHelper::materializeBlock` may meet a npe problem, how about creating the scatter columns use the type in `ColumnWithTypeAndName`?

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
```diff
@@ -57,81 +59,111 @@ void FineGrainedShuffleWriter<StreamWriterPtr>::finishWrite()
...
+    HashBaseWriterHelper::materializeBlock(header);
+    num_columns = header.columns();
+    // fine_grained_shuffle_stream_count is in (0, 1024], and partition_num is uint16_t, so will not overflow.
+    num_bucket = partition_num * fine_grained_shuffle_stream_count;
+    partition_key_containers_for_reuse.resize(collators.size());
+    resetScatterColumns();
```
windtalker, line 73: Looks like `resetScatterColumns` is only used in `prepare`, it's more like `initScatterColumns`?

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Done

dbms/src/Flash/Mpp/FineGrainedShuffleWriter.cpp
yibin87: Renamed to initScatterColumns

windtalker (approved):
LGTM

PR #6188, opened 2022-10-26
remove is_partial_sort

### LINKED ISSUE #6189: Remove is_partial_sort from physical sort operator
## Enhancement
Remove is_partial_sort from physical sort operator, because is_partial_sort will be removed from planner.

fixdb:
is_partial_sort is always true, planner removed it, so we should remove it here in TiFlash too.

Issue Number: close #6189

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


windtalker:
`is_partial_sort` is added for the sort generated by window function, and it means `only sort for each partition data is enough`, if in the future, TiFlash support the normal sort operator, we still need a flag to distinguish normal sort and the sort generated by window function.

LittleFall:
> `is_partial_sort` is added for the sort generated by window function, and it means `only sort for each partition data is enough`, if in the future, TiFlash support the normal sort operator, we still need a flag to distinguish normal sort and the sort generated by window function.

I think the path we support the normal sort is adding a merge exchange before sort, so single TiFlash node will always do a "partial" sort

PR #6193, opened 2022-10-27
KVStore: Force persist when meet Prepare(Finish)Flashback

CalvinNeo:
Signed-off-by: CalvinNeo <calvinneo1995@gmail.com>

Issue Number: close https://github.com/pingcap/tiflash/issues/6194

Problem Summary:

Problems may raise when restart after Prepare(Finish)Flashback has changed RegionLocalState. 

Currently, Proxy will not persist data and apply state in these cases. So there maybe a state with newer RegionLocalState and older ApplyState and data.

If we replay from the old ApplyState with the new RegionLocalState, problems can raise.

In this PR, we:
1. Provide a way to force a persistence when need.
    Notice if this fix can work, we don't actually need to call this. So it is an enhancement.
3. Force persist when Prepare(Finish)Flashback. We previously do not persist, and return None(not persist apply state).

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Transaction/KVStore.cpp
```diff
@@ -348,7 +348,24 @@ bool KVStore::tryFlushRegionData(UInt64 region_id, bool try_until_succeed, TMTCo
         LOG_WARNING(log, "region {} [index: {}, term {}], not exist when flushing, maybe have exec `RemoveNode` first", region_id, index, term);
         return true;
     }
-    return canFlushRegionDataImpl(curr_region_ptr, true, try_until_succeed, tmt, region_task_lock, index, term);
+    if (force_persist)
+    {
+        if (curr_region_ptr == nullptr)
```
lidezhu, line 353: `curr_region_ptr == nullptr` is already checked before.

dbms/src/Storages/Transaction/KVStore.cpp
```diff
@@ -379,27 +396,32 @@ bool KVStore::canFlushRegionDataImpl(const RegionPtr & curr_region_ptr, UInt8 fl
...
-            curr_region.markCompactLog();
-            curr_region.cleanApproxMemCacheInfo();
-            return true;
-        }
-        else
-        {
-            return false;
-        }
+        LOG_DEBUG(log, "{} flush region due to tryFlushRegionData, index {} term {}", curr_region.toString(false), index, term);
+        forceFlushRegionDataImpl(curr_region, try_until_succeed, tmt, region_task_lock, index, term);
```
lidezhu, line 400:
```suggestion
        return forceFlushRegionDataImpl(curr_region, try_until_succeed, tmt, region_task_lock, index, term);
```

lidezhu (approved):

dbms/src/Storages/Transaction/KVStore.cpp
CalvinNeo: fixed

hehechen (approved):

PR #6202, opened 2022-10-28
Exchange receiver decode optimization to do squashing work at the same time

### LINKED ISSUE #6157: Join & Aggregation Fine Grained Partition Optimization

yibin87:

Issue Number: close #6157 

Problem Summary:

Previously, TiFlash adds a Squash transform after ExchangeReceiver, because the block output by ExchangeReceiver might be too small and not efficient to handle. However, the two-stage solution would introduce too many memory operations, like column allocations and de-allocations. So in this PR, the one-stage solution is provided.

For a simple local test, with 10 integer columns, 1024 rows per block, performance improves from: 140ms => 60ms.
 
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


fuzhe1989:
Fine-grained partitioning will exacerbate the chunk fragmentation in exchange and would benefit from this pr much more from normal cases.

yibin87:
> Fine-grained partitioning will exacerbate the chunk fragmentation in exchange and would benefit from this pr much more from normal cases.

Yeah, and for large-scale clusters, it would be benefit also.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
```diff
@@ -128,51 +132,54 @@ class TiRemoteBlockInputStream : public IProfilingBlockInputStream
...
+
+                total_rows += decode_detail.rows;
+                LOG_TRACE(
+                    log,
+                    "recv {} rows from remote for {}, total recv row num: {}",
+                    decode_detail.rows,
+                    result.req_info,
+                    total_rows);
+                if (decode_detail.rows > 0)
+                    return true;
```
SeaRise, line 180: It seems that we need to check all results before returning true

dbms/src/DataStreams/TiRemoteBlockInputStream.h
```diff
@@ -128,51 +132,54 @@ class TiRemoteBlockInputStream : public IProfilingBlockInputStream
...
+                // else continue
```
SeaRise, line 181: This line of comment is a description of `while(true)`.
Need to move to the outer loop.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -692,48 +694,97 @@ DecodeDetail ExchangeReceiverBase<RPCContext>::decodeChunks(
...
     {
         LOG_ERROR(exc_log, "stream_id out of range, stream_id: {}, total_stream_count: {}", stream_id, msg_channels.size());
-        return ExchangeReceiverResult::newError(0, "", "stream_id out of range");
+        results.push_back(ExchangeReceiverResult::newError(0, "", "stream_id out of range"));
+        return results;
     }
-    std::shared_ptr<ReceivedMessage> recv_msg;
-    if (msg_channels[stream_id]->pop(recv_msg) != MPMCQueueResult::OK)
+
+    while (true)
```
SeaRise, line 728: It looks like both `while(true)` and `std::vector<ExchangeReceiverResult>` are unnecessary, we can do this in `TiRemoteBlockInputStream::fetchRemoteResult`

SeaRise: such as
```
auto result = remote_reader->nextResult(block_queue, sample_block, stream_id);
if (result.meet_error)
    throw Exception()
if (result.await)
    continue;
```

dbms/src/DataStreams/TiRemoteBlockInputStream.h
```diff
@@ -193,6 +200,8 @@ class TiRemoteBlockInputStream : public IProfilingBlockInputStream
         execution_summaries.resize(source_num);
         connection_profile_infos.resize(source_num);
         sample_block = Block(getColumnWithTypeAndName(toNamesAndTypes(remote_reader->getOutputSchema())));
+        if constexpr (is_streaming_reader)
+            decoder_ptr = std::make_unique<CHBlockChunkDecodeAndSquash>(sample_block, 8192);
```
SeaRise, line 204: Maybe we can change to `context.getSettingsRef().max_block_bytes`?
I don't know the reason for using the magic number 8192..

dbms/src/DataStreams/TiRemoteBlockInputStream.h
yibin87: Good point, it works now because the nextResult implementation will ensure only the last result having rows > 0. I'll consider it with your merging two "while" suggestion together.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
yibin87: It comes from previous SquashingBlockInputStream's parameter. Not sure its original source。

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "ExchangeReceiver.h"
```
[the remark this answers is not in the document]
yibin87: Will fix it.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
windtalker: +1

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
```diff
@@ -0,0 +1,87 @@
...
+        codec.readBlockMeta(istr, columns, rows);
+
+        auto mutable_columns = accumulated_block->mutateColumns();
+        for (size_t i = 0; i < columns; ++i)
+        {
+            ColumnWithTypeAndName column;
+            codec.readColumnMeta(i, istr, column);
+
+            /// Data
+            MutableColumnPtr read_column = column.type->createColumn();
```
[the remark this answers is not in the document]
yibin87: Ah, removed it now.

dbms/src/Flash/Coprocessor/tests/gtest_ti_remote_block_inputstream.cpp
```diff
@@ -0,0 +1,448 @@
...
+#include <Storages/Transaction/TiDB.h>
+#include <TestUtils/ColumnGenerator.h>
+#include <TestUtils/FunctionTestUtils.h>
+#include <TestUtils/TiFlashTestBasic.h>
+#include <TestUtils/TiFlashTestEnv.h>
+#include <gtest/gtest.h>
+
+#include <Flash/Coprocessor/StreamingDAGResponseWriter.cpp>
+#include <Flash/Mpp/BroadcastOrPassThroughWriter.cpp>
+#include <Flash/Mpp/ExchangeReceiver.cpp>
```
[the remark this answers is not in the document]
yibin87: Done

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
yibin87: Done, introduced the rowsInfoOnly flag to determine if it's the result previous to EOF.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
yibin87: Recover the comment after merging two while loops.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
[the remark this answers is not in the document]
yibin87: Done.

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+CHBlockChunkDecodeAndSquash::CHBlockChunkDecodeAndSquash(
+    const Block & header,
+    size_t rows_limit_)
+    : codec(header)
+    , rows_limit(rows_limit_)
+{
+}
+
+std::optional<Block> CHBlockChunkDecodeAndSquash::decodeAndSquash(const String & str)
```
SeaRise, line 28: It seems to be fine to use `Block`.
https://github.com/pingcap/tiflash/blob/8b123a83aa6150768b746d089f73dd8205314cc8/dbms/src/Core/Block.h#L113-L114

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (accumulated_block && accumulated_block->rows() >= rows_limit)
+    {
+        /// Return accumulated data and reset accumulated_block
+        res.swap(accumulated_block);
+        return res;
+    }
+    return res;
+}
+
+std::optional<Block> CHBlockChunkDecodeAndSquash::flush()
```
SeaRise, line 75: ditto

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -692,48 +692,83 @@ DecodeDetail ExchangeReceiverBase<RPCContext>::decodeChunks(
...
+    if (this->state != DB::ExchangeReceiverState::NORMAL)
+    {
+        return DB::ExchangeReceiverResult::newError(0, DB::ExchangeReceiverBase<RPCContext>::name, DB::constructStatusString(this->state, this->err_msg));
+    }
+    else
+    {
+        /// If there are cached data in squashDecoder, then just push the block and return EOF next iteration
+        if (last_block && last_block->rows() > 0)
+        {
+            const auto & result = DB::ExchangeReceiverResult::newRowsInfoOnly(last_block->rows());
```
SeaRise, line 755: Why not return `ExchangeReceiverResult::newOk()`?
What is the difference between `newOk` and `newRowsInfoOnly`?

dbms/src/DataStreams/TiRemoteBlockInputStream.h
```diff
@@ -155,21 +159,25 @@ class TiRemoteBlockInputStream : public IProfilingBlockInputStream
...
+            if (!result.rowsInfoOnly)
+            {
+                size_t index = 0;
+                if constexpr (is_streaming_reader)
+                    index = result.call_index;
 
-            size_t index = 0;
-            if constexpr (is_streaming_reader)
-                index = result.call_index;
+                ++connection_profile_infos[index].packets;
```
SeaRise, line 170:
```suggestion
                connection_profile_infos[index].packets += decode_detail.packets;
```
And then we don't need to use rowsInfoOnly.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
yibin87: Introduce a new flag to indicate that the call_index and other info in the result is not usable.

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
yibin87: There's a little differences between block's bool and decodeAndSquash return value:
Block's bool just indicate whether there the block has column info, usually used to check if it is the end of a stream. And other corner case might produce it also.
optional value just make it more clear.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
yibin87: Actually, it is dirty here, since the squash logic will seperate block and packets, somehow breaks the outside expection.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
yibin87: Finally, sacrified the accurate of index info, use newOk instead of newRowsInfoOnly to keep protocal simple.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
yibin87: Extract a new constexpr,  not moved to Settings, for not seeing the situation that users/us will tune this parameter for performance usage.

SeaRise: ok

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
SeaRise: ok

dbms/src/Flash/Coprocessor/CHBlockChunkCodec.cpp
```diff
@@ -120,21 +143,81 @@ std::unique_ptr<ChunkCodecStream> CHBlockChunkCodec::newCodecStream(const std::v
...
+
+    for (size_t i = 0; i < columns; ++i)
+    {
+        ColumnWithTypeAndName column;
+        readColumnMeta(i, istr, column);
+
+        /// Data
+        MutableColumnPtr read_column = column.type->createColumn();
+        if (reserve_size > 0)
+            read_column->reserve(reserve_size);
```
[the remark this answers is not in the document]
yibin87: Make sense, Done.

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        auto mutable_columns = accumulated_block->mutateColumns();
+        for (size_t i = 0; i < columns; ++i)
+        {
+            ColumnWithTypeAndName column;
+            codec.readColumnMeta(i, istr, column);
+
+            if (rows) /// If no rows, nothing to read.
+                CHBlockChunkCodec::readData(*column.type, *(mutable_columns[i]), istr, rows);
+        }
```
SeaRise, line 62:
```suggestion
        if (rows) /// If no rows, nothing to read.
        {
            auto mutable_columns = accumulated_block->mutateColumns();
            for (size_t i = 0; i < columns; ++i)
            {
                ColumnWithTypeAndName column;
                codec.readColumnMeta(i, istr, column);
                CHBlockChunkCodec::readData(*column.type, *(mutable_columns[i]), istr, rows);
            }
        }
```

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -692,48 +692,86 @@ DecodeDetail ExchangeReceiverBase<RPCContext>::decodeChunks(
...
+}
+
+template <typename RPCContext>
+ExchangeReceiverResult ExchangeReceiverBase<RPCContext>::handleUnnormalChannel(
+    std::queue<Block> & block_queue,
+    std::unique_ptr<IChunkDecodeAndSquash> & decoder_ptr)
+{
+    std::optional<Block> last_block;
+    if (decoder_ptr)
+        last_block = decoder_ptr->flush();
```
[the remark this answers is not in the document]
yibin87: You're right, it can't be  null here. Changed it.

dbms/src/DataStreams/TiRemoteBlockInputStream.h
```diff
@@ -222,6 +230,7 @@ class TiRemoteBlockInputStream : public IProfilingBlockInputStream
         return execution_summaries_inited[index].load() ? &execution_summaries[index] : nullptr;
     }
 
+    size_t getTotalRows() const { return total_rows; }
```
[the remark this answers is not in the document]
yibin87: Done

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
yibin87: Nice suggestion, make it clearner and better performance. Nice!

windtalker: But if `rows == 0`, we can skip `codec.readColumnMeta`?

SeaRise: Because it will skip `CodecUtils::checkDataTypeName`?

SeaRise: Now I think the original is better...

yibin87: It sounds ok to pass all these, but with the risk of missing potential error?

dbms/src/Flash/Coprocessor/CodecUtils.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+    if (expected != actual)
+        throw Exception(
+            fmt::format("NativeBlockInputStream schema mismatch, expected {}, actual {}.", expected, actual),
+            ErrorCodes::LOGICAL_ERROR);
+}
+
+void checkDataTypeName(size_t column_index, const String & expected, const String & actual)
+{
+    if (expected != actual)
```
SeaRise, line 38:
```suggestion
    if (unlikely(expected != actual))
```

dbms/src/Flash/Coprocessor/CodecUtils.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace ErrorCodes
+{
+extern const int LOGICAL_ERROR;
+} // namespace ErrorCodes
+
+namespace CodecUtils
+{
+void checkColumnSize(size_t expected, size_t actual)
+{
+    if (expected != actual)
```
SeaRise, line 30:
```suggestion
    if (unlikely(expected != actual))
```

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
yibin87: Confirmed with windtalker locally, we agreed to use the new one, since the missing potential error can be ignore.

SeaRise (commented):
others LGTM

dbms/src/Flash/Coprocessor/IChunkDecodeAndSquash.cpp
SeaRise: Both are ok for me

dbms/src/Flash/Coprocessor/CodecUtils.cpp
yibin87: Done.

SeaRise (approved):
LGTM

windtalker (approved):
LGTM

Issue #5816, opened 2022-09-08
Optimize merge for empty segment


breezewish:
## Enhancement

It is possible to produce empty segments, when we logical split the range during ingestion.

Currently, merging empty segment with a non-empty segment requires rewriting the stable layer. However, when we are sure that the segment is empty, it can be just merged without any cost.

breezewish:
Maybe we can first start from a simple case: `snapshot->getRows() == 0`.

PR #6207, opened 2022-10-28
storage: Introduce isDefinitelyEmpty API

breezewish:

Issue Number: ref #5237

Problem Summary:

- Added a new API: `Segment::isDefinitelyEmpty`.

  It returns true when the snapshot is definitely empty.
  It returns false when the snapshot is not definitely empty, e.g. not empty, or may be empty in some rare cases.

  This API is used to determine whether it is safe to perform replaceData for the segment when we trying to use replaceData for the ingest SST process. It could be also helpful for https://github.com/pingcap/tiflash/issues/5816

  This API does not respond an accurate isEmpty result in certain cases, in order to minimize the cost.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lidezhu (approved):
LGTM

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -373,6 +375,77 @@ bool Segment::write(DMContext & dm_context, const RowKeyRange & delete_range)
...
+    auto columns_to_read = std::make_shared<ColumnDefines>();
+    columns_to_read->push_back(getExtraHandleColumnDefine(is_common_handle));
+    auto read_ranges = RowKeyRanges{rowkey_range};
+
+    SkippableBlockInputStreams streams;
+    for (const auto & file : segment_snap->stable->getDMFiles())
+    {
+        DMFileBlockInputStreamBuilder builder(dm_context.db_context);
+        auto stream = builder
+                          .setRowsThreshold(std::numeric_limits<UInt64>::max()) // TODO: May be we could have some better settings
```
JaySon-Huang, line 401: I think we don't need this line as we only try to read one row from the stream

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -373,6 +375,77 @@ bool Segment::write(DMContext & dm_context, const RowKeyRange & delete_range)
...
+        ProfileEvents::increment(ProfileEvents::DMSegmentIsEmptyFastPath);
+        return true;
+    }
+
+    ProfileEvents::increment(ProfileEvents::DMSegmentIsEmptySlowPath);
+
+    // Build read stream
+    auto columns_to_read = std::make_shared<ColumnDefines>();
+    columns_to_read->push_back(getExtraHandleColumnDefine(is_common_handle));
+    auto read_ranges = RowKeyRanges{rowkey_range};
```
JaySon-Huang, line 394: what about using `getInputStreamModeFast`?

dbms/src/Storages/DeltaMerge/Segment.cpp
breezewish: We may also read multiple rows in some rare cases, for example, the whole block may be filtered out by SegmentRange.

dbms/src/Storages/DeltaMerge/Segment.cpp
breezewish: I'm not sure. My intuition was to not rely on fast mode, as fast mode may change behavior at any time. Currently fast mode will additionally do filter for del_mark, while we don't need it as well.

JaySon-Huang: OK

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -373,6 +375,77 @@ bool Segment::write(DMContext & dm_context, const RowKeyRange & delete_range)
...
+    {
+        DMFileBlockInputStreamBuilder builder(dm_context.db_context);
+        auto stream = builder
+                          .setRowsThreshold(std::numeric_limits<UInt64>::max()) // TODO: May be we could have some better settings
+                          .onlyReadOnePackEveryTime()
+                          .build(file, *columns_to_read, read_ranges);
+        streams.push_back(stream);
+    }
+
+    BlockInputStreamPtr stable_stream = std::make_shared<ConcatSkippableBlockInputStream>(streams);
```
JaySon-Huang, line 407: We can only build the stable_stream after there is no data read from `delta_stream`. By doing this we can save some unnecessary IO for min-max index.
But now it is also fine.

JaySon-Huang (approved):
LGTM

dbms/src/Storages/DeltaMerge/Segment.cpp
breezewish: Good catch!

Issue #5985, opened 2022-09-22
cmake --install is not working for debug profile


breezewish:
## Enhancement

Currently cmake always cp tiflash-proxy from `target/release`, which does not work in debug profile.

PR #6226, opened 2022-11-01
cmake: Use gnu objcopy by default

breezewish:

Issue Number: ref #5985

Problem Summary:

In https://github.com/pingcap/tiflash/issues/5985 we changed to store compressed DWARF in the new standard gABI format. However binutils in CentOS 7 does not recognize this format, so this PR brings back the old format.

The support for the old format has been removed in LLVM 15, so we just use the GNU objcopy to do this job.

Note: Go 1.19 changed to use the new standard gABI format as well: https://github.com/golang/go/commit/d151134851554aa0a3f05206019b9b8c1e1ad70f

More references: https://maskray.me/blog/2022-01-23-compressed-debug-sections

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

  ```shell
  objdump -x ./tiflash  # Works fine
  ```

  In CI we have the GNU objdump: https://ci.pingcap.net/blue/organizations/jenkins/tiflash-build-common/detail/tiflash-build-common/14951/pipeline/163

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):

Lloyd-Pottiger (approved):
maybe we should abandon centos7……

breezewish:
> maybe we should abandon centos7……

Unfortunately CentOS 7 has not EOL yet.. But soon!

breezewish:
Update:

I have verified that, when using the gABI format, binutils in CentOS 7 doesn't work (e.g. objdump), but the Linux perf still works. So there are actually fewer blocking stones for us to deprecate the legacy format.

PR #6236, opened 2022-11-02
fix potential failure when allocating space from BlobFile

### LINKED ISSUE #6235: The `hint_max_capacity` of `SpaceMap` may be larger than the actual `max_capacity` of it
## Bug Report

When adapt `PageStorage` to `RaftEngine`, there may some error like the following:
```
[2022/11/02 14:06:11.986 +08:00] [ERROR] [SpaceMapSTDMap.h:246] ["SpaceMap:Not sure why can't found any place to insert.[size=50645] [old biggest_offset=252703622] [old biggest_cap=53284] [new biggest_offset=249354265] [new biggest_cap=50507]"] [thread_id=86]
[2022/11/02 14:06:12.047 +08:00] [ERROR] [Exception.cpp:89] ["void DB::ConsumeWriteBatch(const DB::EngineStoreServerWrap *, DB::RawVoidPtr):Code: 49, e.displayText() = DB::Exception: Get postion from BlobStat failed, it may caused by `sm_max_caps` is no correct. [size=50645] [old_max_caps=53284] [max_caps=50507] [blob_id=30], e.what() = DB::Exception, Stack trace:\n\n\n       0x1ce274e\tDB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) [tiflash+30287694]\n                \tdbms/src/Common/Exception.h:46\n       0x1c8cb3a\tDB::PS::V3::BlobStore<DB::PS::V3::universal::BlobStoreTrait>::getPosFromStats(unsigned long) [tiflash+29936442]\n                \tdbms/src/Storages/Page/V3/BlobStore.cpp:497\n       0x1c8df1f\tDB::PS::V3::BlobStore<DB::PS::V3::universal::BlobStoreTrait>::write(DB::UniversalWriteBatch&, std::__1::shared_ptr<DB::WriteLimiter> const&) [tiflash+29941535]\n                \tdbms/src/Storages/Page/V3/BlobStore.cpp:310\n       0x6f548e9\tDB::UniversalPageStorage::write(DB::UniversalWriteBatch&&, std::__1::shared_ptr<DB::WriteLimiter> const&) const [tiflash+116738281]\n                \tdbms/src/Storages/Page/universal/UniversalPageStorage.cpp:56\n       0x6de0387\tConsumeWriteBatch [tiflash+115213191]\n                \tdbms/src/Storages/Transaction/ProxyFFI.cpp:218\n  0x7f8972ce83e0\traftstore::store::async_io::write::Worker$LT$EK$C$ER$C$N$C$T$GT$::write_to_db::ha47eb534ff999cf7 [libtiflash_proxy.so+28382176]\n  0x7f89726ba23b\tbatch_system::batch::Poller$LT$N$C$C$C$Handler$GT$::poll::h60db684e842f1236 [libtiflash_proxy.so+21901883]\n  0x7f8972f21f8e\tstd::sys_common::backtrace::__rust_begin_short_backtrace::h5a9ee7c88b3ca032 [libtiflash_proxy.so+30715790]\n  0x7f89725b73f7\tcore::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::hdee45bcde14b967f [libtiflash_proxy.so+20841463]\n  0x7f897368c5ca\tstd::sys::unix::thread::Thread::new::thread_start::hd39c5f08bdcda277 [libtiflash_proxy.so+38491594]\n  0x7f8970fc2ea5\tstart_thread [libpthread.so.0+32421]\n                \t/usr/src/debug/glibc-2.17-c758a686/nptl/pthread_create.c:307\n  0x7f89703c6b0d\t<unknown symbol> [libc.so.6+1043213]\n                \t/usr/src/debug////////glibc-2.17-c758a686/misc/../sysdeps/unix/syscall-template.S:83"] [thread_id=86]
```

lidezhu:

Issue Number: close #6235 

Problem Summary:
When free space in `SpaceMapSTDMap` of `PageStorage`, there are two steps:
1. mark the space free;
2. update `hint_biggest_cap` and `hint_biggest_offset`;

And these two steps is processed by two functions and is not protected by lock;

So there is a time gap in which `hint_biggest_cap` is smaller than the accurate `biggest_cap`.
But this is ok because it is just a little pessimistic and won't cause any failure for other activities.

However, `searchInsertOffset` assumes an accurate `hint_biggest_offset`. So if `searchInsertOffset` happens between the above two steps, it will return a wrong `hint_biggest_cap` and cause failure for later allocation.

For example, 
1. Assume the current `hint_biggest_offset` is 50, and `hint_biggest_cap` is also 50. So the current biggest free space is [50, 100);
2. Mark [25, 50) as free, and now the biggest free space is [25, 100). But the hint biggest free space is not updated and is still [50, 100);
4. There is a request to allocate a space of size 50, so [25, 75) is allocated, and it finds that 25 is not the current `hint_biggest_offset`, so it doesn't update `hint_biggest_cap` which is still 50, but the actual biggest capacity is 25 now;
5. There is another request to allocate a space of size 30. The spacemap thought it can satisfy the request because its `hint_biggest_cap` is 50 but fail to do that after an entire traverse of the free space.

When allocate space in `searchInsertOffset`, check whether the allocated free space range contains the range `[hint_biggest_offset, hint_biggest_offset + hint_biggest_cap)`, instead of just check whether the offset of the allocated space is the same as `hint_biggest_offset`.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/spacemap/SpaceMapSTDMap.h
```diff
@@ -252,11 +252,19 @@ class STDMapSpaceMap
 
         // Update return start
         offset = it->first;
+        bool is_champion = it->first <= hint_biggest_offset && hint_biggest_offset < it->first + it->second;
```
JaySon-Huang, line 255:
```suggestion
        // If the chosen free block contains `hint_biggest_offset`, then the free block must be the largest free block
        // in this spacemap
        bool is_champion = it->first <= hint_biggest_offset && hint_biggest_offset < it->first + it->second;
```

JaySon-Huang (approved):
lgtm

breezewish:
Seems that allowing this race condition may cause some tricky cases. Even if we fixed the calculation of champion in this case, the "champion" judgement may be still wrong in some other cases. I'm not sure whether the violation of specific characteristic will cause further problems.

Consider the case:

1. real_max_space = hint_max_space = [50, 100)
2. Free [1000, 2000)
3. Now, real_max_space = [1000, 2000), while hint_max_space = [50, 100)
4. Allocate 40, and we discovered [0, 40)
5. In this case, `is_champion == false`, and we are returning `max_cap = 50` (instead of `max_cap = 1000`) <-- This may be a pitfall. Will it cause further issues?

JaySon-Huang:
> Seems that allowing this race condition may cause some tricky cases.

I think it is ok.
The returning `max_cap` will update `blobstat.sm_max_caps`. If it is smaller than the actual, then we just choose another  blobfile to save the blob data.
https://github.com/pingcap/tiflash/blob/5d53d2d5814fc7a65ffc98629da1dcc105abbb03/dbms/src/Storages/Page/V3/Blob/BlobStat.cpp#L292-L299

And this incorrect is likely caused by the time gap between line 416 to line 437 in `BlobStore::remove`. `stat->recalculateCapacity()` can fix the `blobstat.sm_max_caps` to correct value.
https://github.com/pingcap/tiflash/blob/5d53d2d5814fc7a65ffc98629da1dcc105abbb03/dbms/src/Storages/Page/V3/BlobStore.cpp#L402-L438

breezewish (approved):

Issue #5900, opened 2022-09-15
Improve the code of the TiFlash computing layer


SeaRise:
## Enhancement
1. Extract execution Summary logic from DAGResponseWriter
2. separate DAGUtils
3. Make codec related files into one folder, more tests of codec module.
4. Make TiDB type related files into one folder
5. Add one abstract class for scheduler.
6. More comments of the core classes to express the duty of them.
7. remove DAGQueryBlock and related codes
8. remove sync tunnel and related codes

PR #6250, opened 2022-11-04
.*: Refine filterBlockInputStream

SeaRise:

Issue Number: ref #5900 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/DataStreams/FilterTransformAction.cpp
+++ b/dbms/src/DataStreams/FilterTransformAction.cpp
@@ -31,8 +31,7 @@ extern const int LOGICAL_ERROR;
 FilterTransformAction::FilterTransformAction(
     const Block & header_,
     const ExpressionActionsPtr & expression_,
-    const String & filter_column_name,
-    const String & req_id)
+    const String & filter_column_name)
     : header(header_)
     , expression(expression_)
 {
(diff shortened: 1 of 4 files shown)


dbms/src/DataStreams/FilterTransformAction.cpp
```diff
@@ -0,0 +1,181 @@
...
+      * To do this, we filter out the first non-constant column
+      *  or calculate number of set bytes in the filter.
+      */
+    size_t first_non_constant_column = 0;
+    for (size_t i = 0; i < columns; ++i)
+    {
+        if (!block.safeGetByPosition(i).column->isColumnConst())
+        {
+            first_non_constant_column = i;
+
```
ywqzzy, line 122: Why not add a break？

SeaRise: Are you talking about the breaker on L214?

dbms/src/DataStreams/FilterTransformAction.h
```diff
@@ -0,0 +1,43 @@
...
+    bool alwaysFalse() const;
+    // return false if all filter out.
+    bool transform(Block & block);
+    Block getHeader() const;
+    ExpressionActionsPtr getExperssion() const;
+
+private:
+    Block header;
+    ExpressionActionsPtr expression;
+    ssize_t filter_column;
```
ywqzzy, line 39:
```suggestion
    size_t filter_column;
```
Maybe we can change it to size_t, then the casts in FilterTransformAction.cpp will be removed.

ywqzzy (approved):
Rest LGTM

dbms/src/DataStreams/FilterTransformAction.h
SeaRise: good point!

SeaRise: updated

dbms/src/DataStreams/FilterTransformAction.cpp
```diff
@@ -0,0 +1,181 @@
...
+        if (i == filter_column)
+        {
+            /// The column with filter itself is replaced with a column with a constant `1`, since after filtering, nothing else will remain.
+            /// NOTE User could pass column with something different than 0 and 1 for filter.
+            /// Example:
+            ///  SELECT materialize(100) AS x WHERE x
+            /// will work incorrectly.
+            current_column.column = current_column.type->createColumnConst(filtered_rows, UInt64(1));
+            continue;
+        }
```
Lloyd-Pottiger, line 168: Why always replace the filter column with a constant `1`? What if `select a from t where a > 0`?

SeaRise: Because after filter, the value of filter_column must be `1`

SeaRise: if the data type of `a` isn't bool, the filter column will be `tidb_cast(a, bool)`.

Lloyd-Pottiger: so, for `select a from t where a > 0`, there are two columns in a block, `a` and `tidb_cast(a, bool)`?

SeaRise: yes, the related code is https://github.com/pingcap/tiflash/blob/9d7d060e67637e5c1f8854f6ba5e7b89d8570eae/dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp#L680-L709

Lloyd-Pottiger: get it, thanks

Lloyd-Pottiger (approved):

Issue #6115, opened 2022-10-12
Support push down of regexp_xxx functions


xzhangxian1008:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

Functions need supported:
- [x] regexp_like
- [x] regexp_instr
- [x] regexp_substr
- [ ] regexp_replace

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #6272, opened 2022-11-07
Support regexp_instr function

xzhangxian1008:

Issue Number: ref #6115 

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -111,18 +115,102 @@ inline constexpr bool check_int_type()
...
+{
+    const auto * tmp = reinterpret_cast<const typename ColumnVector<Int32>::Container *>(container);
+    return static_cast<Int64>((*tmp)[idx]);
+}
+
+Int64 getInt64(const void * container, size_t idx)
+{
+    const auto * tmp = reinterpret_cast<const typename ColumnVector<Int64>::Container *>(container);
+    return static_cast<Int64>((*tmp)[idx]);
+}
```
gengliqi, line 184: How about using a template function?

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1114,373 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+                    {
+                        null_map[i] = 1;
+                        continue;
+                    }
+                    null_map[i] = 0;
+                    expr_param.getStringRef(i, expr_ref);
+                    GET_POS_VALUE(i)
+                    GET_OCCUR_VALUE(i)
+                    GET_RET_OP_VALUE(i)
+                    vec_res[i] = regexp->instr(expr_ref.data, expr_ref.size, pos, occur, ret_op);
```
windtalker, line 1329: Why don't check if `regexp` is nullptr?

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -469,5 +472,109 @@ unsigned OptimizedRegularExpressionImpl<thread_safe>::match(const char * subject
...
+}
+
+template <bool thread_safe>
+Int64 OptimizedRegularExpressionImpl<thread_safe>::processEmptyStringExpr(const char * expr, size_t expr_size, size_t pos, Int64 occur)
+{
+    if (occur != 1)
+        return 0;
+
+    StringPieceType expr_sp(expr, expr_size);
+    bool success = RegexType::FindAndConsume(&expr_sp, *re2);
```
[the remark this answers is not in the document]
xzhangxian1008: > Why need to call `FindAndConsume` even if `expr_size == 0`?

Because `^$` could match empty string.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -111,18 +115,55 @@ inline constexpr bool check_int_type()
...
-    Int8,
-    Int16,
-    Int32,
-    Int64
-};
+    switch (field.getType()) {
+    case Field::Types::Which::Int64:
+        return field.safeGet<Int64>();
+    case Field::Types::Which::UInt64:
+        return field.safeGet<UInt64>();
```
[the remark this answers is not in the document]
xzhangxian1008: > This will implicitly cast to Int64, is it the expected behavior?

Yes, it's expected behavior.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1114,373 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+
+        // Const value will not be used when param is not const
+        Int64 pos_const_val = PosT::isConst() ? pos_param.template getInt<Int64>(0) : -1;
+        Int64 occur_const_val = OccurT::isConst() ? occur_param.template getInt<Int64>(0) : -1;
+        Int64 ret_op_const_val = RetOpT::isConst() ? ret_op_param.template getInt<Int64>(0) : -1;
+
+        // Check if args are all const columns
+        if constexpr (ExprT::isConst() && PatT::isConst() && PosT::isConst() && OccurT::isConst() && RetOpT::isConst() && MatchTypeT::isConst())
+        {
+            if (col_size == 0 || expr_param.isNullAt(0) || pat_param.isNullAt(0) || pos_param.isNullAt(0) || occur_param.isNullAt(0) || ret_op_param.isNullAt(0) || match_type_param.isNullAt(0))
```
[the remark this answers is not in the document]
xzhangxian1008: > I think we should not check col_size here, for all constant case, col_size will be 1 at least.

okk

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1114,373 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+
+        constexpr bool has_nullable_col = ExprT::isNullableCol() || PatT::isNullableCol() || PosT::isNullableCol() || OccurT::isNullableCol() || RetOpT::isNullableCol() || MatchTypeT::isNullableCol();
+
+#define GET_POS_VALUE(idx) \
+    do \
+    { \
+        if constexpr (PosT::isConst()) \
+            pos = pos_const_val; \
+        else \
+            pos = get_pos_func(pos_container, idx); \
```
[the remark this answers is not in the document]
xzhangxian1008: > Why not hidden `isConst` inside `get_pos_func`? So we don't need these macros

Because compiler can't inline the function as we use function pointer and we will always get a const value by calling function if we hide `isCOnst` inside `get_pos_func`.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1114,373 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+        Int64 occur;
+        Int64 ret_op;
+        String match_type;
+
+        // Start to execute instr
+        if (canMemorize<PatT, MatchTypeT>())
+        {
+            // Codes in this if branch execute instr with memorized regexp
+
+            const auto & regexp = memorize(pat_param, match_type_param, collator);
```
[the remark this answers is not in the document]
xzhangxian1008: > According to the implementation of `regexp_like`, if `col_size == 0` we should not memorize or it may throw error if pattern is empty.

okk

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > Why don't check if `regexp` is nullptr?

we can ensure regexp is not nullptr here

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1126,380 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+                    null_map[i] = 0;
+                    expr_param.getStringRef(i, expr_ref);
+                    pat = pat_param.getString(i);
+                    if (unlikely(pat.empty()))
+                        throw Exception(EMPTY_PAT_ERR_MSG);
+                    GET_POS_VALUE(i)
+                    GET_OCCUR_VALUE(i)
+                    GET_RET_OP_VALUE(i)
+                    match_type = match_type_param.getString(i);
+                    auto regexp = createRegexpWithMatchType(pat, match_type, collator);
```
windtalker, line 1395: Seems we can not put the regexp into to pool if regexp can not be memoryized, otherwise server may get oom since the ObjectPool actually has no evict mechanism, the regexp put into the pool will be hold forever.

xzhangxian1008: > Seems we can not put the regexp into to pool if regexp can not be memoryized, otherwise server may get oom since the ObjectPool actually has no evict mechanism, the regexp put into the pool will be hold forever.

Done

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -90,12 +93,13 @@ inline int getDefaultFlags()
...
 }
 
+// add '()' outside of the pattern to get the matched substr
 inline String addMatchTypeForPattern(const String & pattern, const String & match_type, TiDB::TiDBCollatorPtr collator)
 {
     String mode = re2Util::getRE2ModeModifiers(match_type, collator);
     if (mode.empty())
-        return pattern;
-    return fmt::format("{}{}", mode, pattern);
+        return fmt::format("({})",pattern);
```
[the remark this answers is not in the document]
xzhangxian1008: > I think just add `()` if we need to get the matched substr is enough, for functions like `regexp_like` we don't need to get the match substr.

Done

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -90,18 +93,29 @@ inline int getDefaultFlags()
     return flags;
 }
 
+template <bool need_subpattern>
```
windtalker, line 96: How about we let the caller to make sure the input pattern is already a subpattern so we don't need to add this template argument for all related functions?

xzhangxian1008: > How about we let the caller to make sure the input pattern is already a subpattern so we don't need to add this template argument for all related functions?

okk

windtalker (commented):
Also need to add some fullstack tests. And since regexp related function have different behavior if the input is empty or not, we need add test of empty input explicitly.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -111,19 +115,79 @@ inline constexpr bool check_int_type()
...
+        return &getInt<Int32>;
+    case IntType::Int64:
+        return &getInt<Int64>;
+    default:
+        throw Exception("Unexpected int type");
+    }
+}
+
+template <typename ResultType>
+void setResultColumnValuesNull(ColumnWithTypeAndName & res_arg, size_t col_size)
```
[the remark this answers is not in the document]
xzhangxian1008: > Where is this function used?

I will delete it.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1150,383 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+        GetIntFuncPointerType get_occur_func = getGetIntFuncPointer(occur_param.getIntType());
+        GetIntFuncPointerType get_ret_op_func = getGetIntFuncPointer(ret_op_param.getIntType());
+
+        // Container will not be used when parm is const
+        const void * pos_container = pos_param.getContainer();
+        const void * occur_container = occur_param.getContainer();
+        const void * ret_op_container = ret_op_param.getContainer();
+
+        // Const value will not be used when param is not const
+        Int64 pos_const_val = PosT::isConst() ? pos_param.template getInt<Int64>(0) : -1;
```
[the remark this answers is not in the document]
xzhangxian1008: > What happens if `col_size == 0`? Why not just use `Int64 pos_const_val = -1` since we already make sure that `pos_const_val` is only used when `PosT::isConst()` is true



> Also need to add some fullstack tests. And since regexp related function have different behavior if the input is empty or not, we need add test of empty input explicitly.

okk

xzhangxian1008:
> Also need to add some fullstack tests. And since regexp related function have different behavior if the input is empty or not, we need add test of empty input explicitly.

okk

dbms/src/Functions/FunctionsRegexp.h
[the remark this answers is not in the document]
xzhangxian1008: > What happens if `col_size == 0`? Why not just use `Int64 pos_const_val = -1` since we already make sure that `pos_const_val` is only used when `PosT::isConst()` is true

nothing will happen when `col_size == 0` as the index parameter passed into `getInt<Int64>` is useless when this is a const value. You can see the implementation of `getInt` in `ParamInt`.

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -973,15 +1136,383 @@ class FunctionStringRegexp : public FunctionStringRegexpBase
...
+        ColumnPtr col_occur;
+        ColumnPtr col_return_option;
+        ColumnPtr col_match_type;
+
+        // Go through cases to get arguments
+        switch (arg_num)
+        {
+        case REGEXP_INSTR_MAX_PARAM_NUM:
+            col_match_type = block.getByPosition(arguments[5]).column;
+        case REGEXP_MIN_PARAM_NUM + 3:
```
windtalker, line 1467: I think it's more understandable by using `case REGEXP_INSTR_MAX_PARAM_NUM - XXX` instead of `case REGEXP_MIN_PARAM_NUM + XXX`

tests/fullstack-test/expr/regexp.test
```diff
@@ -94,3 +94,9 @@ mysql> set tidb_enforce_mpp=1; select regexp_like(data, pattern, match_type) as
 |    1 |
 |    1 |
 +------+
+
+mysql> drop table if exists test.t;
+mysql> create table test.t (data varchar(30), pattern varchar(30), pos int, occur int, ret_op int, match_type varchar(30));
+mysql> alter table test.t set tiflash replica 1;
+func> wait_table test t
+mysql> set tidb_enforce_mpp=1; select regexp_instr("1", "1", pos, occur, 1, match_type) as res from test.t;
```
windtalker, line 102: For empty input, should test the case that no error throw even if `pattern` is empty string.

windtalker: And also add a test that with non-empty input.

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > I think it's more understandable by using `case REGEXP_INSTR_MAX_PARAM_NUM - XXX` instead of `case REGEXP_MIN_PARAM_NUM + XXX`

okk

tests/fullstack-test/expr/regexp.test
xzhangxian1008: > 
> For empty input, should test the case that no error throw even if `pattern` is empty string.

Done

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: > I think it's more understandable by using `case REGEXP_INSTR_MAX_PARAM_NUM - XXX` instead of `case REGEXP_MIN_PARAM_NUM + XXX`

Done

tests/fullstack-test/expr/regexp.test
```diff
@@ -94,3 +94,24 @@ mysql> set tidb_enforce_mpp=1; select regexp_like(data, pattern, match_type) as
...
+mysql> set tidb_enforce_mpp=1; select regexp_instr("1", "", pos, occur, 1, match_type) as res from test.t;
+
+mysql> drop table if exists test.t;
```
windtalker, line 105: Why need to drop the table and recreate a new one?

windtalker (approved):
Others LGTM

tests/fullstack-test/expr/regexp.test
xzhangxian1008: > Why need to drop the table and recreate a new one?

I will delete it.

dbms/src/Common/UTF8Helpers.h
```diff
@@ -83,6 +83,30 @@ inline size_t countCodePoints(const UInt8 * data, size_t size)
...
+//   utf8 position of character 'a' in this string is 4 and byte position is 6.
+static inline Int64 utf8Pos2bytePos(const UInt8 * str, Int64 utf8_pos)
+{
+    Int64 byte_index = 0;
+    utf8_pos--;
+    while (utf8_pos > 0)
+    {
+        byte_index += seqLength(str[byte_index]);
+        utf8_pos--;
+    }
```
SeaRise, line 98:
```suggestion
    while (--utf8_pos > 0)
        byte_index += seqLength(str[byte_index]);
```

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -111,19 +116,65 @@ inline constexpr bool check_int_type()
     return std::is_same_v<T, UInt8> || std::is_same_v<T, UInt16> || std::is_same_v<T, UInt32> || std::is_same_v<T, UInt64> || std::is_same_v<T, Int8> || std::is_same_v<T, Int16> || std::is_same_v<T, Int32> || std::is_same_v<T, Int64>;
 }
 
+Int64 getIntFromField(Field & field)
+{
+    switch (field.getType())
+    {
+    case Field::Types::Which::Int64:
```
SeaRise, line 123:
```suggestion
    case Field::Types::Int64:
```

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -111,19 +116,65 @@ inline constexpr bool check_int_type()
...
+        return field.safeGet<Int64>();
+    case Field::Types::Which::UInt64:
```
SeaRise, line 125:
```suggestion
    case Field::Types::UInt64:
```

SeaRise: like https://github.com/pingcap/tiflash/blob/2beef486de9786c1a39c1600441cdb475700de7f/dbms/src/Core/Field.cpp#L43

dbms/src/Functions/FunctionsRegexp.h
```diff
@@ -135,46 +186,68 @@ class ParamString
...
-        : const_string(str_ref)
+        : const_string_data(nullptr)
+        , const_string_data_size(0)
         , chars(nullptr)
         , offsets(nullptr)
     {
+        // Deep copy
+        const_string_data = new char[str_ref.size];
+        if (const_string_data == nullptr)
+            throw Exception("ParamString constructor get a nullptr");
```
SeaRise, line 206:
```suggestion
```
Seems useless?

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -469,5 +473,71 @@ unsigned OptimizedRegularExpressionImpl<thread_safe>::match(const char * subject
...
+    return pos;
+}
+
+static inline void checkArgs(Int64 utf8_total_len, size_t subject_size, Int64 pos, Int64 ret_op)
+{
+    if (unlikely(ret_op != 0 && ret_op != 1))
+        throw DB::Exception("Incorrect argument to regexp function: return_option must be 1 or 0");
+
+    if (unlikely(pos <= 0 || (pos > utf8_total_len && subject_size != 0)))
+        throw DB::Exception("Index out of bounds in regular function.");
```
SeaRise, line 495: use `RUMTIME_CHECK`?

dbms/src/Functions/FunctionsRegexp.h
xzhangxian1008: I will delete it

dbms/src/Common/UTF8Helpers.h
xzhangxian1008: okk

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -469,5 +473,68 @@ unsigned OptimizedRegularExpressionImpl<thread_safe>::match(const char * subject
...
+Int64 OptimizedRegularExpressionImpl<thread_safe>::processEmptyStringExpr(const char * expr, size_t expr_size, size_t pos, Int64 occur)
+{
+    if (occur != 1)
+        return 0;
+
+    StringPieceType expr_sp(expr, expr_size);
+    bool success = RegexType::FindAndConsume(&expr_sp, *re2);
+    if (!success)
+        return 0;
+    return pos;
```
SeaRise, line 486:
```suggestion
    return RegexType::FindAndConsume(&expr_sp, *re2) ? pos : 0;
```

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -469,5 +473,68 @@ unsigned OptimizedRegularExpressionImpl<thread_safe>::match(const char * subject
...
+    size_t expr_size = subject_size - byte_offset;
+
+    StringPieceType expr_sp(expr, expr_size);
+    StringPieceType matched_str;
+
+    while (occur > 0)
+    {
+        bool success = RegexType::FindAndConsume(&expr_sp, *re2, &matched_str);
+        if (!success)
+            return 0;
```
SeaRise, line 514:
```suggestion
        if (!RegexType::FindAndConsume(&expr_sp, *re2, &matched_str))
            return 0;
```

dbms/src/Common/UTF8Helpers.h
SeaRise: It still looks like the original code?

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: okk

dbms/src/Common/UTF8Helpers.h
xzhangxian1008: resolved

SeaRise (approved):
/run-all-tests

SeaRise:
> /run-all-tests

ERROR 1139 (42000) at line 1: Got error 'Not support binary collation so far' from regexp

xzhangxian1008:
> > /run-all-tests
> 
> ERROR 1139 (42000) at line 1: Got error 'Not support binary collation so far' from regexp

resolved

PR #6346, opened 2022-11-22
feat: auto reload ssl cert

### LINKED ISSUE #5503: Support certificates reloading

ywqzzy:

Issue Number: close #5503

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Context.cpp
```diff
@@ -569,6 +571,19 @@ ConfigurationPtr Context::getUsersConfig()
...
 }
 
+void Context::setSecurityConfig(Poco::Util::AbstractConfiguration & config, const LoggerPtr & log)
+{
+    LOG_INFO(log, "Setting secuirty config.");
+    auto lock = getLock();
+    shared->security_config = TiFlashSecurityConfig(config, log);
+}
+
+TiFlashSecurityConfig & Context::getSecurityConfig()
```
zanmato1984, line 581: Will the following case happen?
1. Some thread gets the reference to `TiFlashSecurityConfig` from the return value of this function;
2. The above thread start using this config, WITHOUT the protection of the lock;
3. Some other thread sets the security config thus make the original one dangling.

dbms/src/Interpreters/Context.cpp
```diff
@@ -569,6 +571,19 @@ ConfigurationPtr Context::getUsersConfig()
...
+void Context::setSecurityConfig(Poco::Util::AbstractConfiguration & config, const LoggerPtr & log)
+{
+    LOG_INFO(log, "Setting secuirty config.");
+    auto lock = getLock();
+    shared->security_config = std::make_shared<TiFlashSecurityConfig>(config, log);
+}
+
+TiFlashSecurityConfigPtr Context::getSecurityConfig()
+{
+    auto lock = getLock();
```
ywqzzy, line 583: Maybe the lock is useless?

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -167,6 +199,20 @@ struct TiFlashSecurityConfig
...
+    String cert_path;
+    String key_path;
+    LoggerPtr log;
+
+    bool redact_info_log = false;
+
+    std::set<String> allowed_common_names;
+
+    bool has_tls_config = false;
+    grpc::SslCredentialsOptions options;
```
ywqzzy, line 214: I will remove it

dbms/src/Server/CMakeLists.txt
```diff
@@ -32,10 +32,11 @@ add_library(server_for_test
 )
 
 target_link_libraries (server_for_test PUBLIC clickhouse_common_io clickhouse_storages_system)
+target_include_directories(server_for_test PRIVATE ${TiFlash_SOURCE_DIR}/contrib/grpc/)
```
ywqzzy, line 35: Is there any problem when building tiflash using grpc built in local machine?

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -167,6 +199,19 @@ struct TiFlashSecurityConfig
...
+    String cert_path;
+    String key_path;
+    LoggerPtr log;
+
+    bool redact_info_log = false;
+
+    std::set<String> allowed_common_names;
+
+    bool has_tls_config = false;
+    mutable std::mutex mu;
```
zanmato1984, line 214: Is this mutex necessary? Seems no member write except in the `init` method which is only called in ctor.

dbms/src/Interpreters/Context.cpp
zanmato1984: No. You still need that lock to prevent read/write race of the `shared_ptr`.

dbms/src/Common/TiFlashSecurity.h
ywqzzy: > Is this mutex necessary? Seems no member write except in the `init` method which is only called in ctor.

Yes, when reloading config `init` will be called too.

dbms/src/Server/Server.cpp
```diff
@@ -526,7 +529,7 @@ class Server::TcpHttpServersHolder
         , server_pool(1, server.config().getUInt("max_connections", 1024))
     {
         auto & config = server.config();
-        auto & security_config = server.security_config;
+        auto security_config = server.global_context->getSecurityConfig();
```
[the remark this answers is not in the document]
ywqzzy: global_context will destructed in `scope_exit`. And We can make sure that when construct TcpHttpServersHolder, global_context will not destructed.

dbms/src/Server/Server.cpp
```diff
@@ -1095,11 +1091,16 @@ int Server::main(const std::vector<std::string> & /*args*/)
     auto main_config_reloader = std::make_unique<ConfigReloader>(
         config_path,
         [&](ConfigurationPtr config) {
+            LOG_INFO(log, "run main config reloader");
```
ywqzzy, line 1094: main_config_reloader will check whether `config.toml` is changed periodically, if `config.toml` is modified, it will call  the lambda function set below.

dbms/src/Common/TiFlashSecurity.h
zanmato1984: Saw it. Thanks~

dbms/src/Server/CertificateReloader.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+} // namespace
+
+/// This is callback for OpenSSL. It will be called on every connection to obtain a certificate and private key.
+int CertificateReloader::setCertificate(SSL * ssl, void * arg)
+{
+    auto * context = static_cast<Context *>(arg);
+    auto security_config = context->getSecurityConfig();
+    auto [cert_path, key_path] = security_config->getCertAndKeyPath();
+    LOG_INFO(log, "setCertificate callback called, cert_path: {}, key_path: {}", cert_path, key_path);
```
SeaRise, line 38: use `LOG_DEBUG`?
Could it be printing too many logs?

dbms/src/Interpreters/Context.cpp
zanmato1984: BTW, thread-safety of `shared_ptr` is in the fact that multiple threads can use multiple `shared_ptr`s pointing to the same object and don't have to worry about the race when multiple threads try to destruct the same object. Other than that, there is no thread-safety about `shared_ptr`, esp. for the case that one thread tries to assign a `shared_ptr` an object and the other is using that `shared_ptr`.

dbms/src/Interpreters/Context.h
```diff
@@ -99,6 +100,8 @@ class ReadLimiter;
 using ReadLimiterPtr = std::shared_ptr<ReadLimiter>;
 using MockMPPServerInfo = DB::tests::MockMPPServerInfo;
 using MockStorage = DB::tests::MockStorage;
+struct TiFlashSecurityConfig;
+using TiFlashSecurityConfigPtr = std::shared_ptr<TiFlashSecurityConfig>;
```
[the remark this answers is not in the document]
zanmato1984: I don't think `unique_ptr` is applicable. Seems there could be co-existing holders of this config: one is the `Context`, the other is `Reloader`. And the life-cycle of a security config is unclear: `Context` holds it and can update it. `Reloader` uses it temporarily. The chance for releasing a config is even. So I think `shared_ptr` here is a proper usage.

dbms/src/Server/CMakeLists.txt
zanmato1984: There is.

dbms/src/Server/FlashGrpcServerHolder.cpp
```diff
@@ -14,6 +14,7 @@
 #include <Flash/EstablishCall.h>
 #include <Server/FlashGrpcServerHolder.h>
 
+#include "../../contrib/grpc/src/cpp/server/secure_server_credentials.h"
```
zanmato1984, line 17: Please add some comment about this irregular include and the following subsequent grpc callback hijacking thing.

dbms/src/Server/CertificateReloader.cpp
ywqzzy: ssl connection build is rare. But I think LOG_DEBUG is good

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -121,9 +145,9 @@ struct TiFlashSecurityConfig
         }
     }
 
-
     bool checkGrpcContext(const grpc::ServerContext * grpc_context) const
     {
+        std::unique_lock lock(mu);
```
birdstorm, line 150: I think this mutex is unnecessary.

dbms/src/Server/Server.cpp
birdstorm: it should be debug level

dbms/src/Common/TiFlashSecurity.h
ywqzzy: It is necessary. When reloading config, other threads may call `checkGrpcContext`, there is a data race.

dbms/src/Server/CMakeLists.txt
zanmato1984: But that is not a big problem.

dbms/src/Server/CertificateReloader.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include <ext/singleton.h>
+
+namespace DB
+{
+
+class CertificateReloader : public ext::Singleton<CertificateReloader>
+{
+public:
+    static void initSSLCallback(Poco::Net::Context::Ptr context, Context * global_context);
```
zanmato1984, line 36: This class is singleton, plus all the callers are using `CertificateReloader::instance().initSSLCallback`. Why is it static?

dbms/src/Server/CertificateReloader.cpp
```diff
@@ -0,0 +1,61 @@
...
+int CertificateReloader::setCertificate(SSL * ssl, void * arg)
+{
+    auto * context = static_cast<Context *>(arg);
+    auto security_config = context->getSecurityConfig();
+    auto [cert_path, key_path] = security_config->getCertAndKeyPath();
+    LOG_DEBUG(log, "setCertificate callback called, cert_path: {}, key_path: {}", cert_path, key_path);
+    Poco::Crypto::X509Certificate cert(cert_path);
+    Poco::Crypto::EVPPKey key("", key_path);
+    SSL_use_certificate(ssl, const_cast<X509 *>(cert.certificate()));
+    SSL_use_PrivateKey(ssl, const_cast<EVP_PKEY *>(static_cast<const EVP_PKEY *>(key)));
```
zanmato1984, line 42: Shall we check the return values of these two calls?

dbms/src/Server/CertificateReloader.h
ywqzzy: Following the clang-tidy option: readability-convert-member-functions-to-static

dbms/src/Server/CertificateReloader.cpp
ywqzzy: > Shall we check the return values of these two calls?

Got, I will check the returen val.

ywqzzy: > > Shall we check the return values of these two calls?
> 
> Got, I will check the returen val.

And IMO, the check below can ensure the error handling safe, but will lose some debug info.

ywqzzy: > > Shall we check the return values of these two calls?
> 
> Got, I will check the returen val.

And currently when cert or key is invalid, the following log will be printed:
```
[2022/11/24 08:33:24.548 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = Cannot read file: Faild to load certificate from: /data1/zzywq/tiup_deploy/dev1/tiflash-6319/tls_old/tiflash.crt, e.what() = Cannot read file"] [source=ServerErrorHandler] [thread_id=58]
[2022/11/24 08:33:54.550 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = OpenSSLException: EVPKey::loadKey(string): error:06000066:public key routines:OPENSSL_internal:DECODE_ERROR: error:0900000c:PEM routines:OPENSSL_internal:ASN.1 encoding routines: error:00000000:invalid library (0):OPENSSL_internal:invalid library (0), e.what() = OpenSSLException"] [source=ServerErrorHandler] [thread_id=58]
```

ywqzzy: > > > Shall we check the return values of these two calls?
> > 
> > 
> > Got, I will check the returen val.
> 
> And currently when cert or key is invalid, the following log will be printed:
> 
> ```
> [2022/11/24 08:33:24.548 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = Cannot read file: Faild to load certificate from: /data1/zzywq/tiup_deploy/dev1/tiflash-6319/tls_old/tiflash.crt, e.what() = Cannot read file"] [source=ServerErrorHandler] [thread_id=58]
> [2022/11/24 08:33:54.550 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = OpenSSLException: EVPKey::loadKey(string): error:06000066:public key routines:OPENSSL_internal:DECODE_ERROR: error:0900000c:PEM routines:OPENSSL_internal:ASN.1 encoding routines: error:00000000:invalid library (0):OPENSSL_internal:invalid library (0), e.what() = OpenSSLException"] [source=ServerErrorHandler] [thread_id=58]
> ```

After I check the return val, the log is like:
```
[2022/11/24 08:39:24.553 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = OpenSSLException: EVPKey::loadKey(string): error:06000066:public key routines:OPENSSL_internal:DECODE_ERROR: error:0900000c:PEM routines:OPENSSL_internal:ASN.1 encoding routines: error:00000000:invalid library (0):OPENSSL_internal:invalid library (0), e.what() = OpenSSLException"] [source=ServerErrorHandler] [thread_id=57]
[2022/11/24 08:39:54.548 +08:00] [ERROR] [Exception.cpp:89] ["Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = Cannot read file: Faild to load certificate from: /data1/zzywq/tiup_deploy/dev1/tiflash-6319/tls_old/tiflash.crt, e.what() = Cannot read file"] [source=ServerErrorHandler] [thread_id=57]
```
No difference since ssl is not null and if the cert or key is invalid, Poco will raise exception.
But error checking is always necessary.

dbms/src/Server/CertificateReloader.h
zanmato1984: If so the call sites could use `CertificateReloader::initSSLCallback` instead of `CertificateReloader::instance().initSSLCallback`?

ywqzzy: > If so the call sites could use `CertificateReloader::initSSLCallback` instead of `CertificateReloader::instance().initSSLCallback`?

Yes, I will change the code style.

dbms/src/Common/TiFlashSecurity.h
windtalker: Maybe we can use a rw lock if we only re-init `TiFlashSecurityConfig` if it actually changes.

dbms/src/Server/Server.cpp
```diff
@@ -1095,11 +1091,16 @@ int Server::main(const std::vector<std::string> & /*args*/)
     auto main_config_reloader = std::make_unique<ConfigReloader>(
         config_path,
         [&](ConfigurationPtr config) {
+            LOG_DEBUG(log, "run main config reloader");
             buildLoggers(*config);
             global_context->setMacros(std::make_unique<Macros>(*config, "macros"));
             global_context->getTMTContext().reloadConfig(*config);
             global_context->getIORateLimiter().updateConfig(*config);
             global_context->reloadDeltaTreeConfig(*config);
+            global_context->getSecurityConfig()->init(*config);
```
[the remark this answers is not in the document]
ywqzzy: > The configure will be overwrite even if it is not changed?

No, only if the config is changed, will we update the config.

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -100,11 +134,14 @@ struct TiFlashSecurityConfig
...
                 redact_info_log = config.getBool("security.redact_info_log");
             }
+            return updated;
         }
+        return false;
     }
 
     void parseAllowedCN(String verify_cns)
     {
+        std::unique_lock lock(mu);
```
windtalker, line 144: Looks like when `parseAllowedCN()` is called inside `update()`, it will be deadlock?

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -32,41 +36,76 @@ extern const int INVALID_CONFIG_PARAMETER;
...
+    std::set<String> allowedCommonNames()
+    {
+        std::unique_lock lock(mu);
+        return allowed_common_names;
+    }
+
+    bool update(Poco::Util::AbstractConfiguration & config)
+    {
+        std::unique_lock lock(mu);
         if (config.has("security"))
```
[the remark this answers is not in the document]
ywqzzy: This should not happen. If we config the cluster with security config, we will not remove it online, since the flashService or http server will always use sslServerCredential.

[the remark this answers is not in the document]
windtalker: So my understand is the security config should not be changed online, neither from non-security to security nor from security to non-security? If yes, I think we should check this and reject this kind of change explicitly?

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -137,17 +185,25 @@ struct TiFlashSecurityConfig
...
-        options.pem_cert_chain = readFile(cert_path);
-        options.pem_private_key = readFile(key_path);
-        inited = true;
-        return options;
+        return allowed_common_names.count(cert.commonName()) > 0;
+    }
+
+    grpc::SslCredentialsOptions readSecurityInfo()
+    {
+        std::unique_lock lock(mu);
```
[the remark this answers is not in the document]
ywqzzy: > The original impl will cache the `options`, why remove this cache?

Now the config will be changed by 2 ways:
1. change the cert file contents
2. change the cert file path
The old cache method can't handle the  first way.
Maybe I can use modification time of cert file to check if the file is changed to handle the first case.

dbms/src/Common/TiFlashSecurity.h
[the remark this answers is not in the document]
ywqzzy: > So my understand is the security config should not be changed online, neither from non-security to security nor from security to non-security? If yes, I think we should check this and reject this kind of change explicitly?

Yes, I will handle it

dbms/src/Common/TiFlashSecurity.h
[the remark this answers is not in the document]
ywqzzy: > > The original impl will cache the `options`, why remove this cache?
> 
> Now the config will be changed by 2 ways:
> 
> 1. change the cert file contents
> 2. change the cert file path
>    The old cache method can't handle the  first way.
>    Maybe I can use modification time of cert file to check if the file is changed to handle the first case.

Further more, In current implementation, setCertificate will read the file every time it called, I didn't find a way to optimize it.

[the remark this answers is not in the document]
windtalker: OK, just keep your impl is fine to me.

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -32,60 +36,119 @@ extern const int INVALID_CONFIG_PARAMETER;
...
         if (config.has("security"))
         {
+            has_security = true;
             bool miss_ca_path = true;
             bool miss_cert_path = true;
             bool miss_key_path = true;
+            String new_ca_path;
+            String new_cert_path;
+            String new_key_path;
+            bool updated = false;
```
ywqzzy, line 97: The updated may useless.

dbms/src/Common/Config/ConfigReloader.cpp
```diff
@@ -73,29 +73,68 @@ void ConfigReloader::run()
     }
 }
 
+bool ConfigReloader::tlsCertUpdated(Poco::AutoPtr<Poco::Util::AbstractConfiguration> config, bool init)
```
windtalker, line 76: looks like for `main_config_reloader`, `init` is always false.

dbms/src/Common/Config/ConfigReloader.cpp
```diff
@@ -73,29 +73,68 @@ void ConfigReloader::run()
...
+}
+
 void ConfigReloader::reloadIfNewer(bool force, bool throw_on_error)
 {
     std::lock_guard lock(reload_mutex);
 
     FilesChangesTracker new_files = getNewFileList();
-    if (force || new_files.isDifferOrNewerThan(files))
+
+    ConfigProcessor config_processor(path);
```
windtalker, line 118: I think we don't need to load the config everytime, if `config` itself is not changed, the tls file path can be found in `cert_files`

dbms/src/Server/Server.cpp
```diff
@@ -1094,12 +1090,21 @@ int Server::main(const std::vector<std::string> & /*args*/)
...
         config_path,
-        [&](ConfigurationPtr config) {
+        [&](ConfigurationPtr config, bool is_new_tls_cert) {
+            LOG_DEBUG(log, "run main config reloader");
             buildLoggers(*config);
             global_context->setMacros(std::make_unique<Macros>(*config, "macros"));
             global_context->getTMTContext().reloadConfig(*config);
             global_context->getIORateLimiter().updateConfig(*config);
             global_context->reloadDeltaTreeConfig(*config);
+            bool updated = global_context->getSecurityConfig()->update(*config);
```
windtalker, line 1100: I think we don't need to pass `is_new_tls_cert`, `global_context->getSecurityConfig()->update(*config);` can return true if tls path or tls file is changed

dbms/src/Common/Config/ConfigReloader.cpp
ywqzzy: No. In ctor of ConfigReloader, the `force` is `true`.

dbms/src/Common/Config/ConfigReloader.cpp
```diff
@@ -73,29 +73,68 @@ void ConfigReloader::run()
...
+                cert_files.addIfExists(config->getString("security.cert_path"));
+                cert_files.addIfExists(config->getString("security.key_path"));
+            }
+            else
+            {
+                FilesChangesTracker new_files;
+                new_files.addIfExists(config->getString("security.ca_path"));
+                new_files.addIfExists(config->getString("security.cert_path"));
+                new_files.addIfExists(config->getString("security.key_path"));
+                bool updated = cert_files.isDifferOrNewerThan(new_files);
```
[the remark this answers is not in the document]
ywqzzy: The result will be the same, but I will change it.

dbms/src/Server/Server.cpp
ywqzzy: > I think we don't need to pass `is_new_tls_cert`, `global_context->getSecurityConfig()->update(*config);` can return true if tls path or tls file is changed

Yes, I choose to keep it because I want a UT of update.
I will find a better way to write the UT

ywqzzy: If we don't pass the `is_new_tls_cert`, the `update()` will check whether the tls file changed twice.

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -141,6 +143,10 @@ struct TiFlashSecurityConfig
                         ca_path = new_ca_path;
```
ywqzzy, line 143: I will refine it later

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -94,13 +164,25 @@ struct TiFlashSecurityConfig
                 parseAllowedCN(verify_cns);
             }
 
-            // redact_info_log = config.getBool("security.redact-info-log", false);
             // Mostly options name are combined with "_", keep this style
             if (config.has("security.redact_info_log"))
             {
                 redact_info_log = config.getBool("security.redact_info_log");
             }
+            return updated;
```
windtalker, line 172: if file content is change, it still returns false?

dbms/src/Common/TiFlashSecurity.h
```diff
@@ -137,17 +219,38 @@ struct TiFlashSecurityConfig
...
-        options.pem_private_key = readFile(key_path);
-        inited = true;
-        return options;
+        return allowed_common_names.count(cert.commonName()) > 0;
+    }
+
+    grpc::SslCredentialsOptions readSecurityInfo()
+    {
+        std::unique_lock lock(mu);
+        grpc::SslCredentialsOptions new_options;
```
windtalker, line 235: Looks like we can cache it now?

dbms/src/Server/Server.cpp
```diff
@@ -1094,12 +1090,21 @@ int Server::main(const std::vector<std::string> & /*args*/)
...
         config_path,
-        [&](ConfigurationPtr config) {
+        [&](ConfigurationPtr config, bool config_object_updated) {
+            LOG_DEBUG(log, "run main config reloader, config_object_updated: {}", config_object_updated);
             buildLoggers(*config);
             global_context->setMacros(std::make_unique<Macros>(*config, "macros"));
             global_context->getTMTContext().reloadConfig(*config);
             global_context->getIORateLimiter().updateConfig(*config);
             global_context->reloadDeltaTreeConfig(*config);
+            bool updated = global_context->getSecurityConfig()->update(*config); // Whether the cert path is updated.
```
windtalker, line 1100: I think `global_context->getSecurityConfig()->update(*config)` itself can return true if `config_object_updated` is true, so we don't need to add the extra parameter?

windtalker (approved):
Others LGTM

PR #6368, opened 2022-11-25
storage: Add lock for segment abandon

breezewish:

Issue Number: ref #5237

Problem Summary:

Add explicit lock argument to abandon segment.

If someone calls abandon without a segment lock, the abandon will cause race conditions with the update operations.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):
lgtm

breezewish:
As suggested by @lidezhu , the abandon should be more precisely protected by the DeltaMergeStore's lock, instead of segment's lock. So I think we could close this PR.

PR #6375, opened 2022-11-28
storage: Add more checks for ingestFiles

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/5237

Problem Summary:

1. Add check: Check files passed in are ordered and not overlapped.

2. Add check: Check files' range is contained by the ingest range (the ingest range is the region's range).

   Reason: The content of a ColumnFileBig can be only stripped by the segment range (at the read time), not any other range. If the ingest range is smaller than the files' range, for example: ingest_range = [0, 100) and file_data_range = [50, 200), [50, 200) will be still ingested, while [0, 100) will be cleared (if `clear_data_in_range` = true).

   To avoid surprising results later, we simply forbid this case now.

   Note: Although we specified a stripped range to `ingestColumnFiles` https://github.com/pingcap/tiflash/blob/09e3a9e2413189070740af58430a9c1da7fe9b72/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#LL849 , the stripped range is only used for delete data, instead of stripping the output.

   **This check is risky.** We need some time for E2E tests to verify whether there are some unintentional edge cases.

   Currently TiKV has similar checks:

   image

3. Introduce `ingestFiles` to the dm test framework, so that we can write related tests easier.

   As a result, test `DeltaMergeStoreRWTest/IngestEmptyFileLists` is migrated to `StoreIngestTest/EmptyFileLists`, simplified using new test framework.

4. Add tests to verify check 1 and check 2, based on the test framework above.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/tests/gtest_dm_simple_pk_test_basic.cpp
```diff
@@ -156,8 +167,11 @@ std::pair<Int64, Int64> SimplePKTestBasic::parseRange(const RowKeyRange & range)
     return {start_key, end_key};
 }
 
-Block SimplePKTestBasic::prepareWriteBlock(Int64 start_key, Int64 end_key, bool is_deleted)
+Block SimplePKTestBasic::fillBlock(FillBlockOptions options)
```
JaySon-Huang, line 170:
```suggestion
Block SimplePKTestBasic::fillBlock(const FillBlockOptions &options)
```

dbms/src/Storages/DeltaMerge/tests/gtest_dm_simple_pk_test_basic.cpp
```diff
@@ -290,6 +304,52 @@ void SimplePKTestBasic::deleteRange(Int64 start_key, Int64 end_key)
...
+    const auto & pk_column = block.getByPosition(0).column;
+    auto min_pk = pk_column->getInt(0);
+    auto max_pk = pk_column->getInt(block.rows() - 1);
+    HandleRange range(min_pk, max_pk + 1);
+    auto handle_range = RowKeyRange::fromHandleRange(range);
+    auto external_file = ExternalDTFileInfo{.id = file_id, .range = handle_range};
+    return external_file;
+}
+
+void SimplePKTestBasic::ingestFiles(IngestFilesOptions options)
```
JaySon-Huang, line 331:
```suggestion
void SimplePKTestBasic::ingestFiles(const IngestFilesOptions &options)
```

JaySon-Huang (approved):
LGTM with minor comment

JaySon-Huang:
Will tikv's multi-rocksdb instances feature break the assumption added here (in the future)?

breezewish:
> Will tikv's multi-rocksdb instances feature break the assumption added here (in the future)?

Currently no, the MultiRaft's data must be strictly matched with the Raft group's boundary.

lidezhu (approved):

PR #6394, opened 2022-11-30
limit the queued task number and queued duration of coprocess task.

LittleFall:

Issue Number: ref https://github.com/pingcap/tiflash/issues/6438

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

  - send 20 mpp tasks every seconds, each will produce 10 remote read request per machine.  
  - after about 2.5min (3000 tasks), stop the load.

  before
  image
  
  
  R6GBzXZE6N
  
  after
  image
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/FlashService.cpp
```diff
@@ -137,7 +137,31 @@ grpc::Status FlashService::Coprocessor(
...
+        if (auto current = GET_METRIC(tiflash_coprocessor_handling_request_count, type_cop).Value(); current > handle_limit)
+        {
+            response->mutable_region_error()->mutable_server_is_busy()->set_reason(fmt::format("tiflash cop pool queued too much, current = {}, limit = {}", current, handle_limit));
+            return grpc::Status::OK;
+        }
+    }
+
+
     grpc::Status ret = executeInThreadPool(*cop_pool, [&] {
+        if (max_queued_duration_seconds > 0)
```
JaySon-Huang, line 157: should it also applied to BatchCoprocessor?

dbms/src/Flash/FlashService.cpp
```diff
@@ -137,7 +137,31 @@ grpc::Status FlashService::Coprocessor(
 
     context->setMockStorage(mock_storage);
 
+    auto settings = context->getSettingsRef();
```
JaySon-Huang, line 140:
```suggestion
    const auto & settings = context->getSettingsRef();
```

windtalker (approved):
LGTM

dbms/src/Flash/FlashService.cpp
LittleFall: I think we needn't, because:
1. most pressure is from remote read, which is cop request.
2. batch cop will be deprecated soon.

gengliqi (approved):

PR #6435, opened 2022-12-06
WIP: split cross join return block if its rows is more than max_block_size

mengxin9014:

Issue Number: close #xxx

Problem Summary:

If we return too many rows in the block in cross join,  we'll split those rows into blocks and return them multiple times to make sure that the rows in each block are close to max_block_size.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


windtalker:
close as duplicate with #7444 and #7479

Issue #6296, opened 2022-11-11
Support MVCC Bitmap Filter


JinheLin:
## Feature Request

- [x] https://github.com/pingcap/tiflash/pull/6457
- [x] https://github.com/pingcap/tiflash/pull/6458
- [x] Read with MVCC bitmap filter:
- [ ] Revise statistic of table scanning:

PR #6458, opened 2022-12-08
Storage: row ID generation for MVCC bitmap filter

JinheLin:

Issue Number: ref #6296 

Problem Summary:

- In order to build a MVCC bitmap filter, we need to scan three columns of handle, version and tag.
- Iterating through the input stream of `Segment::getInputStream`  can obtain an ordered data stream.
- However, the current input stream does not contain the position of the record on the array, so a "virtual column" needs to be added to indicate the record position. This "virtual column" is called `SegmentRowIdCol`.
- In order to generate the `SegmentRowIdCol`, the read interfaces of Delta and Stable need to return the 'row position' of returned data.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/DeltaMerge.h
```diff
@@ -247,14 +267,36 @@ class DeltaMergeBlockInputStream final : public SkippableBlockInputStream
...
+    }
+
+    inline void fillSegmentRowId(const std::vector<UInt32> & row_ids)
+    {
+        for (UInt32 row_id : row_ids)
+        {
+            row_id += stable_rows;
+            seg_row_id_col->insertData(reinterpret_cast<const char*>(&row_id), sizeof(row_id));
+        }
+    }
```
Lloyd-Pottiger, line 299: how about moving the two function to inside `class Block`?

dbms/src/Storages/DeltaMerge/DeltaMerge.h
```diff
@@ -90,13 +91,22 @@ class DeltaMergeBlockInputStream final : public SkippableBlockInputStream
     size_t last_handle_pos = 0;
     size_t last_handle_read_num = 0;
 
+    // Use for calculating MVCC-bitmap-filter when `need_row_id` is true.
+    inline static DataTypeUInt32 seg_row_id_col_type{};
```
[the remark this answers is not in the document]
Lloyd-Pottiger: get, thanks

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetReader.h
```diff
@@ -55,7 +55,7 @@ class ColumnFileSetReader
 
     // Use for DeltaMergeBlockInputStream to read rows from MemTableSet to do full compaction with other layer.
     // This method will check whether offset and limit are valid. It only return those valid rows.
-    size_t readRows(MutableColumns & output_columns, size_t offset, size_t limit, const RowKeyRange * range);
+    size_t readRows(MutableColumns & output_columns, size_t offset, size_t limit, const RowKeyRange * range, std::vector<UInt32> * row_ids = nullptr);
```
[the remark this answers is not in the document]
Lloyd-Pottiger: why need default paremeter? just pass `{}` when `row_id` is not needed?

dbms/src/Storages/DeltaMerge/Delta/Snapshot.cpp
```diff
@@ -104,10 +104,24 @@ size_t DeltaValueReader::readRows(MutableColumns & output_cols, size_t offset, s
...
+    {
+        actual_read += mem_table_reader->readRows(output_cols, mem_table_start, mem_table_end - mem_table_start, range, row_ids);
+    }
+
+    if (row_ids != nullptr)
+    {
+        for (size_t i = persisted_read_rows; i < row_ids->size(); ++i)
+        {
+            (*row_ids)[i] += mem_table_rows_offset;
+        }
```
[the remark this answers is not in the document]
Lloyd-Pottiger: From my previous test, for `vector<bool>`, using the algorithm from `std::algorithm` is much faster than the loop. Not quite sure about `UInt32`

dbms/src/Storages/DeltaMerge/DeltaMerge.h
Lloyd-Pottiger: Since we need fillSegmentRowId before the filters, move inside `class Block` can help reuse the logic.

Lloyd-Pottiger: When implementing late materialization, we need SegmentRowId to generate fitering-bitmap.

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetReader.cpp
```diff
@@ -157,7 +157,17 @@ size_t ColumnFileSetReader::readRows(MutableColumns & output_columns, size_t off
...
+        actual_read += read_rows;
+        if (row_ids != nullptr)
+        {
+            auto rows_before_cur_file = file_index == 0 ? 0 : column_file_rows_end[file_index - 1];
+            auto start_row_id = read_offset + rows_before_cur_file;
+            for (size_t i = 0; i < read_rows; ++i)
+            {
+                row_ids->push_back(start_row_id + i);
+            }
+        }
```
Lloyd-Pottiger, line 170: usually, when MVCC, the remaining rows are more than the filter-out rows, maybe we can record the filter-out row_id?

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetReader.h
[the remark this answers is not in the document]
flowbehappy: Pointer is good for me. `std::optional<std::vector<UInt32>> &` is more complicated.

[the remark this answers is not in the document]
Lloyd-Pottiger: make sense, thank you~

dbms/src/Storages/DeltaMerge/RowKeyFilter.h
```diff
@@ -153,6 +175,10 @@ inline Block filterUnsorted(const RowKeyRanges & rowkey_ranges, Block && block,
         auto & column = block.getByPosition(i);
         column.column = column.column->filter(filter, passed_count);
     }
+    if (block.segmentRowIdCol() != nullptr)
+    {
+        block.segmentRowIdCol()->filter(filter, passed_count);
```
flowbehappy, line 180: It looks like a bug here because the result of `filter` is not used.

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileBig.cpp
```diff
@@ -138,15 +138,24 @@ size_t ColumnFileBigReader::readRowsRepeatedly(MutableColumns & output_cols, siz
         const auto & columns = cached_pk_ver_columns.at(block_index);
         const auto & pk_column = columns[0];
 
-        actual_read += copyColumnsData(columns, pk_column, output_cols, rows_start_in_block, rows_in_block_limit, range);
+        auto [offset, rows] = copyColumnsData(columns, pk_column, output_cols, rows_start_in_block, rows_in_block_limit, range);
+        // For DMFile, records are sorted by row key. Only the first block and the last block will be filtered by range.
```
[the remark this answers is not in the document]
flowbehappy: `p0 p1, p2, p3`
There are possibilities that `range` can filter out `p0` and part of `p1`. Although it won't happen by now, I am afraid it could be in the future after somebody changes the code.
I mean we better don't rely on the assumption that `range` can only filter out part of `p0` or `p3`, it is dangerous.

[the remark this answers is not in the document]
Lloyd-Pottiger: like
`p0: [0, 1000), p1:[1000, 2000), p2:[2000, 3000), p3:[3000, 4000)`
`range: [0, 1003)`
?

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilterBlockInputStream.cpp
```diff
@@ -0,0 +1,94 @@
...
+        {
+            res_filter = &filter;
+        }
+        else
+        {
+            for (auto & col : block)
+            {
+                col.column = col.column->filter(filter, block.rows());
+            }
+        }
```
Lloyd-Pottiger, line 60: optimize here, when we call `bitmap_filter->get` also return the positive count as `positive`, when `positive == block.rows()` skip the filter, and call `col.column = col.column->filter(filter, positive);` instead

dbms/src/Interpreters/Settings.h
```diff
@@ -320,6 +320,7 @@ struct Settings
                                                                                                                                                                                                                                         \
     M(SettingDouble, dt_page_gc_threshold, 0.5, "Max valid rate of deciding to do a GC in PageStorage")                                                                                                                                 \
     M(SettingBool, dt_enable_read_thread, true, "Enable storage read thread or not")                                                                                                                                                   \
+    M(SettingBool, dt_enable_bitmap_filter, true, "Use bitmap filter to read data or not")                                                                                                                                             \
```
flowbehappy, line 323: We should remove this setting before merging this pr.

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void BitmapFilter::set(const UInt32 * data, UInt32 size, const FilterPtr f)
+{
+    if (size == 0)
+    {
+        return;
+    }
+    //size_t max_row_id = *std::max_element(data, data + size);
+    //RUNTIME_CHECK(max_row_id < filter.size(), max_row_id, filter.size());
+    if (f == nullptr)
```
flowbehappy, line 41: `f == nullptr` means all positive seems not human-friendly to me.
Besides, better use `!f` instead of `f == nullptr`.

flowbehappy: You can use another function like `setAll` to make all rows positive. And do this logic outside of this class.

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetReader.h
flowbehappy, line 58: Please add documents about `row_ids`.

dbms/src/Storages/DeltaMerge/DeltaMerge.h
```diff
@@ -491,7 +546,18 @@ class DeltaMergeBlockInputStream final : public SkippableBlockInputStream
...
         // Note that the rows between [use_delta_offset, use_delta_offset + write_rows) are guaranteed sorted,
         // otherwise we won't read them in the same range.
-        auto actual_write = delta_value_reader->readRows(output_columns, use_delta_offset, write_rows, &rowkey_range);
+        size_t actual_write = 0;
+        if constexpr (need_row_id)
+        {
+            delta_row_ids.clear();
+            delta_row_ids.reserve(write_rows);
+            actual_write = delta_value_reader->readRows(output_columns, use_delta_offset, write_rows, &rowkey_range, &delta_row_ids);
+            fillSegmentRowId(delta_row_ids);
```
flowbehappy, line 555: I wonder why delta's `row_ids` cannot be calculated like stable's.

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2224,5 +2270,111 @@ bool Segment::placeDelete(const DMContext & dm_context,
...
+                                                        const ColumnDefines & columns_to_read,
+                                                        const RowKeyRanges & data_ranges,
+                                                        const RSOperatorPtr & filter,
+                                                        UInt64 max_version,
+                                                        size_t expected_block_size)
+{
+    auto & segment_snap = bitmap_filter->snapshot();
+    auto enable_handle_clean_read = !hasColumn(columns_to_read, EXTRA_HANDLE_COLUMN_ID);
+    constexpr auto is_fast_scan = true;
+    auto enable_del_clean_read = !hasColumn(columns_to_read, TAG_COLUMN_ID);
```
Lloyd-Pottiger, line 2324: I think `enable_del_clean_read` is always false. TiDB Planner is not aware of the tag column, so columns_to_read never contain tag column.

Lloyd-Pottiger (commented):
The code is mixed with bitmap logic now.

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilterBlockInputStream.cpp
```diff
@@ -0,0 +1,97 @@
...
+                res_filter = &filter;
+            }
+            else
+            {
+                for (auto & col : block)
+                {
+                    col.column = col.column->filter(filter, block.rows());
+                }
+            }
+        }
```
Lloyd-Pottiger, line 63: what if `all_match = true && return_filter = true`

dbms/src/Storages/DeltaMerge/DeltaMerge.h
```diff
@@ -247,14 +265,40 @@ class DeltaMergeBlockInputStream final : public SkippableBlockInputStream
...
+            v[offset + i] = start + i;
+        }
+    }
+
+    inline void fillSegmentRowId(const std::vector<UInt32> & row_ids)
+    {
+        ColumnUInt32::Container & v = seg_row_id_col->getData();
+        auto offset = v.size();
+        v.resize(v.size() + row_ids.size());
+        for (UInt32 i = 0; i < row_ids.size(); i++)
```
Lloyd-Pottiger, line 297:
```suggestion
        for (UInt32 i = 0; i < row_ids.size(); ++i)
```

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+        {
+            set(blk.segmentRowIdCol(), f);
+        }
+        else
+        {
+            break;
+        }
+    }
+    stream->readSuffix();
+}
```
Lloyd-Pottiger, line 44: Feel confused about this function. When the stream is `DMVersionFilterBlockInputStream`, the return FilterPtr is associated with the block return by `DMRowKeyFilterBlockInputStream` rather than the original block in dmfile, why it can be used for generating bitmap?

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.h
```diff
@@ -0,0 +1,49 @@
...
+
+    SegmentSnapshotPtr & snapshot();
+
+    void runOptimize();
+
+    String toDebugString() const;
+
+private:
+    std::vector<bool> filter;
+    SegmentSnapshotPtr snap;
```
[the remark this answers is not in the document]
flowbehappy: `BitmapFilter` better be a pure data object. It is not necessary to make `snap` a member of it. Besides, `snap` is only here to be returned by `snapshot`, it does not mean anything to this object. 

I think you can consider passing `snap` around outside of this object.

dbms/src/Storages/DeltaMerge/DMVersionFilterBlockInputStream.cpp
```diff
@@ -393,23 +393,32 @@ Block DMVersionFilterBlockInputStream<MODE>::read(FilterPtr & res_filter, bool r
...
             {
-                auto & column = cur_raw_block.getByName(c.name);
-                column.column = column.column->filter(filter, passed_count);
-                res.insert(std::move(column));
+                for (const auto & c : header)
+                {
+                    auto & column = cur_raw_block.getByName(c.name);
+                    column.column = column.column->filter(filter, passed_count);
+                    res.insert(std::move(column));
+                }
```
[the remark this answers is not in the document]
Lloyd-Pottiger: ok

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2224,5 +2255,285 @@ bool Segment::placeDelete(const DMContext & dm_context,
...
+                                                                      columns_to_read,
+                                                                      read_ranges,
+                                                                      filter,
+                                                                      max_version,
+                                                                      expected_block_size,
+                                                                      /*enable_handle_clean_read*/ false,
+                                                                      /*is_fast_scan*/ false,
+                                                                      /*enable_del_clean_read*/ false,
+                                                                      /*read_packs*/ some_packs,
+                                                                      /*need_row_id*/ true);
```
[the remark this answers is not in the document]
Lloyd-Pottiger: We need to fill `segment_row_id_col` with continuous row_id, and the filter twice. We can call `Block read(FilterPtr, true)` both in `DMRowKeyFilterBlockInputStream` and `DMVersionFilterBlockInputStream`, which means that there is no need to fill `segment_row_id_col` with continuous row_id, and filter only once.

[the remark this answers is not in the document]
Lloyd-Pottiger: Since there is no `DeltaMergeBlockInputStream`, the order of rows will not change, so we can do this optimization.

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
Lloyd-Pottiger: get it

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+SegmentSnapshotPtr & BitmapFilter::snapshot()
+{
+    return snap;
+}
+
+void BitmapFilter::runOptimize()
+{
+    all_match = std::find(filter.begin(), filter.end(), false) == filter.end();
+}
```
Lloyd-Pottiger, line 111: if `all_match`, we can just clear `filter`?

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+    String s(filter.size(), '1');
+    for (UInt32 i = 0; i < filter.size(); i++)
+    {
+        if (!filter[i])
+        {
+            s[i] = '0';
+        }
+    }
+    return fmt::format("{}", s);
+}
```
Lloyd-Pottiger, line 124: The return string usually is too long, I think we can just return size and number of positive.

Lloyd-Pottiger: ok, get

dbms/src/Storages/DeltaMerge/Segment.cpp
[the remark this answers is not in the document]
Lloyd-Pottiger: Are there some special reason for not supporting `read(FilterPtr, true)` in `DMRowKeyFilterBlockInputStream`? Maybe   we can achieve it since filter is heavy-cost?

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+void BitmapFilter::set(const UInt32 * data, UInt32 size, const FilterPtr & f)
+{
+    if (size == 0)
+    {
+        return;
+    }
+    //size_t max_row_id = *std::max_element(data, data + size);
+    //RUNTIME_CHECK(max_row_id < filter.size(), max_row_id, filter.size());
```
flowbehappy, line 59:
```suggestion
```

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+bool BitmapFilter::get(IColumn::Filter & f, UInt32 start, UInt32 limit) const
+{
+    RUNTIME_CHECK(start + limit <= filter.size(), start, limit, filter.size());
+    auto begin = filter.cbegin() + start;
+    auto end = filter.cbegin() + start + limit;
+    if (all_match || std::find(begin, end, false) == end)
+    {
+        //static const UInt8 match = 1;
+        //f.assign(static_cast<size_t>(limit), match);
```
flowbehappy, line 93:
```suggestion
```

dbms/src/Storages/DeltaMerge/BitmapFilter/BitmapFilter.h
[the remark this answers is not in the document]
Lloyd-Pottiger: Agree, we can move `BitmapFilter` to `libs/libcommon` or `dbms/src/Common`, and set `avx2` flag during compiling to speed up.

dbms/src/Storages/DeltaMerge/Segment.cpp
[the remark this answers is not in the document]
Lloyd-Pottiger: I think it is worth to do this now. We do not need to `createSegmentRowIdCol` in `ConcatSkippableBlockInputStream`. When the parent of `ConcatSkippableBlockInputStream` is `DeltaMergeBlockInputStream`, there is no need to `createSegmentRowIdCol` since  `DeltaMergeBlockInputStream` will do that. And the parent of `ConcatSkippableBlockInputStream` is `DMRowKeyFilterBlockInputStream`, we can call `read(FilterPtr, true)` to return filter, and we can use filter to generaate bitmap.

flowbehappy (approved):
LGTM

dbms/src/Storages/DeltaMerge/Segment.cpp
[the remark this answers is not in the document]
Lloyd-Pottiger: no, I mean use filter to replace `SegmentRowIdCol`.

`ConcatSkippableBlockInputStream` return the following block:

```
startOffset: 100
col filter
1   1
1   1
2   1
3   1
4   1
5   1
6   1
```

After `DMVersionFilterBlockInputStream`, the block is:

```
startOffset: 100
col filter
1   0
1   1
2   1
3   1
4   1
5   1
6   1
```

After `DMRowKeyFilterBlockInputStream`, the block is:

```
startOffset: 100
col filter
1   0
1   0
2   1
3   1
4   1
5   1
6   0
```

Then we can just set [102, 103, 104, 105] of bitmap to 1.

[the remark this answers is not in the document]
Lloyd-Pottiger: > However, read continuous packs is not necessary since the packs to be read has been specified. We can read multiple packs from DMFile concurrently in this situation in the future or in the cloud native mode, we can return several already cached packs first.

But fill `SegmentRowIdCol` in `ConcatSkippableBlockInputStream` can not handle this case either. Like we can a block in `ConcatSkippableBlockInputStream`, which contain data in `pack2`, `pack9` and `pack88`, how we can know about that?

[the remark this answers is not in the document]
Lloyd-Pottiger: > because it would introduce special code paths and more restrictions but the benefits are small.

fill `SegmentRowIdCol` in `ConcatSkippableBlockInputStream` also introduce code path 🤔 And using `offset` only need little changes in `BitmapFilter`

Lloyd-Pottiger (approved):
The others LGTM. We can add some comments about previous discussion. Leave it to be optimized.

Lloyd-Pottiger:
> > The others LGTM. We can add some comments about previous discussion. Leave it to be optimized.
> 
> Can you write an issue about it? @Lloyd-Pottiger

ok

PR #6461, opened 2022-12-08
Optimize delta compact strategy to reduce iops and write amplification

### LINKED ISSUE #6460: Optimize read iops caused by delta compact
## Enhancement
Currently, delta compact can cause a lot of disk read ops. Because `readBlockForMinorCompaction` use `PageReader::read(const std::vector<PageStorage::PageReadFields> & page_fields)` to read all columns data of the cftiny. And this read api does not combine the continuously reads together (another optimization opportunity).

And write amplification is also very high, especially the write workload is distributed and the data volume is large.

image
image
image
image

flowbehappy:

Issue Number: close #6460 

Problem Summary:

1. Use `PageReader::read(PageId page_id)` instead of `PageReader::read(const std::vector<PageStorage::PageReadFields> & page_fields)` in `ColumnFileTiny::readBlockForMinorCompaction`. So that we only need to do 1 read for each cftiny, instead of the number of columns.
2. Use a better compact strategy: do delta compact only if the *increased* number of cftiny is larger than the limit, instead of the *total* number.

Tested on the worst update workload of TiFlash: large volume data, each transaction only update 1 rows, and the update pk is very distributed. Results:

1. Read iops reduces from 80K to 4K
2. Write amplification reduces from 75x to 25x

Before:
image
image
image
image

After:

image
image
image
image

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce the IOPS and write amplification of TiFlash under high update throughput workloads.
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/Delta/ColumnFilePersistedSet.cpp
+++ b/dbms/src/Storages/DeltaMerge/Delta/ColumnFilePersistedSet.cpp
@@ -263,6 +263,7 @@ MinorCompactionPtr ColumnFilePersistedSet::pickUpMinorCompaction(DMContext & con
             is_all_trivial_move = is_all_trivial_move && is_trivial_move;
             cur_task = {};
         };
+        size_t index = 0;
         for (auto & file : persisted_files)
         {
             if (auto * t_file = file->tryToTinyFile(); t_file)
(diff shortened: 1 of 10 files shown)


dbms/src/Storages/DeltaMerge/Delta/ColumnFilePersistedSet.cpp
```diff
@@ -280,16 +281,20 @@ MinorCompactionPtr ColumnFilePersistedSet::pickUpMinorCompaction(DMContext & con
...
                 pack_up_cur_task();
-                cur_task.addColumnFile(file);
+                cur_task.addColumnFile(file, index);
             }
+
+            ++index;
         }
         pack_up_cur_task();
 
+        RUNTIME_ASSERT(is_all_trivial_move || compaction->getFirsCompactIndex() != std::numeric_limits<size_t>::max());
```
JaySon-Huang, line 296: I think move this check to `DeltaValueSpace::compact` is better

lidezhu (approved):
LGTM

flowbehappy:
Thanks for the ut @JaySon-Huang  ❤️

PR #6430, opened 2022-12-06
Get correct cpu cores in k8s pod

xzhangxian1008:
### What problem does this PR solve?

Issue Number: close #6434

Problem Summary:

max_streams of xxxxInputStream is determined by Context's max_threads which is initialized by `getNumberofPhysicalCPUCores()`. In x86 virtual environment, we get max_threads as 48 which is physical cpu cores while getting max_threads as 128 which is logical cpu cores in arm virtual environment. 128 is much larger than 48, so it will use more threads and exceed the limit more easily.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #6481, opened 2022-12-12
Set max_threads according to the number of physical cpu cores

### LINKED ISSUE #6480: Set max_threads according to the number of physical cpu cores

xzhangxian1008:

Issue Number: close #6480

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/getNumberOfCPUCores.cpp
+++ b/dbms/src/Common/getNumberOfCPUCores.cpp
@@ -43,7 +43,7 @@ void setNumberOfLogicalCPUCores(UInt16 number_of_logical_cpu_cores_)
 void computeAndSetNumberOfPhysicalCPUCores(UInt16 number_of_logical_cpu_cores_, UInt16 number_of_hardware_physical_cores)
 {
     // First of all, we need to take consideration of two situation:
-    //   1. tiflash on physical machine. 
+    //   1. tiflash on physical machine.
     //      In old codes, tiflash needs to set max_threads which is equal to
     //      physical cpu cores, so we need to ensure this behavior is not broken.
     //   2. tiflash on virtual environment.
--- a/dbms/src/Common/getNumberOfCPUCores.cpp
+++ b/dbms/src/Common/getNumberOfCPUCores.cpp
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <Common/getNumberOfLogicalCPUCores.h>
+#include <Common/getNumberOfCPUCores.h>
+#include "common/types.h"
 
 namespace CPUCores
 {
 UInt16 number_of_logical_cpu_cores = std::thread::hardware_concurrency();
+UInt16 number_of_physical_cpu_cores = std::thread::hardware_concurrency() / 2;
 } // namespace CPUCores
 
 
@@ -25,10 +27,21 @@ UInt16 getNumberOfLogicalCPUCores()
     return CPUCores::number_of_logical_cpu_cores;
 }
 
+UInt16 getNumberOfPhysicalCPUCores()
+{
+    return CPUCores::number_of_physical_cpu_cores;
+}
+
 // We should call this function before Context has been created,
 // which will call `getNumberOfLogicalCPUCores`, or we can not
 // set cpu cores any more.
-void setNumberOfLogicalCPUCores(UInt16 max_logical_cpu_cores)
+void setNumberOfLogicalCPUCores(UInt16 number_of_logical_cpu_cores_)
 {
-    CPUCores::number_of_logical_cpu_cores = max_logical_cpu_cores;
+    CPUCores::number_of_logical_cpu_cores = number_of_logical_cpu_cores_;
+}
+
+void setNumberOfPhysicalCPUCores(UInt16 number_of_logical_cpu_cores_, UInt16 number_of_hardware_physical_cores)
+{
+    auto hardware_logical_cpu_cores = std::thread::hardware_concurrency();
+    CPUCores::number_of_physical_cpu_cores = number_of_logical_cpu_cores_ / (hardware_logical_cpu_cores / number_of_hardware_physical_cores);
 }
--- a/dbms/src/Common/getNumberOfCPUCores.cpp
+++ b/dbms/src/Common/getNumberOfCPUCores.cpp
@@ -27,7 +27,7 @@ UInt16 getNumberOfLogicalCPUCores()
     return CPUCores::number_of_logical_cpu_cores;
 }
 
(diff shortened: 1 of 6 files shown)


JaySon-Huang:
Can we remove the contrib/libcpuid ?

xzhangxian1008:
> Can we remove the contrib/libcpuid ?

After decoupling the storage and compute, compute layer will have to get these information by itself, instead of from proxy.

yibin87 (approved):
Please double check the "server_info.cpu_info.physical_cores" value in docker env. Others lgtm

Lloyd-Pottiger:
why physical cpu cores? IMO, `logical_cpu_cores / 2` is better

Lloyd-Pottiger:
> why is physical cpu cores? IMO, `logical_cpu_cores / 2` is better

the number of physical cpu cores in CH is also equal to `logical_cpu_cores / 2`, rather than the number of real hardware cores.

dbms/src/Common/getNumberOfCPUCores.cpp
```diff
@@ -25,10 +27,21 @@ UInt16 getNumberOfLogicalCPUCores()
...
 // We should call this function before Context has been created,
 // which will call `getNumberOfLogicalCPUCores`, or we can not
 // set cpu cores any more.
-void setNumberOfLogicalCPUCores(UInt16 max_logical_cpu_cores)
+void setNumberOfLogicalCPUCores(UInt16 number_of_logical_cpu_cores_)
+{
+    CPUCores::number_of_logical_cpu_cores = number_of_logical_cpu_cores_;
+}
+
+void setNumberOfPhysicalCPUCores(UInt16 number_of_logical_cpu_cores_, UInt16 number_of_hardware_physical_cores)
```
bestwoody, line 43: this function is actually `computeAndSetNumberOfPhysicalCPUCores()` . `setNumberOfPhysicalCPUCores()` is somehow misleading

dbms/src/Common/getNumberOfCPUCores.cpp
```diff
@@ -25,10 +27,21 @@ UInt16 getNumberOfLogicalCPUCores()
...
 {
-    CPUCores::number_of_logical_cpu_cores = max_logical_cpu_cores;
+    auto hardware_logical_cpu_cores = std::thread::hardware_concurrency();
+    CPUCores::number_of_physical_cpu_cores = number_of_logical_cpu_cores_ / (hardware_logical_cpu_cores / number_of_hardware_physical_cores);
```
bestwoody, line 46: it seems` hardware_logical_cpu_cores` and `number_of_logical_cpu_cores_` both are assigned from `std::thread::hardware_concurrency()`. Why not use `x`  instead of `n/(n/x)`? `x` is `number_of_hardware_physical_cores`, `n` is `std::thread::hardware_concurrency()`

xzhangxian1008: > it seems` hardware_logical_cpu_cores` and `number_of_logical_cpu_cores_` both are assigned from `std::thread::hardware_concurrency()`. Why not use `x` instead of `n/(n/x)`? `x` is `number_of_hardware_physical_cores`, `n` is `std::thread::hardware_concurrency()`

First of all, we need to take consideration of two situation:
1. **tiflash on physical machine.** In old codes, tiflash needs to set `max_threads` which is equal to physical cpu cores, so we need to ensure this behavior is not broken.
2. **tiflash on virtual environment** In virtual environment, when setting `max_threads`, we can't directly get physical cpu cores to set this variable because only host machine's physical cpu core can be reached. So, number of physical cpus cores can only be assigned by calculated with logical cpu cores.

Back to comment, 
- `number_of_logical_cpu_cores_` which represents how many logical cpu cores a tiflash could use(no matter in physical or virtual environment) is assigned from `ServerInfo`.
- `hardware_logical_cpu_cores` represents how many logical cpu cores the host physical machine has.
- `number_of_hardware_physical_cores` represent how many physical cpu cores the host physical machine has.
- `(hardware_logical_cpu_cores / number_of_hardware_physical_cores)` means how many logical cpu core a physical cpu core has.
- `number_of_logical_cpu_cores_ / (hardware_logical_cpu_cores / number_of_hardware_physical_cores)` means how many physical cpu cores the tiflash process could use. (Actually, it's needless to get physical cpu cores in virtual environment, bu we must ensure the behavior `1` is not broken)

bestwoody: how about copy those explainings into code as comments.

xzhangxian1008: > how about copy those explainings into code as comments.

okk

dbms/src/Common/getNumberOfCPUCores.cpp
xzhangxian1008: > this function is actually `computeAndSetNumberOfPhysicalCPUCores()` . `setNumberOfPhysicalCPUCores()` is somehow misleading

okk

dbms/src/Common/getNumberOfCPUCores.cpp
```diff
@@ -0,0 +1,62 @@
...
+UInt16 number_of_physical_cpu_cores = std::thread::hardware_concurrency() / 2;
+} // namespace CPUCores
+
+
+UInt16 getNumberOfLogicalCPUCores()
+{
+    return CPUCores::number_of_logical_cpu_cores;
+}
+
+UInt16 computeAndSetNumberOfPhysicalCPUCores()
```
bestwoody, line 30: mistake？

xzhangxian1008: > 

sry, it;s a mistake..

bestwoody (approved):

windtalker (approved):
LGTM

breezewish:
https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/getNumberOfPhysicalCPUCores.cpp#L51

In clickhouse, when cores < 32, it will set max threads according to logical cores instead of physical cores. May be we could follow this pattern?

JaySon-Huang:
The reason of changes in this PR is:

https://github.com/pingcap/tiflash/pull/6430/files#diff-64dad304ef83932f1dc505de02b2bca69271c6af8023291c2a71c61778e84dc2R168-R172
In #6430, SettingsMaxThreads is set to logical CPU cores but not physical cores. After that, TiFlash uses more threads to run some queries, which brings performance regression.

PR #6484, opened 2022-12-13
Minor refine of `TableLock` and `MPPTask::cancel`

SeaRise:

Issue Number: 
ref #5095 
ref #5900

Problem Summary:

- Move the table locks from `BlockInputStream` to `DAGContext`
- Always use non-kill to cancel `BlockInputStream`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/DAGContext.h
+++ b/dbms/src/Flash/Coprocessor/DAGContext.h
@@ -380,6 +380,9 @@ private:
 
 private:
     std::shared_ptr<ProcessListEntry> process_list_entry;
+    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query,
+    /// even if there are no local regions.
+    TableLockHolders table_locks;
     /// profile_streams_map is a map that maps from executor_id to profile BlockInputStreams.
     std::unordered_map<String, BlockInputStreams> profile_streams_map;
     /// executor_id_to_join_id_map is a map that maps executor id to all the join executor id of itself and all its children.
@@ -408,10 +411,6 @@ private:
     // In disaggregated tiflash mode, table_scan in tiflash_compute node will be converted ExchangeReceiver.
     // Record here so we can add to receiver_set and cancel/close it.
     std::optional<std::pair<String, ExchangeReceiverPtr>> disaggregated_compute_exchange_receiver;
-
-    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query,
-    /// even if there are no local regions.
-    TableLockHolders table_locks;
 };
 
 } // namespace DB
--- a/dbms/src/Flash/Coprocessor/DAGContext.h
+++ b/dbms/src/Flash/Coprocessor/DAGContext.h
@@ -380,8 +380,8 @@ private:
 
 private:
     std::shared_ptr<ProcessListEntry> process_list_entry;
-    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query,
-    /// even if there are no local regions.
+    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query, even if there are no local regions.
+    /// TableLockHolders need to be released after the BlockInputStream is destroyed to prevent data read exceptions.
     TableLockHolders table_locks;
     /// profile_streams_map is a map that maps from executor_id to profile BlockInputStreams.
     std::unordered_map<String, BlockInputStreams> profile_streams_map;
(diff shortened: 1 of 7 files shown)


dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -406,6 +408,10 @@ class DAGContext
     // In disaggregated tiflash mode, table_scan in tiflash_compute node will be converted ExchangeReceiver.
     // Record here so we can add to receiver_set and cancel/close it.
     std::optional<std::pair<String, ExchangeReceiverPtr>> disaggregated_compute_exchange_receiver;
+
+    /// We don't want the table to be dropped during the lifetime of this query,
+    /// and sometimes if there is no local region, we will hold the table lock.
```
SeaRise, line 413:
```suggestion
    /// We don't want the table to be dropped during the lifetime of this query,
    /// even though there is no local region here.
```

dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -328,6 +328,8 @@ class DAGContext
     void setProcessListEntry(std::shared_ptr<ProcessListEntry> entry) { process_list_entry = entry; }
     std::shared_ptr<ProcessListEntry> getProcessListEntry() const { return process_list_entry; }
 
+    void addTableLock(const TableLockHolder & lock) { table_locks.push_back(lock); }
```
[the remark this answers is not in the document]
SeaRise: yes, because `addTableLock` will only be called by `DAGStorageInterpreter` in a single thread, there is no data race, in line with `IBlockInputStream::addTableLock`.

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -94,14 +94,12 @@ void MPPTask::abortReceivers()
     receiver_set->cancel();
 }
 
-void MPPTask::abortDataStreams(AbortType abort_type)
+void MPPTask::abortDataStreams()
```
windtalker, line 97: Seems now we can rename `abortDataStreams` to `abortExecutor`?

dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -406,6 +408,10 @@ class DAGContext
     // In disaggregated tiflash mode, table_scan in tiflash_compute node will be converted ExchangeReceiver.
     // Record here so we can add to receiver_set and cancel/close it.
     std::optional<std::pair<String, ExchangeReceiverPtr>> disaggregated_compute_exchange_receiver;
+
+    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query,
+    /// even if there are no local regions.
+    TableLockHolders table_locks;
```
windtalker, line 414: After this pr, the locks may be released before the data streams, before this pr, the locks will be released when data streams are destructed, is it the expected behavior?

dbms/src/Flash/Mpp/MPPTask.cpp
SeaRise: yes, renamed.

windtalker (approved):
LGTM

dbms/src/Flash/Coprocessor/DAGContext.h
SeaRise: There may be a problem with cancel because the asynchronous table scan will not stop until after the UnorderedInputStream is destroyed...

SeaRise: fixed.

dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -378,6 +380,9 @@ class DAGContext
 
 private:
     std::shared_ptr<ProcessListEntry> process_list_entry;
+    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query,
```
ywqzzy, line 383: Comment out why the `table_locks` is inited before streams?
Then we can make the lifetime of each component more clear.

ywqzzy (approved):

dbms/src/Flash/Coprocessor/DAGContext.h
SeaRise: ok

dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -378,6 +380,9 @@ class DAGContext
...
+    /// even if there are no local regions.
```
SeaRise, line 384:
```suggestion
    /// Holding the table lock to make sure that the table wouldn't be dropped during the lifetime of this query, even if there are no local regions.
    /// TableLockHolders need to be released after the BlockInputStream is destroyed to prevent data read exceptions.
```

dbms/src/Flash/Coprocessor/DAGContext.h
SeaRise: done

bestwoody (approved):

PR #6496, opened 2022-12-16
Storage: Avoid wasting CPU with inefficient scheduling in SegmentReadTaskScheduler

### LINKED ISSUE #6495: SegmentReadTaskScheduler waste too many CPU
## Bug Report

image

The reason is that:
- `scheduleSegmentReadTaskPoolUnlock` does not check the limit on the number of active segments. If the number of active segments has reached the limit, this causes `scheduleSegmentUnlock` to fail to schedule a segment.
- When `scheduleSegmentUnlock` scheduling fails, the scheduler will immediately execute the next scheduling because there may be other `segmentReadTaskPool`s to schedule.
- If there are no tasks to schedule, this will cause the scheduling to keep executing between `scheduleSegmentReadTaskPoolUnlock` and `scheduleSegmentUnlock`.

JinheLin:

Issue Number: close #6495 

Problem Summary: 

To determine whether a `SegmentReadTaskPool` needs to be scheduled, we need to determine free block slots and free active segments both greater than 0.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Before this PR:
image
After this PR:
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.h
```diff
@@ -225,7 +228,7 @@ class SegmentReadTaskPool : private boost::noncopyable
 
     int64_t increaseUnorderedInputStreamRefCount();
     int64_t decreaseUnorderedInputStreamRefCount();
```
Lloyd-Pottiger, line 230: also use `Int64`?

Lloyd-Pottiger (approved):
Others LGTM.

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.h
```diff
@@ -183,6 +184,8 @@ class SegmentReadTaskPool : private boost::noncopyable
         , unordered_input_stream_ref_count(0)
         , exception_happened(false)
         , mem_tracker(current_memory_tracker == nullptr ? nullptr : current_memory_tracker->shared_from_this())
+        , block_slot_limit(std::max(num_streams_, 3))
+        , active_segment_limit(std::max(num_streams_, 2))
```
Lloyd-Pottiger, line 188: Thread number of a SegmentReadPool object is the same as the number of CPU logical core of a CPU NUMA node.
Why `active_segment_limit` and `block_slot_limit` is set to `num_streams_` which mainly equal to half of the number of CPU logical core? When there are two NUMA node, they are the same, but what if there are more than two NUMA node?

Lloyd-Pottiger: get

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.h
```diff
@@ -275,6 +279,9 @@ class SegmentReadTaskPool : private boost::noncopyable
     // std::once_flag and std::call_once to prevent duplicated add.
     std::once_flag add_to_scheduler;
 
+    const Int64 block_slot_limit;
+    const Int64 active_segment_limit;
```
JaySon-Huang, line 283: Add some comments about these two var?

dbms/src/Storages/DeltaMerge/SegmentReadTaskPool.h
flowbehappy, line 188: I think we need some comments to explain why `3` and `2` are selected.

JaySon-Huang (approved):

Issue #6507, opened 2022-12-20
TiFlash crash frequently when checkAndApplyPreHandledSnapshot


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

a big table r/w and sync Tiflash replica
```
[2022/12/20 04:10:32.765 +00:00] [WARN] [SegmentReader.cpp:73] ["sched_setaffinity fail, cpus=[116, 106, 96, 124, 58, 114, 48, 104, 38, 122, 56, 112, 46, 102, 36, 120, 54, 110, 44, 100, 34, 62, 119, 52, 109, 42, 99, 32, 127, 60, 117, 50, 107, 40, 97, 125, 59, 115, 49, 105, 39, 123, 57, 113, 47, 103, 37, 121, 55, 111, 45, 101, 35, 63, 53, 43, 33, 61, 118, 51, 108, 41, 98, 126] errno=Invalid argument"] [source=SegmentReader] [thread_id=129]
[2022/12/20 04:11:08.084 +00:00] [WARN] [StoragePool.cpp:687] ["The DTFile is already exists, continute to acquire another ID. [call=std::tuple<String, PageId> DB::DM::DeltaMergeStore::preAllocateIngestFile()][path=/data0/db/data/t_94/stable] [id=561431]"] [source=db_92.t_94] [thread_id=152]
[2022/12/20 04:11:08.084 +00:00] [WARN] [StoragePool.cpp:687] ["The DTFile is already exists, continute to acquire another ID. [call=std::tuple<String, PageId> DB::DM::DeltaMergeStore::preAllocateIngestFile()][path=/data0/db/data/t_94/stable] [id=561432]"] [source=db_92.t_94] [thread_id=152]
[2022/12/20 04:11:10.229 +00:00] [ERROR] [Exception.cpp:89] ["Code: 49, e.displayText() = DB::Exception: Check compare(range.getEnd(), ext_file.range.getEnd()) >= 0 failed, e.what() = DB::Exception, Stack trace:
       0x171df4e  DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) [tiflash+24239950]
                  dbms/src/Common/Exception.h:46
[...]

PR #6538, opened 2022-12-23
Add back ingest file range check

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/6507

Problem Summary:

Add back ingest file range check, and use `toNext()` instead of `toPrefixNext()`.

The `toNext()` works for common handle, but does not work when there is a `0x00` suffix for int handles. This is intentional, as in this case TiFlash will suffer from data lose.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/tests/gtest_key_range.cpp
```diff
@@ -98,6 +98,56 @@ TEST(RowKeyRange_test, RedactRangeFromCommonHandle)
...
+            RecordKVFormat::genKey(1, 0),
+            RecordKVFormat::genKey(1, 21));
+        const auto range = RowKeyRange::fromRegionRange(
+            range_keys,
+            /* table_id */ 1,
+            /* is_common_handle */ false,
+            /* row_key_column_size */ 1);
+        EXPECT_EQ(0, compare(next.toRowKeyValueRef(), range.getEnd()));
+    }
+    // Note: The following does not work, because {20,00} will be regarded as Key=20 in RowKeyRange::fromRegionRange.
```
JaySon-Huang, line 122: It would be better if we could move the check before tiflash sync the data, but it seems hard.

- For pd-server, it knows nothing about "table" and "rowkey", so it can not check the split key is valid or not
- For tiflash-server, it does not (and should not?) do a schema check when applying Region split. So tiflash-server can only do this check when data is receive or query key range receive

* * *

Another approach to resolve this problem maybe to parse key {20, 00} as Key=21 in `RowKeyRange::fromRegionRange`. Can this avoid data loss when decoding snapshot data or handling queries with key range with '\00' suffix?

breezewish: > Another approach to resolve this problem maybe to parse key {20, 00} as Key=21 in RowKeyRange::fromRegionRange. Can this avoid data loss when decoding snapshot data or handling queries with key range with '\00' suffix?

I guess it could work.

My thinking is that, it would be nicer if we could enhance TiFlash's robustness to an extent of supporting **arbitrary** region ranges. `\00` is only one of the case. However this may be hard.

Another direction is to let TiFlash completely forbid arbitrary region ranges, but only accept correct ones (and always fail for incorrect ones). This could also be a good behavior if other components follow it.

It may not be a very good idea to stand in the middle of the two, as this may be not a not well-defined behavior now.

JaySon-Huang: yes, I agree

JaySon-Huang (approved):

lidezhu (approved):

Issue #6225, opened 2022-11-01
Enhance EstablishConnection


xzhangxian1008:
## Enhancement

- [x] Make EstablishConnectionAsync non block
- [ ] Tune the default value of async_pollers_per_cq
- [x] Fix establish connection related metric after enable async server
- [x] Add metrics about write bytes for EstablishConnection call
- [x] Add metrics about grpc async client and server connections
- [x] Add metrics about the status of exchange sender/receiver queue
- [ ] Refine async client in ExchangeReciever, remove the extra reactor thread
  - [x] Use alarm to trigger sleep
  - [ ] Remove reactor thread
- [ ] Refine local tunnel, remove the extra read loop thread

gengliqi:
> Tune the default value of async_pollers_per_cq

ref https://github.com/pingcap/tiflash/issues/5653.
After https://github.com/pingcap/tiflash/pull/6122 makes `EstablishConnectionAsync` nonblocking, the `call_cq` and `notify_cq` can be combined. Also, we can try to use more completion queues to test whether the performance can be better.
Furthermore, since the operation in grpc thread is all nonblocking, the number of grpc thread should not greater than the CPU core number.

Issue #5758, opened 2022-09-01
introduce magic_enum into tiflash


ywqzzy:
## Enhancement
Magic_enum provide static reflection for enums (to string, from string, iteration) for modern C++, work with any enum type without any macro or boilerplate code.
https://github.com/Neargye/magic_enum

Ziy1-Tan:
Just curious, is this a issue about project building? I am interested in this.

ywqzzy:
> Just curious, is this a issue about project building? I am interested in this.

Part of it is related to project building. And we want to replace some of enum toString functions using magic_enum.

Ziy1-Tan:
Looks like a big project?

ywqzzy:
> intro

You are right, just introduce magic_enum, then replace one original enum is ok.
Then we can replace the others later.

ywqzzy:
> Looks like a big project? many places may be modify. Or this issues just intro `magic_enum` to tiflash, But the modification of the original enum and the use of future development are not included in this task?

You are right, just introduce magic_enum, then replace one original enum is ok.
Then we can replace the others later.

Ziy1-Tan:
> > Looks like a big project? many places may be modify. Or this issues just intro `magic_enum` to tiflash, But the modification of the original enum and the use of future development are not included in this task?
> 
> You are right, just introduce magic_enum, then replace one original enum is ok. Then we can replace the others later.

Nice one. I want to change it, I have some experience on MariaDB OLAP Task. Could you give me some guide?very grateful!

ywqzzy:
> > > Looks like a big project? many places may be modify. Or this issues just intro `magic_enum` to tiflash, But the modification of the original enum and the use of future development are not included in this task?
> > 
> > 
> > You are right, just introduce magic_enum, then replace one original enum is ok. Then we can replace the others later.
> 
> Nice one. I want to change it, I have some experience on MariaDB OLAP Features. Could you give me some guide?very grateful!

Wow. You can copy the header of magic_enum into dbms/src/Common. And write some simple test for one of the enum we used in TiFlash. BTW, you can contact me on wechat or qq.

ywqzzy:
> > > > Looks like a big project? many places may be modify. Or this issues just intro `magic_enum` to tiflash, But the modification of the original enum and the use of future development are not included in this task?
> > > 
> > > 
> > > You are right, just introduce magic_enum, then replace one original enum is ok. Then we can replace the others later.
> > 
> > 
[...]

PR #6550, opened 2022-12-26
Add GRPCReceiveQueue

xzhangxian1008:

Issue Number: ref #6225

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,143 @@
...
+            status = Status::WAITING;
+            tag = new_tag;
+            return GRPCReceiveQueueRes::FULL;
+        }
+        default:
+            RUNTIME_ASSERT(false, log, "Result {} is invalid", magic_enum::enum_name(res));
+        }
+    }
+private:
+    MPMCQueue<T> send_queue;
```
gengliqi, line 140: Should rename to `recv_queue`.

gengliqi (commented):
Should add a test like 'gtest_grpc_send_queue'.

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,163 @@
...
+        {
+        case MPMCQueueResult::OK:
+            return GRPCReceiveQueueRes::OK;
+        case MPMCQueueResult::FINISHED:
+            return GRPCReceiveQueueRes::FINISHED;
+        case MPMCQueueResult::CANCELLED:
+            return GRPCReceiveQueueRes::CANCELLED;
+        case MPMCQueueResult::FULL:
+        {
+            // If empty, change status to WAITING.
```
gengliqi, line 147:
```suggestion
            // If full, change status to WAITING.
```

dbms/src/Flash/Mpp/GRPCQueue.h
```diff
@@ -0,0 +1,128 @@
...
+
+private:
+    /// `action` is called before the `tag` is popped from completion queue
+    /// in `FinalizeResult`.
+    std::function<void *()> action;
+};
+
+using GRPCKickFunc = std::function<grpc_call_error(KickTag *)>;
+
+class GRPCQueue
```
gengliqi, line 46:
```suggestion
/// The grpc cpp framework provides a tool named `Alarm` can be used to push a tag into
/// completion queue thus the write can be done in grpc threads. But `Alarm` must need
/// a timeout and it uses a timer to trigger the notification, which is wasteful if we want
/// to trigger it immediately. So we can say `kickCompletionQueue` function is a
/// immediately-triggered `Alarm`.
class GRPCQueue
```

dbms/src/Flash/Mpp/GRPCSendQueue.h
```diff
@@ -74,29 +52,18 @@ enum class GRPCSendQueueRes
 /// to trigger it immediately. So we can say `kickCompletionQueue` function is a
 /// immediately-triggered `Alarm`.
```
gengliqi, line 53: Remove the comments `/// The grpc cpp framework provides...`.

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,163 @@
...
+        return ret;
+    }
+
+    // Push data into the queue.
+    //
+    // Return FULL if the queue is full and `new_tag` is saved.
+    // When the next pop/finish is called, the `new_tag` will be pushed
+    // into grpc completion queue.
+    template <typename U>
+    GRPCReceiveQueueRes push(U && data, void * new_tag)
```
gengliqi, line 111: I think it can not meet our requirements.
It's possible that several grpc threads call `push` to the same `GRPCReceiveQueue.h`. But here only can save one tag. So it can not handle this case.

xzhangxian1008: > I think it can not meet our requirements. It's possible that several grpc threads call `push` to the same `GRPCReceiveQueue.h`. But here only can save one tag. So it can not handle this case.

ok, updated

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,235 @@
...
+        case MPMCQueueResult::CANCELLED:
+            return GRPCReceiveQueueRes::CANCELLED;
+        case MPMCQueueResult::FULL:
+            // Handle this case later.
+            break;
+        default:
+            RUNTIME_ASSERT(false, log, "Result {} is invalid", static_cast<Int32>(res));
+        }
+
+        std::lock_guard lock(mu);
```
SeaRise, line 165: I prefer to refine the code like 
```
MPMCQueueResult res;
for (size_t i = 0; i < 2; ++i)
{
    res = recv_queue.tryPush(data);
    switch (res)
    {
    case MPMCQueueResult::OK:
        return GRPCReceiveQueueRes::OK;
    case MPMCQueueResult::FINISHED:
        return GRPCReceiveQueueRes::FINISHED;
    case MPMCQueueResult::CANCELLED:
        return GRPCReceiveQueueRes::CANCELLED;
    case MPMCQueueResult::FULL:
        break;
    default:
        RUNTIME_ASSERT(false, log, "Result {} is invalid", static_cast<Int32>(res));
    }
}
assert(res == MPMCQueueResult::FULL);
{
    std::lock_guard lock(mu);
    tags.push_back(new_tag);
}
```

SeaRise: And it seems that lock_guard only used for tags

xzhangxian1008: > And it seems that lock_guard only used for tags

The second tryPush must be contained in the lock.

If we don't contain the second tryPush in the lock, the following bug will happen:
Thread1: tryPush fail at first.
Thread1: tryPush fail at second.
Thread2: pop successfully and find that it's needless to kick CompletionQueue.
Thread1: lock and set tags. However, the pop of the Thread2 is the last pop operation, no more pop operation will be executed. So the tag will not be kicked into completion queue forever.

SeaRise: ok, refine the code like
```
MPMCQueueResult res = recv_queue.tryPush(data);
if (res == MPMCQueueResult::FULL)
{
    // add some comment about why add lock where.
    std::lock_guard lock(mu);
	res = recv_queue.tryPush(data);
    if (res == MPMCQueueResult::FULL)
	    tags.push_back(new_tag);
}
switch (res)
{
case MPMCQueueResult::OK:
    return GRPCReceiveQueueRes::OK;
case MPMCQueueResult::FINISHED:
    return GRPCReceiveQueueRes::FINISHED;
case MPMCQueueResult::CANCELLED:
    return GRPCReceiveQueueRes::CANCELLED;
case MPMCQueueResult::FULL:
    return GRPCReceiveQueueRes::FULL;
default:
    RUNTIME_ASSERT(false, log, "Result {} is invalid", magic_enum::enum_name(res);
}
```
we can use magic_enum::enum_name to print enum, ref https://github.com/pingcap/tiflash/issues/5758

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,235 @@
...
+            if (tags.empty())
+                return;
+
+            transferTagToKickReceiveTagWithNoLock();
+        }
+
+        callKickFunc();
+    }
+
+    void handleTheRemainingTags()
```
windtalker, line 204: This function will put the same tag to completion queue multiple times at the same time, is it a well defined behavior?

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,235 @@
...
+        while (!tags.empty())
+        {
+            transferTagToKickReceiveTagWithNoLock();
+            callKickFunc();
+        }
+    }
+
+    void transferTagToKickReceiveTagWithNoLock()
+    {
+        kick_recv_tag.pushTag(tags.front());
```
windtalker, line 216: Why not just put the tag to `kick_recv_tag` directly in `push()`?

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
xzhangxian1008: > ok, refine the code like
> 
> ```
> MPMCQueueResult res = recv_queue.tryPush(data);
> if (res == MPMCQueueResult::FULL)
> {
>     // add some comment about why add lock where.
>     std::lock_guard lock(mu);
>     res = recv_queue.tryPush(data);
>     if (res == MPMCQueueResult::FULL)
> 	tags.push_back(new_tag);
> }
> switch (res)
> {
> case MPMCQueueResult::OK:
>     return GRPCReceiveQueueRes::OK;
> case MPMCQueueResult::FINISHED:
>     return GRPCReceiveQueueRes::FINISHED;
> case MPMCQueueResult::CANCELLED:
>     return GRPCReceiveQueueRes::CANCELLED;
> case MPMCQueueResult::FULL:
>     return GRPCReceiveQueueRes::FULL;
> default:
>     RUNTIME_ASSERT(false, log, "Result {} is invalid", magic_enum::enum_name(res);
> }
> ```
> 
> we can use magic_enum::enum_name to print enum, ref #5758

okk

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
xzhangxian1008: > This function will put the same tag to completion queue multiple times at the same time, is it a well defined behavior?

No, each time we push a tag to KickReceiveTag, the same tag will be erased in GRPCReceiveQueue, so we can ensure that each tag can be pushed into completion queue only once.

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
xzhangxian1008: > Why not just put the tag to `kick_recv_tag` directly in `push()`?

Because we should ensure that each tag can be pushed into completion queue only once.

PR #6556, opened 2022-12-28
Fix big endian issue for StringHashTable

### LINKED ISSUE #6555: StringHashTable only works correctly on little-endian arch
## Bug Report
Current StringHashTable doesn't work correctly on big-endian arch. The problem might occur when aggregate keys' number is 1, and the key is non-nullable string.

yibin87:
Signed-off-by: yibin <huyibin@pingcap.com>

Issue Number: close #6555

Problem Summary:

### Check List

Tests 
Simple fix, no tests.
- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/IO/Endian.h
```diff
@@ -18,6 +18,11 @@
 
 namespace DB
 {
+inline constexpr bool isLittleEndian()
```
bestwoody, line 21: is `constexpr` means it's decided at comiler time?

yibin87: Yes

bestwoody: however, ref to doc, I found `native` is determined at execution environment

bestwoody: The value of order::native shall be:

equal to order::big if the **execution environment** is big endian, otherwise

equal to order::little if the **execution environment** is little endian, otherwise

unequal to both order::little and order::big.

yibin87: I'll check it.

yibin87: From contrib/boost/boost/endian/detail/order.hpp, it looks like the native value can be initialized to either big or little from compilation environments.

bestwoody: ok

yibin87: Oh, there is a third possibility: __ORDER_PDP_ENDIAN__, seems mixture of little and big, which I think we'd not support now.

dbms/src/Common/HashTable/StringHashTable.h
```diff
@@ -296,13 +295,19 @@ class StringHashTable : private boost::noncopyable
             if ((reinterpret_cast<uintptr_t>(p) & 2048) == 0)
             {
                 memcpy(&n[0], p, 8);
-                n[0] &= -1ul >> s;
+                if constexpr (DB::isLittleEndian())
+                    n[0] &= (-1ULL >> s);
```
bestwoody, line 299: it seem the old one use `-1ul`, while the new one use `-1ull`. is it expected?

yibin87: > 

Yes, it should be 8 bytes integer.

bestwoody (approved):
lgtm

dbms/src/Common/HashTable/StringHashTable.h
```diff
@@ -296,13 +295,19 @@ class StringHashTable : private boost::noncopyable
...
+                else
+                    n[0] &= (-1ULL << s);
             }
             else
             {
                 const char * lp = x.data + x.size - 8;
                 memcpy(&n[0], lp, 8);
-                n[0] >>= s;
+                if constexpr (DB::isLittleEndian())
```
mengxin9014, line 307: Whether the same fix is needed in TwoLevelStringHashTable?

yibin87: Good catch, I'll apply it also.

mengxin9014 (approved):
LGTM

ywqzzy (approved):

ywqzzy:
Do we need to check the runtime endian is exact the same as the compilation endian?

Issue #6533, opened 2022-12-23
merge join probe return block if need


mengxin9014:
## Enhancement
Currently, the output block after joined  is close to max_block_size, but in some cases that contain other condition，the final output block may be filtered to a very small size. So we should merge final output blocks again if they are too small.

PR #6565, opened 2022-12-30
check structure when merging blocks

### LINKED ISSUE #6564: check structure when merging blocks
## Enhancement
When merging blocks, we need to check their structure is the same.

mengxin9014:

Issue Number: close #6564

Problem Summary:

add structure check when merging blocks.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Core/Block.cpp
+++ b/dbms/src/Core/Block.cpp
@@ -543,7 +543,7 @@ Block mergeBlocks(Blocks && blocks)
 
     for (size_t i = 1; i < blocks.size(); ++i)
     {
-        if (likely(blocks[i].rows()) > 0)
+        if (likely(blocks[i].rows() > 0))
         {
             if (unlikely(!blocksHaveEqualStructure(first_block, blocks[i])))
             {
--- a/dbms/src/Core/Block.cpp
+++ b/dbms/src/Core/Block.cpp
@@ -545,7 +545,7 @@ Block mergeBlocks(Blocks && blocks)
     {
         if (likely(blocks[i].rows() > 0))
         {
-            RUNTIME_CHECK_MSG(blocksHaveEqualStructure(first_block, blocks[i]), "Cannot merge different structure blocks.");
+            assert(blocksHaveEqualStructure(first_block, blocks[i]));
             for (size_t column = 0; column < blocks[i].columns(); ++column)
             {
                 dst_columns[column]->insertRangeFrom(*blocks[i].getByPosition(column).column, 0, blocks[i].rows());
--- a/dbms/src/Core/Block.cpp
+++ b/dbms/src/Core/Block.cpp
@@ -545,6 +545,10 @@ Block mergeBlocks(Blocks && blocks)
     {
         if (likely(blocks[i].rows()) > 0)
         {
+            if (unlikely(!blocksHaveEqualStructure(first_block, blocks[i])))
+            {
+                throw Exception("Cannot merge different structure blocks.", ErrorCodes::LOGICAL_ERROR);
+            }
             for (size_t column = 0; column < blocks[i].columns(); ++column)
             {
                 dst_columns[column]->insertRangeFrom(*blocks[i].getByPosition(column).column, 0, blocks[i].rows());
--- a/dbms/src/Core/Block.cpp
+++ b/dbms/src/Core/Block.cpp
@@ -545,10 +545,7 @@ Block mergeBlocks(Blocks && blocks)
     {
         if (likely(blocks[i].rows() > 0))
         {
-            if (unlikely(!blocksHaveEqualStructure(first_block, blocks[i])))
-            {
-                throw Exception("Cannot merge different structure blocks.", ErrorCodes::LOGICAL_ERROR);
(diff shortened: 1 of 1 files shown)


dbms/src/Core/Block.cpp
```diff
@@ -543,8 +543,12 @@ Block mergeBlocks(Blocks && blocks)
 
     for (size_t i = 1; i < blocks.size(); ++i)
     {
-        if (likely(blocks[i].rows()) > 0)
+        if (likely(blocks[i].rows() > 0))
         {
+            if (unlikely(!blocksHaveEqualStructure(first_block, blocks[i])))
+            {
+                throw Exception("Cannot merge different structure blocks.", ErrorCodes::LOGICAL_ERROR);
+            }
```
SeaRise, line 551: Maybe we just need to check under debug mode?
```suggestion
            assert(blocksHaveEqualStructure(first_block, blocks[i]));
```

SeaRise: Or use RUNTIME_CHECK
```suggestion
            RUNTIME_CHECK(blocksHaveEqualStructure(first_block, blocks[i]), "Cannot merge different structure blocks.");
```

yibin87 (approved):
LGTM

dbms/src/Core/Block.cpp
mengxin9014: > Or use RUNTIME_CHECK

OK.

dbms/src/Core/Block.cpp
```diff
@@ -543,8 +543,12 @@ Block mergeBlocks(Blocks && blocks)
 
     for (size_t i = 1; i < blocks.size(); ++i)
     {
-        if (likely(blocks[i].rows()) > 0)
+        if (likely(blocks[i].rows() > 0))
         {
+            if (unlikely(!blocksHaveEqualStructure(first_block, blocks[i])))
```
bestwoody, line 548: why the same block case will be encountered? Are there some expected cases that it needs merge identical blocks?

xzhangxian1008: maybe we can use `RUNTIME_CHECK_MSG`

mengxin9014: > why the same block case will be encountered? Are there some expected cases that it needs merge identical blocks?

This is for protection, anyone can call mergeBlocks, if there is a different structure in the input blocks, the result will be problematic.  Its function is  similar here https://github.com/pingcap/tiflash/blob/fececd9375bd3b5de2318788bb08fa84e7c5deed/dbms/src/Interpreters/Aggregator.cpp#L2042

dbms/src/Core/Block.cpp
```diff
@@ -543,8 +543,9 @@ Block mergeBlocks(Blocks && blocks)
 
     for (size_t i = 1; i < blocks.size(); ++i)
     {
-        if (likely(blocks[i].rows()) > 0)
+        if (likely(blocks[i].rows() > 0))
         {
+            RUNTIME_CHECK_MSG(blocksHaveEqualStructure(first_block, blocks[i]), "Cannot merge different structure blocks.");
```
xzhangxian1008, line 548: Why we need to check structure? What will happen when two blocks' structure are different?

dbms/src/Core/Block.cpp
mengxin9014: > Or use RUNTIME_CHECK

done.

dbms/src/Core/Block.cpp
mengxin9014: > why the same block case will be encountered? Are there some expected cases that it needs merge identical blocks?

merging identical blocks you can refer https://github.com/pingcap/tiflash/issues/6533

dbms/src/Core/Block.cpp
mengxin9014: > Why we need to check structure? What will happen when two blocks' structure are different?

This is for protection, anyone can call mergeBlocks, if there is a different structure in the input blocks, the result will be problematic.  Its function is  similar here https://github.com/pingcap/tiflash/blob/fececd9375bd3b5de2318788bb08fa84e7c5deed/dbms/src/Interpreters/Aggregator.cpp#L2042

mengxin9014: > Why we need to check structure? What will happen when two blocks' structure are different?

if the column of block[0] is nullable, block[1] is not nullable.  the merging will transfer the column of block[1] to nullable and insert to the column block[0]. You can't be sure if the column we need will be nullable or not.

dbms/src/Core/Block.cpp
mengxin9014: > Maybe we just need to check under debug mode?

done.

SeaRise (approved):
lgtm

xzhangxian1008:
lgtm

bestwoody (approved):
lgtm

PR #6623, opened 2023-01-11
[WIP]Least string

ywqzzy:

Issue Number: close #xxx

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Functions/LeastGreatest.h
+++ b/dbms/src/Functions/LeastGreatest.h
@@ -124,7 +124,7 @@ private:
 };
 struct LeastStringImpl
 {
-    static void processImplWithCollator(
+    static void processImpl(
         const TiDB::TiDBCollatorPtr & collator,
         size_t a_size,
         size_t b_size,
@@ -133,20 +133,30 @@ struct LeastStringImpl
         ColumnString::Chars_t & c_data,
         ColumnString::Offsets & c_offsets)
     {
-        std::cout << "process with collator" << std::endl;
-        int res;
+        int res = 0;
         StringRef s_ra(&a_data[0], a_size);
         StringRef s_rb(&b_data[0], b_size);
-        res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
-        if (res < 0)
+        if (!collator)
         {
-            std::cout << "res < 0" << std::endl;
-            memcpy(&c_data[c_offsets.back()], &a_data[0], a_size);
-            c_offsets.push_back(c_offsets.back() + a_size + 1);
+            res = memcmp(&a_data[0], &b_data[0], std::min(a_size, b_size));
         }
         else
         {
-            std::cout << "res > 0" << std::endl;
+            res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
+        }
+        if (res < 0)
+        {
+            memcpy(&c_data[c_offsets.back()], &a_data[0], a_size);
+            c_offsets.push_back(c_offsets.back() + a_size + 1);
+        }
+        else if (res == 0)
+        {
+            size_t size = std::min(a_size, b_size);
+            memcpy(&c_data[c_offsets.back()], &b_data[0], size);
+            c_offsets.push_back(c_offsets.back() + size + 1);
+        }
+        else
+        {
             memcpy(&c_data[c_offsets.back()], &b_data[0], b_size);
             c_offsets.push_back(c_offsets.back() + b_size + 1);
         }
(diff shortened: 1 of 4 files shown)


dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,289 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+                fmt::format("Number of arguments for function {} doesn't match: passed {}, should be at least 2.", getName(), arguments.size()),
+                ErrorCodes::NUMBER_OF_ARGUMENTS_DOESNT_MATCH);
+        }
+        using impl = StringOperationImpl;
+        const auto * c0 = block.getByPosition(arguments[0]).column.get();
+        auto c_res = ColumnString::create();
+
+        for (size_t i = 1; i < num_arguments; ++i)
+        {
+            c_res = ColumnString::create();
```
xzhangxian1008, line 357: Instead of repeatedly creating result column, I think we can store the result directly in the result column.
Start: create result column and assign it to `res`
Step 1:  `c0` vs `c1`, store result in `res`
Step 2: `res` vs `c2`, more specifically, `res[0]` vs `c2[0]` store result in `res[0]`,  ... , `res[n]` vs `c2[n]` store result in `res[n]`. 
.....
Step n; `res` vs `cn`, store data in res.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,289 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+            const auto * const c1 = block.getByPosition(arguments[i]).column.get();
+            const auto * c0_string = checkAndGetColumn<ColumnString>(c0);
+            const auto * c1_string = checkAndGetColumn<ColumnString>(c1);
+            const ColumnConst * c0_const = checkAndGetColumnConstStringOrFixedString(c0);
+            const ColumnConst * c1_const = checkAndGetColumnConstStringOrFixedString(c1);
+            if (c0_const && c1_const)
+            {
+                String res;
+                impl::constantConstant(c0_const->getValue<String>(), c1_const->getValue<String>(), collator, res);
+                block.getByPosition(result).column = block.getByPosition(result).type->createColumnConst(c0_const->size(), toField(res));
```
xzhangxian1008, line 371: I think we should scan all the columns beforing processing so that we can know if res is const column at the beginning.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,289 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const ColumnString::Offsets & a_offsets,
+        const ColumnString::Chars_t & b_data,
+        const ColumnString::Offsets & b_offsets,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.reserve(std::max(a_data.size(), b_data.size()));
+        for (size_t i = 0; i < size; ++i)
```
xzhangxian1008, line 258: I think we should put the `i == 0` case out of the for-loop so that we can avoid `if` calse in `process` and make compiler tend to generate vectorization codes.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,289 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const ColumnString::Chars_t & a_data,
+        const ColumnString::Offsets & a_offsets,
+        const ColumnString::Chars_t & b_data,
+        const ColumnString::Offsets & b_offsets,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.reserve(std::max(a_data.size(), b_data.size()));
```
xzhangxian1008, line 257: Here maybe a bug. Reserve is indeed necessary, but we need to resize the container before using it.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,289 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+    static void NO_INLINE stringVectorConstant(
+        const ColumnString::Chars_t & a_data,
+        const ColumnString::Offsets & a_offsets,
+        const std::string & b,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.reserve(std::max(a_data.size(), b.size() * size));
```
xzhangxian1008, line 273: ditto

dbms/src/Functions/LeastGreatest.h
ywqzzy: Nice catch

SeaRise (approved):

gengliqi (approved):
LGTM

PR #6645, opened 2023-01-17
Use ${thread_name}_${thread_id} as RWLock query_id

### LINKED ISSUE #6631: Use thread_id in RWLock instead of thread name
## Enhancement
https://github.com/pingcap/tiflash/blob/9b8503bca9665b66ffcfef39f3c406dcf560c709/dbms/src/Storages/Transaction/PartitionStreams.cpp#L83 use thread name to get RWLock. But the thread name of different threads may be same, because the max length of thread name is 16, but some proxy thread name has exceeded 16.
image
So the thread names of different threads obtained through `getThreadName` may be the same.
I think we should use thread_id in RWLock instead of thread name to avoid exception when get RWlock.

hehechen:
Signed-off-by: hehechen <awd123456sss@gmail.com>

Issue Number: close #6631

Problem Summary:

Use `${thread_name}_${thread_id}` as RWLock query_id to avoid duplicated thread name caused by the thread name's 16 char limited.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CalvinNeo (approved):
lgtm

JaySon-Huang (approved):

PR #6649, opened 2023-01-17
Collect request unit for tiflash compute

### LINKED ISSUE #6678: Collect cpu time second for tiflash compute

SeaRise:

Issue Number: close #6678

Problem Summary:

- add conf `cluster.cluster_id` and held by InstanceLabelHolder
- add metrics `tiflash_compute_cpu_time_second` and add labels `cluster_id`, `process_id`; `process_id = compute_${host_name}_${start_unix_time}`
- add method `collectRequestUnit` for `QueryExecutor` and collect cpu time second for mpp, cop and batchCop
- add method `collectCPUTimeNs` for `IBlockInputStream`
- add `toRU` to convert cpu_time_ns to request unit

SvbAUbrSIe

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -410,6 +410,9 @@ void MPPTask::runImpl()
         {
             err_msg = result.err_msg;
         }
+        auto cpu_time = query_executor_holder->collectCPUTime();
+        LOG_INFO(log, "finish with cpu time: {} [cpu core num * ns]", cpu_time);
```
SeaRise, line 414: use ns, ms or s?

SeaRise:
YK1YhNupzP

dbms/src/Common/InstanceLabelHolder.cpp
```diff
@@ -0,0 +1,58 @@
...
+auto microsecondsUTC()
+{
+    return std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+}
+} // namespace
+
+void InstanceLabelHolder::init(const Poco::Util::LayeredConfiguration & conf)
+{
+    std::lock_guard lock(mu);
+    RUNTIME_ASSERT(!label_got, "Can't init after label_got got");
```
SeaRise, line 34:
```suggestion
    RUNTIME_ASSERT(!label_got, "Can't init after label got");
```

dbms/src/Server/MetricsPrometheus.cpp
```diff
@@ -181,19 +195,12 @@ MetricsPrometheus::MetricsPrometheus(
...
-            std::replace(job_name.begin(), job_name.end(), '.', '_');
-            job_name = "tiflash_" + job_name;
-
-            char hostname[1024];
-            ::gethostname(hostname, sizeof(hostname));
-
-            gateway = std::make_shared<prometheus::Gateway>(host, port, job_name, prometheus::Gateway::GetInstanceLabel(hostname));
+            const String & job_name = "tiflash";
+            const auto & labels = prometheus::Gateway::GetInstanceLabel(conf.getString("flash.service_addr", getHostName()));
+            gateway = std::make_shared<prometheus::Gateway>(host, port, job_name, labels);
```
SeaRise, line 200: Make labels `job_name` and `instance` consistent with pull model

dbms/src/DataStreams/UnionBlockInputStream.h
```diff
@@ -255,6 +255,16 @@ class UnionBlockInputStream final : public IProfilingBlockInputStream
             children[i]->readSuffix();
     }
 
+    uint64_t collectCPUTimeNsImpl(bool /*is_root*/) override
+    {
+        uint64_t cpu_time_ns = 0;
```
windtalker, line 260: Why ignore `is_root` here?

dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp
```diff
@@ -235,4 +235,14 @@ void ParallelAggregatingBlockInputStream::appendInfo(FmtBuffer & buffer) const
     buffer.fmtAppend(", max_threads: {}, final: {}", max_threads, final ? "true" : "false");
 }
 
+uint64_t ParallelAggregatingBlockInputStream::collectCPUTimeNsImpl(bool is_root)
+{
+    uint64_t cpu_time_ns = impl ? impl->collectCPUTimeNs(is_root) : 0;
+    forEachChild([&](IBlockInputStream & child) {
+        cpu_time_ns += child.collectCPUTimeNs(true);
```
windtalker, line 242: So the cpu time is defined as the sum of execution time of all threads?

dbms/src/DataStreams/SharedQueryBlockInputStream.h
```diff
@@ -183,6 +183,16 @@ class SharedQueryBlockInputStream : public IProfilingBlockInputStream
             throw Exception(exception_msg);
     }
 
+    uint64_t collectCPUTimeNsImpl(bool /*is_root*/) override
+    {
+        uint64_t cpu_time_ns = 0;
```
windtalker, line 188: `is_root` is ignored because `SharedQueryBlockInputStream` should never be a root?

dbms/src/Flash/Executor/toCPUTimeSecond.cpp
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Flash/Executor/toCPUTimeSecond.h>
+#include <common/likely.h>
+
+namespace DB
+{
+/**
+ * cpu time second used by tiflash compute = vcores * second
```
windtalker, line 21: The implementation does not use `vcores`?

dbms/src/Server/MetricsPrometheus.cpp
```diff
@@ -181,19 +195,12 @@ MetricsPrometheus::MetricsPrometheus(
...
-            std::replace(job_name.begin(), job_name.end(), ':', '_');
-            std::replace(job_name.begin(), job_name.end(), '.', '_');
-            job_name = "tiflash_" + job_name;
-
-            char hostname[1024];
-            ::gethostname(hostname, sizeof(hostname));
-
-            gateway = std::make_shared<prometheus::Gateway>(host, port, job_name, prometheus::Gateway::GetInstanceLabel(hostname));
+            const String & job_name = "tiflash";
+            const auto & labels = prometheus::Gateway::GetInstanceLabel(conf.getString("flash.service_addr", getHostName()));
```
windtalker, line 199: What's the difference between the instance label here and the instance label in `InstanceLabelHolder`

dbms/src/DataStreams/UnionBlockInputStream.h
SeaRise: `is_root` is a bad name...
I had renamed to `is_thread_runner` and add some comments here.
```
    /** Estimate the cpu time nanoseconds used by block input stream dag.
      * In this method, streams are divided into two categories:
      * - thread-runner: Called directly by a thread
      * - non-thread-runner: Called by a thread-runner
      * Here we should count the execution time of each thread-runner.
      * Note: Because more threads than vcore, and blocking relationships between streams,
      * the result may not be 100% identical to the actual cpu time nanoseconds.
      */

        // UnionBlockInputStream isn't a thread-runner, so is_thread_runner is ignored here.
        // But each of UnionBlockInputStream's children is a thread-runner.
```

dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp
SeaRise: Yes, comments added.
```
    /** Estimate the cpu time nanoseconds used by block input stream dag.
      * In this method, streams are divided into two categories:
      * - thread-runner: Called directly by a thread
      * - non-thread-runner: Called by a thread-runner
      * Here we should count the execution time of each thread-runner.
      * Note: Because more threads than vcore, and blocking relationships between streams,
      * the result may not be 100% identical to the actual cpu time nanoseconds.
      */
```

dbms/src/DataStreams/SharedQueryBlockInputStream.h
SeaRise: comments added
```
        // SharedQueryBlockInputStream isn't a thread-runner, so is_thread_runner is ignored here.
        // But each of SharedQueryBlockInputStream's children is a thread-runner.
```

dbms/src/Flash/Executor/toCPUTimeSecond.cpp
SeaRise: bad comment removed...
New comment added: `// Convert cpu time nanoseconds to cpu time seconds, and round up.`

dbms/src/Server/MetricsPrometheus.cpp
SeaRise: `InstanceLabelHolder` and `InstanceId` are both bad name...
InstanceLabelHolder has been renamed to `ComputeLabelHolder` and `ComputeLabelHolder::instance_id` has been renamed to `ComputeLabelHolder::process_id`.

SeaRise: The `instance` label here is mean that the host_name of tiflash compute node.
The `process_id` label in `ComputeLabelHolder` is used to distinguish between processes that have been started multiple times.

dbms/src/Common/TiFlashMetrics.h
```diff
@@ -96,182 +97,186 @@ namespace DB
...
+    M(tiflash_exchange_queueing_data_bytes, "Total bytes of data contained in the queue", Gauge,                                                    \
+        F(type_send, {{"type", "send_queue"}}),                                                                                                     \
+        F(type_receive, {{"type", "recv_queue"}}))                                                                                                  \
+    M(tiflash_compute_cpu_time_second, "cpu time second used by tiflash compute", Counter,                                                          \
+        F(type_mpp, {{"type", "mpp"}, ComputeLabelHolder::instance().getClusterIdLabel(), ComputeLabelHolder::instance().getProcessIdLabel()}),     \
+        F(type_cop, {{"type", "cop"}, ComputeLabelHolder::instance().getClusterIdLabel(), ComputeLabelHolder::instance().getProcessIdLabel()}),     \
+        F(type_batch, {{"type", "batch"}, ComputeLabelHolder::instance().getClusterIdLabel(), ComputeLabelHolder::instance().getProcessIdLabel()})) \
```
SeaRise, line 279: new metrics added here, others are fmt and not changed.

bestwoody:
What's the differences between `cluster_id` and `instance_id`? Could you explain?

SeaRise:
> What's the differences between `cluster_id` and `instance_id`? Could you explain?

`instance_id` has been renamed to `process_Id`, comments as follows has been added in codes
```
    // the id of tiflash compute cluster
    std::string cluster_id{"unknown"};
    // the id of tiflash compute process, used to distinguish between processes that have been started multiple times.
    // Format: `compute_${host_name}_${start_time_second_utc}`
    std::string process_id{"unknown"};
```

dbms/src/DataStreams/UnionBlockInputStream.h
bestwoody: If `is_root` is equivalent to `is_thread_runner`? if true, we should add comment about it.

dbms/src/Server/MetricsPrometheus.cpp
```diff
@@ -181,19 +195,12 @@ MetricsPrometheus::MetricsPrometheus(
...
-            std::string job_name = service_addr;
-            std::replace(job_name.begin(), job_name.end(), ':', '_');
-            std::replace(job_name.begin(), job_name.end(), '.', '_');
-            job_name = "tiflash_" + job_name;
-
-            char hostname[1024];
-            ::gethostname(hostname, sizeof(hostname));
-
-            gateway = std::make_shared<prometheus::Gateway>(host, port, job_name, prometheus::Gateway::GetInstanceLabel(hostname));
+            const String & job_name = "tiflash";
```
bestwoody, line 198: tiflash_compute?

dbms/src/Server/MetricsPrometheus.cpp
bestwoody, line 199: should the `hostname` need be replaced `":"` and `"."` like what you did on prevous lines?

bestwoody, line 199: does it means that it will use as label getHostName as value if "flash.service_addr" is null?

bestwoody: the `flash.service_addr` won't be null, isn't it? what if service_addr is sth like "0.0.0.0:1234"

SeaRise: > does it means that it will use as label getHostName as value if "flash.service_addr" is null?
yes

> what if service_addr is sth like "0.0.0.0:1234"
good point!
In testing locally, the result of `getHostName` is `P99`, so `flash.service_addr` is used here.
Looks like we need to use `getHostName` when service_addr are `127.0.0.1` and `0.0.0.0`

dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
```diff
@@ -291,4 +291,16 @@ void CreatingSetsBlockInputStream::createOne(SubqueryForSet & subquery)
     }
 }
 
+uint64_t CreatingSetsBlockInputStream::collectCPUTimeNsImpl(bool /*is_thread_runner*/)
```
windtalker, line 294: `CreatingSetsBlockInputStream` is executed in `MPPTask::runImpl`, and I think we don't collect cpu time for that thread, is it the expected behavior?

dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
```diff
@@ -291,4 +291,16 @@ void CreatingSetsBlockInputStream::createOne(SubqueryForSet & subquery)
...
+{
+    // CreatingSetsBlockInputStream isn't a thread-runner, so is_thread_runner is ignored here.
+    // But each of CreatingSetsBlockInputStream's children is a thread-runner.
+    uint64_t cpu_time_ns = 0;
+    forEachChild([&](IBlockInputStream & child) {
+        cpu_time_ns += child.collectCPUTimeNs(true);
```
windtalker, line 300: Actually, the last child is not thread runner.

dbms/src/DataStreams/UnionBlockInputStream.h
SeaRise: > If `is_root` is equivalent to `is_thread_runner`? if true, we should add comment about it.

yes, comment `// The first stream of stream dag is thread-runner.` added.

dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
SeaRise: Yes, I don't think it is necessary to distinguish what thread is, mpptaskImpl threads are treated equally.
Usually the last child is a UnionBlockInputStream which doesn't count its own cpu time but its children's.
IMO, It depends on what stream last child is

dbms/src/DataStreams/CreatingSetsBlockInputStream.cpp
SeaRise: I think it depends on what stream the last child is, if it is a computational stream, it will also consume a lot of cpu on the `mpptaskimpl` thread.

dbms/src/Server/MetricsPrometheus.cpp
SeaRise: > > does it means that it will use as label getHostName as value if "flash.service_addr" is null?
> 
> yes
> 
> > what if service_addr is sth like "0.0.0.0:1234"
> 
> good point! In testing locally, the result of `getHostName` is `h99`, so `flash.service_addr` is used here. Looks like we need to use `getHostName` when service_addr are `127.0.0.1` and `0.0.0.0`

updated.

bestwoody (approved):

windtalker (approved):
LGTM

PR #6650, opened 2023-01-17
support spill to disk in join

### LINKED ISSUE #6907: support spill to disk in join
## Enhancement
Currently, the join executor in TiFlash does not support spilling to disk operation. If the memory used by the join executor exceeds the `max_memory_usage` parameter, the query will be cancelled, and even OOM may appear.
Therefore, the join executor in TiFlash needs to support spilling to disk operation to keep the memory used within a fixed value.

mengxin9014:

Issue Number: close #6907

Problem Summary:

In this feature, we dispatch the build table data to several partitions. When the memory used by join exceeds the limit, one partition is selected for spilling, until the join operation can be successfully completed on other partitions, and then the spilled partition will be restored and the previous operation is repeated.
### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Core/Spiller.cpp
```diff
@@ -211,8 +211,10 @@ BlockInputStreams Spiller::restoreBlocks(UInt64 partition_id, UInt64 max_stream_
         }
         for (UInt64 i = 0; i < spill_file_read_stream_num; ++i)
         {
-            if (likely(!file_infos[i].empty()))
+            if (likely(i < file_infos.size() && !file_infos[i].empty()))
```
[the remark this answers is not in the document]
mengxin9014: Old hack code, restored.

dbms/src/Interpreters/Settings.h
```diff
@@ -294,17 +294,17 @@ struct Settings
...
     M(SettingUInt64, manual_compact_pool_size, 1, "The number of worker threads to handle manual compact requests.")                                                                                                                    \
     M(SettingUInt64, manual_compact_max_concurrency, 10, "Max concurrent tasks. It should be larger than pool size.")                                                                                                                   \
     M(SettingUInt64, manual_compact_more_until_ms, 60000, "Continuously compact more segments until reaching specified elapsed time. If 0 is specified, only one segment will be compacted each round.")                                \
-    M(SettingUInt64, max_cached_data_bytes_in_spiller, 1024ULL * 1024 * 100, "Max cached data bytes in spiller before spilling, 100MB as the default value, 0 means no limit")                                                              \
+    M(SettingUInt64, max_join_bytes, 0, "max bytes used by join")                                                                                                                                                                       \
```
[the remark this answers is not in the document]
mengxin9014: changed.

dbms/src/Interpreters/Join.h
```diff
@@ -172,10 +255,18 @@ class Join
...
-    void finishOneProbe();
+
+    void finishOneBuild();
+    void waitUntilAllBuildFinished() const;
+
+    void finishOneProbe(bool is_canceled = false);
     void waitUntilAllProbeFinished() const;
 
+    void finishOneNonJoin(bool is_canceled = false);
+    void waitUntilAllNonJoinFinished() const;
```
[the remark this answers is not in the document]
mengxin9014: > I think there is no need to wait all non join finish?
removed.

dbms/src/Common/HashTable/HashTable.h
```diff
@@ -1597,16 +1625,21 @@ class ConcurrentHashTable : private boost::noncopyable
         size_t ret = 0;
         for (size_t i = 0; i < segments.size(); i++)
             /// note the return value might not be accurate since it does not use lock, but should be enough for current usage
-            ret += segments[i]->getBufferSizeInBytes();
+            ret += segments[i] ? segments[i]->getBufferSizeInBytes() : 0;
```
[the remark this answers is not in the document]
mengxin9014: > can call `getSegmentBufferSizeInBytes` directly?

done.

dbms/src/DataStreams/NonJoinedBlockInputStream.cpp
```diff
@@ -243,33 +258,51 @@ size_t NonJoinedBlockInputStream::fillColumns(const Map & map,
...
+        return rows_added;
+    }
     /// use pointer instead of reference because `it` need to be re-assigned latter
     auto it = reinterpret_cast<typename Map::SegmentType::HashTable::const_iterator *>(position.get());
     auto end = map.getSegmentTable(current_segment).end();
 
     for (; *it != end || current_segment + step < map.getSegmentSize();)
     {
-        if (*it == end)
+        if (*it == end || (parent.max_bytes_before_external_join && parent.partitions[current_segment]->spill))
```
windtalker, line 286: The spill check is different with L262, I think we should keep the same style, and since `parent.partitions[current_segment]->spill` is false by default even if spill is not enabled, I think just use the check in L262 is enough.

dbms/src/Flash/tests/gtest_join_executor.cpp
```diff
@@ -738,10 +768,11 @@ try
     ColumnsWithTypeAndName left_column_data;
     ColumnsWithTypeAndName right_column_data;
     ColumnsWithTypeAndName common_column_data;
-    size_t table_rows = 102400;
-    size_t common_rows = 20480;
+    size_t table_rows = 51200;
```
windtalker, line 771: why change this

SeaRise:
> /run-sanitizer-test tsan

https://ci.pingcap.net/blue/organizations/jenkins/tiflash-sanitizer-daily/detail/tiflash-sanitizer-daily/936/pipeline

dbms/src/Flash/tests/gtest_join_executor.cpp
mengxin9014: > why change this

The tests are running too slowly，to prevent CI execution timeouts, halve the amount of data.

dbms/src/Flash/tests/gtest_join_executor.cpp
```diff
@@ -136,7 +136,13 @@ try
                                .build(context);
 
             {
+                context.context.setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(0)));
                 executeAndAssertColumnsEqual(request, expected_cols[i * simple_test_num + j]);
+
+                // for spill to disk tests
+                context.context.setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(10000)));
+                ASSERT_THROW(executeAndAssertColumnsEqual(request, expected_cols[i * simple_test_num + j], {1}), Exception);
+                executeAndAssertColumnsEqual(request, expected_cols[i * simple_test_num + j], {2, 5, 10});
```
SeaRise, line 145: Use `executeStreams` directly?
Because `executeAndAssertColumnsEqual` will run query at least three times: disable planner, enable planner, enable pipeline.

SeaRise: and then leave `executeAndAssertColumnsEqual ` as it is.

dbms/src/Flash/tests/gtest_join_executor.cpp
```diff
@@ -738,10 +768,11 @@ try
...
     ColumnsWithTypeAndName right_column_data;
     ColumnsWithTypeAndName common_column_data;
-    size_t table_rows = 102400;
-    size_t common_rows = 20480;
+    size_t table_rows = 51200;
+    size_t common_rows = 10240;
     UInt64 max_block_size = 800;
-    size_t original_max_streams = 20;
+    size_t max_streams_big = 20;
+    size_t max_streams_small = 4;
```
SeaRise, line 775: What is the difference between these two?
It looks like the following test doesn't respect this, using the table_concurrency_hint.

dbms/src/Interpreters/Settings.h
```diff
@@ -291,17 +291,17 @@ struct Settings
...
-    M(SettingUInt64, max_spilled_bytes_per_file, 0, "Max spilled data bytes per spill file, 1GB as the default value, 0 means no limit.")                                                                                               \
-                                                                                                                                                                                                                                        \
+    M(SettingUInt64, max_bytes_before_external_join, 0, "max bytes used by join before spill, 0 as the default value, 0 means no limit")                                                                                                                                                                       \
+    M(SettingInt64, join_restore_concurrency, 0, "join restore concurrency, negative value means restore join serially, 0 means TiFlash choose restore concurrency automatically, 0 as the default value")                            \
```
SeaRise, line 298: Is this configuration useful, it looks like the maximum is the concurrency of the join probe.
It doesn't look like the user is going to adjust this value.

dbms/CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ if (ENABLE_TESTS)
     )
     target_include_directories(gtests_dbms BEFORE PRIVATE ${SPARCEHASH_INCLUDE_DIR})
     target_compile_definitions(gtests_dbms PUBLIC DBMS_PUBLIC_GTEST)
+    target_compile_definitions(dbms PUBLIC DBMS_PUBLIC_GTEST)
```
SeaRise, line 306: why add it?

dbms/src/Common/FailPoint.cpp
```diff
@@ -139,8 +140,9 @@ namespace DB
     M(random_pipeline_model_event_schedule_failpoint)   \
     M(random_pipeline_model_event_finish_failpoint)     \
     M(random_pipeline_model_operator_run_failpoint)     \
-    M(random_pipeline_model_cancel_failpoint)
-
+    M(random_pipeline_model_cancel_failpoint)           \
+    M(random_join_spill_to_disk_failpoint)              \
+    M(random_join_restore_from_disk_failpoint)
```
SeaRise, line 145: how about replacing by `random_spill_to_disk_failpoint` and `random_restore_from_disk_failpoint`. and injecting fail point in `Spiller::spill` and `Spiller::restore`.
Then not just join, but all spill operators are injected with fail points.

dbms/src/DataStreams/HashJoinProbeBlockInputStream.cpp
```diff
@@ -41,34 +47,54 @@ Block HashJoinProbeBlockInputStream::getHeader() const
...
+    /// There is a little bit hack here because cancel will be called in two cases:
+    /// 1. the query is cancelled by the caller or meet error: in this case, wake up all waiting threads is safe
+    /// 2. the query is executed normally, and one of the data stream has read an empty block, the the data stream and all its children
+    ///    will call `cancel(false)`, in this case, there is two sub-cases
+    ///    a. the data stream read an empty block because of EOF, then it means there must be no threads waiting in Join, so cancel the join is safe
+    ///    b. the data stream read an empty block because of early exit of some executor(like limit), in this case, just wake the waiting
+    ///       threads is not 100% safe because if the probe thread is wake up when build is not finished yet, it may produce wrong results, for now
+    ///       it is safe because when any of the data stream read empty block because of early exit, the execution framework ensures that no further
+    ///       data will be used.
+    current_join->cancel();
```
SeaRise, line 82: How about just calling `current_join->cancel();` here?
And check `isCancelled()` in `while (restore_build_stream->read()) {};` and etc.

dbms/src/Interpreters/Join.h
```diff
@@ -94,19 +99,28 @@ class Join
          const String & req_id,
          bool enable_fine_grained_shuffle_,
          size_t fine_grained_shuffle_count_,
+         size_t max_bytes_before_external_join_,
+         const SpillConfig & build_spill_config_,
+         const SpillConfig & probe_spill_config_,
+         Int64 join_restore_concurrency_,
```
SeaRise, line 105: I prefer to remove `join_restore_concurrency` and then the logic of join restoring can be simplified.

dbms/src/Interpreters/Join.cpp
```diff
@@ -2138,9 +2496,415 @@ BlockInputStreamPtr Join::createStreamWithNonJoinedRows(const Block & left_sampl
...
+
+void Join::insertBlockToBuildPartition(Block && block, size_t partition_index)
+{
+    size_t rows = block.rows();
+    size_t bytes = block.bytes();
+    partitions[partition_index]->build_partition.rows += rows;
+    partitions[partition_index]->build_partition.bytes += bytes;
+    partitions[partition_index]->build_partition.blocks.push_back(block);
+    partitions[partition_index]->build_partition.original_blocks.push_back(std::move(block));
+    partitions[partition_index]->memory_usage += bytes;
```
SeaRise, line 2628: Wrap into `Partition` and just call `partitions[partition_index]->insertBlockForBuild`?

dbms/CMakeLists.txt
mengxin9014: > why add it?

it is for join executor gtests.  In join executor spilling  gtests, the amount of data is small, so it is difficult to simulate  spilling, with the compilation option controlling that the spilling only goes to the second round and only half the partition spilling to disk.

dbms/src/Common/FailPoint.cpp
mengxin9014: > how about replacing by `random_spill_to_disk_failpoint` and `random_restore_from_disk_failpoint`. and injecting fail point in `Spiller::spill` and `Spiller::restore`. Then not just join, but all spill operators are injected with fail points.

The purpose of this failPoint is to test the join executor， if injecting fail point in Spiller it is not possible to determine whether the problem is in the join or another executor when the failPoint test fails.

dbms/src/Flash/tests/gtest_join_executor.cpp
```diff
@@ -729,6 +735,30 @@ try
...
+                       .join(context.scan("split_test", "t2"), tipb::JoinType::TypeInnerJoin, {col("a")})
+                       .build(context);
+
+    auto join_restore_concurrences = {-1, 0, 1, 5};
+    const ColumnsWithTypeAndName expect = {toNullableVec<Int32>({1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7, 7, 8, 8, 8, 9, 9, 9, 0, 0, 0}), toNullableVec<Int32>({2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2}), toNullableVec<Int32>({1, 1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7, 7, 8, 8, 8, 9, 9, 9, 0, 0, 0})};
+    context.context.setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(10000)));
+    for (const auto & join_restore_concurrency : join_restore_concurrences)
+    {
+        context.context.setSetting("join_restore_concurrency", Field(static_cast<Int64>(join_restore_concurrency)));
+        ASSERT_THROW(executeAndAssertColumnsEqual(request, expect, {1}, {DEFAULT_BLOCK_SIZE}), Exception);
```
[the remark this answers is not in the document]
SeaRise: In this case, is it better to execute successfully and not spill?
Reporting errors may leave users wondering.

dbms/src/Common/FailPoint.cpp
SeaRise: random fail point test is the chaos test for random fault injection.
IMO, `random_spill_to_disk_failpoint` and `random_restore_from_disk_failpoint` is more suitable for random fail point test.
Not only join, but also other errors in the falling spill operator should be detected.

Finally, in case of a crash, you can determine from the stack log if it was caused by a join.
If it's a hang, then there's no way to tell if it's a join issue even if the fail point is only injected against a join.

dbms/CMakeLists.txt
SeaRise: Isn't it controlled by `context.context.setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(max_bytes_before_external_join)));` at runtime?
Where is this compile option used?

mengxin9014: > Isn't it controlled by `context.context.setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(max_bytes_before_external_join)));` at runtime? Where is this compile option used?

Join.cpp  L1157 and L2576

dbms/src/Interpreters/Settings.h
mengxin9014: > Is this configuration useful? it looks like the maximum is the concurrency of the join probe. It doesn't look like the user is going to adjust this value.

It controls restore partition concurrency, user will not adjust it.

dbms/src/Interpreters/Join.cpp
```diff
@@ -2138,9 +2496,415 @@ BlockInputStreamPtr Join::createStreamWithNonJoinedRows(const Block & left_sampl
...
+}
+
+void Join::insertBlockToProbePartition(Block && block, size_t partition_index)
+{
+    size_t rows = block.rows();
+    size_t bytes = block.bytes();
+    partitions[partition_index]->probe_partition.rows += rows;
+    partitions[partition_index]->probe_partition.bytes += bytes;
+    partitions[partition_index]->probe_partition.blocks.push_back(std::move(block));
+}
```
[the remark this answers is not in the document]
mengxin9014: > Wrap into Partition and just call `partitions[partition_index]->insertBlockForProbe`?

done.

dbms/src/Interpreters/Join.cpp
mengxin9014: > Wrap into `Partition` and just call `partitions[partition_index]->insertBlockForBuild`?

done.

dbms/src/Interpreters/Join.h
mengxin9014: > I prefer to remove `join_restore_concurrency` and then the logic of join restoring can be simplified.

it has already discussed in meeting.

dbms/src/Common/FailPoint.cpp
mengxin9014: done.

dbms/src/Flash/tests/gtest_join_executor.cpp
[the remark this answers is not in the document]
mengxin9014: > In this case, is it better to execute successfully and not spill? Reporting errors may leave users wondering.

change error description.

dbms/src/Common/HashTable/HashTable.h
```diff
@@ -1479,6 +1497,20 @@ class ConcurrentHashTable : private boost::noncopyable
         return segments[segment_index]->getHashTable();
     }
 
+    size_t resetSegmentTable(size_t segment_index)
+    {
+        size_t ret = 0;
+        std::unique_ptr<SegmentType> segment_ptr = nullptr;
+        {
+            /// release the lock before destruct related segment
```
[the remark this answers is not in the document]
mengxin9014: > `RuntimeCheck(segments[segment_index])` here?

I think add assert(segments[segment_index]) is ok.

dbms/src/Common/FailPoint.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <Common/Exception.h>
 #include <Common/FailPoint.h>
+#include <Poco/String.h>
```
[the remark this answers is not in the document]
mengxin9014: > 

removed.

dbms/src/DataStreams/NonJoinedBlockInputStream.cpp
mengxin9014: > The spill check is different with L262, I think we should keep the same style, and since `parent.partitions[current_segment]->spill` is false by default even if spill is not enabled, I think just use the check in L262 is enough.

done.

dbms/src/Flash/tests/gtest_join_executor.cpp
windtalker: `max_streams_small ` is used to test the case that fine_grained_stream_count bigger than max_stream, I will update the test to use `max_streams_small` only fine_grained_stream is enabled.

dbms/src/DataStreams/HashJoinProbeBlockInputStream.cpp
windtalker: You mean don't cancel the related stream? But if `HashJoinProbeBlockInputStream::cancel()` is called, there is no guarantee that `HashJoinProbeBlockInputStream::readImpl()` will be called again.

dbms/src/Flash/tests/gtest_join_executor.cpp
mengxin9014: done.

dbms/src/TestUtils/ExecutorTestUtils.h
```diff
@@ -128,7 +131,9 @@ class ExecutorTest : public ::testing::Test
 private:
     void executeExecutor(
         const std::shared_ptr<tipb::DAGRequest> & request,
-        std::function<::testing::AssertionResult(const ColumnsWithTypeAndName &)> assert_func);
+        std::function<::testing::AssertionResult(const ColumnsWithTypeAndName &)> assert_func,
+        std::vector<size_t> concurrencies = {1, 2, 10},
+        std::vector<size_t> block_sizes = {1, 2, DEFAULT_BLOCK_SIZE});
```
SeaRise, line 136: useless change here?

dbms/src/Flash/tests/gtest_join_executor.cpp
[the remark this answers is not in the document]
SeaRise: Does that mean exception wouldn't thrown here and just printing wanring log?

dbms/src/TestUtils/ExecutorTestUtils.h
mengxin9014: > useless change here?

done.

dbms/src/Interpreters/Join.h
```diff
@@ -400,18 +537,48 @@ class Join
...
+struct RestoreInfo
+{
+    JoinPtr join;
+    BlockInputStreamPtr non_joined_stream;
+    BlockInputStreamPtr build_stream;
+    BlockInputStreamPtr probe_stream;
+
+    RestoreInfo() = default;
+    RestoreInfo(JoinPtr join_, BlockInputStreamPtr non_joined_data_stream_, BlockInputStreamPtr build_stream_, BlockInputStreamPtr probe_stream_)
+        : join(std::move(join_))
```
windtalker, line 572: why use `std::move` here?

dbms/src/Interpreters/Join.h
```diff
@@ -271,6 +324,60 @@ class Join
...
+            memory_usage += bytes;
+        }
+
+        void insertBlockForProbe(Block && block)
+        {
+            size_t rows = block.rows();
+            size_t bytes = block.bytes();
+            probe_partition.rows += rows;
+            probe_partition.bytes += bytes;
+            probe_partition.blocks.push_back(std::move(block));
```
windtalker, line 372: Although not used in probe stage, but I think we should update the `memory_usage` here?

dbms/src/Interpreters/Join.h
```diff
@@ -154,17 +192,31 @@ class Join
...
     void waitUntilAllProbeFinished() const;
 
+    void finishOneNonJoin(size_t partition_index);
+
     size_t getBuildConcurrency() const
     {
-        if (unlikely(build_concurrency == 0))
-            throw Exception("Logical error: `setBuildConcurrencyAndInitPool` has not been called", ErrorCodes::LOGICAL_ERROR);
-        return build_concurrency;
+        std::shared_lock lock(rwlock);
```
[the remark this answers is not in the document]
windtalker: The master branch already remove `getBuildConcurrencyInternal` and rewrite `getBuildConcurrency` always return `build_concurrency` without acquire the lock, maybe we should not change it back?

dbms/src/DataStreams/HashJoinProbeBlockInputStream.cpp
```diff
@@ -41,34 +47,52 @@ Block HashJoinProbeBlockInputStream::getHeader() const
...
     /// When the probe stream quits probe by cancelling instead of normal finish, the Join operator might still produce meaningless blocks
     /// and expects these meaningless blocks won't be used to produce meaningful result.
-    try
+
+    JoinPtr current_join;
+    RestoreInfo restore_info;
     {
-        finishOneProbe();
+        std::lock_guard lock(mutex);
+        current_join = join;
```
windtalker, line 64: why not just construct the `restore_info` here?

dbms/src/DataStreams/HashJoinProbeBlockInputStream.cpp
```diff
@@ -94,56 +118,226 @@ void HashJoinProbeBlockInputStream::readSuffixImpl()
...
+    {
+        while (true)
+        {
+            switch (status)
+            {
+            case ProbeStatus::WAIT_BUILD_FINISH:
+                join->waitUntilAllBuildFinished();
+                /// after Build finish, always go to Probe stage
+                if (join->isRestoreJoin())
+                    current_probe_stream->readSuffix();
```
windtalker, line 241: Should be `current_probe_stream->readPrefix()`

dbms/src/Interpreters/Join.cpp
```diff
@@ -862,25 +1106,116 @@ void recordFilteredRows(const Block & block, const String & filter_column, Colum
...
     std::shared_lock lock(rwlock);
-    assert(stream_index < getBuildConcurrency());
+    assert(stream_index < getBuildConcurrencyInternal());
+    total_input_build_rows += block.rows();
 
     if (unlikely(!initialized))
         throw Exception("Logical error: Join was not initialized", ErrorCodes::LOGICAL_ERROR);
     Block * stored_block = nullptr;
+
+    //    LOG_INFO(log, "insert one block, enable spill {}, max_bytes_before_external_join {}, current bytes {}", isEnableSpill(), max_bytes_before_external_join, getTotalByteCount());
```
windtalker, line 1116: remove useless code

dbms/src/Interpreters/Join.cpp
```diff
@@ -862,25 +1106,116 @@ void recordFilteredRows(const Block & block, const String & filter_column, Colum
...
+            blocks.push_back(block);
+            stored_block = &blocks.back();
+            original_blocks.push_back(block);
+        }
+        insertFromBlockInternal(stored_block, stream_index);
     }
-    insertFromBlockInternal(stored_block, stream_index);
+    else
+    {
+        //        LOG_INFO(log, "enable spill, max_bytes_before_external_join {}, current bytes {}", max_bytes_before_external_join, getTotalByteCount());
```
windtalker, line 1129: Ditto

dbms/src/Interpreters/Join.cpp
```diff
@@ -862,25 +1106,116 @@ void recordFilteredRows(const Block & block, const String & filter_column, Colum
...
+            size_t i = j % build_concurrency;
+            if (!dispatch_blocks[i].rows())
+            {
+                continue;
+            }
+            Blocks blocks_to_spill;
+            {
+                const auto & join_partition = partitions[i];
+                std::unique_lock partition_lock(join_partition->partition_mutex);
+                //                insertBlockToBuildPartition(std::move(dispatch_blocks[i]), i);
```
windtalker, line 1171: Ditto

dbms/src/Interpreters/Join.cpp
```diff
@@ -2444,6 +2752,14 @@ void Join::finishOneBuild()
...
     {
+        if (isEnableSpill())
+        {
+            if (hasPartitionSpilled())
+            {
+                trySpillBuildPartitions(true);
+                tryMarkBuildSpillFinish();
+            }
+        }
         workAfterBuildFinish();
```
windtalker, line 2763: I think we can move the `trySpillxxx` into `workAfterBuildFinish`

dbms/src/Interpreters/Join.cpp
```diff
@@ -2463,30 +2779,70 @@ void Join::workAfterBuildFinish()
...
-    build_cv.wait(lock, [&]() {
-        return meet_error || active_build_concurrency == 0;
+    if (active_probe_concurrency == 1)
+    {
+        FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::exception_mpp_hash_probe);
+    }
+    --active_probe_concurrency;
+    if (active_probe_concurrency == 0)
+    {
+        if (isEnableSpill())
```
windtalker, line 2802: maybe add a new function `workAfterProbeFinish()`?

dbms/src/Interpreters/Join.h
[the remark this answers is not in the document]
mengxin9014: > getBuildConcurrencyInternal

done.

dbms/src/Interpreters/Join.h
```diff
@@ -288,10 +394,13 @@ class Join
     size_t build_concurrency;
     size_t active_build_concurrency;
 
+    mutable std::mutex probe_mutex;
```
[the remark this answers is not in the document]
mengxin9014: > Seems it is not used?

removed.

dbms/src/Interpreters/Join.h
mengxin9014: > why use `std::move` here?

done.

dbms/src/Interpreters/Join.h
mengxin9014: > Although not used in probe stage, but I think we should update the `memory_usage` here?

done.

dbms/src/Interpreters/Join.cpp
mengxin9014: done.

dbms/src/Interpreters/Join.cpp
mengxin9014: done.

dbms/src/Interpreters/Join.cpp
mengxin9014: > I think we can move the `trySpillxxx` into `workAfterBuildFinish`

done

dbms/src/Interpreters/Join.cpp
mengxin9014: > maybe add a new function `workAfterProbeFinish()`?

done.

dbms/src/Interpreters/Join.cpp
```diff
@@ -2451,6 +2756,15 @@ void Join::finishOneBuild()
 
 void Join::workAfterBuildFinish()
 {
+    if (isEnableSpill())
```
windtalker, line 2759: The spill clean staff should be the last step of `workAfterBuildFinish`

windtalker (approved):
Others LGTM

dbms/src/Interpreters/Join.cpp
mengxin9014: > The spill clean staff should be the last step of `workAfterBuildFinish`

done.

SeaRise (approved):

dbms/src/DataStreams/HashJoinProbeBlockInputStream.h
```diff
@@ -38,38 +37,108 @@ class HashJoinProbeBlockInputStream : public IProfilingBlockInputStream
...
-    size_t probe_index;
+    const bool need_output_non_joined_data;
+    size_t current_non_joined_stream_index;
+    BlockInputStreamPtr current_probe_stream;
+    UInt64 max_block_size;
     ProbeProcessInfo probe_process_info;
     BlockInputStreamPtr non_joined_stream;
-    ProbeStatus status{ProbeStatus::PROBE};
+    BlockInputStreamPtr restore_build_stream;
+    BlockInputStreamPtr restore_probe_stream;
```
SeaRise, line 136: how about replacing by `RestoreInfo`?

dbms/src/DataStreams/HashJoinProbeBlockInputStream.h
```diff
@@ -38,38 +37,108 @@ class HashJoinProbeBlockInputStream : public IProfilingBlockInputStream
...
+    ProbeStatus status{ProbeStatus::WAIT_BUILD_FINISH};
     size_t joined_rows = 0;
     size_t non_joined_rows = 0;
-    std::atomic<bool> probe_finished = false;
+    std::list<JoinPtr> parents;
```
SeaRise, line 140: And `std::list<RestoreInfoPtr> parents;` here?

dbms/src/DataStreams/HashJoinProbeBlockInputStream.h
SeaRise: I think it's better not to expose `restore_join`, `restore_stream` to `HashJoinProfbeBlockInputStream`.
The `HashJoinProfbeBlockInputStream` will only see the `RestoreInfo` and will only operate on the `RestoreInfo`

SeaRise: Like `RestoreInfo::restoreBuild` and `RestoreInfo::restoreProbe`?

PR #6662, opened 2023-01-19
least/greatest function for string

### LINKED ISSUE #6660: support least(string, string....) function pushdown

ywqzzy:

Issue Number: close #6660 

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Coprocessor/DAGUtils.cpp
```diff
@@ -162,13 +162,13 @@ const std::unordered_map<tipb::ScalarFuncSig, String> scalar_func_map({
...
     {tipb::ScalarFuncSig::GreatestReal, "tidbGreatest"},
-    {tipb::ScalarFuncSig::GreatestString, "greatest"},
+    {tipb::ScalarFuncSig::GreatestString, "tidbGreatestString"},
     {tipb::ScalarFuncSig::GreatestDecimal, "greatest"},
     {tipb::ScalarFuncSig::GreatestTime, "greatest"},
 
     {tipb::ScalarFuncSig::LeastInt, "tidbLeast"},
     {tipb::ScalarFuncSig::LeastReal, "tidbLeast"},
-    {tipb::ScalarFuncSig::LeastString, "least"},
+    {tipb::ScalarFuncSig::LeastString, "tidbLeastString"},
```
Lloyd-Pottiger, line 171: the permission of file changed.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +120,325 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        int res = 0;
+        StringRef s_ra(&a_data[0], a_size);
+        StringRef s_rb(&b_data[0], b_size);
+        if constexpr (!use_collator)
+            res = memcmp(&a_data[0], &b_data[0], std::min(a_size, b_size));
+        else
+            res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
```
Lloyd-Pottiger, line 141:
```suggestion
        if constexpr (use_collator)
            res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
        else
            res = memcmp(&a_data[0], &b_data[0], std::min(a_size, b_size));
```

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +120,325 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const ColumnString::Offsets & a_offsets,
+        const ColumnString::Chars_t & b_data,
+        const ColumnString::Offsets & b_offsets,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.reserve(std::max(a_data.size(), b_data.size()));
+        c_data.resize(std::max(a_data.size(), b_data.size()));
```
Lloyd-Pottiger, line 274: no need to call `reserve` https://cplusplus.com/reference/vector/vector/resize/

dbms/src/Flash/Coprocessor/DAGUtils.cpp
ywqzzy: Wow~

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +120,325 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        size_t num_arguments = arguments.size();
+
+        using impl = StringOperationImpl<least, use_collator>;
+        const auto * c0 = block.getByPosition(arguments[0]).column.get();
+        auto c_res = ColumnString::create();
+        for (size_t i = 1; i < num_arguments; ++i)
+        {
+            if (i > 1)
+            {
+                c_res = ColumnString::create();
```
yibin87, line 393: Try avoid create new column every time?

yibin87 (commented):
Add fullstack sql test please.

dbms/src/Functions/LeastGreatest.h
gengliqi: @windtalker reminds me that we can refer to the `concat` implementation. https://github.com/pingcap/tiflash/blob/a2979078c9548de22697863e4beb2adc42afb89d/dbms/src/Functions/FunctionsString.cpp#L1205-L1211
It uses vectorized implementation for 2 arguments and polymorphic implementation for more arguments.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        size_t a_size = a_offsets[i] - a_offsets[i - 1] - 1;
+        processImpl(collator, a_size, b_size, &a_data[a_offsets[i - 1]], &b_data[0], c_data, c_offsets, i);
+    }
+
+    // constant_constant
+    static void process(
+        const TiDB::TiDBCollatorPtr & collator,
+        const std::string & a,
+        const std::string & b,
+        std::string & c)
```
gengliqi, line 227: How about this? We can reduce unnecessary memory allocation and copy.
```suggestion
    static void process(
        const TiDB::TiDBCollatorPtr & collator,
        std::string & a,
        StringRef & b)
```
If the result is a, no memory allocation and copy. Else if the result is b, `a = b.toString()`.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        {
+            if (const_columns.size() == 1)
+            {
+                const_res = const_columns[0]->getValue<String>();
+            }
+            else
+            {
+                auto res = const_columns[0]->getValue<String>();
+                for (size_t i = 1; i < const_columns.size(); ++i)
+                    impl::constantConstant(const_columns[i]->getValue<String>(), res, collator, res);
```
gengliqi, line 398: Like I said above, we can call `impl::constantConstant(res, const_columns[i]->getDataAt(0), collator);`

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+                string_columns.emplace_back(c_string);
+        }
+
+        String const_res;
+        // for const columns
+        if (!const_columns.empty())
+        {
+            if (const_columns.size() == 1)
+            {
+                const_res = const_columns[0]->getValue<String>();
```
gengliqi, line 392: I think `getValue<String>` is slower than `getDataAt(0).toString()` because it's heavy to construct `Field`.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+            else
+            {
+                auto res = const_columns[0]->getValue<String>();
+                for (size_t i = 1; i < const_columns.size(); ++i)
+                    impl::constantConstant(const_columns[i]->getValue<String>(), res, collator, res);
+                if (string_columns.empty()) // fill result column
+                {
+                    block.getByPosition(result).column = block.getByPosition(result).type->createColumnConst(const_columns[0]->size(), Field(res));
+                    return;
+                }
```
gengliqi, line 403: This fast code path can be moved outside this branch for the case `const_columns.size() == 1`.

dbms/src/Functions/LeastGreatest.h
gengliqi: Oh. I think we can also use `StringRef` for `a` to avoid any memory allocation.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+            }
+        }
+        else // no string columns, use const columns result
+        {
+            auto col_str = ColumnString::create();
+            col_str->insert(Field(const_res));
+            auto const_res_col = ColumnConst::create(col_str->getPtr(), const_columns[0]->size());
+            block.getByPosition(result).column = std::move(const_res_col);
+            return;
+        }
```
xzhangxian1008, line 457: I think it's impossible to reach here.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+                {
+                    const_res = res;
+                }
+            }
+        }
+
+        // for string columns
+        std::vector<ColumnString::MutablePtr> results;
+        results.emplace_back(ColumnString::create());
+        results.emplace_back(ColumnString::create());
```
xzhangxian1008, line 414: I think these are useless, when string_columns.size() == 1? Maybe we can create them in the second if-branch.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+{
+    static void processImpl(
+        const TiDB::TiDBCollatorPtr & collator,
+        size_t a_size,
+        size_t b_size,
+        const unsigned char * a_data,
+        const unsigned char * b_data,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets,
+        int i)
```
gengliqi, line 135: We can use `StringRef` for `c` to avoid memory copy.

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const ColumnString::Offsets & a_offsets,
+        const std::string & b,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.resize(std::max(a_data.size(), b.size() * size));
+        const auto * b_data = reinterpret_cast<const unsigned char *>(b.data());
+        if (size == 0)
```
xzhangxian1008, line 296: ditto

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const ColumnString::Offsets & a_offsets,
+        const ColumnString::Chars_t & b_data,
+        const ColumnString::Offsets & b_offsets,
+        const TiDB::TiDBCollatorPtr & collator,
+        ColumnString::Chars_t & c_data,
+        ColumnString::Offsets & c_offsets)
+    {
+        size_t size = a_offsets.size();
+        c_data.resize(std::max(a_data.size(), b_data.size()));
+        if (size == 0)
```
xzhangxian1008, line 278: size == 0 could be judged after size is assigned, so that we can avoid many useless operation

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +121,363 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+            res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
+
+        else
+            res = memcmp(&a_data[0], &b_data[0], std::min(a_size, b_size));
+
+        if constexpr (least)
+        {
+            if (res < 0)
+            {
+                memcpy(&c_data[pre_offset], &a_data[0], a_size);
```
xzhangxian1008, line 151: string's terminating zero is not added?

ywqzzy: Yes, we don't need it.

dbms/src/Functions/LeastGreatest.h
ywqzzy: Yes, tidb will not push down this case, but I tested it in UT

xzhangxian1008: > Yes, tidb will not push down this case, but I tested it in UT

We can reach here only when `string_columns.size() == 0`. When `string_columns.size() == 0`, we can know that columns are all const columns, and the function has returned in line 402. So, it's impossible to hit this branch. By the way, which ut case could hit this branch?

dbms/src/Functions/LeastGreatest.h
xzhangxian1008: > Yes, we don't need it.

The original column strings are terminated with zero, why we don't need it here?

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,484 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        const unsigned char * a_data,
+        const unsigned char * b_data,
+        std::vector<StringRef> & res_ref,
+        size_t i)
+    {
+        int res = 0;
+        if constexpr (use_collator)
+            res = collator->compare(reinterpret_cast<const char *>(&a_data[0]), a_size, reinterpret_cast<const char *>(&b_data[0]), b_size);
+        else
+            res = memcmp(&a_data[0], &b_data[0], std::min(a_size, b_size));
```
Lloyd-Pottiger, line 206: `mem_utils::CompareStrView` can deliver better performance than `memcmp`

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,484 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+        {
+            if (res < 0)
+            {
+                memcpy(&c_data[pre_offset], &a_data[0], a_size);
+                c_offsets[i] = pre_offset + a_size + 1;
+            }
+            else if (res == 0)
+            {
+                size_t size = std::min(a_size, b_size);
+                memcpy(&c_data[pre_offset], &b_data[0], size);
```
xzhangxian1008, line 155: `mem_utils::avx2_inline_memcpy` may be helpful.

Lloyd-Pottiger: Actually, we have replace it globally, check `libs/libmemcpy/memcpy.cpp`

ywqzzy: > avx2_inline_memcpy

Nice, I will use inline_memcpy

dbms/src/Functions/LeastGreatest.h
ywqzzy: > > Yes, tidb will not push down this case, but I tested it in UT
> 
> We can reach here only when `string_columns.size() == 0`. When `string_columns.size() == 0`, we can know that columns are all const columns, and the function has returned in line 402. So, it's impossible to hit this branch. By the way, which ut case could hit this branch?

Ops, Removed

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,475 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+                = block
+                      .getByPosition(result)
+                      .type->createColumnConst(const_columns[0]->size(), Field(const_res.toString()));
+            return;
+        }
+
+        // 2. calculate result column for string columns
+        auto string_columns_size = string_columns.size();
+        std::vector<StringRef> result_string_refs;
+        result_string_refs.resize(string_columns[0]->size());
```
gengliqi, line 522: if `string_columns_size` is 1, this memory allocation is useless. I suggest that `result_string_refs` can be moved to `if (string_columns_size >= 2)` clause.

xzhangxian1008:
lgtm

SeaRise (approved):
LGTM by xzhangxian

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,466 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+            {
+                res_ref[i] = StringRef(&b_data[0], b_size);
+            }
+            else if (res == 0)
+            {
+                size_t size = std::max(a_size, b_size);
+                if (a_size > b_size)
+                    res_ref[i] = StringRef(&a_data[0], size);
+                else
+                    res_ref[i] = StringRef(&b_data[0], size);
```
gengliqi, line 236:
```suggestion
                if (a_size > b_size)
                    res_ref[i] = StringRef(&a_data[0], a_size);
                else
                    res_ref[i] = StringRef(&b_data[0], b_size);
```

dbms/src/Functions/LeastGreatest.h
```diff
@@ -120,5 +122,466 @@ class FunctionVectorizedLeastGreatest : public IFunction
...
+                    const_res,
+                    col_str->getChars(),
+                    col_str->getOffsets());
+                block.getByPosition(result).column = std::move(col_str);
+                return;
+            }
+        }
+
+        // 3B. merge result columns of const columns and result_col
+        if (string_columns_size == 1 && !const_columns.empty())
```
gengliqi, line 568: Move to upper `if (string_columns_size == 1)` ? Actually, if `string_columns_size` is 1, `const_columns` must be non-empty because the arguments size is greater than 2.

gengliqi (approved):

PR #6707, opened 2023-02-01
Optimize cross join impl

### LINKED ISSUE #6706: CrossJoin implementation performance improvement
## Enhancement
Current tiflash cross join implementation is not good enough: 
1. Column size is not welled-reserved while its final size can be calculated accurately
2. Column elements are inserted using insertFrom instead of insertRangeFrom, which is not efficient

yibin87:
Signed-off-by: yibin <huyibin@pingcap.com>

Issue Number: close #6706 

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Join.cpp
```diff
@@ -1731,27 +1731,38 @@ struct CrossJoinAdder;
...
 {
-    static void addFound(MutableColumns & dst_columns, size_t num_existing_columns, ColumnRawPtrs & src_left_columns, size_t num_columns_to_add, size_t start_offset, size_t i, const BlocksList & blocks, IColumn::Filter * is_row_matched, IColumn::Offset & current_offset, IColumn::Offsets * expanded_row_size_after_join)
+    static size_t calTotalRightRows(const BlocksList & blocks)
     {
-        size_t expanded_row_size = 0;
+        size_t total_rows = 0;
         for (const Block & block_right : blocks)
         {
             size_t rows_right = block_right.rows();
             if constexpr (STRICTNESS == ASTTableJoin::Strictness::Any)
```
windtalker, line 1740: I think we should update `total_rows` as `total_rows = std::min(total_rows, 1)` after total_rows calculation is done? Otherwise, for `Any` strictniss, the total_rows is the total non-empty block size.

yibin87: You're right, I'll fixed it.

yibin87: > I think we should update `total_rows` as `total_rows = std::min(total_rows, 1)` after total_rows calculation is done? Otherwise, for `Any` strictniss, the total_rows is the total non-empty block size.

Done

gengliqi (approved):
LGTM.
Actually, cross join + STRICTNESS::Any should not happen. 🤣

dbms/src/Interpreters/Join.cpp
```diff
@@ -1731,27 +1731,40 @@ struct CrossJoinAdder;
...
```
windtalker, line 1740: Looks like don't need to update the right_rows here, because it will be updated after `total_rows` is calculated.

windtalker (approved):
Others LGTM

dbms/src/Interpreters/Join.cpp
yibin87: Fix it.

Issue #6699, opened 2023-01-31
Data inconsistency when tpcc workload and consistency check


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

<!-- a step by step guide for reproducing the bug. -->
tpcc workload and consistency check. the correct result is 30000000
```
mysql> begin;
Query OK, 0 rows affected (0.05 sec)

mysql> set tidb_isolation_read_engines='tiflash';
Query OK, 0 rows affected (0.05 sec)

mysql> select count(*) from customer;
+----------+
| count(*) |
+----------+
| 29999991 |
+----------+
1 row in set (0.11 sec)

mysql> select count(*) from customer;
+----------+
| count(*) |
+----------+
| 29999989 |
+----------+
1 row in set (0.10 sec)

mysql> explain analyze select count(*) from customer;
+------------------------------+-------------+----------+--------------+----------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------+---------+------+
| id                           | estRows     | actRows  | task         | access object  | execution info                                                                                                                                                                                                                                                                                                                                                                                        | operator info                              | memory  | disk |
+------------------------------+-------------+----------+--------------+----------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------+---------+------+
| HashAgg_27                   | 1.00        | 1        | root         |                | time:574.1ms, loops:2, partial_worker:{wall_time:573.997579ms, concurrency:5, task_num:1, tot_wait:2.869544929s, tot_exec:18.431µs, tot_time:2.869595816s, max:573.970407ms, p95:573.970407ms}, final_worker:{wall_time:574.030217ms, concurrency:5, task_num:1, tot_wait:2.869907181s, tot_exec:36.645µs, tot_time:2.86994861s, max:574.003103ms, p95:574.003103ms}                                  | funcs:count(Column#39)->Column#23          | 9.86 KB | N/A  |
[...]

PR #6724, opened 2023-02-02
Fix incorrect string value when doing int_handle.toNext

breezewish:

Issue Number: close https://github.com/pingcap/tiflash/issues/6699

Problem Summary:

Consider an Int Handle in variable `foo` which is 0xCCAB. It is encoded as `{ 80 00 00 00 00 00 CC AB }` in bytes.

In #6538:

```
foo.toNext:       int_value = 0xCCAC, bytes_value = { 80 00 00 00 00 00 CC AB 00 }
foo.toPrefixNext: int_value = 0xCCAC, bytes_value = { 80 00 00 00 00 00 CC AC }
```

The implementation of `toNext()` is incorrect because it violates the rule that the int_value and bytes_value should be matching. 

Let's see what will happen when this rule is not respected:

- Consider we construct a new Int Handle in variable `bar` from int 0xCCAC. It will be built as `int_value = 0xCCAC, bytes_value = { 80 00 00 00 00 00 CC AC }`. Now `bar` and `foo`'content are not identical any more.

- The `max`, `min` and several other basic functions are implemented by comparing bytes values:

  ```c++
  inline const RowKeyValue & max(const RowKeyValue & a, const RowKeyValue & b)
  {
      return a.value->compare(*b.value) >= 0 ? a : b;
  }

  inline const RowKeyValue & min(const RowKeyValue & a, const RowKeyValue & b)
  {
      return a.value->compare(*b.value) < 0 ? a : b;
  }

  inline bool none() const
  {
      return start.value->compare(*end.value) >= 0;
  }

  inline bool intersect(const RowKeyRange & other) const
  {
      return max(other.start, start).value->compare(*min(other.end, end).value) < 0;
  }

  bool operator==(const RowKeyRange & rhs) const
  {
      return start.value->compare(*rhs.start.value) == 0 && end.value->compare(*rhs.end.value) == 0;
  }
  ```

- As a result, buggy things happen, like `foo.toNext() != bar`. Everything related with equality or orders may have wrong results.

Encode Int handle using `toPrefixNext()` in `toNext()`, so that the int_value and bytes_value are still matching.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/tests/gtest_key_range.cpp
+++ b/dbms/src/Storages/DeltaMerge/tests/gtest_key_range.cpp
@@ -148,7 +148,7 @@ TEST(RowKey, ToNextKeyCommonHandle)
     EXPECT_EQ(0, compare(my_next.toRowKeyValueRef(), next.toRowKeyValueRef()));
 }
 
-TEST(RowKey, NextIntHandle)
+TEST(RowKey, NextIntHandleCompare)
 {
     auto int_max = RowKeyValue::INT_HANDLE_MAX_KEY;
     auto int_max_i64 = RowKeyValue::fromHandle(Handle(std::numeric_limits<HandleID>::max()));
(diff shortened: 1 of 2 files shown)


JaySon-Huang:
can you elaborate more about why this cause data lost in #6699?

breezewish:
> can you elaborate more about why this cause data lost in #6699?

I updated the PR description, please take a look again, thanks!

dbms/src/Storages/DeltaMerge/tests/gtest_key_range.cpp
```diff
@@ -148,6 +148,37 @@ TEST(RowKey, ToNextKeyCommonHandle)
...
+{
+    auto v0 = RowKeyValue::fromHandle(Handle(1178400));
+    auto v0_next = v0.toNext();
+    auto v1 = RowKeyValue::fromHandle(Handle(1178401));
+
+    EXPECT_EQ(v0, min(v0, v1));
+    EXPECT_EQ(v0, min(v0, v0_next));
+
+    EXPECT_EQ(v1, max(v0, v1));
+    EXPECT_EQ(v1, max(v0, v0_next));
```
hehechen, line 179: It's confused. `EXPECT_EQ(v1, v0_next)` is better ?

hehechen (approved):

JaySon-Huang (approved):
lgtm

Issue #6726, opened 2023-02-03
htap ddl daily tests failed: `DB::Exception: Reading with fields but  entry size is 0`


solotzg:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

```
[2023/02/03 04:36:34.792 +08:00] [ERROR] [SegmentReader.cpp:122] ["ErrMsg: Reading with fields but entry size is 0. StackTrace 
       0x192865e    DB::Exception::Exception(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) [tiflash+26379870]
                    dbms/src/Common/Exception.h:46
       0x7b8fb78    DB::PS::V3::BlobStore::read(std::__1::vector<DB::PS::V3::BlobStore::FieldReadInfo, std::__1::allocator<DB::PS::V3::BlobStore::FieldReadInfo> >&, std::__1::shared_ptr<DB::ReadLimiter> const&) [tiflash+129563512]
                    dbms/src/Storages/Page/V3/BlobStore.cpp:585
       0x7bb6a5b    DB::PS::V3::PageStorageImpl::readImpl(unsigned long, std::__1::vector<std::__1::pair<unsigned long, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > >, std::__1::allocator<std::__1::pair<unsigned long, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > > > > const&, std::__1::shared_ptr<DB::ReadLimiter> const&, std::__1::shared_ptr<DB::PageStorageSnapshot>, bool) [tiflash+129722971]
                    dbms/src/Storages/Page/V3/PageStorageImpl.cpp:240
       0x7be40a0    DB::PageReaderImplNormal::read(std::__1::vector<std::__1::pair<unsigned long, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > >, std::__1::allocator<std::__1::pair<unsigned long, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > > > > const&) const [tiflash+129908896]
                    dbms/src/Storages/Page/PageStorage.cpp:109
[...]

PR #6729, opened 2023-02-03
Fix an exception thrown while reading from ColumnFileTiny after add column

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/6726

#### Testing workload
1. Run TPCC workload
2. Add a column
3. Only read the newly added column

This might lead to an exception with error message `Reading with fields but entry size is 0`

Problem Summary:

Because only the newly added column is read, `fields.length() == 0`. It violates an assumption in BlobStore::read and an exception is thrown

https://github.com/pingcap/tiflash/blob/4206a522521d19d918ae0483a482874f727da30e/dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileTiny.cpp#L62-L80

https://github.com/pingcap/tiflash/blob/4206a522521d19d918ae0483a482874f727da30e/dbms/src/Storages/Page/V3/BlobStore.cpp#L582-L586

* In `ColumnFileTiny::readFromDisk`, skip reading from disk if all are filled by default column
* In `BlobStore::read` return an PageMap with empty Page instead of throwing exceptions
* Add some checks to ensure we will not save an empty block/empty column

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that reading the newly added column after an `ADD COLUMN` DDL might cause error
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetSnapshot.h
+++ b/dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetSnapshot.h
@@ -39,12 +39,13 @@ public:
         , block_offset(offset_)
     {}
     explicit BlockOrDelete(const RowKeyRange & delete_range_)
-        : delete_range(delete_range_)
+        : block_offset(0)
+        , delete_range(delete_range_)
     {}
 
-    bool isBlock() { return (bool)block; }
+    bool isBlock() { return static_cast<bool>(block); }
     auto & getBlock() { return block; };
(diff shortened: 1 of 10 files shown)


Lloyd-Pottiger (approved):
LGTM

lidezhu (approved):
LGTM

hongyunyan (approved):
LGTM

PR #6268, opened 2022-11-07
*: Refine LimitInputStream

ywqzzy:
### What problem does this PR solve?

Issue Number: ref #5900 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #6745, opened 2023-02-06
fix the problem that offset in limit query for tiflash system tables doesn't take effect

### LINKED ISSUE #6747: Offset in limit query for tiflash system tables doesn't take effect
## Bug Report
It causes problem like https://github.com/pingcap/tidb/issues/40988.

lidezhu:

Issue Number: close #6747 

Problem Summary: When query tiflash system tables from tidb, tidb will send query like `select ... limit offset, 1024`. And the query only stops when the num of return rows is less than `length`. However, the `offset` parameter is not repected by tiflash, and will always return the same result with any `offset` value. So if the system table has more than 1024 rows, the query will hang forever.

Port the implementation from https://github.com/ClickHouse/ClickHouse/blob/19.5/dbms/src/DataStreams/LimitBlockInputStream.cpp#L21

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
1. deploy a cluster and import data to make the tiflash system tables `dt_segments` with more than 1024 rows;
2. run the query `select count(*) from information_schema.tiflash_segments` and make sure it succeeds;
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the problem that query on tiflash system tables may hang under large amount of data
```


dbms/src/DataStreams/LimitBlockInputStream.h
```diff
@@ -46,7 +47,10 @@ class LimitBlockInputStream : public IProfilingBlockInputStream
 
 private:
     LoggerPtr log;
-    LocalLimitTransformAction action;
```
windtalker, line 49: This actually remove our current implementaion of `LimitBlockInputStream`, I think it is not recommended, cc @SeaRise

SeaRise: `LocalLimitTransformAction` now is only used in `LimitBlockInputStream`, So class `LocalLimitTransformAction` also needs to be deleted.

lidezhu: Removed.

JaySon-Huang:
Does this bug also affect the v5.x versions or only the v6.x versions?

lidezhu:
> Does this bug also affect the v5.x versions or only the v6.x versions?

It afffects all versions. But it seems there is no more minor version on v5.x.

windtalker (approved):
LGTM

gengliqi (approved):
LGTM

lidezhu:
> Does this bug also affect the v5.x versions or only the v6.x versions?

After verification, it just affect the master branch.

JaySon-Huang:
> > Does this bug also affect the v5.x versions or only the v6.x versions?
> 
> After verification, it just affect the master branch.

OK, I think it is introduced by this PR: https://github.com/pingcap/tiflash/pull/6268

PR #6811, opened 2023-02-13
Remove reactor threads in ExchangeReceiver

xzhangxian1008:

Issue Number: ref #6225

Problem Summary:

- We remove the reactor threads with the implementation of GRPCReceiveQueue and non-blocking AsyncRequestHandler.
- Retain the reactor threads  mode by adding switch

##### How to implement the non-blocking AsyncRequestHandler?

We split the `ReceiverChannelWriter` into `ReceiverChannelBase`, `ReceiverChannelWriter` and `ReceiverChannelTryWriter`. `ReceiverChannelTryWriter`will hold the data which was pushed unsuccessfully when the queue is full, and it provides a function `tryReWrite` to re-write them.

`AsyncRequestHandler` holds the `ReceiverChannelTryWriter` and push data through functions provided by the `ReceiverChannelTryWriter`. When MPMCQueue is full, `ReceiverChannelTryWriter` will return a value to tell the `AsyncRequestHandler` what happened and `AsyncRequestHandler` will push himself into the AsyncRequestHandlerWaitQueue and wait to be executed again.

What does AsyncRequestHandlerWaitQueue do? It stores tags representing `AsyncRequestHandler` which may push data unsuccessfully when the `MPMCQueue` is full. `GRPCReceiveQueue` will get tags from this queue and put the tag into `GRPCCompletionQueue` to wake `AsyncRequestHandler` up.

In a word, `AsyncRequestHandler` puts himself into queue when `MPMCQueue` is full and GRPCReceiveQueue gets tags from it then `MPMCQueue` is available to be pushed.

  | Origin | Refined
-- | -- | --
Q1 | 2.85 | 2.85
Q2 | 1.17 | 1.11
Q3 | 2.72 | 2.85
Q4 | 2.45 | 2.32
Q5 | 6.21 | 6.21
Q6 | 0.64 | 0.64
Q7 | 2.85 | 2.79
Q8 | 4.26 | 4.33
Q9 | 17.35 | 17.35
Q10 | 3.79 | 3.79
Q11 | 0.84 | 0.84
Q12 | 1.58 | 1.51
Q13 | 3.05 | 3.12
Q14 | 0.77 | 0.77
Q15 | 1.64 | 1.64
Q16 | 0.84 | 0.84
Q17 | 6.14 | 6.14
Q18 | 6.95 | 6.95
Q19 | 1.78 | 1.78
Q20 | 1.51 | 1.44
Q21 | 8.89 | 8.49
Q22 | 0.77 | 0.77
Sum | 79.05 | 78.53

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


ywqzzy:
Can we make bench_exchange.cpp great again?

xzhangxian1008:
> Can we make bench_exchange.cpp great again?

Okk, I will fix it

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,336 @@
...
+
+    // execute will be called by RPC framework so it should be as light as possible.
+    // Do not do anything after processXXX functions.
+    void execute(bool & ok) override
+    {
+        switch (stage)
+        {
+        case AsyncRequestStage::WAIT_RETRY:
+            // debug
+            LOG_INFO(log, "Profiling: enter WAIT_RETRY");
```
SeaRise, line 92: Seems useless log?

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -370,9 +370,19 @@ class LocalTunnelSenderV2 : public TunnelSender
...
+    // of local tunnel is always valid(valid means pushing data to an alive reveiver).
+    void closeLocalConnection() const
+    {
+        local_request_handler.closeConnection();
+    }
+
     size_t source_index;
     LocalRequestHandler local_request_handler;
     std::atomic_bool is_done;
+    std::mutex mu;
```
SeaRise, line 385: Seems useless?

dbms/src/Flash/Mpp/MPPTunnel.h
```diff
@@ -370,9 +370,19 @@ class LocalTunnelSenderV2 : public TunnelSender
         }
     }
 
+    // It should only be called in the destructor.
+    //
+    // This function is used to hold the destruction of receiver so that the push operation
+    // of local tunnel is always valid(valid means pushing data to an alive reveiver).
+    void closeLocalConnection() const
```
SeaRise, line 377: Where is this function used?

dbms/src/Common/ConcurrentIOQueue.h
```diff
@@ -48,9 +48,17 @@ class ConcurrentIOQueue
...
     {
-        return mpmc_queue.push(std::move(data));
+        return mpmc_queue.push(std::forward<U>(data));
+    }
+
+    /// Just like MPMCQueue::tryPush.
+    template <typename U>
+    MPMCQueueResult tryPush(U && data)
+    {
+        return mpmc_queue.tryPush(std::forward<U>(data));
```
SeaRise, line 61: why change here?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > Seems useless log?

I will delete it in before merge

dbms/src/Flash/Mpp/MPPTunnel.h
xzhangxian1008: > Seems useless?

I will delete it.

dbms/src/Flash/Mpp/ReceiverChannelWriter.cpp
```diff
@@ -87,7 +48,8 @@ bool ReceiverChannelWriter::writeFineGrain(
             std::move(chunks[i]));
         success = (write_func(i, std::move(recv_msg)) == MPMCQueueResult::OK);
 
-        injectFailPointReceiverPushFail(success, mode);
+        success = (*msg_channels)[i]->push(std::move(recv_msg)) == MPMCQueueResult::OK;
```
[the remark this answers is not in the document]
xzhangxian1008: > why push recv_msg twice? recv_msg has been pushed in `success = (write_func(i, std::move(recv_msg)) == MPMCQueueResult::OK);`.

This is a mistake after merged. I will delete it.

dbms/src/Flash/Mpp/ReceiverChannelBase.h
```diff
@@ -0,0 +1,159 @@
...
+    {
+    case ReceiverMode::Local:
+        fiu_do_on(FailPoints::random_receiver_local_msg_push_failure_failpoint, push_succeed = false);
+        break;
+    case ReceiverMode::Sync:
+        fiu_do_on(FailPoints::random_receiver_sync_msg_push_failure_failpoint, push_succeed = false);
+        break;
+    default:
+        throw Exception(fmt::format("Invalid ReceiverMode: {}", magic_enum::enum_name(mode)));
+    }
```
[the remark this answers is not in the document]
xzhangxian1008: > Seems this function only used in `ChannelWriter.cpp`, why not moving to `ChannelWriter.cpp`?

I will move it.

dbms/src/Flash/Mpp/ReceiverChannelBase.h
```diff
@@ -0,0 +1,159 @@
...
+{
+    return (res == GRPCReceiveQueueRes::OK || res == GRPCReceiveQueueRes::FULL);
+}
+
+// result can not be changed from FULL to OK
+inline void updateResult(GRPCReceiveQueueRes & dst, GRPCReceiveQueueRes & src)
+{
+    if (likely(!(dst == GRPCReceiveQueueRes::FULL && src == GRPCReceiveQueueRes::OK)))
+        dst = src;
+}
```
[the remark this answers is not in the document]
xzhangxian1008: > Seems these functions only used in `RetryWriter.cpp`, why not moving to `RetryWriter.cpp`?

I will move it

dbms/src/Common/ConcurrentIOQueue.h
xzhangxian1008: > why change here?

I will revoke the change of `push`.

For `tyPush`, not always we wan't to pass value by std::move, as tryPush may fail when queue is full. So, sometimes we want to pass value by reference when using tryPush

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.h
```diff
@@ -0,0 +1,66 @@
...
+    {
+        // debug
+        // LOG_INFO(log, "Profiling: rct_cons {}", reinterpret_cast<size_t>(this));
+    }
+
+    // ~ReceiverChannelTryWriter()
+    // {
+    //     // debug
+    //     LOG_INFO(log, "Profiling: rct_des {}", reinterpret_cast<size_t>(this));
+    // }
```
SeaRise, line 46: Please remove this debugging code

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,405 @@
...
+    NEED_INIT,
+    WAIT_MAKE_READER,
+    WAIT_BATCH_READ,
+    WAIT_FINISH,
+    WAIT_RETRY,
+    WAIT_REWRITE,
+    FINISHED,
+};
+
+using Clock = std::chrono::system_clock;
```
SeaRise, line 39:
```suggestion
using Clock = std::chrono::steady_clock;
```
just like https://github.com/pingcap/tiflash/blob/aa9174f6c54ebc428e601b2611680817803fbcea/dbms/src/Common/MPMCQueue.h#L243

dbms/src/Flash/Mpp/ExchangeReceiver.h
```diff
@@ -213,6 +226,11 @@ class ExchangeReceiverBase
     DAGSchema schema;
 
     std::vector<MsgChannelPtr> msg_channels;
+    std::vector<GRPCReceiveQueue<RecvMsgPtr>> grpc_recv_queue;
+    AsyncRequestHandlerWaitQueuePtr async_wait_rewrite_queue;
+
+    std::vector<std::unique_ptr<AsyncRequestHandler<RPCContext, true>>> async_handler_fine_grained_ptrs;
+    std::vector<std::unique_ptr<AsyncRequestHandler<RPCContext, false>>> async_handler_no_fine_grained_ptrs;
```
SeaRise, line 233: How about
```
class AsyncRequestHandlerBase
{};

template<typename A, bool flag>
class AsyncRequestHandler
{};
```
And then one vector only need here.
```
std::vector<std::unique_ptr<AsyncRequestHandlerBase>> async_handler_ptrs;
```

dbms/src/Flash/Mpp/ReceiverChannelBase.cpp
```diff
@@ -0,0 +1,65 @@
...
+// limitations under the License.
+
+#include <Flash/Mpp/GRPCReceiveQueue.h>
+#include <Flash/Mpp/ReceiverChannelBase.h>
+
+#include <utility>
+
+namespace DB
+{
+bool ReceiverChannelBase::splitPacketIntoChunks(size_t source_index, mpp::MPPDataPacket & packet, std::vector<std::vector<const String *>> & chunks)
```
SeaRise, line 22:
```suggestion
bool ReceiverChannelBase::splitFineGrainedShufflePacketIntoChunks(size_t source_index, mpp::MPPDataPacket & packet, std::vector<std::vector<const String *>> & chunks)
```

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,405 @@
...
+                // LOG_INFO(log, "Profiling: enter WAIT_REWRITE");
+                processWaitReWrite();
+                break;
+            case AsyncRequestStage::WAIT_FINISH:
+                // debug
+                // LOG_INFO(log, "Profiling: enter WAIT_FINISH");
+                processWaitFinish();
+                break;
+            default:
+                __builtin_unreachable();
```
SeaRise, line 126:
```suggestion
               RUNTIME_ASSERT(false, "Unexpected stage {}", magic_enum::enum_name(stage));
```

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.h
xzhangxian1008: > Please remove this debugging code

okk

dbms/src/Flash/Mpp/ExchangeReceiver.h
xzhangxian1008: > How about
> 
> ```
> class AsyncRequestHandlerBase
> {};
> 
> template<typename A, bool flag>
> class AsyncRequestHandler
> {};
> ```
> 
> And then one vector only need here.
> 
> ```
> std::vector<std::unique_ptr<AsyncRequestHandlerBase>> async_handler_ptrs;
> ```

okk, It's a good idea.

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+inline bool loopJudge(GRPCReceiveQueueRes res)
+{
+    return (res == GRPCReceiveQueueRes::OK || res == GRPCReceiveQueueRes::FULL);
+}
+
+// result can not be changed from FULL to OK
+inline void updateResult(GRPCReceiveQueueRes & dst, GRPCReceiveQueueRes & src)
+{
+    if (likely(!(dst == GRPCReceiveQueueRes::FULL && src == GRPCReceiveQueueRes::OK)))
```
SeaRise, line 32: It seems this is not unlikely?

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
```diff
@@ -0,0 +1,156 @@
...
+            source_index,
+            req_info,
+            tracked_packet,
+            error_ptr,
+            resp_ptr,
+            std::move(chunks[i]));
+
+        GRPCReceiveQueueRes write_res = tryWriteImpl(i, std::move(recv_msg));
+
+        updateResult(res, write_res);
```
SeaRise, line 80: how about
```suggestion
        if (auto cur_res = tryWriteImpl(i, std::move(recv_msg); cur_res != GRPCReceiveQueueRes::OK)
            res = cur_res;
```

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,389 @@
...
+    std::shared_ptr<AsyncReader> reader;
+    TrackedMppDataPacketPtr packet;
+    Status finish_status;
+    LoggerPtr log;
+
+    ReceiverChannelTryWriter channel_try_writer;
+    AsyncRequestHandlerWaitQueuePtr async_wait_rewrite_queue;
+    KickReceiveTag kick_recv_tag;
+
+    std::mutex mu;
```
SeaRise, line 380: Better to set a meaningful name, such as `make_reader_mu`?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+
+    void start()
+    {
+        stage = AsyncRequestStage::WAIT_MAKE_READER;
+
+        // Use lock to ensure async reader is unreachable from grpc thread before this function returns
+        std::lock_guard lock(mu);
+        rpc_context->makeAsyncReader(request, reader, cq, thisAsUnaryCallback());
```
SeaRise, line 279: I'm a bit confused, will `execute` be called before `makeAsyncReader` returns to callback?

dbms/src/Interpreters/Settings.h
```diff
@@ -304,9 +304,12 @@ struct Settings
...
     M(SettingBool, enable_planner, true, "Enable planner")                                                                                                                                                                              \
     M(SettingBool, enable_pipeline, false, "Enable pipeline model")                                                                                                                                                                     \
+    M(SettingUInt64, local_tunnel_version, 2, "1: not refined, 2: refined")                                                                                                                                                             \
+    M(SettingUInt64, async_recv_version, 2, "1: reactor mode, 2: no additional threads")                                                                                                                                                \
+    M(SettingUInt64, recv_queue_size, 0, "size of ExchangeReceiver queue, 0 means the size is set to tiflash_num * 50")                                                                                                                 \
```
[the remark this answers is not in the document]
xzhangxian1008: > pipeline_io_task_thread_pool_size

okk

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
xzhangxian1008: > how about

I think it's good.

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
xzhangxian1008: > It seems this is not unlikely?

`updateResult` will be modified, according to the following comment.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > Better to set a meaningful name, such as `make_reader_mu`?

okk

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > I'm a bit confused, will `execute` be called before `makeAsyncReader` returns to callback?

Yes, execute may be called when reader in the init() is being assigned.

https://github.com/pingcap/tiflash/blob/f9d16c46e0ed6b66494d764954641d20cd4efb19/dbms/src/Flash/Mpp/GRPCReceiverContext.cpp#L105-L113

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+inline bool loopJudge(GRPCReceiveQueueRes res)
+{
+    return (res == GRPCReceiveQueueRes::OK || res == GRPCReceiveQueueRes::FULL);
+}
+
+// result can not be changed from FULL to OK
+inline void updateResult(GRPCReceiveQueueRes & dst, GRPCReceiveQueueRes & src)
+{
+    if (unlikely(src != GRPCReceiveQueueRes::OK))
```
SeaRise, line 32: For `{channelA, channelB, channelC}`, it's possible for `channelA` and `channelC` to return `OK`, and `channelB` to return `FULL`, it's just not allowed to set the final result to `channelC`'s `OK` right?

SeaRise: If it's just not allowed to update the result, rather than this situation being impossible, "unlikely" shouldn't be used.

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+namespace
+{
+inline bool loopJudge(GRPCReceiveQueueRes res)
+{
+    return (res == GRPCReceiveQueueRes::OK || res == GRPCReceiveQueueRes::FULL);
+}
+
+// result can not be changed from FULL to OK
+inline void updateResult(GRPCReceiveQueueRes & dst, GRPCReceiveQueueRes & src)
```
SeaRise, line 30: It seems that wrapping this method is not necessary.
Maybe
```
GRPCReceiveQueueRes write_res = tryWriteImpl(i, std::move(recv_msg));
if (write_res != GRPCReceiveQueueRes::OK)
    res = write_res;
```
and
```
       if (write_res == GRPCReceiveQueueRes::OK)
        {
            auto tmp_iter = iter;
            ++iter;
            rewrite_msgs.erase(tmp_iter);
        }
        else
        {
            ++iter;
            res = write_res;
        }
            
```
is better.

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
xzhangxian1008: > For `{channelA, channelB, channelC}`, it's possible for `channelA` and `channelC` to return `OK`, and `channelB` to return `FULL`, it's just not allowed to set the final result to `channelC`'s `OK` right?

yes

xzhangxian1008: > t's just not allowe

okk

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
xzhangxian1008: > It seems that wrapping this method is not necessary. Maybe
> 
> ```
> GRPCReceiveQueueRes write_res = tryWriteImpl(i, std::move(recv_msg));
> if (write_res != GRPCReceiveQueueRes::OK)
>     res = write_res;
> ```
> 
> and
> 
> ```
>        if (write_res == GRPCReceiveQueueRes::OK)
>         {
>             auto tmp_iter = iter;
>             ++iter;
>             rewrite_msgs.erase(tmp_iter);
>         }
>         else
>         {
>             ++iter;
>             res = write_res;
>         }
>             
> ```
> 
> is better.

okk

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    return res;
+}
+
+GRPCReceiveQueueRes ReceiverChannelTryWriter::tryWriteImpl(size_t index, std::shared_ptr<ReceivedMessage> && msg)
+{
+    GRPCReceiveQueueRes res = grpc_recv_queues[index].push(msg);
+    if (res == GRPCReceiveQueueRes::FULL)
+        rewrite_msgs.insert({index, std::move(msg)});
```
SeaRise, line 139: What will happen when the index is repeated? Or will it not happen? How about adding an assert?

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    std::queue<AsyncRetryConnection> popAll()
+    {
+        std::lock_guard lock(mu);
+        if (wait_retry_queue.empty())
+            return std::queue<AsyncRetryConnection>{};
+
+        std::queue<AsyncRetryConnection> ret_conn = std::move(wait_retry_queue);
+        return ret_conn;
```
SeaRise, line 99: I am a little worried about whether using move will cause problems with `wait_retry_queue`.
```suggestion
        std::queue<AsyncRetryConnection> ret;
        {
            std::lock_guard lock(mu);
            std::swap(ret, wait_retry_queue);
        }
        return wait_retry_queue;
```

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,250 @@
...
+        wait_retry_queue.push(conn);
+    }
+
+    AsyncRetryConnection pop()
+    {
+        std::lock_guard lock(mu);
+        if (wait_retry_queue.empty())
+            return AsyncRetryConnection(nullptr, nullptr);
+
+        AsyncRetryConnection ret_conn = wait_retry_queue.front();
```
SeaRise, line 86:
```suggestion
        AsyncRetryConnection ret_conn = std::move(wait_retry_queue.front());
```

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,250 @@
...
+    {
+        return recv_queue->getCancelReason();
+    }
+
+    // Cancel the send queue, and set the cancel reason
+    bool cancelWith(const String & reason)
+    {
+        auto ret = recv_queue->cancelWith(reason);
+        if (ret)
+            handleRemainingTags();
```
SeaRise, line 152: It seems that for cancel, the remaining messages can not be handled.

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
```diff
@@ -0,0 +1,250 @@
...
+
+        putTagIntoCompletionQueue(retry_conn);
+    }
+
+    void handleRemainingTags()
+    {
+        std::queue<AsyncRetryConnection> remaining_tags = conn_wait_queue->popAll();
+        while (!remaining_tags.empty())
+        {
+            AsyncRetryConnection conn = remaining_tags.front();
```
SeaRise, line 228:
```suggestion
            AsyncRetryConnection conn = std::move(remaining_tags.front());
```

dbms/src/Interpreters/Settings.h
```diff
@@ -291,7 +291,10 @@ struct Settings
...
     M(SettingUInt64, pipeline_cpu_task_thread_pool_size, 0, "The size of cpu task thread pool. 0 means using number_of_logical_cpu_cores.")                                                                                             \
     M(SettingUInt64, pipeline_io_task_thread_pool_size, 0, "The size of io task thread pool. 0 means using number_of_logical_cpu_cores.")                                                                                               \
-    M(SettingUInt64, local_tunnel_version, 2, "1: not refined, 2: refined")
+    M(SettingUInt64, local_tunnel_version, 2, "1: not refined, 2: refined")                                                                                                                                                             \
+    M(SettingUInt64, async_recv_version, 2, "1: reactor mode, 2: no additional threads")                                                                                                                                                \
+    M(SettingUInt64, recv_queue_size, 0, "size of ExchangeReceiver queue, 0 means the size is set to tiflash_num * 50")
```
SeaRise, line 296: `tiflash_num` makes me a little confused, maybe `data_source_mpp_task_num` is better?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -465,27 +483,81 @@ void ExchangeReceiverBase<RPCContext>::setUpConnection()
...
+                    reactor<false>(async_requests);
+            });
+
+            ++thread_count;
+            connection_uncreated_num -= async_conn_num;
+        }
+    }
+    else
+    {
+        LOG_INFO(exc_log, "enable async_recv_version 2");
```
SeaRise, line 523:
```suggestion
        LOG_DEBUG(exc_log, "enable async_recv_version 2");
```

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -967,21 +1042,38 @@ void ExchangeReceiverBase<RPCContext>::connectionLocalDone()
...
+    if (grpc_recv_queue.empty())
+    {
+        for (auto & msg_channel : msg_channels)
+            msg_channel->cancel();
+    }
+    else
+    {
+        for (auto & channel : grpc_recv_queue)
+            channel.cancel();
+    }
```
SeaRise, line 1076: Looks like `grpc_recv_queue` won't be empty is it?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -967,21 +1042,38 @@ void ExchangeReceiverBase<RPCContext>::connectionLocalDone()
...
+    if (grpc_recv_queue.empty())
+    {
+        for (auto & msg_channel : msg_channels)
+            msg_channel->finish();
+    }
+    else
+    {
+        for (auto & channel : grpc_recv_queue)
+            channel.finish();
+    }
```
SeaRise, line 1061: Looks like `grpc_recv_queue` won't be empty is it?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -413,9 +424,16 @@ void ExchangeReceiverBase<RPCContext>::prepareMsgChannels()
...
+            msg_channels.push_back(std::make_shared<ConcurrentIOQueue<RecvMsgPtr>>(max_buffer_size));
     else
-        msg_channels.push_back(std::make_shared<ConcurrentIOQueue<std::shared_ptr<ReceivedMessage>>>(max_buffer_size));
+        msg_channels.push_back(std::make_shared<ConcurrentIOQueue<RecvMsgPtr>>(max_buffer_size));
+}
+
+template <typename RPCContext>
+void ExchangeReceiverBase<RPCContext>::prepareGRPCReceiveQueue()
+{
+    for (auto & msg_channel : msg_channels)
```
SeaRise, line 435: how about adding `assert(!msg_channels.empty());`

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -914,48 +988,49 @@ void ExchangeReceiverBase<RPCContext>::connectionDone(
...
-            if (state == ExchangeReceiverState::NORMAL)
-                state = ExchangeReceiverState::ERROR;
-            if (err_msg.empty())
-                err_msg = local_err_msg;
-            else
-                first_err_msg = err_msg;
-        }
+    // We must protect the following codes with lock.
+    // Because once live_connections == 0, ExchangeReceiver may be destructed immediately
+    // and the data in ExchangeReceiver is no longer alive.
```
SeaRise, line 995: Or like `local_tunnel_v2`, add a `live_async_connections`?

SeaRise: or call `async_handler_ptrs.wait()` in `~Exchange()` just like `thread_manager.wait()`?

dbms/src/Flash/Mpp/tests/gtest_grpc_send_queue.cpp
```diff
@@ -28,7 +28,7 @@ class TestGRPCSendQueue : public testing::Test
 protected:
     TestGRPCSendQueue()
         : tag(nullptr)
-        , queue(10, [this](KickTag * t) -> grpc_call_error {
+        , queue(10, [this](KickSendTag * t) -> grpc_call_error {
```
SeaRise, line 31: Is there a similar unit test for `grpc_receiver_queue`?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,389 @@
...
+            return true;
+        }
+        else
+        {
+            closeConnection(std::move(done_msg));
+            return false;
+        }
+    }
+
+    GRPCReceiveQueueRes sendPackets()
```
SeaRise, line 301:
```suggestion
    GRPCReceiveQueueRes sendPackets() noexcept
```

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,389 @@
...
+            LOG_ERROR(log, "Some errors happen in the AsyncRequestHandler.execute()");
+            closeConnection("Exception is thrown in AsyncRequestHandler");
+        }
+    }
+
+private:
+    void processWaitMakeReader(bool ok)
+    {
+        // Use lock to ensure reader is created already in reactor thread
+        std::lock_guard lock(make_reader_mu);
```
SeaRise, line 128: Will this lock block block for a long time?
Is it necessary to use a non-blocking way of checking?
```
if !(is_ready)
    return;
```

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.cpp
xzhangxian1008: > What will happen when the index is repeated? Or will it not happen? How about adding an assert?

Same index is impossible to appear here, as only all write fail data has been written successfully the new data will goes on.
okk, I will add an assert

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
xzhangxian1008: > I am a little worried about whether using move will cause problems with `wait_retry_queue`.



> I am a little worried about whether using move will cause problems with `wait_retry_queue`.

Repeatedly using a moved object seems to violate the semantic, I will change it.

dbms/src/Interpreters/Settings.h
xzhangxian1008: > `tiflash_num` makes me a little confused, maybe `data_source_mpp_task_num` is better?

okk, I will change it.

dbms/src/Flash/Mpp/GRPCReceiveQueue.h
xzhangxian1008: > It seems that for cancel, the remaining messages can not be handled.

Not for processing the remaining messages, but for awaking the remaining `AsyncRequestHandler` to exit and notify the `ExchangeReceiver` that a connection has been closed, or the receiver will hang up.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
xzhangxian1008: > Looks like `grpc_recv_queue` won't be empty is it?

Yes, and I will change the logic related to `grpc_recv_queue`.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
xzhangxian1008: > how about adding `assert(!msg_channels.empty());`

I think it's unnecessary as `msg_channels` is impossible to be empty.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > 

I think try-catch is better, and we can exist the tiflash when exception is thrown.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > Will this lock block block for a long time? Is it necessary to use a non-blocking way of checking?
> 
> ```
>     void start()
>     {
>         stage = AsyncRequestStage::WAIT_MAKE_READER;
> 
>         rpc_context->makeAsyncReader(request, reader, cq, thisAsUnaryCallback());
>         is_reader_ready = true; // std::atomic_bool
>     }
> ```
> 
> ```
>     void processWaitMakeReader(bool ok)
>     {
>         // Use lock to ensure reader is created already in reactor thread
>         std::lock_guard lock(make_reader_mu);
>         if (!ok)
>         {
>             reader.reset();
>             LOG_WARNING(log, "MakeReader fail. retry time: {}", retry_times);
>             retryOrDone("Exchange receiver meet error : send async stream request fail");
>         }
>         else
>         {
>             if (!is_reader_ready)
>                 return;
>             startAsyncRead();
>         }
>     }
> ```

First, data race here is very hard to trigger. Second, only when failing to creating connection, the data race will happen, and performance is unnecessary to focus on in this circumstance.

xzhangxian1008: > > Will this lock block block for a long time? Is it necessary to use a non-blocking way of checking?
> > ```
> >     void start()
> >     {
> >         stage = AsyncRequestStage::WAIT_MAKE_READER;
> > 
> >         rpc_context->makeAsyncReader(request, reader, cq, thisAsUnaryCallback());
> >         is_reader_ready = true; // std::atomic_bool
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> >     void processWaitMakeReader(bool ok)
> >     {
> >         // Use lock to ensure reader is created already in reactor thread
> >         std::lock_guard lock(make_reader_mu);
> >         if (!ok)
> >         {
> >             reader.reset();
> >             LOG_WARNING(log, "MakeReader fail. retry time: {}", retry_times);
> >             retryOrDone("Exchange receiver meet error : send async stream request fail");
> >         }
> >         else
> >         {
> >             if (!is_reader_ready)
> >                 return;
> >             startAsyncRead();
> >         }
> >     }
> > ```
> 
> First, data race here is very hard to trigger. Second, only when failing to creat connection, the data race will happen, and performance is unnecessary to focus on in this circumstance.

dbms/src/Flash/Mpp/tests/gtest_grpc_send_queue.cpp
SeaRise: ok

dbms/src/Flash/Mpp/AsyncRequestHandler.h
SeaRise: ok

dbms/src/Flash/Mpp/AsyncRequestHandler.h
SeaRise: ok

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,398 @@
...
+            int copy_retry_times = retry_times;
+
+            if (!retryOrDone(fmt::format("Exchange receiver meet error : {}", finish_status.error_message())))
+            {
+                LOG_WARNING(
+                    copy_log,
+                    "Finish fail. err code: {}, err msg: {}, retry time {}",
+                    copy_finish_status.error_code(),
+                    copy_finish_status.error_message(),
+                    copy_retry_times);
```
SeaRise, line 172: It seems that if we move to https://github.com/pingcap/tiflash/pull/6811/files#diff-8f3f21fe5c9b41fa0b0ea3e1e802d3aeeca41d4a83bb01217fc9e745d7fb69aaR296, there will be no need to copy.

SeaRise: And then `retryOrDone` doesn't need to return a bool.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
xzhangxian1008: > Or like `local_tunnel_v2`, add a `live_async_connections`?

okk, I will change it.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > It seems that if we move to https://github.com/pingcap/tiflash/pull/6811/files#diff-8f3f21fe5c9b41fa0b0ea3e1e802d3aeeca41d4a83bb01217fc9e745d7fb69aaR296, there will be no need to copy.

okk, I will move it.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -280,22 +283,22 @@ class AsyncRequestHandler : public AsyncRequestHandlerBase
         rpc_context->makeAsyncReader(request, reader, cq, thisAsUnaryCallback());
     }
 
-    bool retryOrDone(String done_msg)
+    void retryOrDone(String && done_msg, const String & log_msg)
     {
         if (retriable())
         {
+            LOG_WARNING(log, log_msg);
```
SeaRise, line 290: Move to L287 and don't need two `LOG_WARNING(log, log_msg)`?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -394,5 +397,8 @@ class AsyncRequestHandler : public AsyncRequestHandlerBase
 
     // try-catch may call close_conn, and we need to ensure that close_conn is called for only once.
     bool is_close_conn_called;
+
+    std::mutex mu;
```
SeaRise, line 401: It looks like `std::mutex mu` and `std::mutex make_reader_mu` can be merged into one

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > Move to L287 and don't need two `LOG_WARNING(log, log_msg)`?

okk

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > It looks like `std::mutex mu` and `std::mutex make_reader_mu` can be merged into one

No, or it will be dead lock as we will call `retryOrDone` in `processWaitMakeReader`. I think we can rename mu to `is_close_conn_called_mu`.

SeaRise: ok

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -961,28 +1042,76 @@ void ExchangeReceiverBase<RPCContext>::connectionDone(
...
+    //     LOG_DEBUG(log, "All threads end in ExchangeReceiver");
+    //     cv.notify_all();
+    // }
+
+    // if (meet_error || live_connections == 0)
+    // {
+    //     auto log_level = meet_error ? Poco::Message::PRIO_WARNING : Poco::Message::PRIO_INFORMATION;
+    //     LOG_IMPL(exc_log, log_level, "Finish receiver channels, meet error: {}, error message: {}", meet_error, first_err_msg);
+    //     finishAllMsgChannels();
+    // }
```
SeaRise, line 1091: Seems useless codes

SeaRise (approved):
Others LGTM

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,408 @@
...
+            case AsyncRequestStage::WAIT_FINISH:
+                processWaitFinish();
+                break;
+            default:
+                RUNTIME_ASSERT(false, "Unexpected stage {}", magic_enum::enum_name(stage));
+            }
+        }
+        catch (...)
+        {
+            LOG_ERROR(log, "Some errors happen in the AsyncRequestHandler.execute()");
```
ywqzzy, line 131: log exception more precisely?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,408 @@
...
+            closeConnection("Exception is thrown in AsyncRequestHandler");
+        }
+    }
+
+    void wait() override
+    {
+        std::unique_lock<std::mutex> ul(is_close_conn_called_mu);
```
ywqzzy, line 138: maybe use lock_guard?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,408 @@
...
+    WAIT_RETRY,
+    WAIT_REWRITE,
+    FINISHED,
+};
+
+using SystemClock = std::chrono::system_clock;
+
+constexpr Int32 max_retry_times = 10;
+constexpr Int32 retry_interval_time = 1; // second
+constexpr Int32 batch_packet_count = 16;
```
ywqzzy, line 53: How about making this configurable?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,408 @@
...
+                "Exchange receiver meet error : send async stream request fail",
+                fmt::format("Make reader fail. retry time: {}", retry_times));
+        }
+        else
+        {
+            startAsyncRead();
+        }
+    }
+
+    void processWaitBatchRead(bool ok)
```
ywqzzy, line 160: I am wondering if we add inline in function definition can have some effect?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -45,7 +46,13 @@ String constructStatusString(ExchangeReceiverState state, const String & error_m
     return fmt::format("Receiver state: {}, error message: {}", magic_enum::enum_name(state), error_message);
 }
 
-enum class AsyncRequestStage
+size_t getMaxBufferSize(Int32 source_num, Int32 recv_queue_size)
+{
+    size_t size = recv_queue_size == 0 ? static_cast<size_t>(source_num) * 50 : static_cast<size_t>(recv_queue_size);
+    return std::min(1000, size);
```
ywqzzy, line 52: Why choose 50*source_num?

dbms/src/Flash/Mpp/AsyncRequestHandler.h
```diff
@@ -0,0 +1,408 @@
...
+
+constexpr Int32 max_retry_times = 10;
+constexpr Int32 retry_interval_time = 1; // second
+constexpr Int32 batch_packet_count = 16;
+
+class AsyncRequestHandlerBase : public UnaryCallback<bool>
+{
+public:
+    virtual void wait() = 0;
+};
```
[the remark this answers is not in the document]
xzhangxian1008: > Why introduce this base class?

In ExchangeReceiver, we need to save pointer for `AsyncRequestHandler<RPCContext, true>` or `AsyncRequestHandler<RPCContext, false>`, base class could represent both of them.

dbms/src/Flash/Mpp/ReceiverChannelTryWriter.h
```diff
@@ -0,0 +1,57 @@
...
+namespace DB
+{
+
+using RecvMsgPtr = std::shared_ptr<ReceivedMessage>;
+
+class ReceiverChannelTryWriter : public ReceiverChannelBase
+{
+public:
+    ReceiverChannelTryWriter(std::vector<GRPCReceiveQueue<RecvMsgPtr>> & grpc_recv_queues_, const String & req_info_, const LoggerPtr & log_, std::atomic<Int64> * data_size_in_queue_, ReceiverMode mode_)
+        : ReceiverChannelBase(grpc_recv_queues_.size(), req_info_, log_, data_size_in_queue_, mode_)
```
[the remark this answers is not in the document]
xzhangxian1008: > The `ReceiverMode` must be async?

So far, `ReceiverMode` must be async, however I can ensure it will always be this in the future.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > log exception more precisely?

okk

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > maybe use lock_guard?

lock passed into condition_variable must be unique_lock

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > How about making this configurable?

`batch_packet_count` has been set as configurable, other parameters are no need I think, as they are unlikely to be used.

dbms/src/Flash/Mpp/AsyncRequestHandler.h
xzhangxian1008: > I am wondering if we add inline in function definition can have some effect?

Performance should not be the reason to use inline attribute as the compiler will try it best to inline anything.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
xzhangxian1008: > Why choose 50*source_num?

It's just a experience value.

gengliqi (commented):
It's better to move some logic of `AsyncRequestHandler` and `ReceiverChannelTryWriter` to `GRPCReceiveQueue` to enforce adherence to the principles of high cohesion and low coupling. 
For example, we can save the data and tag in `GRPCReceiveQueue` if `push` returns `FULL`. After a `pop` is called, the data can be pushed into the queue and the tag can be kicked. 
The case is more complicated for the fine-grained shuffle. There should be several queues in `GRPCReceiveQueue` and the tag should be kicked only when all packets of data have been sent to different queues.

ywqzzy (approved):

Issue #6528, opened 2022-12-22
Support spill to disk for memory intensive operator


windtalker:
## Enhancement
Support spill to disk for memory intensive operator like  `join/aggregation/sort/topN/window`.

PR #6820, opened 2023-02-15
add support for auxiliary memory bound in mpmc_queue

windtalker:

Issue Number: ref #6528

Problem Summary:
In TiFlash runtime, some executors(like ExchangeReceiver, MPPTunnelSender, UnionBlockInputStream) use a MPMC queue to buffer the data, but the queues are only limited by the number of elements. If the element occupies a large amount of memory, the buffer queue may occupy too much memory, this pr add `auxiliary_memory_bound` to mpmc queue, so the memory usage by the queue can be limited.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/MPMCQueue.h
+++ b/dbms/src/Common/MPMCQueue.h
@@ -306,6 +306,7 @@ private:
             if (!wait(lock, reader_head, node, pred, deadline))
                 is_timeout = true;
         }
+        /// double check status after potential wait
         if (!isCancelled() && read_pos < write_pos)
         {
             auto & obj = getObj(read_pos);
@@ -465,8 +466,9 @@ private:
     {
         if constexpr (read)
         {
-            current_auxiliary_memory_usage -= element_auxiliary_memory[pos % capacity];
-            element_auxiliary_memory[pos % capacity] = 0;
+            auto & elem_value = element_auxiliary_memory[pos % capacity];
+            current_auxiliary_memory_usage -= elem_value;
+            elem_value = 0;
         }
         else
         {
--- a/dbms/src/Common/MPMCQueue.h
+++ b/dbms/src/Common/MPMCQueue.h
@@ -73,20 +73,20 @@ class MPMCQueue
 public:
     using Status = MPMCQueueStatus;
     using Result = MPMCQueueResult;
-    using ElementAuxiliaryMemoryUsageFunc = std::function<size_t(const T & element)>;
+    using ElementAuxiliaryMemoryUsageFunc = std::function<Int64(const T & element)>;
 
     explicit MPMCQueue(size_t capacity_)
         : capacity(capacity_)
-        , max_auxiliary_memory_usage(std::numeric_limits<size_t>::max())
+        , max_auxiliary_memory_usage(std::numeric_limits<Int64>::max())
         , get_auxiliary_memory_usage([](const T &) { return 0; })
         , data(capacity * sizeof(T))
     {
     }
 
-    /// max_auxiliary_memory_usage_ = 0 means no limit on auxiliary memory usage
-    MPMCQueue(size_t capacity_, size_t max_auxiliary_memory_usage_, ElementAuxiliaryMemoryUsageFunc && get_auxiliary_memory_usage_)
+    /// max_auxiliary_memory_usage_ <= 0 means no limit on auxiliary memory usage
+    MPMCQueue(size_t capacity_, Int64 max_auxiliary_memory_usage_, ElementAuxiliaryMemoryUsageFunc && get_auxiliary_memory_usage_)
         : capacity(capacity_)
-        , max_auxiliary_memory_usage(max_auxiliary_memory_usage_ == 0 ? std::numeric_limits<size_t>::max() : max_auxiliary_memory_usage_)
+        , max_auxiliary_memory_usage(max_auxiliary_memory_usage_ <= 0 ? std::numeric_limits<Int64>::max() : max_auxiliary_memory_usage_)
         , get_auxiliary_memory_usage(std::move(get_auxiliary_memory_usage_))
         , data(capacity * sizeof(T))
     {
@@ -463,7 +463,7 @@ private:
     /// auxiliary memory usage, then the user should provide the function to calculate the auxiliary memory usage
     /// Note: unlike capacity, max_auxiliary_memory_usage is actually a soft-limit because we need to make at least
     /// one element can be pushed to the queue even if its auxiliary memory exceeds max_auxiliary_memory_usage
-    const size_t max_auxiliary_memory_usage;
+    const Int64 max_auxiliary_memory_usage;
     const ElementAuxiliaryMemoryUsageFunc get_auxiliary_memory_usage;
 
     mutable std::mutex mu;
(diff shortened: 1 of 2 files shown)


dbms/src/Common/MPMCQueue.h
```diff
@@ -73,9 +73,21 @@ class MPMCQueue
 public:
     using Status = MPMCQueueStatus;
     using Result = MPMCQueueResult;
+    using ElementAuxiliaryMemoryUsageFunc = std::function<size_t(const T & element)>;
 
     explicit MPMCQueue(size_t capacity_)
         : capacity(capacity_)
+        , max_auxiliary_memory_usage(std::numeric_limits<size_t>::max())
+        , get_auxiliary_memory_usage([](const T &) { return 0; })
```
SeaRise, line 81: How about using templates?
```
template <typename T>
struct DefaultGetAuxiliaryMemoryUsage
{
    size_t get(const T &) { return 0; }
};
template <typename T, typename GetAuxiliaryMemoryUsage = DefaultGetAuxiliaryMemoryUsage>
class MPMCQueue
```

dbms/src/Common/MPMCQueue.h
```diff
@@ -289,26 +302,33 @@ class MPMCQueue
...
             auto & obj = getObj(read_pos);
             res = std::move(obj);
             destruct(obj);
+            current_auxiliary_memory_usage -= get_auxiliary_memory_usage(res);
 
             /// update pos only after all operations that may throw an exception.
             ++read_pos;
+            if (read_pos == write_pos)
+                /// if there is no element, the current_auxiliary_memory_usage must be zero
+                current_auxiliary_memory_usage = 0;
```
SeaRise, line 318:
```suggestion
       assert(read_pos != write_pos || current_auxiliary_memory_usage == 0);
```

dbms/src/Common/MPMCQueue.h
```diff
@@ -353,6 +375,8 @@ class MPMCQueue
             notifyNext(reader_head);
             return Result::OK;
         }
+        if (is_timeout)
+            return Result::TIMEOUT;
```
SeaRise, line 379: why move to here?

dbms/src/Common/MPMCQueue.h
windtalker: Seems don't have too much benifits to use template here?

dbms/src/Common/MPMCQueue.h
windtalker: I think it's ok to set `current_auxiliary_memory_usage =0` here, because even if `current_auxiliary_memory_usage` is not 0 due to some bugs, it still works fine after we reset it to zero.

dbms/src/Common/MPMCQueue.h
windtalker: Consider a case that `wait` is timeout, and at the meantime, there is a notifiy on the condition variable, in the original implementation, the notification is actually lost, so we double check the condition even if `wait` return false.

dbms/src/Common/MPMCQueue.h
SeaRise: ok

dbms/src/Common/MPMCQueue.h
SeaRise: What about both assert and set 0?
This will catch `current_auxiliary_memory_usage`'s bug in the daily run

dbms/src/Common/MPMCQueue.h
```diff
@@ -436,6 +472,7 @@ class MPMCQueue
     Int64 write_pos = 0;
     Status status = Status::NORMAL;
     String cancel_reason;
+    size_t current_auxiliary_memory_usage = 0;
```
yibin87, line 475: Use Int64 here to avoid incorrect get_auxiliary_memory_usage case?

dbms/src/Common/MPMCQueue.h
windtalker: OK, updated.

dbms/src/Common/MPMCQueue.h
```diff
@@ -289,26 +302,34 @@ class MPMCQueue
...
             ///    This need carefully procesing in `assignObj`.
             /// 2. If we do not remove the next writer, only obtain its pointer and notify it later,
             ///    deadlock can be possible because different readers may notify one writer.
-            notifyNext(writer_head);
+            if (current_auxiliary_memory_usage < max_auxiliary_memory_usage)
+                notifyNext(writer_head);
             return Result::OK;
         }
+        if (is_timeout)
+            return Result::TIMEOUT;
```
gengliqi, line 332:
```suggestion
        if constexpr (need_wait) 
        {
            if (is_timeout)
            return Result::TIMEOUT;
        }
```

dbms/src/Common/MPMCQueue.h
```diff
@@ -353,6 +376,8 @@ class MPMCQueue
             notifyNext(reader_head);
             return Result::OK;
         }
+        if (is_timeout)
+            return Result::TIMEOUT;
```
gengliqi, line 380: Ditto.

dbms/src/Common/MPMCQueue.h
```diff
@@ -289,26 +302,34 @@ class MPMCQueue
...
             if (!wait(lock, reader_head, node, pred, deadline))
-                return Result::TIMEOUT;
+                is_timeout = true;
         }
         if (!isCancelled() && read_pos < write_pos)
         {
             auto & obj = getObj(read_pos);
             res = std::move(obj);
             destruct(obj);
+            current_auxiliary_memory_usage -= get_auxiliary_memory_usage(res);
```
gengliqi, line 312: It's a little wasteful to call `get_auxiliary_memory_usage` twice for one object.
And also it requires that these two return values of `get_auxiliary_memory_usage` must be the same.

windtalker: OK, save `auxiliary_memory` in another vector.

dbms/src/Common/MPMCQueue.h
windtalker: Done

gengliqi (approved):

dbms/src/Common/MPMCQueue.h
```diff
@@ -426,8 +460,35 @@ class MPMCQueue
         return false;
     }
 
+    template <bool read>
+    ALWAYS_INLINE void updateElementAuxiliaryMemory(size_t pos)
+    {
+        if constexpr (read)
+        {
+            current_auxiliary_memory_usage -= element_auxiliary_memory[pos % capacity];
+            element_auxiliary_memory[pos % capacity] = 0;
```
SeaRise, line 469:
```suggestion
            auto & elem_value = element_auxiliary_memory[pos % capacity];
            current_auxiliary_memory_usage -= elem_value;
            elem_value = 0;
```

dbms/src/Common/MPMCQueue.h
```diff
@@ -289,26 +304,37 @@ class MPMCQueue
                 return read_pos < write_pos || !isNormal();
             };
             if (!wait(lock, reader_head, node, pred, deadline))
-                return Result::TIMEOUT;
+                is_timeout = true;
         }
         if (!isCancelled() && read_pos < write_pos)
```
SeaRise, line 309: add comment same as `assignObj`
```suggestion
        /// double check status after potential wait
        if (!isCancelled() && read_pos < write_pos)
```

SeaRise (approved):
Others LGTM

Issue #6426, opened 2022-12-05
Reduce TiFlash's memory usage about the write path


flowbehappy:
## Enhancement
A TiFlash node with 400 GiB data consumes over 20 GiB memory under distributed write workload. We need to reduce the memory consumption.

- [ ] https://github.com/pingcap/tiflash/issues/6401
- [ ] https://github.com/pingcap/tiflash/issues/6399
- [ ] https://github.com/pingcap/tiflash/issues/6333
- [ ] https://github.com/pingcap/tiflash/issues/6332

PR #6861, opened 2023-02-20
Change mark_cache size calculation to cut down memory usage

hongyunyan:

Issue Number: ref #6426

Problem Summary:

Currently, we calculate the mark_cache memory usage with marks.size() * sizeof(MarkInCompressedFile);
However, PODArray will do roundUpToPowerOfTwoOrZero when allocate element memory usage.
Thus, the memory usage is quite bigger then the calculation ways, which is close in twice time of the calculation ways.
To ensure the mark_cache only use the memory as we allocate, we change the memory usage calculation approach.

Then, we take the memory of allocation for MarkInCompressedFile for consideration. I think when we just initialization MarkInCompressedFile with the element number, we don't need to make the memory roundUpToPowerOfTwoOrZero. Thus, we change the alloc in  PODArray(size n), to cut down the usage.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

I compare the cluster with master tiflash and tiflash of this PR, to run a typical Workload.
We can find for the master tiflash, the memory will reach 13.0GB steeply and then continue slowing growth. 
While for the tilfash of this PR, the memory will reach 11.5GB steeply and then continue slowing growth.
Thus, in a typical workload, we can cut down nearly 1.5GB.

image
image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce the resident memory usage of mark_cache.
```


dbms/src/Common/PODArray.h
```diff
@@ -116,7 +116,7 @@ class PODArrayBase : private boost::noncopyable
 
     void alloc_for_num_elements(size_t num_elements)
     {
-        alloc(roundUpToPowerOfTwoOrZero(minimum_memory_for_elements(num_elements)));
+        alloc(minimum_memory_for_elements(num_elements));
```
flowbehappy, line 119: Plz add some comments to this method. About why don't we use roundUpToPowerOfTwoOrZero.

breezewish (approved):
The reduce of memory seems to be not as notable as expected. But anyway it's an improvement. Congrats!

PR #6869, opened 2023-02-21
Support rate limiting for Local Page Cache

breezewish:
When active queries need to occupy too much space in Local Page Cache, queries will be blocked until space is sufficient.


dbms/src/Storages/DeltaMerge/Remote/RemoteReadTask.cpp
```diff
@@ -421,36 +400,32 @@ RemoteSegmentReadTaskPtr RemoteSegmentReadTask::buildFrom(
         table_id,
         proto);
 
+    // Note: The segment_snap here is not readable.
```
flowbehappy, line 403: What do you mean by "not readable"? Can you explain more specifically?

dbms/src/Storages/DeltaMerge/Remote/RemoteReadTask.h
```diff
@@ -254,12 +252,12 @@ class RemoteSegmentReadTask
...
-    std::vector<UInt64> pending_page_ids;
-    size_t total_num_cftiny;
+    // std::vector<UInt64> pending_page_ids;
+    //    size_t total_num_cftiny;
 
 public:
-    std::atomic<size_t> num_msg_to_consume;
-    std::atomic<size_t> num_msg_consumed;
+    std::atomic<size_t> num_msg_to_consume = 0;
+    std::atomic<size_t> num_msg_consumed = 0;
```
flowbehappy, line 260:
```suggestion
    std::atomic<size_t> num_msg_to_consume{0};
    std::atomic<size_t> num_msg_consumed{0};
```

dbms/src/Storages/DeltaMerge/Remote/RemoteReadTask.cpp
breezewish: We should not read from the segment_snap at this moment, because it has not been connected to an underlying storage.

I will enhance the comment when cherry picking to master.

breezewish:
Close in favor of https://github.com/pingcap/tiflash/pull/6961

PR #6879, opened 2023-02-22
Fix compile with option TIFLASH_ENABLE_AVX_SUPPORT=OFF

### LINKED ISSUE #6877: Compile error when building with option `TIFLASH_ENABLE_AVX_SUPPORT=OFF`
## Enhancement

Compile error when compiling without `AVX2` feature support using cmake option `-DTIFLASH_ENABLE_AVX_SUPPORT=OFF`

```
In file included from /home/dev/workspace/tiflash-upstream/libs/libcommon/src/StringRef.cpp:17:
In file included from /home/dev/workspace/tiflash-upstream/libs/libcommon/include/common/StringRef.h:19:
/home/dev/workspace/tiflash-upstream/libs/libcommon/include/common/mem_utils_opt.h:87:17: error: no member named 'memcmp' in namespace 'std'; did you mean simply 'memcmp'?
    return 0 == std::memcmp(lhs.data(), rhs.data(), lhs.size());
                ^~~~~~~~~~~
                memcmp
/usr/include/string.h:64:12: note: 'memcmp' declared here
extern int memcmp (const void *__s1, const void *__s2, size_t __n)
           ^
1 error generated.
[3576/4521] Building CXX object libs/libcommon/CMakeFiles/common.dir/src/DateLUT.cpp.o
ninja: build stopped: subcommand failed.
```

yongman:

Issue Number: close #6877 

Problem Summary:

Add `TIFLASH_ENABLE_AVX_SUPPORT` check in `ToBits64` and `CountBytesInFilter`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/StringUtils/StringUtils.h
+++ b/dbms/src/Common/StringUtils/StringUtils.h
@@ -18,6 +18,7 @@
 #include <cstring>
 #include <string>
 #include <string_view>
+#include <tuple>
 
 
 namespace detail
--- a/cmake/cpu_features.cmake
+++ b/cmake/cpu_features.cmake
@@ -94,7 +94,7 @@ elseif (ARCH_AMD64)
     # and note that the compile will succeed even if the host machine does not support the instruction set
     # so we do not set the flags to avoid core dump in old machines
     option (TIFLASH_ENABLE_AVX_SUPPORT "Use AVX/AVX2 instructions on x86_64" ON)
-    option (TIFLASH_ENABLE_AVX512_SUPPORT "Use AVX512 instructions on x86_64" ON)
+    option (TIFLASH_ENABLE_AVX512_SUPPORT "Use AVX512 instructions on x86_64" OFF)
     
     # `haswell` was released since 2013 with cpu feature avx2, bmi2. It's a practical arch for optimizer
     option (TIFLASH_ENABLE_ARCH_HASWELL_SUPPORT "Use instructions based on architecture `haswell` on x86_64" ON)
--- a/cmake/cpu_features.cmake
+++ b/cmake/cpu_features.cmake
@@ -99,8 +99,8 @@ elseif (ARCH_AMD64)
     # `haswell` was released since 2013 with cpu feature avx2, bmi2. It's a practical arch for optimizer
     option (TIFLASH_ENABLE_ARCH_HASWELL_SUPPORT "Use instructions based on architecture `haswell` on x86_64" ON)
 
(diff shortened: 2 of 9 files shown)


dbms/src/Common/StringUtils/StringUtils.h
```diff
@@ -18,6 +18,7 @@
 #include <cstring>
 #include <string>
 #include <string_view>
+#include <tuple>
```
Lloyd-Pottiger, line 22: why add this?

Lloyd-Pottiger (approved):

solotzg:
You can cmake option `NO_SSE42_OR_HIGHER` to disable features avx, avx2, ... and later.

solotzg:
@Lloyd-Pottiger `NO_SSE42_OR_HIGHER` it looks like a typo. We still need sse4.2 by default.

Lloyd-Pottiger:
> You can cmake option `NO_SSE42_OR_HIGHER` to disable features avx, avx2, ... and later.

when we set flag `-mavx2` to a file, the `__AVX2__  = 1` whether the machine support `avx2` or not. We may still need `TIFLASH_ENABLE_AVX_SUPPORT` to dispatch.

> @Lloyd-Pottiger `NO_SSE42_OR_HIGHER` it looks like a typo. We still need sse4.2 under AMD64 platforms by default.

yes, it is a typo. @yongman could you please help rename `NO_SSE42_OR_HIGHER` to `NO_AVX_OR_HIGHER` in this pr?

solotzg:
-march=haswell or later platform  can also define macro __AVX2__. Since avx2 is the default required feature of tiflash, I think macro __AVX2__ is enough.

solotzg:
We need to guarantee that if the platform used for compiling can not support avx2, NO_SSE42_OR_HIGHER should be set to ON.

yongman:
@solotzg  Set `NO_SSE42_OR_HIGHER` to `ON` also has some errors for now, I will check if it works after another fix PR.

Lloyd-Pottiger:
> We need to guarantee that if the platform used for compiling can not support avx2, NO_SSE42_OR_HIGHER should be set to ON.

[check_cxx_source_runs](https://cmake.org/cmake/help/latest/module/CheckCXXSourceRuns.html) may help, but it only exists in v3.26.0 or higher of cmake.

solotzg:
> @solotzg Set `NO_SSE42_OR_HIGHER` to `ON` also has some errors for now, I will check if it works after another fix PR.

To be fixed by #6879

Issue #6233, opened 2022-11-02
Refine code or doc


LittleFall:
## Enhancement

just a stub issue, welcome to ref this in your pr when you think it needn't open a new issue.

Don't abuse it!

PR #6884, opened 2023-02-23
Storage: Decouple compile units

breezewish:

Problem Summary:

1. `WriteBatch.h`, `WriteBatches.h`, `UniversalWriteBatch.h` and `WriteBatchWrapper.h` are changed into `WriteBatchImpl.h`, `WriteBatchesImpl.h`, etc, to explicitly indicate that it includes many implementations and the content is very likely to change.

2. Change to only import these `XxxImpl.h` in `XxxImpl.h` or `Xxx.cpp` (not in `Xxx.h` files), so that changing the content of `XxxImpl.h` will not cause too many files to recompile.

3. Introduce `PageStorage_fwd.h`, `StoragePool_fwd.h`, `PathPool_fwd.h`, which only includes forward declarations. We usually use these classes in the form of `XxxPtr` so that we can just include `Xxx_fwd.h` in normal header files `Xxx.h`. For example, in most of headers, `PageStorage.h` is no longer included, instead `PageStorage_fwd.h` is included. In this way, even if we have changed the interface of `PageStorage`, the changing will not cause a wide range of files to be recompiled. Only `.cpp` files directly imported `PageStorage.h` will be recompiled.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/Universal/UniversalPageId.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <Storages/Page/V3/Universal/UniversalPageId.h>
+#include <Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h>
+
+namespace DB::details
+{
+
+String UniversalPageIdFormatHelper::format(const DB::UniversalPageId & value)
+{
+    auto prefix = DB::UniversalPageIdFormat::getFullPrefix(value);
+    return fmt::format("{}.{}", Redact::keyToHexString(prefix.data(), prefix.size()), DB::UniversalPageIdFormat::getU64ID(value));
```
JaySon-Huang, line 24:
```suggestion
    return fmt::format("0x{}.{}", Redact::keyToHexString(prefix.data(), prefix.size()), DB::UniversalPageIdFormat::getU64ID(value));
```

dbms/src/Storages/Page/V3/PageEntry.h
```diff
@@ -79,25 +79,30 @@ using PageEntriesV3 = std::vector<PageEntryV3>;
...
+
+        FmtBuffer fmt_buf;
+        fmt_buf.joinStr(
+            entry.field_offsets.begin(),
+            entry.field_offsets.end(),
+            [](const auto & offset_checksum, FmtBuffer & fb) {
+                fb.fmtAppend("{}", offset_checksum.first);
+            },
+            ",");
+        return format_to(ctx.out(), "PageEntryV3{{file: {}, offset: 0x{:X}, size: {}, checksum: 0x{:X}, tag: {}, field_offsets: [{}]}}", entry.file_id, entry.offset, entry.size, entry.checksum, entry.tag, fmt_buf.toString());
```
JaySon-Huang, line 106:
```suggestion
        return format_to(ctx.out(), "PageEntry{{file: {}, offset: 0x{:X}, size: {}, checksum: 0x{:X}, tag: {}, field_offsets: [{}]}}", entry.file_id, entry.offset, entry.size, entry.checksum, entry.tag, fmt_buf.toString());
```

JaySon-Huang (approved):
LGTM with small comment

lidezhu (approved):

Issue #6827, opened 2023-02-15
TiFlash directly read/write S3


JinheLin:
New configuration introduced by https://github.com/pingcap/tiflash/pull/6828:

```
[storage.s3]
endpoint = "127.0.0.1:8080"
bucket = "s3_bucket"
```

and these two config are read from env

* `AWS_ACCESS_KEY_ID`
* `AWS_SECRET_ACCESS_KEY`

JaySon-Huang:
ref https://github.com/pingcap/tiflash/issues/6882

PR #6900, opened 2023-02-26
Add checkpoint utilities

breezewish:

Problem Summary:

This PR introduces these utilities to work with writing and reading PageStorage checkpoints:

- DB::PS::V3::CPDataFileWriter
- DB::PS::V3::CPFilesWriter
- DB::PS::V3::CPManifestFileWriter
- DB::PS::V3::CPManifestFileReader

Usually, you need to write ManifestFile and DataFile via `CPFilesWriter`, and then, read the ManifestFile via `CPManifestFileReader` and read the DataFile via ReadBufferFromFile or something similar.

The manifest file is compressed and the data file is **not**. Manifest file contains Protobuf data and the file format is designed to be friendly for streaming, i.e. you can read / write data piece by piece (instead of all at once) to avoid OOMs.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    // TODO: getMaterializedBytes only works for FramedChecksumWriteBuffer, but does not work for a normal WriteBufferFromFile.
+    //       There must be something wrong and should be fixed.
+    // uint64_t file_offset = file_writer->getMaterializedBytes();
+    // file_writer->write(data, n);
+    // uint64_t write_n = file_writer->getMaterializedBytes() - file_offset;
+
+    uint64_t file_offset = file_writer->count();
+    file_writer->write(data, n);
+    uint64_t write_n = file_writer->count() - file_offset;
```
flowbehappy, line 43: RUNTIME_CHECK_MSG(write_n == n);

dbms/src/Storages/Page/V3/CheckpointFile/Proto/data_file.proto
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.#pragma once
+
+syntax = "proto3";
+
+package DB.PS.V3.CheckpointProto;
+
+import "common.proto";
+
+// This prefix is only for debugging purpose.
```
flowbehappy, line 21: Why "only for debugging purpose"? It is used in many places.

dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+void CPDataFileWriter::writeSuffix()
+{
+    if (write_stage == WriteStage::WritingFinished)
+        return; // writeSuffix can be called multiple times without causing issues.
+    if (write_stage != WriteStage::WritingRecords)
+        RUNTIME_CHECK_MSG(false, "unexpected write stage {}", magic_enum::enum_name(write_stage));
+
+    std::string json;
+    google::protobuf::util::MessageToJsonString(file_suffix, &json);
+    writeStringBinary(json, *file_writer);
```
flowbehappy, line 68: Why use json format? I'm concerned about the ser/deser performance.

breezewish: The suffix part of the data file is not used. It is appended only for debug purpose, for example, we can learn or parse the content of the data file easily without relying on a manifest file when we only have this data file.

dbms/src/Storages/Page/V3/CheckpointFile/Proto/data_file.proto
breezewish: The prefix of the data file is not used. In the read path we should just read data from the data file using a specified seek offset.

dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileWriter.cpp
flowbehappy: OK. Please add some comments about those messages.

dbms/src/Storages/Page/V3/CheckpointFile/Proto/data_file.proto
flowbehappy: OK. Can you put more comments here about the information?

dbms/src/Storages/Page/V3/PageEntryCheckpointInfo.h
```diff
@@ -0,0 +1,83 @@
...
+
+    std::shared_ptr<const std::string> data_file_id;
+
+    uint64_t offset_in_file = 0;
+    uint64_t size_in_file = 0;
+
+    CheckpointProto::EntryDataLocation toProto() const
+    {
+        CheckpointProto::EntryDataLocation proto_rec;
+        proto_rec.set_data_file_id(*data_file_id);
```
flowbehappy: There will be many duplicate `data_file_id` and consume lots of disk space. Can we use a dictionary to optimize?

dbms/src/Storages/Page/V3/CheckpointFile/Proto/common.proto
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.#pragma once
+
+syntax = "proto3";
+
+package DB.PS.V3.CheckpointProto;
+
+message WriterInfo {
+    uint64 store_id = 1;
+    string version = 2;
+    string version_git = 3;
```
JaySon-Huang, line 22: why we need this version_git?

dbms/src/Storages/Page/V3/PageEntryCheckpointInfo.h
breezewish: The manifest file is written over a LZ4 compressed format. Duplicated data file ids will be eliminated away.

I tested with 10000 records with the same 76 byte long DataFileId. The generated manifest file is 92KB which means the materialized size is only 9.2 bytes / record.

dbms/src/Storages/Page/V3/CheckpointFile/Proto/common.proto
breezewish: We can fill the git commit of the version. It may be helpful for us to troubleshoot issues by precisely pinpoint the exact code commit when we have this file.

dbms/src/Storages/Page/V3/CheckpointFile/CPFilesWriter.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+namespace DB::PS::V3
+{
+
+CPFilesWriter::CPFilesWriter(CPFilesWriter::Options options)
+    : manifest_file_id(options.manifest_file_id)
+    , data_writer(CPDataFileWriter::create({
+          .file_path = options.data_file_path,
+          .file_id = options.data_file_id,
+      }))
```
JaySon-Huang, line 25: `CPDataFileWriter::Options::file_id` should be the lockkey to the CheckpointDataFile, which is generated by `S3FilenameView:: getLockKey`.

What about renaming it to `CPDataFileWriter::Options::lock_file`?

https://github.com/pingcap/tiflash/blob/da5c10dd76cc76ceba4cfad2c538df0de40d7001/dbms/src/Storages/S3/S3Filename.h#L88-L89

dbms/src/Storages/Page/V3/PageEntryCheckpointInfo.h
```diff
@@ -0,0 +1,58 @@
...
+namespace CheckpointProto
+{
+using StringsInternMap = std::unordered_map<std::string_view, std::shared_ptr<const std::string>>;
+}
+
+struct CheckpointLocation
+{
+    // This struct is highly coupled with manifest_file.proto -> EditEntry.
+
+    std::shared_ptr<const std::string> data_file_id;
```
JaySon-Huang, line 31: ... and CheckpointLocation.data_file_id -> CheckpointLocation.lock_key

dbms/src/Storages/Page/V3/PageEntryCheckpointInfo.h
flowbehappy: The compression ratio looks good.

Since we need to put all those file_ids into a temporary object `ManifestFileEditsPart`, and the file_ids are already normalized, another thing is the memory consumption could be large. E.g. 10 000 000 records with 76 bytes each takes 724MiB. It looks acceptable.

@lidezhu Maybe we can also using LZ4 like here to generate GC dump file during PageStorage Full GC.

flowbehappy: @breezewish BTW, is it possible to use a streaming way to generate `ManifestFileEditsPart` in the future? I'm concerned that the page number could be larger in extreme cases.

breezewish: > Since we need to put all those file_ids into a temporary object ManifestFileEditsPart, and the file_ids are already normalized, another thing is the memory consumption could be large. E.g. 10 000 000 records with 76 bytes each takes 724MiB.

It's true. Our API allows caller to split into multiple calls to reduce temp memory overhead as well. For example, we may split into multiple writes with each write consumes 10000 records from the memory. This effectively reduces peak temp memory overhead to only 760KB.

dbms/src/Storages/Page/V3/CheckpointFile/CPManifestFileReader.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    RUNTIME_CHECK_MSG(read_stage == ReadStage::ReadingPrefix, "unexpected read stage {}", magic_enum::enum_name(read_stage));
+
+    CheckpointProto::ManifestFilePrefix ret;
+    details::readMessageWithLength(*compressed_reader, ret);
+    read_stage = ReadStage::ReadingEdits;
+    return ret;
+}
+
+std::optional<universal::PageEntriesEdit> CPManifestFileReader::readEdits(CheckpointProto::StringsInternMap & strings_map)
```
flowbehappy, line 33: It is possible to use multiple threads to call this method in parallel in the future?

flowbehappy: Looks like we need to store the `offset` and `length` of each batch somewhere.

dbms/src/Storages/Page/V3/CheckpointFile/CPFilesWriter.cpp
breezewish: I'm not sure. Seems that the relationship is that, the lock key should be identical to the data file id, instead of in the reserve way.

Data file id should be directly tied with Data file path, as a replacement (or shorten, or a format easy to deal with) for the Data file path.

dbms/src/Storages/Page/V3/CheckpointFile/Proto/manifest_file.proto
```diff
@@ -0,0 +1,86 @@
...
+A manifest file format V1 consists of the following parts:
+
+- Len|Prefix
+- Len|EditsPart
+- Len|LocksPart
+...
+
+ */
+
+message ManifestFilePrefix {
```
JaySon-Huang, line 32: do we need to add some fields about the compress algorithm so that we can change the algorithm later?

dbms/src/Storages/Page/V3/CheckpointFile/tests/gtest_file_read_write.cpp
```diff
@@ -0,0 +1,634 @@
...
+try
+{
+    const auto delegator = std::make_shared<DB::tests::MockDiskDelegatorMulti>(std::vector{dir});
+    const auto file_provider = DB::tests::TiFlashTestEnv::getContext().getFileProvider();
+    auto blob_store = BlobStore<universal::BlobStoreTrait>(getCurrentTestName(), file_provider, delegator, BlobConfig{});
+
+    auto edits = universal::PageEntriesEdit{};
+    {
+        UniversalWriteBatch wb;
+        wb.putPage("page_foo", 0, "The flower carriage rocked");
```
lidezhu, line 357: How about add a page with `data_sizes`, and make sure the `fields_offset` is correct from manifest file?

dbms/src/Storages/Page/V3/CheckpointFile/Proto/manifest_file.proto
breezewish: Yes, we can add some settings there when we want to change compression algorithm.

dbms/src/Storages/Page/V3/CheckpointFile/CPManifestFileReader.cpp
breezewish: May be better to consider split into multiple sub-manifest files, when one file is too large, if we want to increase parallelism.

flowbehappy: OK. I think we can do split job later.

dbms/src/Storages/Page/V3/CheckpointFile/tests/gtest_file_read_write.cpp
breezewish: Done. Please take a look again, thanks!

JaySon-Huang (approved):
LGTM

lidezhu (approved):
LGTM

Issue #6908, opened 2023-02-27
The task in BackgroundProcessingPool is not run by fixed interval


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Notice when testing https://github.com/pingcap/tiflash/pull/6899, also reported in https://github.com/pingcap/tiflash/pull/6816#discussion_r1108205099

### 2. What did you expect to see? (Required)
The task which always returns `false` run with a fixed interval

### 3. What did you see instead (Required)
The task get executed frequently by different threads in BackgroundProcessingPool

### 4. What is your TiFlash version? (Required)
master

JaySon-Huang:
It is because the task handle with min_time will be poped to multiple worker threads. Though the task with `multi=false` can prevent a task being run concurrently, it can not prevent those worker threads one by one run the task multiple times within an interval.

https://github.com/pingcap/tiflash/blob/e1418dc066f12983c74a4112932f3fca16c56cb4/dbms/src/Storages/BackgroundProcessingPool.cpp#L173-L190

JaySon-Huang:
With a background pool with 10 threads, and one task expecting to run at 2 seconds intervals. It actually gets executed more than 35 times within 10 seconds.

<details>
<summary>click expand to see the unit test</summary>

```
TEST(BackgroundProcessingPoolTest, FixedInterval)
{
    BackgroundProcessingPool pool(10, "test");

    using Clock = std::chrono::system_clock;
    using TimePoint = std::chrono::time_point<Clock>;

    TimePoint last_update_timepoint = Clock::now();

    using namespace std::chrono_literals;
    const auto sleep_seconds = 10s;
    const Int64 expect_interval_ms = 2 * 1000;
    const auto num_expect_called = 5;

    Int64 num_actual_called = 0;
    Int64 min_diff_ms = std::numeric_limits<Int64>::max();
    auto task = pool.addTask(
        [&]() {
            num_actual_called += 1;
            auto diff_ms = std::chrono::duration_cast<std::chrono::milliseconds>(Clock::now() - last_update_timepoint).count();
            if (diff_ms < expect_interval_ms / 2)
            {
                LOG_ERROR(Logger::get(), "Unexpected frequent call, actual interval={}ms", diff_ms);
            }
            min_diff_ms = std::min(min_diff_ms, diff_ms);

            last_update_timepoint = Clock::now();
            return false; // expected to be run with a fixed interval
        },
[...]

PR #6910, opened 2023-02-28
Fix the tasks in bkg pool may be called more frequently than expected

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/6908

Problem Summary:
When we add a task into BackgroundProcessingPool which always return false, we expect that will be run at a fixed interval. (Almost fixed within small random the BackgroundProcessingPool random part)

However, a task may be popped into multiple threads and they can run the task one by one. That leads to more frequent calls than expected.
We have to add extra logic to avoid the task being unexpectedly called too frequently. That is annoying.

* this is a known isssue
https://github.com/pingcap/tiflash/blob/e1418dc066f12983c74a4112932f3fca16c56cb4/dbms/src/Storages/BackgroundProcessingPool.h#L92-L98

* extra codes to handle unexpected frequent calls
https://github.com/pingcap/tiflash/blob/e1418dc066f12983c74a4112932f3fca16c56cb4/dbms/src/Storages/Page/V3/Universal/UniversalPageStorageService.cpp#L38-L45
https://github.com/pingcap/tiflash/blob/e1418dc066f12983c74a4112932f3fca16c56cb4/dbms/src/Storages/DeltaMerge/StoragePool.cpp#L146-L156

Change one member of Task from `atomic<bool> occupied` into `size_t concurrent_executors`. `concurrent_executors` 
indicate how many worker threads are holding this task.
- If `task.multi == false`, only when `task.concurrent_executor == 0`, can the task pop to a worker thread for executing
- Changing to `task.concurrent_executor` is protected under the lock of `pool.tasks_mutex`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
  - port these changes to s3gc branch and check the frequency of etcd keepalive and s3gc being called
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang:
https://ci.pingcap.net/blue/organizations/jenkins/tiflash-sanitizer-daily/detail/tiflash-sanitizer-daily/917/pipeline

flowbehappy (approved):
LGTM

lidezhu (approved):
LGTM

Issue #6882, opened 2023-02-23
Support using S3 as main storage.


flowbehappy:
## Feature Request

* Separate TiFlash instances into read (compute) nodes and write nodes
* Read nodes are stateless nodes, responsible for answering queries requests
* Write nodes are small state nodes, responsible for writing data to S3
* All data is stored on S3, and can be read by all nodes

flowbehappy:
Design doc: https://github.com/pingcap/tiflash/blob/master/docs/design/2023-02-23-cloud-native-architecture.md

flowbehappy:
Addressed in 7.0.

PR #6930, opened 2023-03-02
Support remote read on uni ps

lidezhu:

Issue Number: ref #6882 

Problem Summary: Support read remote page on S3 directly inside Uni PS.

1. Add a new write type `WriteBatchWriteType::PUT_REMOTE` which just include the location info on S3 of the page;
2. Add a new type `EditRecordType::UPDATE_REMOTE` which is used to update local cache info for remote pages;
3. Determine whether to read locally or read from S3 by the `is_local_data_reclaimed` member inside `CheckpointInfo`;
4. When dump snapshot, do LZ4 compression on the record data;

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/PageEntriesEdit.h
```diff
@@ -90,6 +90,7 @@ enum class EditRecordType
     VAR_REF,
     VAR_EXTERNAL,
     VAR_DELETE,
+    UPDATE_REMOTE,
```
flowbehappy, line 93: For `UPDATE_REMOTE`, do we need to distinguish between snapshot types and in-memory types?

lidezhu: Don't get what is "snapshot types" and "in-memory types".

lidezhu: It won't reside in `PageDirectory`, it is just used to update the local cache info for `VAR_ENTRY` type.

flowbehappy: Then maybe we should put it after UPSERT

dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileReader.h
```diff
@@ -0,0 +1,58 @@
...
+class S3Client;
+}
+
+namespace DB::PS::V3
+{
+using UniversalPageMap = std::map<UniversalPageId, Page>;
+using UniversalPageIdAndEntry = std::pair<UniversalPageId, PS::V3::PageEntryV3>;
+using UniversalPageIdAndEntries = std::vector<UniversalPageIdAndEntry>;
+
+class CPDataFileReader : private Allocator<false>
```
flowbehappy, line 35: Please add documents about this class. The usage and the basic idea behind it.

lidezhu: Added.

dbms/src/Storages/Page/V3/PageEntriesEdit.h
lidezhu: The order of `EditRecordType` can't be changed. Because the order determines the underlying value for each enum value which will be written to the WAL.
Just add some comments in it to distinguish the new type between `VAR_` types.

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -258,10 +282,26 @@ BlobStore<Trait>::write(typename Trait::WriteBatch & wb, const WriteLimiterPtr &
...
+                PageEntryV3 entry;
+                entry.file_id = INVALID_BLOBFILE_ID;
+                entry.tag = write.tag;
+                entry.checkpoint_info = CheckpointInfo{
+                    .data_location = *write.data_location,
+                    .is_local_data_reclaimed = true,
+                };
+                if (!write.offsets.empty())
+                {
+                    entry.field_offsets.swap(write.offsets);
```
flowbehappy, line 300: If we use `swap` here, better use `Trait::WriteBatch && wb` because we changed `wb` already.

dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileReader.h
flowbehappy: I wonder whether `CPDataFileReader ` is a proper name or not. Because it does not contain any logic about "CheckPoint". It is more like a `S3PageReader`.

dbms/src/Storages/Page/V3/PageEntriesEdit.h
flowbehappy: OK

dbms/src/Storages/Page/V3/BlobStore.cpp
lidezhu: Fixed.

dbms/src/Storages/Page/V3/CheckpointFile/CPDataFileReader.h
flowbehappy: I think this class can be used to read Pages not only from checkpoints.  For example, we will implement evict local Page data later.

lidezhu: Fixed.

flowbehappy (approved):
LGTM

dbms/src/Storages/Page/V3/WAL/serialize.cpp
```diff
@@ -332,17 +388,61 @@ String Serializer<PageEntriesEdit>::serializeTo(const PageEntriesEdit & edit)
...
-    if (version != 1)
+    CompressedReadBuffer compressed_buf(buf);
+    switch (version)
+    {
+    case WALSerializeVersion::Plain:
+        DB::PS::V3::deserializeFrom(buf, edit);
+        break;
+    case WALSerializeVersion::LZ4:
+        DB::PS::V3::deserializeFrom(compressed_buf, edit);
+        break;
```
JaySon-Huang, line 442:
```suggestion
    switch (version)
    {
    case WALSerializeVersion::Plain:
        DB::PS::V3::deserializeFrom(buf, edit);
        break;
    case WALSerializeVersion::LZ4:
    {
        CompressedReadBuffer compressed_buf(buf);
        DB::PS::V3::deserializeFrom(compressed_buf, edit);
        break;
    }
```

dbms/src/Storages/Page/V3/Universal/UniversalWriteBatchImpl.h
```diff
@@ -101,6 +103,19 @@ class UniversalWriteBatch : private boost::noncopyable
         putPage(page_id, tag, buffer_ptr, data.size(), data_sizes);
     }
 
+    void putRemotePage(const UniversalPageId & page_id, UInt64 tag, const PS::V3::CheckpointLocation & data_location, PageFieldOffsetChecksums && offset_and_checksums)
```
JaySon-Huang, line 106: I think we also need to add `putRemoteExternal(const UniversalPageId & page_id, const PS::V3::CheckpointLocation & data_location)` to ingest a remote dmfile into UniPS and restore it after restart?

dbms/src/Storages/Page/WriteBatchImpl.h
```diff
@@ -41,6 +42,10 @@ enum class WriteBatchWriteType : UInt8
     // In V2, it is the same as `PUT`; In V3, we treated it as a different type from `PUT`
     // to get its lifetime management correct.
     PUT_EXTERNAL = 4,
+    // Put a remote page which data is stored on remote storage
+    PUT_REMOTE = 5,
+    // Update local cache for remote page
+    UPDATE_REMOTE = 6,
```
JaySon-Huang, line 48: Suggest renaming this to be `UPDATE_DATA_FROM_REMOTE`

dbms/src/Storages/Page/tools/PageCtl/PageStorageCtlV2.cpp
```diff
@@ -277,6 +277,7 @@ void dump_all_entries(PageFileSet & page_files, int32_t mode)
                     break;
                 default:
                     throw DB::Exception(DB::ErrorCodes::LOGICAL_ERROR, "illegal type: {}", magic_enum::enum_name(record.type));
+                    break;
```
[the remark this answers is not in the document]
lidezhu: It is not necessary. It is modified accidentaly. I left it here just for consistency, because sometimes we add this extra line: https://github.com/pingcap/tiflash/blob/e51175250dbd97b22bac40a35d89713616da8b87/dbms/src/Storages/Page/V2/PageFile.cpp#L125

dbms/src/Storages/Page/WriteBatchImpl.h
lidezhu: Renamed.

dbms/src/Storages/Page/V3/Universal/UniversalWriteBatchImpl.h
lidezhu: I suggest put it in the later pr about FAP. Because then we can determine how to implement `putRemoteExternal`.

dbms/src/Storages/Page/V3/WAL/serialize.cpp
lidezhu: It seems we cannot write it in this way because my editor give me an error like "bypass variable initialization".

dbms/src/Storages/Page/V3/Universal/UniversalWriteBatchImpl.h
JaySon-Huang: OK

dbms/src/Storages/Page/V3/WAL/serialize.cpp
```diff
@@ -114,13 +153,20 @@ void serializePutTo(const EditRecord & record, WriteBuffer & buf)
...
     writeIntBinary(record.being_ref_count, buf);
 
-    serializeEntryTo(record.entry, buf);
+    if constexpr (std::is_same_v<EditRecord, u128::PageEntriesEdit::EditRecord>)
+    {
+        serializeEntryTo(record.entry, buf);
+    }
+    else if constexpr (std::is_same_v<EditRecord, universal::PageEntriesEdit::EditRecord>)
+    {
+        serializeEntryWithCheckpointInfoTo(record.entry, buf);
```
JaySon-Huang, line 162: maybe we can put

- has_checkpoint_info or not
- local data reclaimed

as two bits in the `flags` field?

dbms/src/Storages/Page/V3/WAL/serialize.cpp
```diff
@@ -82,6 +101,26 @@ inline void deserializeEntryFrom(ReadBuffer & buf, PageEntryV3 & entry)
...
+    {
+        CheckpointInfo checkpoint_info;
+        UInt32 is_local_data_reclaimed;
+        readIntBinary(is_local_data_reclaimed, buf);
+        checkpoint_info.is_local_data_reclaimed = is_local_data_reclaimed;
+        readIntBinary(checkpoint_info.data_location.offset_in_file, buf);
+        readIntBinary(checkpoint_info.data_location.size_in_file, buf);
+        String data_file_id;
+        readStringBinary(data_file_id, buf);
+        checkpoint_info.data_location.data_file_id = std::make_shared<String>(data_file_id);
```
JaySon-Huang, line 119:
```suggestion
         // TODO: different entries' data_file_id could be highly duplicated, try to reuse the ptr to reduce memory overhead later.
        checkpoint_info.data_location.data_file_id = std::make_shared<String>(data_file_id);
```

dbms/src/Storages/Page/V3/PageEntry.h
```diff
@@ -64,7 +67,16 @@ struct PageEntryV3
                                         field_offsets.size()),
                             ErrorCodes::LOGICAL_ERROR);
         else if (index == field_offsets.size() - 1)
-            return size - field_offsets.back().first;
+        {
+            if (checkpoint_info.has_value() && checkpoint_info->is_local_data_reclaimed)
+            {
+                return checkpoint_info->data_location.size_in_file - field_offsets.back().first;
```
JaySon-Huang, line 73:
```suggestion
                // entry.size is not reliable under this case, use the size_in_file in checkpoint_info instead
                return checkpoint_info->data_location.size_in_file - field_offsets.back().first;
```

dbms/src/Storages/Page/V3/PageEntry.h
```diff
@@ -77,7 +89,16 @@ struct PageEntryV3
                 fmt::format("Try to getFieldOffsets with invalid index [index={}] [fields_size={}]", index, field_offsets.size()),
                 ErrorCodes::LOGICAL_ERROR);
         else if (index == field_offsets.size() - 1)
-            return {field_offsets.back().first, size};
+        {
+            if (checkpoint_info.has_value() && checkpoint_info->is_local_data_reclaimed)
+            {
+                return {field_offsets.back().first, checkpoint_info->data_location.size_in_file};
```
JaySon-Huang, line 95:
```suggestion
                // entry.size is not reliable under this case, use the size_in_file in checkpoint_info instead
                return {field_offsets.back().first, checkpoint_info->data_location.size_in_file};
```

dbms/src/Storages/Page/V3/Universal/S3PageReader.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+std::pair<UniversalPageMap, UniversalPageMap> S3PageReader::read(const FieldReadInfos & to_read)
+{
+    UniversalPageMap complete_page_map;
+    size_t read_fields_size = 0;
+    for (const auto & read_info : to_read)
+    {
+        const auto & page_entry = read_info.entry;
+        complete_page_map.emplace(read_info.page_id, read(std::make_pair(read_info.page_id, page_entry)));
```
JaySon-Huang, line 70:
```suggestion
        // read the whole page from S3 and save it as `complete_page`
        complete_page_map.emplace(read_info.page_id, read(std::make_pair(read_info.page_id, page_entry)));
```

dbms/src/Storages/Page/V3/WAL/serialize.cpp
lidezhu: Great suggestion. And I find we don't even need to store the info about whether local data is reclaimed. We can determine it just from `entry.file_id`.

JaySon-Huang (approved):
The rest LGTM

PR #6112, opened 2022-10-11
Introduce DevContainers

breezewish:
### What problem does this PR solve?

Issue Number: ref #5991

Problem Summary:

### What is changed and how it works?

A [DevContainer](https://containers.dev) that can build TiFlash.

You may use one of the following ways to utilize this DevContainer:

1. Use VsCode

   Install the [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) extension pack, and use VsCode to open a local or even remote (via SSH) TiFlash source code directory. A prompt will be shown to ask whether you want to use the DevContainer:

   image

   By Clicking "Reopen in Container", now you go! You will have a develop environment with the following features **with zero configuration**:

   - C++ Code Completion in VsCode (powered by Clangd)

     image

   - One click to configure (using cmake) and build TiFlash (using Ninja), or even switch different build targets and profiles:
     
     image

   - And even test selection via UI, if `gtests_dbms` is built:

   - The VsCode terminal will also work, in case of you want to invoke customized build commands:

     image

   - Cache just works, cache is persisted even if the docker container is restarted or rebuilt, including: Rust toolchains, Rust dependencies from Cargo, ccache.

2. Use Docker

   The DevContainer is just a standard docker image with dev tools configured. So you can also just use Docker, for example, in a Linux x86_64 machine, you could just:

   ```shell
   $ cd tiflash/.devcontainer
   $ make dev   # Build tiflash using debug profile
   ```

**Base image:**

CentOS 7 (x86_64)

The built artifact is portable for most Linux x86_64 distributions because it is built against CentOS 7 which has a rather old glibc version.

**Built-in components:**

- Clang 15.0.7
- Very recent git, zsh, cmake, ccache, Python3, GitHub CLI

**Work items:**

- [x] Open-source the image builder
- [x] Use docker-compose so that mapping volumes will just work even without using VsCode or DevContainer CLI
- [x] Support x86_64
- [ ] Support ARM
- [ ] Support memoizing bash and zsh histories
- [ ] Include more handy tools in the image, e.g. fzf, fd, rg
- [ ] Publish the image to the CI's image registry
- [ ] Change CI release to be based on this image
- [ ] Support easy-to-run clang-format and clang-tidy in the DevContainer

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
[...]

PR #6938, opened 2023-03-03
README: add instructions for  installing llvm clang in macOS

### LINKED ISSUE #5265: Clang 15 support

Lloyd-Pottiger:

Issue Number: close #5265 

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


README.md
```diff
@@ -125,6 +125,27 @@ xcode-select --install
...
+# to make sure llvm first in your PATH
+export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
+export CC="/opt/homebrew/opt/llvm/bin/clang"
+export CXX="/opt/homebrew/opt/llvm/bin/clang++"
+
+# or use CMAKE_C_COMPILER and CMAKE_CXX_COMPILER to specify the compiler, like this:
+mkdir cmake-build-debug  # The directory name can be customized
+cd cmake-build-debug
+cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG -DCMAKE_C_COMPILER=/opt/homebrew/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/opt/homebrew/opt/llvm/bin/clang++
+ninja tiflash
```
breezewish, line 146: Maybe we should move this part to the `### 3. Build`.

breezewish (approved):
Working!

README.md
```diff
@@ -125,6 +125,27 @@ xcode-select --install
...
 
+The default version of clang in MacOS 13.0 is Apple clang 14.0.0. If your operating system is lower than MacOS 13.0, you can install llvm clang.
+
+```shell
+brew install llvm
+
+# check llvm version
+clang --version # should be 15.0.0 or higher
+
+# Add the following line to your ~/.zshrc
```
flowbehappy, line 136:
```suggestion
# Add the following line to your shell environment, e.g. ~/.bash_profile
```

README.md
```diff
@@ -125,6 +125,27 @@ xcode-select --install
 brew install ninja cmake openssl@1.1 ccache
 ```
 
+The default version of clang in MacOS 13.0 is Apple clang 14.0.0. If your operating system is lower than MacOS 13.0, you can install llvm clang.
+
+```shell
+brew install llvm
+
+# check llvm version
+clang --version # should be 15.0.0 or higher
```
flowbehappy, line 134:
```suggestion
clang --version # should be 15.0.0 or higher
```
```

Comment (breezewish) [inline README.md:134]: How about just installing `llvm@15` explicitly?

File: README.md (hunk 1 above)
Comment (Lloyd-Pottiger) [reply README.md]: moved

File: README.md [hunk 4]
```diff
@@ -157,6 +166,19 @@ cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG \
   -DCMAKE_CXX_COMPILER=/usr/bin/clang++-14
 ```
 
+In MacOS, if you install llvm clang, you need to explicitly specify to use llvm clang.
```
breezewish, line 169: I recommend to only give 1 working solution, instead of multiple possible solutions and let user decide, which is not necessary.

For example, the followings should be always working (in MacOS):

1. Always instruct user to `brew install llvm@15`
2. Use the following to compile:

```
export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
 export CC="/opt/homebrew/opt/llvm/bin/clang"
 export CXX="/opt/homebrew/opt/llvm/bin/clang++"

mkdir cmake-build-debug  # The directory name can be customized
cd cmake-build-debug
cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG
ninja tiflash
```

README.md
Lloyd-Pottiger: done

README.md
breezewish: BTW this explicitly introduces LLVM 15, which should help us adopt more CXX 20 features.

README.md
```diff
@@ -157,6 +166,19 @@ cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG \
...
+# Add the following line to your shell environment, e.g. ~/.bash_profile
+# to make sure llvm first in your PATH
+export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
+export CC="/opt/homebrew/opt/llvm/bin/clang"
+export CXX="/opt/homebrew/opt/llvm/bin/clang++"
+
+# or use CMAKE_C_COMPILER and CMAKE_CXX_COMPILER to specify the compiler, like this:
+cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG -DCMAKE_C_COMPILER=/opt/homebrew/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/opt/homebrew/opt/llvm/bin/clang++
+```
+
```
flowbehappy, line 181:
```suggestion
Add the following line to your shell environment, e.g. `~/.bash_profile`.
```shell
# Make sure llvm comes first in your PATH
export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
export CC="/opt/homebrew/opt/llvm/bin/clang"
export CXX="/opt/homebrew/opt/llvm/bin/clang++"
```

Or use `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` to specify the compiler. For example:
```shell
cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG -DCMAKE_C_COMPILER=/opt/homebrew/opt/llvm/bin/clang -DCMAKE_CXX_COMPILER=/opt/homebrew/opt/llvm/bin/clang++
```

```

File: README.md (hunk 4 above)
Comment (Lloyd-Pottiger) [reply README.md:169]: The maintainer: "We should let the developers use minimum cost to achieve building tiflash, and support Apple clang can achieve that."

Comment (Lloyd-Pottiger) [reply README.md:169]: llvm 14 and apple clang 14 both are enough.

File: README.md [hunk 6]
```diff
@@ -157,6 +166,20 @@ cmake .. -GNinja -DCMAKE_BUILD_TYPE=DEBUG \
   -DCMAKE_CXX_COMPILER=/usr/bin/clang++-14
 ```
 
+In MacOS, if you install llvm clang, you need to explicitly specify to use llvm clang.
+
+Add the following line to your shell environment, e.g. ~/.bash_profile
```
flowbehappy, line 171:
```suggestion
Add the following line to your shell environment, e.g. `~/.bash_profile`.
```

README.md
breezewish: `brew install llvm@15` does not seems to be something hard I think. It even doesn't need any compiling. On the other side, enforce the use of LLVM 15 unlocks newer CXX 20 features.

Lloyd-Pottiger: agree, but left it until we can use clang-15 in centos7, there are some problems with compiling and glibc...

breezewish: May be we can drive devcontainer https://github.com/pingcap/tiflash/pull/6112. It solves CentOS 7 glibc issues.

flowbehappy (approved):
LGTM

PR #6961, opened 2023-03-06
Add local page cache

breezewish:

Problem Summary:

This PR introduces `RNLocalPageCache` module. It supports the following features:

1. Guard / unguard: Keep specified keys not being evicted, when "guard object"'s life is alive.

2. Max size based admission: When total size of **(live) guarding keys** overflows, next request will be waiting, until its size request can be fulfilled.

3. Data is kept in a PageStorage instance, uses disk to store the cache data.

4. LRU based eviction policy: When more and more data is put into the cache, old data will be automatically cleaned using a LRU policy.

#### Implementation

An item is either "guarding", or "evictable":
- Guarding means it is referenced by a living guard object, thus must not be evicted.
- Evictable means it's data could be possibly deleted in order to free up some space.

The LocalPageCache is build on top of:

- A HashMap of living, guarding keys: To know which keys are still guarding, and how many size we have guarded
- A standard LRU: Stores evictable keys
- PageStorage: Stores data for both guarding keys and evictable keys

When keys are guarded, these keys are removed from the LRU (note: their data is still kept in the PageStorage). When keys are unguarded (and not referenced by any guard), these keys will be added to the LRU.

When LRU overflows and evict happens, we will remove the keys from the LRU and from the PageStorage.

The max size of LRU is changed dynamically according to the current size we are guarding. For example, consider that our LocalPageCache's max_size is 5GiB, and currently we are guarding 4GiB data: in this case, the LRU should only use <=1GiB data, evict all old keys that overflows the 1GiB limit.

### Check List

Tests 

- [ ] Unit test: WIP
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V1/PageFile.cpp
```diff
@@ -58,7 +59,7 @@ using Checksum = UInt64;
 static const size_t PAGE_META_SIZE = sizeof(PageId) + sizeof(PageTag) + sizeof(PageOffset) + sizeof(PageSize) + sizeof(Checksum);
 
 /// Return <data to write into meta file, data to write into data file>.
-std::pair<ByteBuffer, ByteBuffer> genWriteData( //
+std::pair<std::span<char>, std::span<char>> genWriteData( //
```
flowbehappy, line 62: I wonder what is the benefit of `std::span` over `ByteBuffer`?

flowbehappy (approved):
LGTM

JaySon-Huang (approved):
lgtm

Lloyd-Pottiger (approved):

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFilePersisted.h
```diff
@@ -19,6 +19,9 @@
 
 namespace DB
 {
+
+struct ByteBuffer;
+
```
breezewish, line 24:
```suggestion
```

dbms/src/Storages/Page/V1/PageFile.cpp
breezewish: Oh, details are documented here: https://github.com/pingcap/tiflash/pull/6972

PR #6972, opened 2023-03-06
Remove ByteBuffer

breezewish:

Problem Summary:

ByteBuffer is:

- Not immutable, where actually it should be immutable
- The name is misleading, does not indicate it has no ownership
- Does not implement standard compare operators, hard to use
- Is not printable by default in google test
- Hard to construct easily

`ByteBuffer` is replaced by `std::string_view`, which is an immutable view to a buffer and does not own the memory of the buffer.

Note: In some places we need a mutable `char *` in order to free the memory later. For such cases, `std::span<char>` is used instead.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lidezhu (approved):

JaySon-Huang (approved):
lgtm

PR #6974, opened 2023-03-07
Protect language boundary between C++ and Rust codes

### LINKED ISSUE #6973: Protect language boundary between C++ and Rust
## Enhancement

Some FFI function don't catch exceptions before return to Rust.

CalvinNeo:

Issue Number: close #6973

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):
LGTM

dbms/src/Storages/Transaction/ProxyFFI.cpp
```diff
@@ -161,51 +161,107 @@ uint8_t TryFlushData(EngineStoreServerWrap * server, uint64_t region_id, uint8_t
 
 RawCppPtr CreateWriteBatch(const EngineStoreServerWrap * dummy)
 {
-    UNUSED(dummy);
-    return GenRawCppPtr(new UniversalWriteBatch(), RawCppPtrTypeImpl::WriteBatch);
+    try
```
lidezhu, line 164: I don't think we should do try catch in these methods. As we didn't always do try catch in every ffi method, `GenCppRawString`, `GcRawCppPtr` and `SetStore` are all examples.
I think the principle is that we should only try catch the ffi which involves IO operations. But these ffis only involves memory operation and are really simple which shouldn't be try-catched.

CalvinNeo: I agree that `CreateWriteBatch` not be protect with try catch. However this is because the function only includes operations that rarely raises exceptions.
I think if a operation may raise exceptions, it should be protected by try-catch on the language boundary. Otherwise we can not find an easy way to examine why the program aborted abruptly.

lidezhu: Yes, that's why I try catch in methods like `HandleConsumeWriteBatch` and `HandleReadPage`. But these methods just involves operations like allocate memory, push into a vector, clear a vector and other simple ops.

lidezhu (approved):

CalvinNeo:
@lidezhu I profiled with the following simple codes, seems that exceptions don't harm too much on normal path?
```
#include <iostream>
#include <stdexcept>
#include <exception>
#include <thread>
#include <chrono>
#include <cstdio>

using std::chrono::high_resolution_clock;
using std::chrono::duration_cast;
using std::chrono::duration;
using std::chrono::milliseconds;

int exception_call_inner() {
    // rarely triggerd
    return 1;
}

int exception_call(int & a) {
    a = (a + 1) % 10000;
    try
    {
        exception_call_inner();
    } catch (...) {
        printf("error");
    }
    return 1;
}

int no_exception_call(int & a) {
    a = (a + 1) % 10000;
    return 1;
}

template<typename F>
int loop_main(F f) {
    int a = 1;
    for(int i = 0; i < 209999900; i++) {
        f(a);
    }
    printf("%d\n", a);
}

int main() {
    {
        auto t1 = high_resolution_clock::now();
        loop_main(no_exception_call);
        auto t2 = high_resolution_clock::now();
        auto ms_int = duration_cast<milliseconds>(t2 - t1);
        printf("no exception cost %lld\n", ms_int.count());
    }
    {
        auto t1 = high_resolution_clock::now();
        loop_main(exception_call);
        auto t2 = high_resolution_clock::now();
        auto ms_int = duration_cast<milliseconds>(t2 - t1);
        printf("exception cost %lld\n", ms_int.count());
    }
}
```

CalvinNeo:
The reason why a rarely hit try catch has low cost https://stackoverflow.com/questions/16784601/does-try-catch-block-decrease-performance

PR #6978, opened 2023-03-07
Allow specifying data provider for ColumnFiles

breezewish:

Problem Summary:

Now ColumnFileTiny only supports read from DM::StoragePool's Snapshot.

This PR abstracts it away with a `IColumnFileDataProvider`, so that later we will allow ColumnFileTiny to read from somewhere else.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileDataProvider.cpp
```diff
@@ -0,0 +1,38 @@
...
+    if (!fields.has_value())
+        return storage_snap->log_reader->read(page_id);
+
+    auto page_map = storage_snap->log_reader->read({{page_id, *fields}});
+    return page_map.at(page_id);
+}
+
+size_t ColumnFileDataProviderLocalStoragePool::getTinyDataSize(IColumnFileDataProvider::PageId page_id) const
+{
+    return storage_snap->log_reader->getPageEntry(page_id).size;
```
lidezhu, line 35: If it is a remote page, the size of `PageEntry` will be 0. The actual size is in checkpoint info. So you should check this case here. (Maybe I should reconsider the related logic to provide some abstracation here.

breezewish: After discussion with @lidezhu , we consider to keep the data size field in the `PageEntry`, even if it's local data is reclaimed.

flowbehappy (approved):
LGTM

dbms/src/Storages/DeltaMerge/Delta/Snapshot.cpp
```diff
@@ -39,9 +40,10 @@ DeltaSnapshotPtr DeltaValueSpace::createSnapshot(const DMContext & context, bool
     snap->delta = this->shared_from_this();
 
     auto storage_snap = std::make_shared<StorageSnapshot>(context.storage_pool, context.getReadLimiter(), context.tracing_id, /*snapshot_read*/ true);
-    snap->persisted_files_snap = persisted_file_set->createSnapshot(storage_snap);
+    auto read_from_storage_snap = ColumnFileDataProviderLocalStoragePool::create(storage_snap);
+    snap->persisted_files_snap = persisted_file_set->createSnapshot(read_from_storage_snap);
+    snap->mem_table_snap = mem_table_set->createSnapshot(read_from_storage_snap, for_update);
```
JaySon-Huang, line 45:
```suggestion
    auto data_from_storage_snap = ColumnFileDataProviderLocalStoragePool::create(storage_snap);
    snap->persisted_files_snap = persisted_file_set->createSnapshot(data_from_storage_snap);
    snap->mem_table_snap = mem_table_set->createSnapshot(data_from_storage_snap, for_update);
```
`read_from_storage_snap` sounds like a function name to me, `data_from_storage_snap` is better

dbms/src/Storages/DeltaMerge/tests/gtest_dm_column_file.cpp
```diff
@@ -220,11 +221,12 @@ try
         wbs.writeAll();
     }
     auto storage_snap = std::make_shared<StorageSnapshot>(dmContext().storage_pool, nullptr, "", true);
+    auto read_from_storage_snap = ColumnFileDataProviderLocalStoragePool::create(storage_snap);
```
JaySon-Huang, line 224: ditto

JaySon-Huang (approved):
LGTM with some comments

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileDataProvider.cpp
JaySon-Huang: > After discussion with @lidezhu , we consider to keep the data size field in the `PageEntry`, even if it's local data is reclaimed.

Then some codes as following should keep this behavior (in later PR) @lidezhu 
https://github.com/pingcap/tiflash/blob/64dab48d5e7bf03344208d0643f899615345cbd9/dbms/src/Storages/Page/V3/Universal/UniversalWriteBatchImpl.h#L106-L110

PR #6991, opened 2023-03-08
Encode delta index epoch

breezewish:

Problem Summary:

Introduce `delta_index_epoch`, which indicates that this delta index instance cannot be used for incremental update. Currently only `flush()` will cause `delta_index_epoch++`.

TODO: May be a better way is that we could detect missing updates and drop the index internally?

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/Delta/DeltaValueSpace.cpp
```diff
@@ -343,8 +343,14 @@ bool DeltaValueSpace::flush(DMContext & context)
 
         /// Update delta tree
         if (new_delta_index)
+        {
             delta_index = new_delta_index;
 
+            // Indicate that the index with old epoch should not be used anymore.
+            // This is useful in disaggregated mode which will invalidate the delta index cache in RN.
+            delta_index_epoch += 1;
```
JaySon-Huang, line 351: Seems there is no synchronization between `DeltaValueSpace::createSnapshot` (for_update=false) and `DeltaValueSpace::flush`. Is there a chance that we still get an incorrect delta-index?

breezewish: Good catch. We should ensure the delta_index and delta_index_epoch are either updated, or either not updated.

JaySon-Huang (approved):
LGTM

Issue #6518, opened 2022-12-21
Support pipeline model


SeaRise:
## Enhancement
tiflash support pipeline model

#### todo list
- [x] public desgin doc
    - https://github.com/pingcap/tiflash/pull/7612 
- [x] pre-refactor
    - https://github.com/pingcap/tiflash/issues/5900 
- [ ] pipeline model
    - [x] support task scheduler and push model operator
        - [x] support task scheduler and simple operator(limit, projection, filter, mock(_scan/exchange_sender/receiver))
            - https://github.com/pingcap/tiflash/pull/6520 
        - [x] support complex operator
            - [x] aggregate
                - https://github.com/pingcap/tiflash/pull/6855
                - https://github.com/pingcap/tiflash/pull/7089
                - https://github.com/pingcap/tiflash/pull/7220
                - https://github.com/pingcap/tiflash/pull/7318
            - [x] join
              - https://github.com/pingcap/tiflash/pull/7333
              - https://github.com/pingcap/tiflash/pull/7833
            - [x] sort/window
              - https://github.com/pingcap/tiflash/pull/6963
              - https://github.com/pingcap/tiflash/pull/7492
            - [x] topn
              - https://github.com/pingcap/tiflash/pull/6805
              - https://github.com/pingcap/tiflash/pull/7216
         - [x] support fine grained shuffle
             - https://github.com/pingcap/tiflash/pull/6934
    - [x] support io executor
        - https://github.com/pingcap/tiflash/pull/7195
    - [x] support exchange / scan async io
        - [x] async table scan
          - https://github.com/pingcap/tiflash/pull/6831 
          - https://github.com/pingcap/tiflash/pull/7225
          - https://github.com/pingcap/tiflash/pull/7474
          - https://github.com/pingcap/tiflash/pull/7480
          - https://github.com/pingcap/tiflash/pull/7552
          - https://github.com/pingcap/tiflash/pull/7723
        - [x] async exchange sender/receiver
          - https://github.com/pingcap/tiflash/pull/6835
          - https://github.com/pingcap/tiflash/pull/6854
    - [ ] task schedule
        - [x] Multi-level feedback queue
            - https://github.com/pingcap/tiflash/pull/7393
        - [x] io_priority_queue
            - https://github.com/pingcap/tiflash/pull/7722
        - [ ] resource group
            - https://github.com/pingcap/tiflash/issues/7660
    - [x] other
        - [x] support explain analyze/mpp tracing
[...]

PR #6995, opened 2023-03-08
*: Refine execution summary with ExecutorStatisticsCollector

ywqzzy:

Issue Number: ref #5900 #6518 

Problem Summary:

In previous implementation, `ExecutionSummaryCollector` traverse through all profilingStreams to collect local execution info.
While the `ExecutorsStatisticsCollector` do the same job for mpp tracing. 
As a result, we can use `ExecutorsStatisticsCollector` to collect local execution info instead.

The change contains:
1. use `ExecutorsStatisticsCollector` to fill execution summary.
2. allow `ExecutorsStatisticsCollector` to collect profile for list based executors and cop tasks.
3. Refine `JoinImpl` to add join build time.
4. add more tests for `list-based` executor (from tispark) and `tree-based` executor (from tidb), add test for execution summary with `planner` and `QueryBlockInterpreter`.
5. fwd declare some `DAGContext` to make the compile faster.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
use test;
create table t(a int);
insert into t values(1),(2),(3),(4);
create table t2(a int);
insert into t2 values(1),(2);
alter table t2 set tiflash replica 1;

set tidb_enforce_mpp=true;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;
insert into t select * from t;

explain analyze select * from t, t2;
explain analyze select * from t;
explain analyze select count(a) from t group by a;
explain analyze select count(a) from t;
explain analyze select * from t, t2 where t.a = t2.a;
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Coprocessor/DAGContext.h
```diff
@@ -311,6 +319,8 @@ class DAGContext
     /// thus we need to pay attention to scan_context_map usage that time.
     std::unordered_map<String, DM::ScanContextPtr> scan_context_map;
 
+    ExecutorStatisticsCollectorPtr executor_statistics_collector;
```
SeaRise, line 322: how about `ExecutorStatisticsCollector executor_statistics_collector`?

dbms/src/Flash/Statistics/ExchangeReceiverImpl.h
```diff
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include <Flash/Coprocessor/RemoteExecutionSummary.h>
```
SeaRise, line 17: seems useless import.

dbms/src/Flash/Planner/ExecutorIdGenerator.h
```diff
@@ -43,37 +43,37 @@ class ExecutorIdGenerator
...
+            return fmt::format("join_{}", ++current_id);
         default:
             throw TiFlashException(
                 fmt::format("Unsupported executor in DAG request: {}", executor.DebugString()),
                 Errors::Planner::Unimplemented);
         }
     }
 
-    UInt32 current_id = 0;
+    UInt32 current_id = -1;
```
SeaRise, line 76: Is it expected? overflow will occur

dbms/src/Flash/Statistics/JoinImpl.cpp
```diff
@@ -61,9 +49,14 @@ void JoinStatistics::collectExtraRuntimeDetail()
             }
         }
     }
+
+    /// In TiFlash, a hash join's build side is finished before probe side starts,
+    /// so the join probe side's running time does not include hash table's build time,
+    /// when construct ExecutionSummaries, we need add the build cost to probe executor
+    base.execution_time_ns += join_build_base.execution_time_ns;
```
SeaRise, line 56: This treatment is wrong
From https://github.com/pingcap/tiflash/pull/6995/files#diff-8afda9cf273246b0b526e551688e59abad0ac41f2803a82e1342eb15bf1c1a4bL88-L109, here also need to add the execute time of all join build streams that executed before.

dbms/src/Flash/Statistics/JoinImpl.h
```diff
@@ -44,8 +44,6 @@ class JoinStatistics : public JoinStatisticsBase
     bool is_spill_enabled = false;
     bool is_spilled = false;
 
-    BaseRuntimeStatistics non_joined_base;
```
ywqzzy, line 47: remove since nonJoinStream removed

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
```diff
@@ -65,22 +80,121 @@ void ExecutorStatisticsCollector::initialize(DAGContext * dag_context_)
...
         }
         return true;
     });
+
+    fillListBasedExecutorsChild();
+}
+
+void ExecutorStatisticsCollector::fillListBasedExecutorsChild()
+{
+    if (!dag_context->return_executor_id)
```
SeaRise, line 97: `return_executor_id` seems to be a bad name at the moment.
- For tree base executor, there must be executor_id
- For list base executor, there must be no executor_id
In fact `is_list_base_executor` or `executor_dag_type` would be better.

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
```diff
@@ -65,22 +80,121 @@ void ExecutorStatisticsCollector::initialize(DAGContext * dag_context_)
...
+        const auto & executors = dag_context->dag_request->executors();
+        String child;
+        for (int i = 0; i < size; ++i)
+        {
+            const auto & executor_id = executors[i].executor_id();
+            if (i != 0)
+                profiles[executor_id]->setChild(child);
+            child = executor_id;
+        }
+    }
```
SeaRise, line 111: I prefer to set child id in the constructor of `ExecutorStatistics`, like
```
if (dag_context->is_list_base_executor)
{
    std::optional<String> child_id;
	auto children = child_id ? {child_id.get()} : {};
    traverseExecutorListReverse(dag_context->dag_request, [&](const tipb::Executor & executor) {
        if (!append<
            AggStatistics,
            ExchangeReceiverStatistics,
            TableScanStatistics,
            TopNStatistics,
            WindowStatistics,
            ExpandStatistics>(executor_id, &executor, children))
        {
            throw TiFlashException(
                fmt::format("Unknown executor type, executor_id: {}", executor_id),
                Errors::Coprocessor::Internal);
        }
		child_id = executor_id;
        return true;
	};
}
else
{
    traverseExecutorTree(dag_context->dag_request, [&](const tipb::Executor & executor) {
        if (!append<
            AggStatistics,
            ExchangeReceiverStatistics,
            TableScanStatistics,
            TopNStatistics,
            WindowStatistics,
            ExpandStatistics>(executor_id, &executor, getChildren(&executor)))
        {
            throw TiFlashException(
                fmt::format("Unknown executor type, executor_id: {}", executor_id),
                Errors::Coprocessor::Internal);
        }
        return true;
    };
}
```

SeaRise: IMO, the code for the same function should be in the same place.

dbms/src/Flash/Statistics/JoinImpl.cpp
```diff
@@ -48,14 +43,6 @@ void JoinStatistics::collectExtraRuntimeDetail()
         build_side_child = join_execute_info.build_side_root_executor_id;
         is_spill_enabled = join_execute_info.join_ptr->isEnableSpill();
         is_spilled = join_execute_info.join_ptr->isSpilled();
-        for (const auto & non_joined_stream : join_execute_info.non_joined_streams)
```
SeaRise, line 51: https://github.com/pingcap/tiflash/blob/ce52266d9d87c4b3b449d97a977ef0cfd0405fa3/dbms/src/Flash/Coprocessor/ExecutionSummaryCollector.cpp#L88-L109
Has this code been migrated?

dbms/src/Flash/Planner/ExecutorIdGenerator.h
```diff
@@ -43,29 +43,29 @@ class ExecutorIdGenerator
         switch (executor.tp())
         {
         case tipb::ExecType::TypeSelection:
-            return fmt::format("{}_selection", ++current_id);
+            return fmt::format("selection_{}", current_id++);
```
yibin87, line 46: changed from ++current_id to current_id++, to allow 0 for executor_id?

dbms/src/Flash/Coprocessor/DAGContext.cpp
```diff
@@ -173,14 +179,19 @@ void DAGContext::initOutputInfo()
     keep_session_timezone_info = encode_type == tipb::EncodeType::TypeChunk || encode_type == tipb::EncodeType::TypeCHBlock;
 }
 
-String DAGContext::getRootExecutorId()
+void DAGContext::initListBasedExecutors()
```
SeaRise, line 182: rename to `initExecutorDAG` and then check `executor.has_executor_id` for tree base executor here?

dbms/src/Flash/Statistics/ExecutorStatistics.h
```diff
@@ -115,5 +127,31 @@ class ExecutorStatistics : public ExecutorStatisticsBase
...
+                auto it = dag_context.getJoinExecuteInfoMap().find(join_executor_id);
+                if (it != dag_context.getJoinExecuteInfoMap().end())
+                {
+                    UInt64 time = 0;
+                    for (const auto & join_build_stream : it->second.join_build_streams)
+                    {
+                        if (auto * p_stream = dynamic_cast<IProfilingBlockInputStream *>(join_build_stream.get()); p_stream)
+                            time = std::max(time, p_stream->getProfileInfo().execution_time);
+                    }
+                    process_time_for_join_build += time;
```
yibin87, line 151: sum all join_build's time together might not be very useful for users to analyze performance. It would be nice if record build time info for each hash join.

dbms/src/Flash/Coprocessor/DAGContext.cpp
```diff
@@ -173,14 +179,19 @@ void DAGContext::initOutputInfo()
...
-        : (list_based_executors_order.empty()
-               ? ""
-               : list_based_executors_order.back());
+    if (!return_executor_id)
+    {
+        ExecutorIdGenerator id_generator;
+        traverseExecutorsReverse(dag_request, [&](const tipb::Executor & executor) {
+            const auto & executor_id = id_generator.generate(executor);
+            list_based_executors_order.push_back(executor_id);
+            auto * mutable_executor = const_cast<tipb::Executor *>(&executor);
```
SeaRise, line 190: a little hack.
how about foreach `dag_request.executors` directly...
for (size_t i = 0; i < dag_request.executors_size(); i++)
{
    const auto & executor_id = id_generator.generate(dag_request.executors[i]);
    blabla
}

SeaRise: img_v2_0e65c29e-3a54-4659-b030-b6ed617d77dg
Although `executors` is const, the elements of `executors` are non-const.

dbms/src/Flash/Planner/ExecutorIdGenerator.h
ywqzzy: > changed from ++current_id to current_id++, to allow 0 for executor_id?

Yes

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
ywqzzy: No, must  follow the logic in DAGContext.cpp:117.
```
{
    RUNTIME_CHECK(dag_request->has_root_executor() && dag_request->root_executor().has_executor_id());
    return_executor_id = dag_request->root_executor().has_executor_id() || dag_request->executors(0).has_executor_id();

    initOutputInfo();
    initListBasedExecutors();
}
```

dbms/src/Flash/Coprocessor/DAGContext.cpp
```diff
@@ -63,6 +65,7 @@ DAGContext::DAGContext(const tipb::DAGRequest & dag_request_, TablesRegionsInfo
     if (return_executor_id)
         root_executor_id = root_executor.executor_id();
     initOutputInfo();
+    initListBasedExecutors();
 }
```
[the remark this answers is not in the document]
ywqzzy: No, since StatisticCollector is holding DAGContext, and this is no need for `DAGContext` to access StatisticCollector

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
ywqzzy: I prefer to move the add child process outside of the constructor to make it clear that we need to init every executors' children. And we don't need to add a 'if else' block here.

dbms/src/Flash/Statistics/ExecutorStatistics.h
ywqzzy: > sum all join_build's time together might not be very useful for users to analyze performance. It would be nice if record build time info for each hash join.

Good idea, maybe I can support it in Pipeline model

yibin87: Yeah, especially, join_builds can run concurrently, sum them up would be mis-understanding.

dbms/src/Flash/Coprocessor/DAGContext.cpp
ywqzzy: > img_v2_0e65c29e-3a54-4659-b030-b6ed617d77dg Although `executors` is const, the elements of `executors` are non-const.

dag_request is marked as const. So it's necessary to remove const here.

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
SeaRise: If so, we should move https://github.com/pingcap/tiflash/pull/6995/files#diff-3aff7923729080b9f5821c18130031eca8fa0504b560e8a256893d05c778828fR48-R56 outside the constructor too?

dbms/src/Flash/Statistics/JoinImpl.cpp
SeaRise: my mistake

dbms/src/Flash/Coprocessor/DAGContext.cpp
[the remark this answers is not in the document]
SeaRise: ok

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
xzhangxian1008, line 97: Maybe we can do this judge in a new function and the reader could get the purpose of this judgement by function name.

dbms/src/Flash/Statistics/ExecutionSummaryHelper.cpp
xzhangxian1008, line 1: May be it's better to put these codes into ExecutionSummary.coo.

xzhangxian1008:
lgtm

dbms/src/Flash/Statistics/ExecutionSummaryHelper.cpp
ywqzzy: > May be it's better to put these codes into ExecutionSummary.cpp.

I will keep it, in case I will use it later

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.h
```diff
@@ -26,37 +29,64 @@ class DAGContext;
 class ExecutorStatisticsCollector
 {
 public:
+    explicit ExecutorStatisticsCollector(const String & req_id,
+                                         bool fill_executor_id_ = false)
```
SeaRise, line 33:
```suggestion
                                         bool force_fill_executor_id_ = false)
```

dbms/src/Flash/Coprocessor/ExecutionSummary.cpp
```diff
@@ -34,6 +35,14 @@ void ExecutionSummary::merge(const tipb::ExecutorExecutionSummary & other)
     scan_context->merge(other.tiflash_scan_context());
 }
 
+void ExecutionSummary::set(const BaseRuntimeStatistics & other)
+{
+    time_processed_ns = other.execution_time_ns;
+    num_produced_rows += other.rows;
+    num_iterations += other.blocks;
+    concurrency += other.concurrency;
+}
```
SeaRise, line 44: It looks like `ExecutionSummary::set` will only be called once, maybe should be 
```
void ExecutionSummary::fill(const BaseRuntimeStatistics & other)
{
    time_processed_ns = other.execution_time_ns;
    num_produced_rows = other.rows;
    num_iterations = other.blocks;
    concurrency = other.concurrency;
}
```
?

SeaRise (commented):
Others LGTM

dbms/src/Flash/Coprocessor/ExecutionSummary.cpp
ywqzzy: > It looks like `ExecutionSummary::set` will only be called once, maybe should be

Good point

SeaRise (approved):

SeaRise (approved):
lgtm

yibin87 (approved):
LGTM

PR #6998, opened 2023-03-09
Split disaggregated shared globals

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/7000

See also: https://github.com/pingcap/tiflash/issues/6827

Problem Summary:

Introduce `SharedContextDisagg`, where we should put all disaggregated shared modules.

When we change `SharedContextDisagg`, only a few files need to be recompiled.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Server/Server.h
```diff
@@ -65,7 +65,7 @@ class Server : public BaseDaemon
     std::string getDefaultCorePath() const override;
 
 private:
-    std::unique_ptr<Context> global_context;
+    ContextPtr global_context;
```
JaySon-Huang, line 68: I think it is dangerous to change global_context from `std::unique_ptr` to be `shared_ptr`. It could easily cause
circular reference

lidezhu (approved):

PR #7003, opened 2023-03-09
Add file cache of DTFile

JinheLin:

Issue Number: ref #6827

This PR add file cache of dtfile in cloud native mode:
1. Cache granularity: column file.
2. Cache missing: if get column file from file cache failed, the read request is read directly from S3, and the file cache asynchronously downloads the corresponding file from the background and caches it. 
3. Cache evicting:
    1. Distinguish cache priority according to file type. The larger the file type, the lower the priority.
    2. Each type of file is maintained in a LRUFileTable to facilitate access to files that have not been used recently.
    3. Basic eviction idea:
        * First, try to evict files which not been used recently with the same type. => Try to evict old files.
        * Second, try to evict files with lower priority. => Try to evict lower-priority files.
6. Add configuration `storage.remote.cache.dtfile_cache_min_age_seconds` that prevent files being read within `dtfile_cache_min_age_seconds` seconds from evicting by files with the same priority. Default value is 1800.

TODO:
- [ ] Pass file size and data type to file cache. (Currently, we use a estimate size to reserve space. This can cause some troublesome problems.)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/S3/FileCache.cpp
```diff
@@ -0,0 +1,552 @@
...
+    }
+    bg_downloading_count.fetch_sub(1, std::memory_order_relaxed);
+    LOG_DEBUG(log, "downloading count {} => s3_key {} finished", bg_downloading_count.load(std::memory_order_relaxed), s3_key);
+}
+
+void FileCache::bgDownload(const String & s3_key, FileSegmentPtr & file_seg)
+{
+    bg_downloading_count.fetch_add(1, std::memory_order_relaxed);
+    LOG_DEBUG(log, "downloading count {} => s3_key {} start", bg_downloading_count.load(std::memory_order_relaxed), s3_key);
+    IOThreadPool::get().scheduleOrThrowOnError(
```
flowbehappy, line 426: Do we need some protection about too many download tasks that `IOThreadPool` will never catch up?  Maybe we need a limit of max waiting tasks.

dbms/src/Storages/S3/FileCache.cpp
```diff
@@ -0,0 +1,552 @@
...
+        tryEvictFile(reserve_for, min_evict_size);
+        return reserveSpaceImpl(reserve_for, size, /*try_evict*/ false);
+    }
+    return false;
+}
+
+// The basic evict logic:
+// 1. Distinguish cache priority according to file type. The larger the file type, the lower the priority.
+// 2. Files with higher priority can be evicted from files with lower priority.
+// 3. Files of the same priority can only evict files that have not been accessed recently.
```
flowbehappy, line 203: Looks like this algorithm will leave many useless files with small `FileType` cannot get evicted. Because we always evict the files with largest `FileType`

dbms/src/Storages/S3/FileCache.cpp
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    auto file_seg = std::make_shared<FileSegment>(toLocalFilename(s3_key), FileSegment::Status::Empty, estimzted_size, file_type);
+    table.set(s3_key, file_seg);
+    bgDownload(s3_key, file_seg);
+
+    return nullptr;
+}
+
+// Remove `local_fname` from disk and remove parent directory if parent directory is empty.
```
flowbehappy, line 123: Will there be any concurrent issues? For example, while a thread is trying to remove parent dir and another thread is trying to put files in it.

dbms/src/Storages/S3/FileCache.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+void FileCache::prepareParentDir(const String & local_fname)
+{
+    auto pos = local_fname.find_last_of('/');
+    RUNTIME_CHECK_MSG(pos != std::string::npos, "local filename {} is invalid", local_fname);
+    auto parent_dir = local_fname.substr(0, pos);
+    prepareDir(parent_dir);
+}
+
+void FileCache::restore()
```
flowbehappy, line 495: I think we need to use multiple threads to run this method. Or use background task to run the restore process.

flowbehappy: Let's do it later.

flowbehappy (approved):

JaySon-Huang (approved):

PR #7005, opened 2023-03-09
Cut down mark_cache default size and add the corresponding metrics in grafana

hongyunyan:

Issue Number: ref #6426

Problem Summary:

1. Change the default of mark_cache size from 5GB to 1GB
2. Add the panel in tiflash_summary.json to support pay attention to effectiveness of mark_cache. It also can help us to find whether the mark_cache is large enough.

image

for the typical workload test, we find that when the memory reach 3-4G, it then slowing growths. 
image

when the default mark_cache size is 1GB, the memory will reach 11.5GB steeply and then continue slowing growth.
Because of the size of minmax index cache is totally same as mark_cache size, So we cut down 7-8G in this workload.

Besides, due to the default value of mark_cache size is set in the code repo of tiup, so it only takes effect after we change the default value in tiup also.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce resident memory usage.
```


breezewish:
How would the memory grows when we change from 5GB to 1GB in the previous workloads?

hongyunyan:
> How would the memory grows when we change from 5GB to 1GB in the previous workloads?

I will do the same workload for this PR.

metrics/grafana/tiflash_summary.json
```diff
@@ -7416,6 +7416,110 @@
...
+            {
+              "expr": "max(tiflash_system_profile_event_MarkCacheMisses{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}) by (instance)",
+              "legendFormat": "mark cache misses",
+              "interval": "",
+              "exemplar": true,
+              "queryType": "randomWalk",
+              "refId": "A"
+            },
+            {
+              "expr": "max(tiflash_system_profile_event_MarkCacheHits{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\"}) by (instance)",
```
breezewish, line 7467: How about using a percent notation? Like 80% effectiveness.

breezewish (approved):

metrics/grafana/tiflash_summary.json
hongyunyan: done ~

Lloyd-Pottiger (approved):

PR #7051, opened 2023-03-12
Compaction for CheckpointDataFile

JaySon-Huang:

Problem Summary:

* Compaction for CheckpointDataFile
  * collect the valid size when running `PageDirectory::gcInMemory` and save them in memory cache struct `CPDataFilesStatCache`
  * when upload checkpoint, collect the valid size and total size of CheckpointData files from S3
  * get the compaction candidates by remote gc threshold
  * rewrite the page data to new CheckpointData file in `CPFilesWriter::writeEditsAndApplyCheckpointInfo`
* New settings for remote gc threshold
  * `remote_gc_ratio` - default to 0.5
  * `remote_gc_small_size` - default to 128 KiB
  * > S3 Standard-IA and S3 One Zone-IA storage have a minimum billable object size of 128 KB. Smaller objects may be stored but will be charged for 128 KB of storage at the appropriate storage class rate.
* Other fixes
  * the very first `S3LockLocalManager::initStoreInfo` will return the latest manifest prefix to restore the `last_checkpoint_sequence`
  * avoid always allocating new upload_sequence while checkpoint upload is actually skip
  * only log down warning/error level from AWS SDK -- the info/debug level of AWS SDK are too noisy
  * Add some check for `storage.s3.root`/`storage.s3.endpoint`/`flash.disaggregated_mode`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/Universal/UniversalPageStorage.cpp
```diff
@@ -420,7 +435,9 @@ void UniversalPageStorage::dumpIncrementalCheckpoint(const UniversalPageStorage:
         .sequence = snap->sequence,
         .last_sequence = last_checkpoint_sequence,
     });
-    bool has_new_data = writer->writeEditsAndApplyCheckpointInfo(edit_from_mem);
+    // TODO: make this magic number configurable
+    const auto rewrite_file_ids = getFileIdsNeedRewrite(0.2);
```
flowbehappy, line 439: Please use a config value.

dbms/src/Storages/Page/V3/CheckpointFile/CPFilesWriter.cpp
```diff
@@ -89,9 +89,16 @@ bool CPFilesWriter::writeEditsAndApplyCheckpointInfo(universal::PageEntriesEdit
...
+            const auto file_id = *rec_edit.entry.checkpoint_info.data_location.data_file_id;
+            if (!file_ids_to_compact.contains(file_id))
+            {
+                // for example, the s3 fullpath that was written in the previous uploaded CheckpointDataFile
+                locked_files.emplace(file_id);
+                continue;
+            }
+            // else we rewrite this entry data to the data file generated by this checkpoint, so that
+            // the old, fragmented data will be compacted. The outdated CheckpointDataFiles will be
+            // removed by S3GCManager after the new manifest uploaded.
```
JaySon-Huang, line 101: TODO: Add a test for whether we can override the checkpoint_info.data_location in PageDirectory

dbms/src/Storages/Page/V3/Universal/UniversalPageStorage.cpp
```diff
@@ -461,4 +479,59 @@ void UniversalPageStorage::dumpIncrementalCheckpoint(const UniversalPageStorage:
...
+    {
+        if (stat.total_size == 0)
+            continue;
+        double valid_rate = 1.0 * stat.valid_size / stat.total_size;
+        if (valid_rate < gc_threshold)
+        {
+            rewrite_files.emplace(file_id);
+        }
+    }
+    return rewrite_files;
```
JaySon-Huang, line 534: TODO: when the file size is less than 128KiB, then rewrite it
https://aws.amazon.com/cn/s3/pricing/

dbms/src/Interpreters/Settings.h
```diff
@@ -240,6 +241,8 @@ struct Settings
     \
     M(SettingInt64, remote_checkpoint_interval_seconds, 30, "The interval of uploading checkpoint to the remote store. Unit is second.")                                                                                                \
     M(SettingInt64, remote_gc_interval_seconds, 3600, "The interval of running GC task on the remote store. Unit is second.")                                                                                                           \
+    M(SettingDouble, remote_gc_ratio, 0.2, "The files with valid rate less than this threshold will be compacted")                                                                                                                      \
```
flowbehappy, line 244: Hmmm it is too small IMO. I think 0.5 is more suitable. If it is proven to be not cost-efficient, we can tune it later.

flowbehappy (approved):
LGTM with a comment left

lidezhu (approved):
LGTM

dbms/src/Interpreters/Settings.h
JaySon-Huang: change to 0.5 by default

PR #7050, opened 2023-03-12
S3GC with lifecycle config

JaySon-Huang:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tiflash/issues/6827

Problem Summary:

AWS S3 support lifecycle with object tagging, we can use it to reduce S3 `LIST` calls

### What is changed and how it works?

* Fix the label filter in StorageDisaggRemote
* Deleting the subfiles for DTFiles on S3
* Read locks from multiple preserved manifests
* Add delete-tagging to delmark in S3LockService
* Rewrite the object with delete-tagging after delmark created successfully

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #7047, opened 2023-03-11
Remove DMFile after upload S3

JinheLin:
### What problem does this PR solve?

Issue Number: ref #6827

### What is changed and how it works?

1. DMFile switching to S3 and removing local DMFile after upload S3 successfully.
2. If S3 enabled, restore from S3.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
  - ./tiflash dtworkload --testing_type=daily_perf --verify_round=10 --s3_bucket=dtw --s3_endpoint=172.16.5.85:9000 --s3_access_key_id=minioadmin --s3_secret_access_key=minioadmin --ps_run_mode=4 --log_level=debug 
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #7060, opened 2023-03-13
Fix several issues under disagg mode

JaySon-Huang:

Problem Summary:

Fix some bugs introduced by 
* https://github.com/pingcap/tiflash/pull/7047
* https://github.com/pingcap/tiflash/pull/7050
* If some S3 clients are release after `Aws::ShutdownAPI(aws_options);` they will make crash. So don't keep long live S3 clients as class member

Workaround a tiup bug
* https://github.com/pingcap/tiup/issues/2139

* Use `1 PiB` as store capacity in PD when S3 is enabled
* Make BackgroundInitStores exception safe
* Fix lifecycle rule settings, report warning when the rule is not defined
  * Creating a new rule by AWS SDK is always fail, don't know why
* Make AWS SDK info logging as debug level
* Disagg Compute node still need to init `ps_write` because it needs to start proxy
* Disable RegionPersister and its background task on disagg compute node
* Disable ReadIndexRunner on disagg compute node
* Remove long-living S3 clients, they will cause crashes if they are released later than `S3::ClientFactory` ends
* Reduce some verbose logging in `ProxyFFIStatusService` `SchemaSyncService`

The following changes are replaced by a standalone PR: https://github.com/pingcap/tiflash/pull/7068
* Config default values changed:
  * Inferr `storage.format_version` to be 5 when S3 is enabled and user don't set it explicitly
  * Add check for `storage.format_version` must be 5 when explicitly set when S3 is enabled
  * Change `flash.use_autoscaler` default to be `false`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Server/Server.cpp
```diff
@@ -941,6 +880,11 @@ int Server::main(const std::vector<std::string> & /*args*/)
     {
         LOG_INFO(log, "Using format_version={} (default settings).", STORAGE_FORMAT_CURRENT.identifier);
     }
+    if (storage_config.s3_config.isS3Enabled() && storage_config.format_version != STORAGE_FORMAT_V5.identifier)
+    {
+        LOG_WARNING(log, "'storage.format_version' must be set to 5 when S3 is enabled!");
+        throw Exception("'storage.format_version' must be set to 5 when S3 is enabled!");
+    }
```
JaySon-Huang, line 887: This would be better if we advance the format_version to 6 later, we don't need to let the user remove the storage.format_version configuration item from the configuration file? @flowbehappy 
```
    if (storage_config.format_version)
    {
        if (storage_config.s3_config.isS3Enabled() && storage_config.format_version != STORAGE_FORMAT_V5.identifier)
        {
            LOG_WARNING(log, "'storage.format_version' must be set to 5 when S3 is enabled!");
            throw Exception("'storage.format_version' must be set to 5 when S3 is enabled!");
        }
        setStorageFormat(storage_config.format_version);
        LOG_INFO(log, "Using format_version={} (explicit storage format detected).", storage_config.format_version);
    }
    else
    {
        if (storage_config.s3_config.isS3Enabled())
        {
            // If the user does not explicitly set format_version in the config file but
            // enables S3, then we set up a proper format version to support S3.
            setStorageFormat(5);
            LOG_INFO(log, "Using format_version={} (inferr by S3 is enabled).", STORAGE_FORMAT_V5.identifier);
        }
        else
        {
            // Use the default settings
            LOG_INFO(log, "Using format_version={} (default settings).", STORAGE_FORMAT_CURRENT.identifier);
        }
    }
```

dbms/src/Server/Server.cpp
```diff
@@ -1232,6 +1176,8 @@ int Server::main(const std::vector<std::string> & /*args*/)
     }
 
     if (global_context->getSharedContextDisagg()->isDisaggregatedStorageMode()
+        // disagg compute node without auto scaler will start proxy, which needs UniPS
+        || (global_context->getSharedContextDisagg()->isDisaggregatedComputeMode() && !global_context->getSharedContextDisagg()->use_autoscaler)
         || global_context->getSharedContextDisagg()->notDisaggregatedMode())
```
breezewish, line 1181: How about simply remove this `if`?

JaySon-Huang: when it is deployed on tidb cloud with autoscaler, we don't need the `ps_wirte` instance :(

dbms/src/Server/Server.cpp
flowbehappy: LGTM

JaySon-Huang: changed

dbms/src/Server/Server.cpp
JaySon-Huang: simply by
https://github.com/pingcap/tiflash/blob/f1781b038fdc4e137160e7d1a9e2ccca97ce97d5/dbms/src/Server/Server.cpp#L1189-L1195

breezewish (approved):

JaySon-Huang:
waiting for https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/11807/pipeline/85

PR #6987, opened 2023-03-07
[WIP] Add fap support

lidezhu:
### What problem does this PR solve?

Issue Number: close #xxx

Problem Summary: depends on #6984 

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #7075, opened 2023-03-15
Support disagg read after data ingest by FAP

JaySon-Huang:

Problem Summary: After FastAddPeer (https://github.com/pingcap/tiflash/pull/6987), the data of one TiFlash write node could ingest some data (ColumnFileTiny/DMFile) that is generated by another TiFlash write node. This break the previous consumption. WNs should return the real remote location to the RN so that RN can fetch the data from remote store (e.g. S3) directly.

* Change the checkpoint interval from 60 seconds to 30 seconds
* Serialize the remote key (S3 path) for DMFile to `RemotePb::ColumnFileBig` so that RN can get the correct S3 key for reading
* Note that if a ColumnFileTiny is ingested by FAP and later read by RN,
  * Now it is still read by WN at the first time, then written back to WN's UniPS
  * When RN receives the page data, it will put it into its LocalPageCache

> Now it is still read by WN at the first time, then written back to WN's UniPS
> This might increase some latency at the first time reading data after FAP, but we can optimize it later

Other
* Q: Why not let the RN parse the WN's manifest on S3 to get the location info of given PageIDs on that store_id?
* A: The latest segment info could be not in the manifest on S3 yet.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

* Deploy a cluster with 1 disagg WN and 1 disagg RN
* Load some data to WN
* Scale-out another WN and read data through RN
* Scale-in the origin WN totally and read data through RN

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Server/StorageConfigParser.cpp
+++ b/dbms/src/Server/StorageConfigParser.cpp
@@ -550,7 +550,7 @@ void StorageS3Config::parse(const String & content, const LoggerPtr & log)
     readConfig(table, "request_timeout_ms", request_timeout_ms);
     RUNTIME_CHECK(request_timeout_ms > 0);
     readConfig(table, "root", root);
-    RUNTIME_CHECK_MSG(root == "/", "Only support using '/' as the S3 key root now"); // working on in another PR
+    RUNTIME_CHECK(!root.empty());
 
     auto read_s3_auth_info_from_env = [&]() {
         access_key_id = Poco::Environment::get(S3_ACCESS_KEY_ID, /*default*/ "");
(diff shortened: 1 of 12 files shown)


lidezhu (approved):
LGTM

dbms/src/Server/StorageConfigParser.cpp
```diff
@@ -550,7 +550,7 @@ void StorageS3Config::parse(const String & content, const LoggerPtr & log)
     readConfig(table, "request_timeout_ms", request_timeout_ms);
     RUNTIME_CHECK(request_timeout_ms > 0);
     readConfig(table, "root", root);
-    RUNTIME_CHECK(!root.empty());
+    RUNTIME_CHECK_MSG(root == "/", "Only support using '/' as the S3 key root now"); // working on in another PR
```
JaySon-Huang, line 553:
```suggestion
    RUNTIME_CHECK(!root.empty());
```

PR #7076, opened 2023-03-15
Support sharing the same S3 bucket for multiple clusters

JaySon-Huang:

Problem Summary:

The number of buckets in Amazon S3 is limited, we need to share the same bucket for different TiDB clusters on Cloud. https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingBucket.html

> You can store any number of objects in a bucket and can have up to 100 buckets in your account.

In order to share the same bucket among different TiDB clusters, we should setup different `storage.s3.root`. The `storage.s3.root` will be prepended to each S3 key. This PR makes the `storage.s3.root` effective.

When sending a request to S3, we use `TiFlashS3Client::setBucketAndKeyWithRoot` to prepend the `root` to the S3 key.
And when getting an object key like using `LIST`, we copy the object struct with `root` cut from the key properly.

Another way to do this is to prepend each S3 key to where we need to store a S3 key. This could waste many disk/memory and make `S3Filename` coupled with the S3Config. Only add it when sending requests/receiving responses from S3 could be better.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Disaggregated/MockS3LockClient.h
+++ b/dbms/src/Flash/Disaggregated/MockS3LockClient.h
@@ -30,13 +30,7 @@ class MockS3LockClient : public IS3LockClient
 {
 public:
     explicit MockS3LockClient(std::shared_ptr<TiFlashS3Client> c)
-        : MockS3LockClient(c, c->bucket())
-    {
-    }
-
-    MockS3LockClient(std::shared_ptr<Aws::S3::S3Client> c, const String & bucket_)
         : s3_client(std::move(c))
-        , bucket(bucket_)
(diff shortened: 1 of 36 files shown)


flowbehappy (approved):

Lloyd-Pottiger (approved):

CalvinNeo (approved):
lgtm

hongyunyan (approved):

PR #7078, opened 2023-03-15
Disable prepare stage

breezewish:

Problem Summary:

Currently prepare (delta index) does not have benefits, because we always use a high concurrency to read (in order to migrate the latency of reading from S3), and during the reading the delta index will be built. Thus, our placing delta index step, which is a time consuming step, is already done in parallel.

We may plan to do something more complicated in prepare, for example, trying to prioritise placing delta index for segments that have cached data, so that such segments can start reading as soon as possible. However this optimization may be too complicated and its benefit may be small, considering that actually placing delta index will not cost much time when we have a delta index cache.

I'm a fan of Go's concurrency philosophy, that is [“Do not communicate by sharing memory; instead, share memory by communicating.”](https://golang.org/doc/effective_go.html#concurrency). As a practice of this philosophy, actually we could have:

- A series of PageFetchRPC threads (maybe 200 threads), which takes request from PageFetchRequests Channel, do RPC call, and put the result in PageFetchResponses Channel.
- A series of SegmentTaskPrepare threads (= CPU cores), which takes results from the PageFetchResponses channel, decode the result, forms a PreparedSegmentTask and put the result in PreparedSegmentsChannel.
- A series of Read threads (maybe CPU cores * N), which takes a Segment from PreparedSegmentsChannel, build a read stream and read from this Segment.

When we do want to prepare index in advance, it could be:

- A series of PageFetchRPC threads (maybe 200 threads), which takes request from PageFetchRequests Channel, do RPC call, and put the result in PageFetchResponses Channel.
- A series of SegmentTaskPrepare threads (= CPU cores), which takes results from the PageFetchResponses channel, decode the result, **check whether we could prepare delta index without waiting and place the delta index in this case**, then forms a PreparedSegmentTask and put the result in PreparedSegmentsChannel.
- A series of Read threads (maybe CPU cores * N), which takes a Segment from PreparedSegmentsChannel, build a read stream and read from this Segment.

In this way, we only need MPMC queues. No more conditional variables and no more state managements. This also allows hot segments to respond earlier (when Delta Index prepare is indeed time consuming). Anyway I don't think delta index prepare optimization is necessary, because when Segment's stable is hot, its delta index should be also hot. 

The shining point is that, this workflow is simple, understandable, and easy to opt-in the optimization without touching a lot of code base. Moreover, when we switched MPMC queue from our implementation to a more advanced one (for example, lock-free implementations), we get performance improvements naturally. It also let us easier to switch to coroutine based implementation, because we only need to deal with the Thread (transit to coroutine) and MPMC.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/StorageDisaggregatedRemote.cpp
+++ b/dbms/src/Storages/StorageDisaggregatedRemote.cpp
@@ -361,7 +361,7 @@ void StorageDisaggregated::buildRemoteSegmentInputStreams(
         log->identifier(),
         executor_id);
 
-    bool do_prepare = true;
+    bool do_prepare = false;
 
     // Build the input streams to read blocks from remote segments
     auto [column_defines, extra_table_id_index] = genColumnDefinesForDisaggregatedRead(table_scan);
--- a/tests/tidb-ci/new_collation_fullstack/function_collator.test
+++ b/tests/tidb-ci/new_collation_fullstack/function_collator.test
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#TODO: Fix this test https://github.com/pingcap/tiflash/issues/7079
+#RETURN
+
 mysql> drop table if exists test.t1
 mysql> drop table if exists test.t2
 mysql> create table test.t1(col_varchar_20_key_signed varchar(20) COLLATE utf8mb4_general_ci, col_varbinary_20_key_signed varbinary(20), col_varbinary_20_undef_signed varbinary(20));


JaySon-Huang (approved):
LGTM

flowbehappy (approved):

CalvinNeo (approved):
lgtm

PR #7080, opened 2023-03-15
Refactor cache policy for FAP

lidezhu:

Issue Number: ref #6827 

Problem Summary: The previous cache policy for FAP causes too much unnecessary request to S3.

After change, the current process for FAP is as following:
1. Get the store ids which may contains the target region;
2. Try to parse the region info from the temp ps in cache first, if not success, then try to fetch latest manifest from S3;
3. Memorize the checked upload sequence for each region, and only try to parse again when there is new manifest file with newer upload sequence;
4. When read segment info from checkpoint, the first region will read all segments meta info for a table and put a map from segment_end_key to segment id in the cache, later regions can just read from the cache.

This pr also fix some bugs:
1. We should use `IDataStore` to prepare DMFile instead of `DMFile::restore` when create it from checkpoint;
2. Fix the range to scanning raft logs of a region;

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Transaction/FastAddPeerCache.h
```diff
@@ -0,0 +1,112 @@
...
+    std::map<DM::RowKeyValue, UInt64, KeyComparator> end_key_to_segment_id;
+};
+using SegmentEndKeyCachePtr = std::shared_ptr<SegmentEndKeyCache>;
+
+class ParsedCheckpointDataHolder
+{
+public:
+    UniversalPageStoragePtr temp_ps;
+
+    std::vector<String> paths = {};
```
flowbehappy, line 91: Better use private and use `getter()`. Looks better.

lidezhu: Fixed.

dbms/src/Storages/Transaction/FastAddPeerCache.h
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+private:
+    std::vector<String> paths = {};
+
+    UniversalPageStoragePtr temp_ps;
+
+    std::mutex mu; // protect segment_end_key_cache
+    using SegmentEndKeyCacheMap = std::unordered_map<TableIdentifier, SegmentEndKeyCachePtr>;
+    SegmentEndKeyCacheMap segment_end_key_cache_map;
```
flowbehappy, line 111: How about `end_to_segment_id`?

dbms/src/Storages/Transaction/FastAddPeer.cpp
```diff
@@ -50,382 +51,267 @@ FastAddPeerContext::FastAddPeerContext(uint64_t thread_count)
...
-    std::unique_lock lock(ps_cache_mu);
-    auto iter = temp_ps_cache.find(store_id);
-    if (iter != temp_ps_cache.end() && iter->second.first >= upload_seq)
-    {
-        return iter->second.second;
-    }
-    return nullptr;
+    std::unique_lock lock(mu);
+    if (!is_ready)
+        cv.wait(lock, [&] { return is_ready; });
```
flowbehappy, line 58: I don't think we need a `condition_variable` here. It is complicated and dangerous. For example, `setParsedCheckpointData` could throw exceptions, then `is_ready` won't never be true.

A much simpler way is using a lock to protect a method named `parseCheckpointData`:

```
lock(mutext);
if parsed_checkpoint_data exist
then 
    return parsed_checkpoint_data;
else
   do parse job;
```

dbms/src/Storages/Transaction/FastAddPeerCache.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <Storages/Transaction/TMTContext.h>
+#include <fmt/core.h>
+
+namespace DB
+{
+UInt64 SegmentEndKeyCache::getSegmentIdContainingKey(const DM::RowKeyValue & key)
+{
+    std::unique_lock lock(mu);
+    if (!is_ready)
+        cv.wait(lock, [&] { return is_ready; });
```
flowbehappy, line 43: Also `cv.wait` is not necessary. Please use `lock` instead.

dbms/src/Storages/Transaction/FastAddPeerCache.h
```diff
@@ -0,0 +1,116 @@
...
+} // namespace std
+
+namespace DB
+{
+class UniversalPageStorage;
+using UniversalPageStoragePtr = std::shared_ptr<UniversalPageStorage>;
+
+struct AsyncTasks;
+
+class SegmentEndKeyCache
```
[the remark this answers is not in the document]
lidezhu: Added.

dbms/src/Storages/Transaction/FastAddPeerCache.cpp
```diff
@@ -0,0 +1,164 @@
...
+    auto iter = end_key_to_segment_id.upper_bound(key);
+    RUNTIME_CHECK(iter != end_key_to_segment_id.end(), key.toDebugString(), end_key_to_segment_id.rbegin()->first.toDebugString());
+    return iter->second;
+}
+
+void SegmentEndKeyCache::build(const std::vector<std::pair<DM::RowKeyValue, UInt64>> & end_key_and_segment_ids)
```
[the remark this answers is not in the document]
lidezhu: Use vector instead.

dbms/src/Storages/Transaction/FastAddPeer.cpp
lidezhu: Fixed.

dbms/src/Storages/Transaction/FastAddPeerCache.cpp
lidezhu: Fixed.

flowbehappy (approved):
LGTM

dbms/src/Storages/Page/V3/Universal/RaftDataReader.cpp
```diff
@@ -37,18 +37,31 @@ void RaftDataReader::traverse(const UniversalPageId & start, const UniversalPage
...
-    traverse(start, end, [&](const UniversalPageId & page_id, const DB::Page & page) {
-        if (page_id.hasPrefix(start))
-            acceptor(page_id, page);
-    });
+    auto end = UniversalPageIdFormat::toFullRaftLogScanEnd(region_id);
+    auto snapshot = uni_ps.getSnapshot(fmt::format("scan_r_{}_{}", start, end));
+    const auto page_ids = uni_ps.page_directory->getAllPageIdsInRange(start, end, snapshot);
+    for (const auto & page_id : page_ids)
+    {
+        RUNTIME_CHECK(page_id.size() == 20, page_id.size());
```
JaySon-Huang, line 60: better add some explain or link about the magic number 20. And a TODO is it could be changed under keyspace

dbms/src/Storages/Page/V3/Universal/RaftDataReader.cpp
```diff
@@ -37,18 +37,31 @@ void RaftDataReader::traverse(const UniversalPageId & start, const UniversalPage
...
-void RaftDataReader::traverseRaftLogForRegion(UInt64 region_id, const std::function<void(const UniversalPageId & page_id, DB::Page page)> & acceptor)
+void RaftDataReader::traverseRemoteRaftLogForRegion(UInt64 region_id, const std::function<void(const UniversalPageId & page_id, const PS::V3::CheckpointLocation & location)> & acceptor)
 {
     auto start = UniversalPageIdFormat::toFullRaftLogPrefix(region_id);
-    auto end = UniversalPageIdFormat::toFullRaftLogPrefix(region_id + 1);
-    traverse(start, end, [&](const UniversalPageId & page_id, const DB::Page & page) {
-        if (page_id.hasPrefix(start))
-            acceptor(page_id, page);
-    });
+    auto end = UniversalPageIdFormat::toFullRaftLogScanEnd(region_id);
```
JaySon-Huang, line 55: why here must use `toFullRaftLogScanEnd(region_id)` but not `toFullRaftLogPrefix(region_id + 1)`?

dbms/src/Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h
```diff
@@ -121,17 +122,30 @@ struct UniversalPageIdFormat
         return buff.releaseStr();
     }
 
-    // LOCAL_PREFIX REGION_RAFT_PREFIX region_id RAFT_LOG_SUFFIX
+    // RAFT_PREFIX LOCAL_PREFIX REGION_RAFT_PREFIX region_id RAFT_LOG_SUFFIX
     static String toFullRaftLogPrefix(UInt64 region_id)
     {
         WriteBufferFromOwnString buff;
         writeChar(0x01, buff);
+        writeChar(0x01, buff);
```
JaySon-Huang, line 130: dose it break the existing data deployed by the older version on master?

JaySon-Huang (approved):
The rest LGTM

dbms/src/Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h
lidezhu: No, because the data is written by proxy.

dbms/src/Storages/Page/V3/Universal/RaftDataReader.cpp
lidezhu: Added.

PR #7086, opened 2023-03-16
Fix evicting files of file cache.

### LINKED ISSUE #6827: TiFlash directly read/write S3

JinheLin:

Issue Number: close #6827

- First problem:
  - When evicting files, we need to traverse the LRU queue and delete files that are not held by other threads.
  - If the deletion is successful, we need to use the iterator returned by the deletion function since the current iterator is deleted and invalid.
  - If the deletion fails, the current iterator is used to continue the iteration.
  - The problem is that the current implementation thinks that when the returned free space is 0, it is a deletion failure. But in fact, if we are deleting an empty file, the return of 0 may also be a deletion success, leading to the invalidation of the iterator.
  - **FIXED**: return -1 when deletion fails.

- Second problem:
  - `FileTable::get` will update LRU queue when traversing the LRU queue and getting file from FileTable and causing inaccurate traversal.
  - **FIXED**: not update LRU queue when traversing.

- Third problem:
  - `std::filesystem::file_size` throws exception if file does not exist. 
  - **FIXED**: check file exists at the beginning.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/S3/FileCache.h
```diff
@@ -125,16 +125,19 @@ using FileSegmentPtr = std::shared_ptr<FileSegment>;
...
             return nullptr;
         }
         auto & [file_seg, lru_itr] = itr->second;
-        // Move the key to the end of the queue. The iterator remains valid.
-        lru_queue.splice(lru_queue.end(), lru_queue, lru_itr);
+        if (update_lru)
+        {
+            // Move the key to the end of the queue. The iterator remains valid.
+            lru_queue.splice(lru_queue.end(), lru_queue, lru_itr);
+        }
```
Lloyd-Pottiger, line 140: just curious, the capacity of the Lru cache is not limited?

flowbehappy (approved):

Lloyd-Pottiger (approved):
Others LGTM

dbms/src/Storages/S3/FileCache.h
Lloyd-Pottiger: get

breezewish (approved):
Verified no more crash in test env for now.

PR #7097, opened 2023-03-18
Speedup ingest when S3 is enabled && fix bug caused by store_id not inited

JaySon-Huang:

Problem Summary:

* In the previous PR, when handling ingesting SST/apply snapshot, we put the DMFile to S3 in the `DeltaMergeStore::ingestFiles`, which could be slow because the apply threads is quite limited compared to the predecode threads. This greatly slow down the speed of ingesting snapshot from TiKV/restoring from BR
* Some background services that cause Segment creation could be failed when `store_id` is not inited
  * GCManager - background delta-merge and segment-merge
  * SchemaSyncService - create a new DeltaMergeStore and its very first segment
  * If these services create segment before `store_id` is inited, they will try put some data to S3 with `store_id==0`, which will be deleted by S3GCManager

* Move upload to S3 to `SSTFilesToDTFilesOutputStream::writeSuffix`, so that different DMFiles could be upload parallel by pre-decode threads
* For the `store_id` problem
  * Try to read the `StoreIdent` from UniPS when disagg is enabled
  * If `StoreIdent` can be restored from UniPS, it means the node has been bootstrapped. Then we setup the `store_id` to KVStore right after TMTContext is created
  * If `StoreIdent` can not be restored from UniPS, it means the node is not bootstrapped. Then we postpone the schema sync with TiDB after proxy bootstrap and setup the `store_id`
  * Add more checks for `store_id` when the background service involve segment rewrite

#### Further improvements
* [ ] We could create an async task once a DMFile is generated in `SSTFilesToDTFilesOutputStream::finalizeDTFileStream` to take more resources to speed up ingestSST/applySnapshot later
* [ ] We need to categorize the dependencies between all global instances and tiflash-proxy, try to bootstrap the node more eailier

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Server/Bootstrap.cpp
+++ b/dbms/src/Server/Bootstrap.cpp
@@ -1,0 +1,42 @@
+// Copyright 2023 PingCAP, Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Common/Exception.h>
+#include <Common/RedactHelpers.h>
+#include <Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h>
+#include <Storages/Page/V3/Universal/UniversalPageStorage.h>
+#include <raft_serverpb.pb.h>
+
+namespace DB
+{
+
+std::optional<raft_serverpb::StoreIdent>
+tryGetStoreIdent(const UniversalPageStoragePtr & wn_ps)
+{
+    auto page = wn_ps->read(
+        UniversalPageIdFormat::getStoreIdentId(),
+        nullptr,
+        {},
+        /*throw_on_not_exist*/ false);
+    if (!page.isValid())
+    {
+        return std::nullopt;
+    }
+    raft_serverpb::StoreIdent store_ident;
+    bool ok = store_ident.ParseFromString(String(page.data));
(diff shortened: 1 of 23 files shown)


dbms/src/Server/Bootstrap.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <Common/RedactHelpers.h>
+#include <Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h>
+#include <Storages/Page/V3/Universal/UniversalPageStorage.h>
+#include <raft_serverpb.pb.h>
+
+namespace DB
+{
+
+std::optional<raft_serverpb::StoreIdent>
+tryGetStoreIdent(const UniversalPageStoragePtr & wn_ps)
```
CalvinNeo, line 25: I'd better make getting meta states such like StoreIdent/RaftApplyState/RegionLocalState common. This is because in v2, they are stored in RaftEngine rather than KVEngine. So in unips we must access them by a different prefix.
I think maybe a better way is we can call an FFI interface, and let Proxy to fetch the result to decouple the complex logic.

JaySon-Huang: `RaftApplyState`/`RegionLocalState` is for a specified region, I think they should not belong here. This file and function is mainly for checking whether a cluster is already bootstrap.
And later we may split `TMTContext` into smaller parts and do the bootstrap more earlier than now.

If we want to adapt `RaftApplyState`/`RegionLocalState` with raft-store V2, we could do it in later PRs.

flowbehappy (approved):

dbms/src/Storages/DeltaMerge/tests/gtest_dm_delta_merge_store_fast_add_peer.cpp
```diff
@@ -270,6 +273,29 @@ try
...
+                    dm_file,
+                    S3::DMFileOID{
+                        .store_id = store_id,
+                        .table_id = store->physical_table_id,
+                        .file_id = file_id.id,
+                    },
+                    true);
+            }
+        }
         store->ingestFiles(dm_context, range, file_ids, false);
```
JaySon-Huang, line 299: The behavior has changed /cc @lidezhu

PR #7098, opened 2023-03-18
Merge small files into one or more larger one before upload to S3

JinheLin:

Issue Number: ref #6827

Problem Summary:
The current DMFile implementation has many small files, resulting in many calls to S3. 

To reduce S3 calls, combine small files into one or more large files before uploading S3.

For write operattion:
- After finalizeColumns and before finalizeMetaV2, call finalizeSmallColumnDataFiles to merge small files and generate metadata.
- The file name of the merged file is {number}.merged. The number increments from 0.
- The location of small files is saved to the meta file by function writeMergedSubFilePositionsToBuffer。
- When uploading, use listFilesForUpload to list metadata file, merged files, and column files that are not in the merged files.

For read operation:
- When reading, it is necessary to pass the information of the merged file to S3RandomAccessFile. It is troublesome to pass parameters layer by layer. So use thread_local global variable to pass parameters. We can refine these codes later.

New configurations:
- `profiles.default.dt_small_file_size_threshold` : When S3 is enabled, file size less than dt_small_file_size_threshold will be merged before uploading to S3
- `profiles.default.dt_merged_file_max_size` : Small files are merged into one or more files not larger than dt_merged_file_max_size.
- `profiles.default.dt_filecache_max_downloading_count_scale` : Max downloading task count of FileCache = io thread count * dt_filecache_max_downloading_count_scale.
- `profiles.default.dt_filecache_min_age_seconds` : Files of the same priority can only be evicted from files that were not accessed within dt_filecache_min_age_seconds seconds.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/File/MergedFile.h
```diff
@@ -0,0 +1,61 @@
...
+
+    void serializeToBuffer(WriteBuffer & buf) const
+    {
+        writeStringBinary(fname, buf);
+        writeIntBinary(number, buf);
+        writeIntBinary(offset, buf);
+        writeIntBinary(size, buf);
+    }
+
+    void parseFromBuffer(ReadBuffer & buf)
```
flowbehappy, line 53: I would prefer make `parseFromBuffer` as static. And remove `explicit MergedSubFileInfo(ReadBuffer & buf)`. Because using constructor to modify `ReadBuffer & buf` looks weird.

```
static MergedSubFileInfo parseFromBuffer(ReadBuffer & buf);
```

flowbehappy (approved):
LGTM with a comment

dbms/src/Storages/S3/S3RandomAccessFile.h
```diff
@@ -71,11 +75,30 @@ class S3RandomAccessFile final : public RandomAccessFile
...
+        read_file_info = std::move(read_file_info_);
+        return ext::make_scope_guard([]() {
+            read_file_info.reset();
+        });
+    }
+
 private:
     void initialize();
 
+    inline static thread_local std::optional<ReadFileInfo> read_file_info;
```
flowbehappy, line 97: Put some comments here to explain why we use thread_local

dbms/src/Storages/DeltaMerge/Remote/DataStore/DataStoreS3.cpp
```diff
@@ -41,7 +41,7 @@ void DataStoreS3::putDMFile(DMFilePtr local_dmfile, const S3::DMFileOID & oid, b
     auto s3_client = S3::ClientFactory::instance().sharedTiFlashClient();
 
     std::vector<std::future<void>> upload_results;
-    for (const auto & fname : local_files)
+    for (const auto & [fname, fsize] : local_files)
```
JaySon-Huang, line 44: maybe we need an `UNUSED(fsize)` in the loop to avoid the compiler warning?

JaySon-Huang (approved):
LGTM

PR #7110, opened 2023-03-20
Disagg read work with autoscaler

JaySon-Huang:

Problem Summary:

Depends on:
* https://github.com/pingcap/kvproto/pull/1081

When deployed with auto scaler, the compute node dose not know s3.bucket/s3.root/s3.endpoint when it starts. We add a RPC for the compute nodes fetching the config from write node.

For the compute node deployed with auto scaler enabled, the S3 client init will be postponed to the first time `ClientFactory::sharedTiFlashClient` is called. And it will use client-c to find one write node and call `GetDisaggConfig` to fetch the config to initlize the S3 client.

Another small fix: Ignore `NoSuchLifecycleConfiguration` error when user does not set lifecycle at all. Log a warning of `The lifecycle rule with filter has not been set`

**Behavior change**
After this PR, when auto scaler is enabled, S3 is always enabled!

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


bestwoody (approved):
LGTM，based on real tests

PR #7132, opened 2023-03-22
Disagg support resolve lock

breezewish:

Problem Summary:

- [x] Merge https://github.com/pingcap/kvproto/pull/1087 first.
- [x] When meeting lock exception, RN will resolve lock.
- [ ] Support query without waiting lock for large txn. (Currently unsupported)

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/StorageDisaggregatedRemote.cpp
```diff
@@ -125,150 +130,178 @@ BlockInputStreams StorageDisaggregated::readFromWriteNode(
...
+                "Received EstablishDisaggregated response with retryable error: Key is locked, addr={} lock={}",
+                batch_cop_task.store_addr,
+                error.locked().DebugString());
+
+            // TODO: Better to construct the Backoff according to remaining backoff time.
+            kv::Backoffer bo(kv::copNextMaxBackoff);
+            kv::LockPtr lock = std::make_shared<kv::Lock>(error.locked());
+            std::vector<uint64_t> pushed;
+            std::vector<kv::LockPtr> locks{lock};
+            auto before_expired = cluster->lock_resolver->resolveLocks(bo, sender_target_mpp_task_id.query_id.start_ts, locks, pushed);
```
JaySon-Huang, line 244: Only retrying one LockException could greatly slow the query time when there are many unclean locks among all Regions. We need to change the logic in `LearnerRead` to return multiple LockException and resolve locks in a batch

dbms/src/Storages/Transaction/LearnerRead.cpp
```diff
@@ -55,18 +55,18 @@ struct UnavailableRegions
...
     }
 
     void tryThrowRegionException(bool batch_cop)
     {
         // For batch-cop request, all unavailable regions, include the ones with lock exception, should be collected and retry next round.
         // For normal cop request, which only contains one region, LockException should be thrown directly and let upper layer(like client-c, tidb, tispark) handle it.
-        if (!batch_cop && region_lock)
-            throw LockException(region_lock->first, std::move(region_lock->second));
+        if (!batch_cop && !region_locks.empty())
+            throw LockException(std::move(region_locks));
```
JaySon-Huang, line 69: When disagg read, we will call `doLearnerRead(... /*for_batch_cop=*/true, ...)`. So I think it will not run into this if branch and no LockException is thrown under disagg read now.
This is sufficient for batch cop or MPP, since they handle `RegionException` and `LockException` in the same way. But disagg read need to acquire the LockInfo in `FlashService::EstablishDisaggTask` and send it back to RN. So I think the changes now is not sufficient.

https://github.com/pingcap/tiflash/blob/c83a185453b4b5a3cb8ad43307ee761ae725fada/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp#L558-L566

P.S. Can we add a unit test to ensure the behavior that Region lock is thrown correctly to the upper level?

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -561,7 +561,8 @@ LearnerReadSnapshot DAGStorageInterpreter::doBatchCopLearnerRead()
         {
             // We can also use current thread to resolve lock, but it will block next process.
             // So, force this region retry in another thread in CoprocessorBlockInputStream.
-            force_retry.emplace(e.region_id);
+            for (const auto & lock : e.locks)
+                force_retry.emplace(lock.first);
```
JaySon-Huang, line 565: This will only set the region id into `region_retry_from_local_region` and swallowed up the `LockInfo`, making `FlashService::EstablishDisaggTask` only get `RegionException` but not `LockException`.
Maybe we need another function for disagg read.

dbms/src/Storages/StorageDisaggregatedRemote.cpp
```diff
@@ -125,150 +130,178 @@ BlockInputStreams StorageDisaggregated::readFromWriteNode(
...
+            // Try to resolve all locks.
+            kv::Backoffer bo(kv::copNextMaxBackoff);
+            std::vector<uint64_t> pushed;
+            std::vector<kv::LockPtr> locks{};
+            for (const auto & lock_info : error.locked())
+                locks.emplace_back(std::make_shared<kv::Lock>(lock_info));
+            auto before_expired = cluster->lock_resolver->resolveLocks(bo, sender_target_mpp_task_id.query_id.start_ts, locks, pushed);
+
+            // TODO: Use `pushed` to bypass large txn.
+            LOG_DEBUG(log, "Finished resolve locks, n_locks={} pushed.size={} before_expired={}", locks.size(), pushed.size(), before_expired);
```
JaySon-Huang, line 248: Better add some metrics about the time consumed in resolving locks

dbms/src/Storages/Transaction/LearnerRead.cpp
breezewish: How about changing to batch cop first, and then I update this PR? Currently it is hard for me to check whether the change "will" work under batch cop.

JaySon-Huang: now it is already running with "batchcop =true"

JaySon-Huang: if (dag_context.isBatchCop() || dag_context.isMPPTask() || dag_context.is_disaggregated_task)

JaySon-Huang: https://github.com/pingcap/tiflash/blob/ea748e8c4b60902f6d18d642cd279708d26afbf3/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp#L379-L382

JaySon-Huang (approved):
LGTM

flowbehappy (approved):

PR #7143, opened 2023-03-22
Fix failing to read from blob store when upload checkpoint

lidezhu:

Issue Number: ref #6827 

Problem Summary: When upload checkpoint, it is possible to read page which are already deleted and may fail to read it.

Previously, we always use the output of `dumpSnapshotToEdit`, but it may contain deleted page id which have not been gced. And if the page id is gced during the process of uploading checkpoint, the read may fail with all kinds of error;
Now we create a temp page directory use the output from `dumpSnapshotToEdit`, then do gc on it, and `dumpSnapshotToEdit` again to filter out all deleted page id to eliminate this kind of error.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/Universal/UniversalPageStorage.cpp
```diff
@@ -430,6 +430,15 @@ PS::V3::CPDataWriteStats UniversalPageStorage::dumpIncrementalCheckpoint(const U
...
         sequence = options.override_sequence.value();
 
+    {
+        // The output of `PageDirectory::dumpSnapshotToEdit` may contain page ids which are logically deleted but have not been gced yet.
+        // These page ids may be gcced when dump snapshot, so we cannot read data of these page ids.
+        // So we create a clean temp page_directory here and use it to dump edits with all visible page ids for `snap`.
+        PS::V3::universal::PageDirectoryFactory factory;
+        auto temp_page_directory = factory.dangerouslyCreateFromEditWithoutWAL(fmt::format("{}_{}", storage_name, sequence), edit_from_mem);
+        edit_from_mem = temp_page_directory->dumpSnapshotToEdit();
+    }
```
JaySon-Huang, line 440: This way is a bit wasteful of memory and CPU resources, and also increases the time consumed for uploading. Let's think about whether there is a better solution first.

lidezhu: OK, I also hope for a better solution.

lidezhu:
PTAL @JaySon-Huang @JinheLin

JaySon-Huang (approved):
LGTM

dbms/src/Storages/Page/V3/Universal/UniversalPageStorage.cpp
JaySon-Huang: There are no better solutions so far. Get this merge to fix the exception first.

Issue #6852, opened 2023-02-20
Optimize count distinct to 3 stages of distributed agg


AilinKid:
## Enhancement
redirect issue to
https://github.com/pingcap/tidb/issues/34704

open this for `do-not-merge/needs-linked-issue`

PR #6545, opened 2022-12-26
plan, executor: implement Expand operator for grouping sets

AilinKid:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tiflash/issues/6852, https://github.com/pingcap/tidb/issues/16581 https://github.com/pingcap/tidb/issues/34704

Problem Summary: this PR is a basic implementation of the Expand operator (according to grouping sets definition) to facilitate the multi-distinct aggregate (now only for count(distinct xxx)), utilizing multi tiflash nodes to accelerate the computing process rather than single-node bottleneck before. details see the related TiDB PR.

### What is changed and how it works?
1: Receive one more defined physical operator, interpret it and build a related executor --- Expand
2: Implement the Expand execution logic according to the grouping sets definition.
3: Consider physical plan re-conversion and old cop-execution mode.

TiDB PR: https://github.com/pingcap/tidb/pull/39973
TiPB PR: https://github.com/pingcap/tipb/pull/283

Example
image

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
plan, executor: implement Expand operator for grouping sets
```

PR #7169, opened 2023-03-27
planner, executor: refactor expand logic from analyzing grouping sets to level projections

AilinKid:

Issue Number: close https://github.com/pingcap/tiflash/issues/6852

Problem Summary:

In previous PR https://github.com/pingcap/tiflash/pull/6545, we introduced Expand operator to analyze raw grouping-sets pb, and generating the implicit projection logic (basically expand and change the value in the block itself)

While for the better implementation of ROLL UP syntax, which means more dynamic generated columns and more complicated grouping sets definition, we choose to move the analyzing work to TiDB planner, and let the execution logic in tiflash side as clean and simple as possible. 

In this PR, TiFlash should receive the level-projections pb structure from TiDB planner, and what it only need to concern is to just project it out (one level projection for outputting a individual block itself)

In the new implementation, we take expand as projection, so it will be also seen as a datasource node in query block division. Additionally, we introduced a ExpandBlockInputStream to control the block cache and current projected index position which is being done the same in ExpandTransformAction.

the new pb structure will be like below:
```
message ProjectExpr {
	repeated Expr exprs = 1;
}

message Expand2{
	repeated ProjectExpr proj_exprs = 1;         // for grouping sets generated projection levels, like [a, b, c, 1#gen_col1, 2#gen_col2]
	repeated string generated_output_names = 2;  // for output names for generated cols like grouping_id etc, like "gen_col1", "gen_col2" here
	optional Executor child = 3;
}
```

TiPB link: https://github.com/pingcap/tipb/pull/300

**Note: for now, since TiDB still relay on the old Expand logic (but not public to user yet, the default switch is `off` now), after we refactored the old logic in Multi Distinct Aggregate from `expand` to `expand2`, I will remove all the old logic completely without concern about the compatibility in the rolling update.**

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
planner, executor: refactor expand logic from analyzing grouping sets to level projections
```


dbms/src/DataStreams/ExpandBlockInputStream.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    children.push_back(input);
+    i_th_project = 0;
+}
+
+Block ExpandBlockInputStream::getHeader() const
+{
+    Block res(names_and_types);
+    return res;
+}
```
SeaRise, line 38: `ColumnPtr` is required for the Block returned by `getHeader`.
`ColumnPtr` is not needed only for `getSampleBlock`.

dbms/src/Interpreters/ExpressionActions.h
```diff
@@ -78,6 +81,9 @@ struct ExpressionAction
     std::string result_name;
     DataTypePtr result_type;
 
+    /// For CHANGE_NULLABLE
+    std::string col_to_nullable;
```
xzhangxian1008, line 85:
```suggestion
    String col_need_be_nullable;
```

dbms/src/DataStreams/ExpandTransformAction.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+Block ExpandTransformAction::getHeader() const
+{
+    return header;
+}
+
+bool ExpandTransformAction::try_output(Block & block)
+{
+    if (i_th_project >= expand->getLevelProjectionNum())
```
SeaRise, line 39:
```suggestion
    if (!block_cache || i_th_project >= expand->getLevelProjectionNum())
```

dbms/src/DataStreams/ExpandBlockInputStream.cpp
SeaRise: like https://github.com/pingcap/tiflash/blob/1dc2d1af13e67f8c0b51414329aef71cbc5209ef/dbms/src/DataStreams/MockExchangeReceiverInputStream.h#L31-L33

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+Block Expand2::next(const Block & block_cache, size_t i_th_project)
+{
+    /// step1: clone a new block
+    ColumnsWithTypeAndName cloned_columns_and_type;
+    auto num_columns = block_cache.columns();
+    MutableColumns res(num_columns);
+    for (size_t i = 0; i < num_columns; ++i)
+    {
+        res[i] = block_cache.getColumns()[i]->cloneResized(block_cache.rows());
```
xzhangxian1008, line 34: For the columns that must be null, I think we should not clone their full content,

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,101 @@
...
+        ColumnWithTypeAndName column = cloned_block.getByName(name);
+        if (!alias.empty())
+            column.name = alias;
+        new_block.insert(std::move(column));
+    }
+
+    /// step4: unfold the constant column which is not meaningful in global scope.
+    for (size_t i = 0; i < new_block.getColumnsWithTypeAndName().size(); i++)
+    {
+        auto new_col = new_block.getColumnsWithTypeAndName()[i];
```
xzhangxian1008, line 63: Maybe rename `new_col` to col, as it has been in the context of `new_block`.

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (new_col.column->isColumnConst())
+        {
+            // if it's a new literal constant column, unfold it, and if it's a not an origin constant column, unfold it.
```
xzhangxian1008, line 66:
```suggestion
            // if it's a new literal constant column, unfold it, and if it's not an origin constant column, unfold it.
```

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,101 @@
...
+            // eg: grouping id projection(1) and grouping set column projection(null)
+            if (!block_cache.has(new_col.name) || !block_cache.getByName(new_col.name).column->isColumnConst())
+                new_block.safeGetByPosition(i).column = new_col.column->convertToFullColumnIfConst();
```
xzhangxian1008, line 69: Why block_cache may not has new_col.name?

dbms/src/DataStreams/ExpandBlockInputStream.cpp
AilinKid: addressed, https://github.com/pingcap/tiflash/pull/7169/commits/2f2f3255e2f2a1b20572e086d3be9096292ac6fe

dbms/src/Interpreters/ExpressionActions.cpp
```diff
@@ -149,6 +149,14 @@ ExpressionAction ExpressionAction::expandSource(GroupingSets grouping_sets_)
     return a;
 }
 
+ExpressionAction ExpressionAction::addNullable(const std::string & col_name)
```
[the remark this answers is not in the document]
AilinKid:
```
    static ExpressionAction addColumn(const ColumnWithTypeAndName & added_column_);
    static ExpressionAction removeColumn(const std::string & removed_name);
    static ExpressionAction copyColumn(const std::string & from_name, const std::string & to_name);
    static ExpressionAction project(const NamesWithAliases & projected_columns_);
    static ExpressionAction project(const Names & projected_columns_);
    static ExpressionAction ordinaryJoin(std::shared_ptr<const Join> join_, const NamesAndTypesList & columns_added_by_join_);
    static ExpressionAction expandSource(GroupingSets grouping_sets);

this one：
    static ExpressionAction convertToNullable(const std::string & col_name);
```
how about convertToNullable？maybe more regular here

dbms/src/Interpreters/Expand2.cpp
AilinKid: in old version you reviewed, projection of literal "1", leading const column won't exist in the source block cache.  
in new version, the before_expand_action will create const column ahead, so I removed it in new codes.

dbms/src/Interpreters/Expand2.cpp
AilinKid: make sense.
we should dig out which col in this level is projected as null, then, do something elimination of copy work here. For now, multi_level_actions is seen as black box here, no too much analyzing work, just use `level_action.execute(block)` to get more simplicity

dbms/src/Interpreters/ExpressionActions.cpp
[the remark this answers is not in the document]
xzhangxian1008: > ```
>     static ExpressionAction addColumn(const ColumnWithTypeAndName & added_column_);
>     static ExpressionAction removeColumn(const std::string & removed_name);
>     static ExpressionAction copyColumn(const std::string & from_name, const std::string & to_name);
>     static ExpressionAction project(const NamesWithAliases & projected_columns_);
>     static ExpressionAction project(const Names & projected_columns_);
>     static ExpressionAction ordinaryJoin(std::shared_ptr<const Join> join_, const NamesAndTypesList & columns_added_by_join_);
>     static ExpressionAction expandSource(GroupingSets grouping_sets);
> 
> this one：
>     static ExpressionAction convertToNullable(const std::string & col_name);
> ```
> 
> how about convertToNullable？maybe more regular here

okk

dbms/src/Interpreters/tests/gtest_expand_projection.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace DB
+{
+namespace tests
+{
+class ExpandProjectionTest : public DB::tests::ProjectionTest
+{
+};
+
+TEST_F(ExpandProjectionTest, literal_projection)
```
xzhangxian1008, line 32: use camel case please

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        cloned_columns_and_type.emplace_back(std::move(res[i]), origin_column_and_types[i].type, origin_column_and_types[i].name, origin_column_and_types[i].column_id, origin_column_and_types[i].default_value);
+    }
+    Block cloned_block(cloned_columns_and_type);
+
+    /// step2: execute a new block.
+    auto ith_projection = leveled_projections_actions[i_th_project];
+    ith_projection->execute(cloned_block);
+
+    /// step3: change raw name to unified alias name.
```
xzhangxian1008, line 48: Actually, step3 picks some columns from the cloned_block, so this comment makes reader confused.

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,107 @@
...
+    Block new_block;
+    auto names_with_alias = leveled_alias_projections[i_th_project];
+    for (auto & one_alias : names_with_alias)
+    {
+        const std::string & name = one_alias.first;
+        const std::string & alias = one_alias.second;
+        ColumnWithTypeAndName column = cloned_block.getByName(name);
+        if (!alias.empty())
+            column.name = alias;
+        new_block.insert(std::move(column));
```
xzhangxian1008, line 58: Only useful columns will be picked up in step3, so I think we can defer the operation of cloning columns in step1 and trigger the clone operation in step3. More specifically, each time we want to pick up a column that from clone_block, we can check if it should be cloned.

dbms/src/Interpreters/tests/gtest_expand_projection.cpp
xzhangxian1008, line 32: camel case please

dbms/src/Interpreters/Expand2.cpp
xzhangxian1008, line 48: Step3 actually picks some columns up and insert them into a new block, so I think this comment may confuse the reader.

dbms/src/Interpreters/Expand2.cpp
xzhangxian1008, line 58: Maybe we can defer the operation of cloning the original columns' content. We could know which columns are needed at here, and unnecessary cloning mentioned at the previous comments could be avoided, I think.

xzhangxian1008:
Any update?

dbms/src/Interpreters/Expand2.cpp
AilinKid: now the case is: 
clone a brand-new origin block, apply the projection actions on them, we don't take too much role into what the projection actions do in detail (and at last, pick the result columns and rename)

we have already got `beforeExpandActions`, whch is quite like `prependProjectionActions` to prune those unneeded columns as a whole, then doing each level projection individually. 

suggestions indicates we may need a fine-granted `copyFilter` before each level projection to find what column-ref in this level projection real need (for example, literal null didn't need a real col ref in current level projection), then copy those necessary source columns. 

good suggestion, a bit complicated, I will left it to next pull request as an enhancement.

dbms/src/Interpreters/Expand2.cpp
xzhangxian1008: > make sense. we should dig out which col in this level is projected as null (varies from levels), then, do something elimination of copy work here. For now, multi_level_actions is seen as black box here, no too much analyzing work, just use `level_action.execute(block)` to get more simplicity

projection action will add null attribute on some columns in L46. I think we can record which columns are null before and after executing. By comparing the difference we can know which columns are the columns that is projected to null and we can avoid the cloning on these columns.

xzhangxian1008: Please add todo for further performance optimization.

dbms/src/TestUtils/ProjectionTestUtil.h
```diff
@@ -0,0 +1,92 @@
+// Copyright 2022 PingCAP, Ltd.
```
xzhangxian1008, line 1:
```suggestion
// Copyright 2023 PingCAP, Ltd.
```

dbms/src/TestUtils/ProjectionTestUtil.h
```diff
@@ -0,0 +1,92 @@
...
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    const ColumnNumbers & column_literal_numbers,
+    std::vector<Field> val_fields,
+    const ColumnNumbers & column_ref_numbers);
+
+std::pair<ExpressionActionsPtr, std::vector<String>> buildLiteralProjection(
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    std::vector<Field> val_fields);
```
xzhangxian1008, line 49:
```suggestion
    const std::vector<Field> & val_fields);
```

dbms/src/TestUtils/ProjectionTestUtil.cpp
```diff
@@ -0,0 +1,174 @@
...
+        result_names.push_back(result_name);
+    }
+    chain.finalize();
+    return std::make_pair(last_step.actions, result_names);
+}
+
+std::pair<ExpressionActionsPtr, std::vector<String>> buildLiteralProjection(
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    const std::vector<Field> val_fields)
```
xzhangxian1008, line 90:
```suggestion
    const std::vector<Field> & val_fields)
```

dbms/src/TestUtils/ProjectionTestUtil.cpp
```diff
@@ -0,0 +1,174 @@
...
+        last_step.actions->add(ExpressionAction::convertToNullable(columns[column_ref_number].name));
+    }
+    return last_step.actions;
+}
+
+std::pair<ExpressionActionsPtr, std::vector<String>> buildProjection(
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    const ColumnNumbers & column_literal_numbers,
+    const std::vector<Field> val_fields,
```
xzhangxian1008, line 51:
```suggestion
    const std::vector<Field> & val_fields,
```

dbms/src/TestUtils/ProjectionTestUtil.h
```diff
@@ -0,0 +1,92 @@
...
+ExpressionActionsPtr buildChangeNullable(
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    const ColumnNumbers & column_nullable_numbers);
+
+std::pair<ExpressionActionsPtr, std::vector<String>> buildProjection(
+    Context & context,
+    const ColumnsWithTypeAndName & columns,
+    const ColumnNumbers & column_literal_numbers,
+    std::vector<Field> val_fields,
```
xzhangxian1008, line 43:
```suggestion
    const std::vector<Field> & val_fields,
```

dbms/src/Interpreters/ExpressionActions.cpp
```diff
@@ -349,6 +361,15 @@ void ExpressionAction::execute(Block & block) const
         break;
     }
 
+    case CONVERT_TO_NULLABLE:
+    {
+        // for expand usage, when original col is const non-null value, the inserted null value will break its const attribute in global scope.
+        block.getByName(col_need_to_nullable).column->convertToFullColumnIfConst();
```
xzhangxian1008, line 367: Why converted full column is not assigned to another column?

dbms/src/Interpreters/ExpressionActions.cpp
```diff
@@ -257,6 +263,12 @@ void ExpressionAction::prepare(Block & sample_block)
         sample_block.insert({nullptr, expand->grouping_identifier_column_type, expand->grouping_identifier_column_name});
         break;
     }
+    case CONVERT_TO_NULLABLE:
+    {
```
xzhangxian1008, line 267: `convertToFullColumnIfConst` is needless here?

dbms/src/Interpreters/Expand2.h
```diff
@@ -0,0 +1,69 @@
...
+
+class Expand2
+{
+public:
+    /// Expand2 is used to level-project child block according projection expressions.
+    /// In grouping sets usage, multi-distinct-agg cases, Rollup and Cute syntax scenarios,
+    /// all of them resort to grouping-sets way to replicate the data to feed different groups,
+    /// underlying which it's mainly supported by expand operator.
+    ///
+    /// eg: select count(a) from t group by a,b,c rollup
```
xzhangxian1008, line 33:
```suggestion
    /// eg: select count(a) from t group by a,b,c with rollup
```

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,107 @@
...
+    buffer.append("]");
+    return buffer.toString();
+}
+
+size_t Expand2::getLevelProjectionNum() const
+{
+    return leveled_projections_actions.size();
+}
+
+ExpressionActionsPtr & Expand2::getBeforeExpandActions()
```
xzhangxian1008, line 103: We should avoid to return reference as possible as we can and the type is already a smart pointer, so I think return `ExpressionActionsPtr` is enough.

dbms/src/Interpreters/Expand2.h
```diff
@@ -0,0 +1,69 @@
...
+    /// Since tiflash made the raw name for every expression it received in getActions(expr,cols),
+    /// we should also cast all the leveled column to the same output alias name after evaluation.
+    ///
+    /// \param leveled_projections_actions
+    /// \param level_alias_projections
+    explicit Expand2(ExpressionActionsPtrVec projections_actions_, ExpressionActionsPtr before_expand_actions_, NamesWithAliasesVec projections_);
+    Block next(const Block & block_cache, size_t i_th_projection);
+    String getLevelProjectionDes() const;
+    size_t getLevelProjectionNum() const;
+    ExpressionActionsPtr & getBeforeExpandActions();
```
xzhangxian1008, line 50: ditto

dbms/src/DataStreams/ExpandTransformAction.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // empty block should be output too.
+    if (unlikely(!block))
+        return;
+
+    // when calling this function, that means we read a new block from source OP, and pull it up through lower to upper OP incrementally.
+    // so just caching this new block here and handle expansion, notice: Expand never accumulate blocks.
+    expand->getBeforeExpandActions()->execute(block);
+    block_cache = block;
+    i_th_project = 0;
+    auto res_block = expand->next(block_cache, i_th_project++);
```
xzhangxian1008, line 57:
```suggestion
    i_th_project = 1;
    auto res_block = expand->next(block_cache, i_th_project);
```
I think this is better

xzhangxian1008:
Any update?

dbms/src/DataStreams/ExpandBlockInputStream.cpp
```diff
@@ -0,0 +1,55 @@
...
+    const BlockInputStreamPtr & input,
+    const Expand2Ptr & expand2_,
+    const Block & header_,
+    const String & req_id)
+    : expand2(expand2_)
+    , header(header_)
+    , log(Logger::get(req_id))
+{
+    children.push_back(input);
+    i_th_project = 0;
```
guo-shaoge, line 31: init `i_th_project` in init list

dbms/src/Interpreters/ExpressionActions.cpp
AilinKid: make sense, addressed

dbms/src/DataStreams/ExpandBlockInputStream.cpp
AilinKid: Not exactly, when pipeline fetch a new block below Expand OP,  we should cache this new mother block and **rewind the current projection offset** for the convenience of later level projections (each time output a new projected block in function tryOutput(...))

dbms/src/Interpreters/Expand2.cpp
AilinKid: image
actually, we wanna pass it to this function which receive a & ref of expressionActions, a ExpressionActionsPtr & will be more clear. Otherwise, we need a make a var to get the temporary value of what getBeforeExpandActions returned.

xzhangxian1008 (approved):
LGTM

dbms/src/Interpreters/Expand2.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+Block Expand2::next(const Block & block_cache, size_t i_th_project)
+{
+    /// step1: clone a new block
+    ColumnsWithTypeAndName cloned_columns_and_type;
+    auto num_columns = block_cache.columns();
+    MutableColumns res(num_columns);
+    for (size_t i = 0; i < num_columns; ++i)
+    {
+        res[i] = block_cache.getColumns()[i]->cloneResized(block_cache.rows());
```
xzhangxian1008, line 35: Please add todo here

guo-shaoge (approved):

PR #7228, opened 2023-04-06
Fix data race in join when `Join::cancel()` is called

windtalker:

Issue Number: ref #6528

Problem Summary:
There are two kinds of data race
- When `Join::cancel()` is called when some threads are wait on `Join::waitUntilAllBuildFinished`, the waiting threads will be waken up, and begin reading from hash table while some build threads are still inserting data into the same hash table
- When a right outer join spill is triggered, and `Join::cancel()` is called when some threads are wait on `Join::waitUntilAllBuildFinished`, the waiting threads will be waken up, and call `Join::finishOneNonJoin` at the end, and inside `Join::finishOneNonJoin`, it will clear the hash map, at the same time, it is possible that other threads are still accessing that hash map.
 
1. Only clear hash map if there is no active build/probe threads
2. For Join Probe, return empty block if there are still active build threads
3. For NonJoinedBlockInputStream, return empty block if there are still active build/probe threads
4. rename `Join::cancel` to `Join::wakeUpAllWaitingThreads`
### Check List

Tests 

- [ ] Unit test
- [x] Integration test
       add random failpoints tests
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/Join.cpp
+++ b/dbms/src/Interpreters/Join.cpp
@@ -2862,6 +2862,7 @@ Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
 {
     if unlikely (active_build_threads != 0)
     {
+        /// build is not finished yet, the query must be cancelled, so just return {}
         LOG_WARNING(log, "JoinBlock without non zero active_build_threads, return empty block");
         return {};
     }
--- a/dbms/src/Interpreters/Join.cpp
+++ b/dbms/src/Interpreters/Join.cpp
@@ -2807,7 +2807,7 @@ void Join::waitUntilAllBuildFinished() const
 {
     std::unique_lock lock(build_probe_mutex);
     build_cv.wait(lock, [&]() {
-        return active_build_threads == 0 || meet_error || is_canceled;
+        return active_build_threads == 0 || meet_error || skip_wait;
     });
     if (meet_error)
         throw Exception(error_message);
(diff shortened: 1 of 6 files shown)


dbms/src/Interpreters/Join.cpp
```diff
@@ -2841,21 +2841,31 @@ void Join::waitUntilAllProbeFinished() const
...
+            }
+            partition_index += build_concurrency;
         }
-        partition_index += build_concurrency;
     }
 }
 
 Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
 {
+    if unlikely (active_build_threads != 0)
```
yibin87, line 2863: Do we have other ways to check cancel situation, like isCancelledOrThrowIfKilled?
Others LGTM

windtalker: We can't use `isCancelledOrThrowIfKilled` here because when one thread meet eof, it will call `cancel()`, and in this case, other threads should not affected by that `cancel()`

yibin87 (approved):
LGTM

dbms/src/Interpreters/Join.cpp
```diff
@@ -2841,21 +2841,38 @@ void Join::waitUntilAllProbeFinished() const
...
+            partition_index += build_concurrency;
         }
-        partition_index += build_concurrency;
     }
 }
 
-Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
+Block Join::joinBlock(ProbeProcessInfo & probe_process_info, bool dry_run) const
 {
+    if (dry_run)
```
SeaRise, line 2863:
```suggestion
    if unlikely (dry_run)
```

SeaRise (approved):
Others LGTM

Issue #4809, opened 2022-04-29
Enhance default config behavior


breeswish:
## Enhancement

Currently many TiFlash config default values are evaluated in a wrong way. For example, given the following two config files, the evaluated final config are different!

**Config-A:** Final `enable_ps_v3 = false`

```toml
tcp_port = 9000
tmp_path = "/tmp/tiflash/data/tmp"
path = "/tmp/tiflash/data/db"

[storage]

[logger]
console = 1
level = "trace"
```

**Config-B:** Final `enable_ps_v3 = true`

```toml
tcp_port = 9000
tmp_path = "/tmp/tiflash/data/tmp"
path = "/tmp/tiflash/data/db"

# [storage]  <---- No storage section

[logger]
console = 1
level = "trace"
```

The root cause of **this specific issue** is that, `enable_ps_v3` is evaluated in the following way:

```cpp
if (config.has("storage"))
{
    storage_config.parseMisc(config.getString("storage"), log);
}

// In parseMisc:

if (table->contains("enable_ps_v3"))
{
    ...
}
else
{
    ...
    enable_ps_v3 = true;
}
```

Thus, if `[storage]` is not defined, the default value of `enable_ps_v3` will be evaluated different.

Throughout the whole code base, not only `enable_ps_v3`, but also other config items have similar problems.

We can improve the config architecture, to achieve these goals:

1. In the code base, there should be **only one** place defining the default value of config item.
2. The default value of each config is effective as long as this config item is not explicitly defined in the config file.

PR #7251, opened 2023-04-08
Allow TiFlash server run without a config file

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/4809

Problem Summary:

**Major change:**

Allow TiFlash to start without a config file (and allow passing necessary config via args instead). In this way, we could start a TiFlash with a reuseable customize config in TiUP in this way:

```bash
tiup playground nightly --tiflash.config <path_to_a_reuseable_config>
```

**Detailed changes:**

- [x] Support start TiFlash server without a config file.

  A minimal working sample:

  ```shell
  dbms/src/Server/tiflash server -- --path=./data --raft.pd_addr=http://127.0.0.1:2379
  ```

- [x] Fix boost compile failure in PocoHTTPClient in Clang15.
- [x] Fix the issue that we cannot specify a relative path in `path`.
- [x] Support `./tiflash server --help`.
- [x] Output the reason when TiFlash proxy is not started.
- [x] Disable writing `xxx_preprocessed.toml` because it is not used.
- [x] Add colors when output logs in the terminal directly:    

  image

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Server/Server.cpp
```diff
@@ -275,38 +293,41 @@ struct TiFlashProxyConfig
...
-                args_map[engine_store_address] = config.getString("flash.service_addr");
-            else
-                args_map[engine_store_advertise_address] = args_map[engine_store_address];
+                args_map[key] = config.getString("flash.proxy." + key);
 
-            args_map[engine_label] = getProxyLabelByDisaggregatedMode(disaggregated_mode);
+            args_map["pd-endpoints"] = config.getString("raft.pd_addr");
+            args_map["engine-version"] = TiFlashBuildInfo::getReleaseVersion();
+            args_map["engine-git-hash"] = TiFlashBuildInfo::getGitHash();
+            if (!args_map.count("engine-addr"))
```
Lloyd-Pottiger, line 321:
```suggestion
            if (args_map.contains("engine-addr"))
```

Lloyd-Pottiger (approved):
LGTM

dbms/src/Common/UnifiedLogFormatter.cpp
```diff
@@ -31,30 +50,63 @@ void UnifiedLogFormatter::format(const Poco::Message & msg, std::string & text)
 
     // Timestamp
     {
+        if (enable_color)
```
JaySon-Huang, line 53: It would be better if we make `enable_color` as a template param and use `if constexpr (enable_color)`. But it does not block this PR

JaySon-Huang (approved):
LGTM with a comment

dbms/src/Server/RaftConfigParser.cpp
```diff
@@ -35,8 +35,14 @@ TiFlashRaftConfig TiFlashRaftConfig::parseSettings(Poco::Util::AbstractConfigura
...
-    if (!config.has("raft"))
-        return res;
+    {
+        // Check by `raft` prefix instead of check by `config.has("raft")`,
+        // because when sub keys are set from cli args, `raft` will not exist.
+        Poco::Util::AbstractConfiguration::Keys keys;
+        config.keys("raft", keys);
+        if (keys.empty())
+            return res;
+    }
```
JaySon-Huang, line 45: Note that before this PR, the config file must contain a line `[raft]`, otherwise tiflash will starts as a standalone mode just for mock test, without connecting to pd and tikv.

After this PR, the raft.pd_addr can be set from cli args and tiflash can connect to pd and tikv correctly.

PR #7253, opened 2023-04-09
Region retry for partitioned table in disaggrgated read

breezewish:

Problem Summary:

Drop regions when retry for partitioned table scan.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test

```sql
create table t1(id int, age int, key(id)) partition by range(id) (
  partition p0 values less than (100),
  partition p1 values less than (200),
  partition p2 values less than (300),
  partition p3 values less than (400));

insert into t1 values (1, 10);
insert into t1 values (10, 10);

alter table t1 set tiflash replica 1;

set @@session.tidb_isolation_read_engines='tiflash';
set @@session.tidb_partition_prune_mode = 'dynamic';

analyze table t1;

select * from information_schema.tiflash_replica;

select count(*) from t1 where id < 150;

-- Trigger a region ver change manually

split table t1 by (4), (7);

select count(*) from t1 where id < 150;
```

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/StorageDisaggregatedRemote.cpp
+++ b/dbms/src/Storages/StorageDisaggregatedRemote.cpp
@@ -234,36 +234,29 @@ void StorageDisaggregated::buildDisaggTask(
                 batch_cop_task.store_addr,
                 retry_regions);
 
-            for (const auto & region : req->regions())
-            {
-                if (retry_regions.contains(region.region_id()))
+            auto retry_from_region_infos = [&retry_regions, &cluster](const google::protobuf::RepeatedPtrField<::coprocessor::RegionInfo> & region_infos) {
+                for (const auto & region : region_infos)
                 {
-                    auto region_ver_id = pingcap::kv::RegionVerID(
-                        region.region_id(),
-                        region.region_epoch().conf_ver(),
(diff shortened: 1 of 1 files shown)


dbms/src/Storages/StorageDisaggregatedRemote.cpp
```diff
@@ -234,17 +234,33 @@ void StorageDisaggregated::buildDisaggTask(
...
+                        cluster->region_cache->dropRegion(region_ver_id);
+                        retry_regions.erase(region.region_id());
+                        if (retry_regions.empty())
+                            break;
+                    }
                 }
+            };
+
+            retry_from_region_infos(req->regions());
+            for (const auto & table_region : req->table_regions())
```
JaySon-Huang, line 255:
```suggestion
            // non-partition table
            retry_from_region_infos(req->regions());
            // partition table
            for (const auto & table_region : req->table_regions())
```

JaySon-Huang (approved):
LGTM

Lloyd-Pottiger (approved):

PR #6529, opened 2022-12-22
merge output blocks if need in hash join

mengxin9014:
### What problem does this PR solve?

Issue Number: close #6533

Problem Summary:

### What is changed and how it works?
if the output block of hash join is too small, we will save it temporarily and merge multiple return output blocks to the final output block so that the size of final block is closer to max block size.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
merge output blocks if need in hash join.
```

PR #7268, opened 2023-04-11
merge join result block if needed

### LINKED ISSUE #6533: merge join probe return block if need

windtalker:

Issue Number: close #6533

Problem Summary:

This pr is a re-implementaion of #6529, #6529 is reverted because it will make performance regression. Compared to #6529, this pr reduce the potential performance regression by
- Use a more reasonable threshold to trigger block merge
- Only merge the result blocks inside one probe block 

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Join.cpp
```diff
@@ -823,147 +823,173 @@ void Join::handleOtherConditions(Block & block, std::unique_ptr<IColumn::Filter>
...
-    recordFilteredRows(block, non_equal_conditions.left_filter_column, null_map_holder, null_map);
-
-    size_t existing_columns = block.columns();
-
-    /** If you use FULL or RIGHT JOIN, then the columns from the "left" table must be materialized.
+    std::vector<Block> result_blocks;
+    size_t result_rows = 0;
+    while (true)
+    {
+        probe_process_info.updateStartRow();
```
windtalker, line 830: Actually wrap the original `Join::joinBlockHash()` by a `while` loop, the code insde the loop is not changed.

SeaRise: how about wrap the code like
```
Block doJoinBlockHash();

Block joinBlockHash()
{
    while (true)
    {
        doJoinBlockHash();
    }
}
```
?

windtalker: Done

dbms/src/Interpreters/Join.cpp
```diff
@@ -966,6 +968,37 @@ Block Join::joinBlockHash(ProbeProcessInfo & probe_process_info) const
...
+    assert(!result_blocks.empty());
+    if (result_blocks.size() == 1)
+    {
+        return result_blocks[0];
+    }
+    else
+    {
+        return vstackBlocks(std::move(result_blocks));
+    }
+}
```
SeaRise, line 1000:
```suggestion
    vstackBlocks(std::move(result_blocks));
```

SeaRise (commented):
Others LGTM

dbms/src/Interpreters/Join.cpp
```diff
@@ -966,6 +968,37 @@ Block Join::joinBlockHash(ProbeProcessInfo & probe_process_info) const
...
+        /// 1. probe_process_info.all_rows_joined_finish is true, which means all the rows in current block is processed
+        /// 2. result_rows exceeds the min_result_block_size
+        if (probe_process_info.all_rows_joined_finish || result_rows >= probe_process_info.min_result_block_size)
+        {
+            if (result_blocks.empty())
+                return block;
+            result_blocks.push_back(std::move(block));
+            break;
+        }
+        result_blocks.push_back(std::move(block));
```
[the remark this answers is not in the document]
windtalker: Done

dbms/src/Interpreters/Join.cpp
```diff
@@ -966,6 +969,26 @@ Block Join::joinBlockHash(ProbeProcessInfo & probe_process_info) const
...
+        /// exit the while loop if
+        /// 1. probe_process_info.all_rows_joined_finish is true, which means all the rows in current block is processed
+        /// 2. the block may be expanded after join and result_rows exceeds the min_result_block_size
+        if (probe_process_info.all_rows_joined_finish || (may_probe_side_expanded_after_join && result_rows >= probe_process_info.min_result_block_size))
+            break;
+    }
+    assert(!result_blocks.empty());
+    return vstackBlocks(std::move(result_blocks));
+}
```
gengliqi, line 990: In which cases should the return `Blocks` by `doJoinBlockHash` be merged? 
In my opinion, if `doJoinBlockHash` returns without finishing, the return `Block` should be large enough to output.

windtalker: If there is other condition，the block returned by doJoinBlockHash may be small

gengliqi (approved):

PR #7271, opened 2023-04-11
Support capture group in regexp_replace

### LINKED ISSUE #7270: Need to support capture group in regexp_replace
## Enhancement

Capture group is not supported in regexp_replace now, and the tiflash is unable to get the correct answer with the following query.

`SELECT REGEXP_REPLACE(Referer, '^https?://(?:www\.)?([^/]+)/.*$', '$1') AS k, AVG(length(Referer)) AS l, COUNT(*) AS c, MIN(Referer) FROM hits WHERE Referer <> '' GROUP BY k HAVING COUNT(*) > 100000 ORDER BY l DESC LIMIT 25;`

xzhangxian1008:

Issue Number: close #7270

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -26,8 +25,15 @@
 #include <optional>
 
 #define MIN_LENGTH_FOR_STRSTR 3
-#define MAX_SUBPATTERNS 5
+#define MAX_SUBPATTERNS 10
```
windtalker, line 28: Is this value must be the same as `max_captures`?

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -604,48 +609,56 @@ std::optional<StringRef> OptimizedRegularExpressionImpl<thread_safe>::substrImpl
...
+
+    // Copy characters that before position
+    res_data.resize(res_data.size() + byte_offset);
+    memcpy(&res_data[res_offset], subject, byte_offset);
+    res_offset += byte_offset;
 
     while (true)
     {
-        bool success = RegexType::FindAndConsume(&expr_sp, *re2, &matched_str);
+        // bool success = RegexType::FindAndConsume(&expr_sp, *re2, &matched_str);
```
windtalker: Remove useless code

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -725,18 +737,80 @@ void OptimizedRegularExpressionImpl<thread_safe>::replace(
...
+            {
+                if (!literals.empty())
+                {
+                    instructions.emplace_back(literals);
+                    literals = "";
+                }
+                instructions.emplace_back(repl.data[i + 1] - '0');
+            }
+            else
+                literals += repl.data[i + 1]; /// Escaping
```
windtalker, line 776: This implicitly assume that `$` is the escaping char, is it the expected behavior?

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -725,18 +737,80 @@ void OptimizedRegularExpressionImpl<thread_safe>::replace(
     Int64 occur)
 {
     Int64 utf8_total_len = DB::UTF8::countCodePoints(reinterpret_cast<const UInt8 *>(subject), subject_size);
-    ;
+
     FunctionsRegexp::checkArgsReplace(utf8_total_len, subject_size, pos);
     FunctionsRegexp::makeReplaceOccurValid(occur);
 
+    auto num_captures = std::min(re2->NumberOfCapturingGroups() + 1, max_captures);
+    Instructions instructions = getInstructions(repl, num_captures);
```
windtalker, line 744: Looks like even if `pattern` column is a constant string ,`getInstructions` will be called for every row?

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > Is this value must be the same as `max_captures`?

Yes, and I will combine them together.

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > This implicitly assume that `$` is the escaping char, is it the expected behavior?

Yes, so does MySQL do.

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > Looks like even if `pattern` column is a constant string ,`getInstructions` will be called for every row?

Yes, and I will memorize them in appropriate place.

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -318,13 +324,13 @@ OptimizedRegularExpressionImpl<thread_safe>::OptimizedRegularExpressionImpl(cons
...
+            throw Poco::Exception(fmt::format("OptimizedRegularExpression: cannot compile re2: {}, error: {}", regexp_, re2->error()));
 
         if (!is_no_capture)
         {
-            number_of_subpatterns = re2->NumberOfCapturingGroups();
-            if (number_of_subpatterns > MAX_SUBPATTERNS)
-                throw Poco::Exception("OptimizedRegularExpression: too many subpatterns in regexp: " + regexp_);
+            capture_num = re2->NumberOfCapturingGroups();
+            if (capture_num > MAX_CAPTURES)
+                throw Poco::Exception(fmt::format("OptimizedRegularExpression: too many subpatterns in regexp: {}", regexp_));
```
windtalker, line 333: I think we can rename all the `subpatterns` to `capture_groups`

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -512,16 +518,15 @@ void OptimizedRegularExpressionImpl<thread_safe>::processReplaceEmptyStringExpr(
 
     StringPieceType expr_sp(subject, subject_size);
     StringPieceType matched_str;
```
windtalker, line 520: Looks like this is not used anymore?

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -604,48 +609,55 @@ std::optional<StringRef> OptimizedRegularExpressionImpl<thread_safe>::substrImpl
 }
 
 template <bool thread_safe>
-void OptimizedRegularExpressionImpl<thread_safe>::replaceAllImpl(const char * subject, size_t subject_size, DB::ColumnString::Chars_t & res_data, DB::ColumnString::Offset & res_offset, const StringRef & repl, Int64 byte_pos)
+void OptimizedRegularExpressionImpl<thread_safe>::replaceAllImpl(const char * subject, size_t subject_size, DB::ColumnString::Chars_t & res_data, DB::ColumnString::Offset & res_offset, Int64 byte_pos, const Instructions & instructions)
 {
     size_t byte_offset = byte_pos - 1; // This is a offset for bytes, not utf8
     StringPieceType expr_sp(subject + byte_offset, subject_size - byte_offset);
     StringPieceType matched_str;
```
windtalker, line 616: Ditto

dbms/src/Common/OptimizedRegularExpression.inl.h
```diff
@@ -604,48 +609,55 @@ std::optional<StringRef> OptimizedRegularExpressionImpl<thread_safe>::substrImpl
...
+
+    // Copy characters that before position
+    res_data.resize(res_data.size() + byte_offset);
+    memcpy(&res_data[res_offset], subject, byte_offset);
+    res_offset += byte_offset;
 
     while (true)
     {
-        bool success = RegexType::FindAndConsume(&expr_sp, *re2, &matched_str);
+        bool success = re2->Match(expr_sp, start_pos, expr_len, re2_st::RE2::Anchor::UNANCHORED, matches, capture_num + 1);
```
windtalker, line 628: Since `matches` will be reused, I'm not sure if it need to be reset before `re2->Match`?

dbms/src/Functions/FunctionsRegexpCommon.h
```diff
@@ -102,19 +102,34 @@ inline int getDefaultFlags()
...
+template <bool need_subpattern = false>
 inline String addMatchTypeForPattern(const String & pattern, const String & match_type, TiDB::TiDBCollatorPtr collator)
 {
+    if (pattern.empty())
+        throw Exception("Length of the pattern argument must be greater than 0.");
+
     String mode = re2Util::getRE2ModeModifiers(match_type, collator);
+    String final_pattern;
+    if constexpr (need_subpattern)
+        final_pattern = fmt::format("({})", pattern);
```
windtalker, line 114: What happens if `pattern` itself is already inside parentheses? Like `pattern = (xxx)`

dbms/src/Functions/FunctionsRegexpReplace.h
```diff
@@ -220,7 +224,7 @@ class FunctionStringRegexpReplace : public FunctionStringRegexpBase
             std::unique_ptr<Regexps::Regexp> regexp;
             if (col_size > 0)
             {
-                regexp = memorize<true>(pat_param, match_type_param, collator);
+                regexp = memorize<false>(pat_param, match_type_param, collator, replace_default_flag);
```
windtalker, line 227: why change `need_subpattern` to false here?

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > I think we can rename all the `subpatterns` to `capture_groups`

okk

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > Looks like this is not used anymore?

I will delete it.

dbms/src/Common/OptimizedRegularExpression.inl.h
xzhangxian1008: > Since `matches` will be reused, I'm not sure if it need to be reset before `re2->Match`?

I think it's needless as the implementation in ch doesn't reset the `matches`

dbms/src/Functions/FunctionsRegexpCommon.h
xzhangxian1008: > What happens if `pattern` itself is already inside parentheses? Like `pattern = (xxx)`

When we use the `FindAndConsume` function in re2, we need this pattern, or there is no way to get the matched string.

dbms/src/Functions/FunctionsRegexpReplace.h
xzhangxian1008: > why change `need_subpattern` to false here?

At the beginning, we use `FindAndConsume` to implement the `regexp_replace`, true is needed to add pattern. And now, we implement the `regexp_replace` with `Match`, so the added pattern is no more need.

dbms/src/Common/OptimizedRegularExpression.h
```diff
@@ -95,14 +115,15 @@ class OptimizedRegularExpressionImpl
...
-        return match(subject, subject_size, matches, number_of_subpatterns + 1);
+        return match(subject, subject_size, matches, capture_num + 1);
     }
 
     bool match(const char * subject, size_t subject_size) const;
     bool match(const char * subject, size_t subject_size, Match & match) const;
     unsigned match(const char * subject, size_t subject_size, MatchVec & matches, unsigned limit) const;
 
-    unsigned getNumberOfSubpatterns() const { return number_of_subpatterns; }
+    unsigned getNumberOfSubpatterns() const { return capture_num; }
```
windtalker, line 125: Why still use `Subpatterns` here?

windtalker (approved):
Others LGTM

dbms/src/Common/OptimizedRegularExpression.h
xzhangxian1008: > Why still use `Subpatterns` here?

I will change it.

dbms/src/Common/OptimizedRegularExpression.h
```diff
@@ -58,6 +58,26 @@ struct Match
 };
 } // namespace OptimizedRegularExpressionDetails
 
+struct Instruction
+{
+    explicit Instruction(int substitution_num_)
+        : substitution_num(substitution_num_)
+    {}
+    explicit Instruction(String literal_)
```
yibin87, line 66: String && ？

yibin87 (approved):
Others LGTM

dbms/src/Common/OptimizedRegularExpression.h
xzhangxian1008: > String && ？

The value passed into this constructor will be repetedly used, I think this will break the `move`'s semantic.

PR #7291, opened 2023-04-13
refine code to avoid redundant calculation when join probe split happens

windtalker:

Issue Number: ref #6233

Problem Summary:
Currently, for each `Join::doJoinBlockHash`, it will do some preparations before call `JoinPartition::probeBlock`, if split happens, these preparations are redundant if the input block is unchanged.
Refine code to avoid redundant calculation when join probe split happens
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/JoinUtils.h
+++ b/dbms/src/Interpreters/JoinUtils.h
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include <Columns/ColumnNullable.h>
 #include <Core/Block.h>
 #include <Parsers/ASTTablesInSelectQuery.h>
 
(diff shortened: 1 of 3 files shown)


dbms/src/Interpreters/JoinUtils.h
```diff
@@ -71,13 +72,25 @@ struct ProbeProcessInfo
...
+    std::unique_ptr<IColumn::Offsets> offsets_to_replicate = nullptr;
+
     explicit ProbeProcessInfo(UInt64 max_block_size_)
         : max_block_size(max_block_size_)
         , min_result_block_size((max_block_size + 1) / 2)
         , all_rows_joined_finish(true){};
 
     void resetBlock(Block && block_, size_t partition_index_ = 0);
     void updateStartRow();
+    void prepareForProbe(const Names & key_names, const String & filter_column, ASTTableJoin::Kind kind, ASTTableJoin::Strictness strictness);
```
SeaRise, line 93: It looks like this method doesn't work for cross joins, does it?

windtalker: Currently, cross join does not split the block during probe stage, so it is fine for now. I'm working on split cross join probe, and will update this method for cross join in that pr.

SeaRise (approved):
lgtm

dbms/src/Interpreters/JoinUtils.h
SeaRise: ok

Issue #7306, opened 2023-04-18
TiFlash with s3 can't evict remote cache


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

insert some table and select count(*) using tiflash. deploy tiflash with s3 using minio
set read node remote cache 800G
### 2. What did you expect to see? (Required)
The remote cache can be evict and tiflash run stable
### 3. What did you see instead (Required)
The remote cache can't not be evicted
```
[2023/04/18 13:29:38.322 +08:00] [WARN] [PathCapacityMetrics.cpp:306] ["No available space for path: /data0/tiflash/cache, capacity: 800.00 GiB, used: 802.70 GiB"] [thread_id=2313]
[2023/04/18 13:29:48.323 +08:00] [WARN] [PathCapacityMetrics.cpp:306] ["No available space for path: /data0/tiflash/cache, capacity: 800.00 GiB, used: 802.70 GiB"] [thread_id=2313]
[2023/04/18 13:29:58.324 +08:00] [WARN] [PathCapacityMetrics.cpp:306] ["No available space for path: /data0/tiflash/cache, capacity: 800.00 GiB, used: 802.70 GiB"] [thread_id=2313]
[2023/04/18 13:30:08.325 +08:00] [WARN] [PathCapacityMetrics.cpp:306] ["No available space for path: /data0/tiflash/cache, capacity: 800.00 GiB, used: 802.70 GiB"] [thread_id=2313]
[2023/04/18 13:30:18.326 +08:00] [WARN] [PathCapacityMetrics.cpp:306] ["No available space for path: /data0/tiflash/cache, capacity: 800.00 GiB, used: 802.70 GiB"] [thread_id=2313]
```
TiFlash seems hang when the remote cache capacity is full
![image](https://user-images.githubusercontent.com/7677963/232691527-e72aa776-3bc7-4284-9d68-3850bc398dce.png)

### 4. What is your TiFlash version? (Required)
master

JaySon-Huang:
This is an expected behavior that TiFlash read node will try to maximize the use of disk capacity. When the disk capacity is no enough for downloading new files from S3, read node will evict some local cached file to download new files automatically 
https://github.com/pingcap/tiflash/blob/a3e48e4b734ac4d25e1642557208ab89fd294794/dbms/src/Storages/S3/FileCache.cpp#L245-L261

[...]

PR #7323, opened 2023-04-19
Refine remote cache capacity.

JinheLin:

Issue Number: ref #7306 

1. Seperating capacity metrics of page cache and dtfile cache in `PathCapacityMetrics`, so that we can easily to see which cache usage is out of limit.
2. Add reserved capacity for remote cache to avoid cache usage out of limit.
3. Add configuration `storage.remote.cache.reserved_rate`, default value is 0.1.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):
LGTM

dbms/src/Interpreters/Context.cpp
```diff
@@ -1465,8 +1465,8 @@ void Context::initializePathCapacityMetric( //
     const std::vector<size_t> & main_capacity_quota, //
     const Strings & latest_data_paths,
     const std::vector<size_t> & latest_capacity_quota,
-    const String & remote_cache_data_path,
-    size_t remote_cache_capacity)
+    const Strings & remote_cache_paths,
```
flowbehappy, line 1468: Instead of a vector, better use a struct which is more clear.

Lloyd-Pottiger (approved):

Issue #7290, opened 2023-04-13
TiFlash crash when dumping incremental checkpoint


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

deploy tiflash under disagg mode

### 3. What did you see instead (Required)
A write node crash with this stack
```
[2023/04/13 08:47:02.568 +00:00] [ERROR] [BaseDaemon.cpp:376] [########################################] [source=BaseDaemon] [thread_id=645]
[2023/04/13 08:47:02.568 +00:00] [ERROR] [BaseDaemon.cpp:377] ["(from thread 417) Received signal Segmentation fault(11)."] [source=BaseDaemon] [thread_id=645]
[2023/04/13 08:47:02.568 +00:00] [ERROR] [BaseDaemon.cpp:407] ["Address: 0x80695f720a00"] [source=BaseDaemon] [thread_id=645]
[2023/04/13 08:47:02.568 +00:00] [ERROR] [BaseDaemon.cpp:413] ["Access: read."] [source=BaseDaemon] [thread_id=645]
[2023/04/13 08:47:02.568 +00:00] [ERROR] [BaseDaemon.cpp:422] ["Address not mapped to object."] [source=BaseDaemon] [thread_id=645]
[2023/04/13 08:47:02.583 +00:00] [ERROR] [BaseDaemon.cpp:569] ["
       0x751ada1    faultSignalHandler(int, siginfo_t*, void*) [tiflash+122793377]
                    libs/libdaemon/src/BaseDaemon.cpp:220
  0x7f28199e7d90    <unknown symbol> [libc.so.6+347536]
       0x1b33ae3    DB::PS::V3::PageDirectory<DB::PS::V3::universal::PageDirectoryTrait>::dumpSnapshotToEdit(std::__1::shared_ptr<DB::PS::V3::PageDirectorySnapshot>) [tiflash+28523235]
                    dbms/src/Storages/Page/V3/PageDirectory.cpp:2066
       0x75b6b84    DB::UniversalPageStorage::dumpIncrementalCheckpoint(DB::UniversalPageStorage::DumpCheckpointOptions const&) [tiflash+123431812]
                    dbms/src/Storages/Page/V3/Universal/UniversalPageStorage.cpp:433
[...]

PR #7356, opened 2023-04-24
storage: Fix UAFs in PageStorage

breezewish:

Issue Number: close https://github.com/pingcap/tiflash/issues/7290

Problem Summary:

`iter`, `iter->first`, `iter->second` will be all invalidated when iter is erased:

```c++
int main() {
    std::map<int, int> a;
    a.emplace(1, 2);
    auto it = a.find(1);
    a.erase(1);
    std::cout << it->second << std::endl;
    return 0;
}

==36182==ERROR: AddressSanitizer: heap-use-after-free on address 0x000105600cb0 at pc 0x0001025d5dac bp 0x00016d82b2b0 sp 0x00016d82b2a8
READ of size 4 at 0x000105600cb0 thread T0
    #0 0x1025d5da8 in main main.cpp:22
    #1 0x102971084 in start+0x200 (dyld:arm64+0x5084) (BuildId: fbb89662e6f23434b542f75185ac5e7432000000200000000100000000030c00)
    #2 0xb308fffffffffffc  (<unknown module>)

```

Thus we must only access iter inside a lock.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/Page/V3/PageDirectory.cpp
+++ b/dbms/src/Storages/Page/V3/PageDirectory.cpp
@@ -1748,8 +1748,6 @@ PageDirectory<Trait>::getEntriesByBlobIds(const std::vector<BlobFileId> & blob_i
 
         while (true)
         {
-            // `iter` is an iter that won't be invalid cause by `apply`/`gcApply`.
-            // do scan on the version list without lock on `mvcc_table_directory`.
             fiu_do_on(FailPoints::pause_before_full_gc_prepare, {
                 if constexpr (std::is_same_v<Trait, u128::PageDirectoryTrait>)
                 {
--- a/dbms/src/Storages/Page/V3/PageDirectory.cpp
+++ b/dbms/src/Storages/Page/V3/PageDirectory.cpp
@@ -1888,18 +1888,18 @@ void PageDirectory<Trait>::copyCheckpointInfoFromEdit(PageEntriesEdit & edit)
 
         // TODO: Improve from O(nlogn) to O(n).
 
-        typename MVCCMapType::iterator iter;
+        VersionedPageEntriesPtr entries;
         {
             std::shared_lock read_lock(table_rw_mutex);
-            iter = mvcc_table_directory.find(rec.page_id);
+            auto iter = mvcc_table_directory.find(rec.page_id);
             if (iter == mvcc_table_directory.end())
                 // There may be obsolete entries deleted.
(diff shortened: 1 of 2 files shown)


breezewish:
Looks like we need to cherry pick these fixes for non-S3 versions as well, for example, when there is a GC, getByIDImpl may UAF. /cc @jayson-huang

lidezhu (approved):
LGTM

JaySon-Huang:
> Looks like we need to cherry pick these fixes for non-S3 versions as well, for example, when there is a GC, getByIDImpl may UAF. /cc @JaySon-Huang

In theory, the page_id in MVCCMapType is only erased by the GC process. And the page_id accessed by getByIDImpl should be protected by the related `snapshot`, so the GC process will not invalidate the iterator.
But sure it is safer to get the k,v under lock instead of an iterator because we may break the assumption `the page_id in MVCCMapType is only erased by the GC process` one day.

JaySon-Huang (approved):
LGTM

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -1731,20 +1734,22 @@ PageDirectory<Trait>::getEntriesByBlobIds(const std::vector<BlobFileId> & blob_i
...
             if (iter == mvcc_table_directory.end())
                 return {blob_versioned_entries, total_page_size};
+            page_id = iter->first;
+            version_entries = iter->second;
         }
 
         while (true)
         {
             // `iter` is an iter that won't be invalid cause by `apply`/`gcApply`.
             // do scan on the version list without lock on `mvcc_table_directory`.
```
JaySon-Huang, line 1752: remove these two lines

breezewish:
> > Looks like we need to cherry pick these fixes for non-S3 versions as well, for example, when there is a GC, getByIDImpl may UAF. /cc @JaySon-Huang
> 
> In theory, the page_id in MVCCMapType is only erased by the GC process. And the page_id accessed by getByIDImpl should be protected by the related `snapshot`, so the GC process will not invalidate the iterator. But sure it is safer to get the k,v under lock instead of an iterator because we may break the assumption `the page_id in MVCCMapType is only erased by the GC process` one day.

Good to know. Let's only cherry back for S3.

JaySon-Huang:
ref https://github.com/pingcap/tiflash/pull/7345 for backporting some fixes

Issue #7256, opened 2023-04-10
Refactor KVStore


CalvinNeo:
## Enhancement

This is the first storage of supporting raftstore v2. 
We will later make a KVStore v2.

PR #7402, opened 2023-04-27
Integrate some functions in KVStore and rename them to what they actually do

CalvinNeo:

Issue Number: ref #7256

Problem Summary:

Some methods are delegators of `writeBlockByRegion` which transforms row-format data into columnar-format data, and write them to cache of DeltaMergeStore.
There methods don't always do a flush. **Actually, they ALWAYS NOT DO a flush**, the only exception is when they are called from `tryFlushRegions`, and that is purely a scene of debugger.

1. flushRegion
2. tryPersist this is some un-locked version of persistRegion, we change its name to tryPersistRegion, and make it coworks with persistRegion.
4. tryFlushRegion
5. tryFlushRegions

So these names are quite misleading, I think we have to change them to what they actually do. 

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Debug/MockRaftStoreProxy.cpp
+++ b/dbms/src/Debug/MockRaftStoreProxy.cpp
@@ -547,7 +547,7 @@ void MockRaftStoreProxy::doApply(
         return;
 
     auto old_applied = kvs.getRegion(region_id)->appliedIndex();
-    auto old_applied_term = kvs.getRegion(region_id)->appliedIndexTerm();
+    auto old_applied_term = 0; //kvs.getRegion(region_id)->appliedIndexTerm();
     if (cmd.has_raw_write_request())
     {
         // TiFlash write
--- a/dbms/src/Debug/MockRaftStoreProxy.cpp
+++ b/dbms/src/Debug/MockRaftStoreProxy.cpp
@@ -547,7 +547,7 @@ void MockRaftStoreProxy::doApply(
         return;
 
     auto old_applied = kvs.getRegion(region_id)->appliedIndex();
-    auto old_applied_term = 0; //kvs.getRegion(region_id)->appliedIndexTerm();
+    auto old_applied_term = kvs.getRegion(region_id)->appliedIndexTerm();
     if (cmd.has_raw_write_request())
     {
         // TiFlash write
--- a/dbms/src/Debug/dbgFuncRegion.cpp
+++ b/dbms/src/Debug/dbgFuncRegion.cpp
@@ -116,7 +116,7 @@ void dbgFuncTryFlushRegion(Context & context, const ASTs & args, DBGInvoker::Pri
     auto region_id = static_cast<RegionID>(safeGet<UInt64>(typeid_cast<const ASTLiteral &>(*args[0]).value));
 
     TMTContext & tmt = context.getTMTContext();
-    tmt.getRegionTable().tryFlushRegion(region_id);
+    tmt.getRegionTable().tryWriteBlockByRegionAndFlush(region_id);
 
     output(fmt::format("region_table try flush region {}", region_id));
 }
--- a/dbms/src/Debug/dbgFuncRegion.cpp
+++ b/dbms/src/Debug/dbgFuncRegion.cpp
@@ -101,7 +101,7 @@ void dbgFuncPutRegion(Context & context, const ASTs & args, DBGInvoker::Printer 
 void dbgFuncTryFlush(Context & context, const ASTs &, DBGInvoker::Printer output)
 {
     TMTContext & tmt = context.getTMTContext();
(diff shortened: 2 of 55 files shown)


dbms/src/Debug/MockRaftStoreProxy.cpp
```diff
@@ -547,7 +547,7 @@ void MockRaftStoreProxy::doApply(
         return;
 
     auto old_applied = kvs.getRegion(region_id)->appliedIndex();
-    auto old_applied_term = kvs.getRegion(region_id)->appliedIndexTerm();
```
CalvinNeo, line 550: typo

dbms/src/Debug/dbgFuncRegion.cpp
```diff
@@ -101,7 +101,7 @@ void dbgFuncPutRegion(Context & context, const ASTs & args, DBGInvoker::Printer
 void dbgFuncTryFlush(Context & context, const ASTs &, DBGInvoker::Printer output)
 {
     TMTContext & tmt = context.getTMTContext();
-    tmt.getRegionTable().tryFlushRegions();
```
lidezhu, line 104: I think function name should describe what job it intend to do not how to do the job. So I prefer the previous one.
And this function is actually deprecated, so I think there is no need to change its name.

CalvinNeo: I don't know what job dbgFuncTryFlush intends to do. However, if dbgFuncTryFlush is deprecated, we can remove dbgFuncTryFlush and tryFlushRegions

lidezhu (approved):
LGTM

dbms/src/Storages/Transaction/RegionTable.h
```diff
@@ -211,7 +216,10 @@ class RegionTable : private boost::noncopyable
     InternalRegion & insertRegion(Table & table, const Region & region);
     InternalRegion & doGetInternalRegion(KeyspaceTableID ks_tb_id, RegionID region_id);
 
-    RegionDataReadInfoList flushRegion(const RegionPtrWithBlock & region, bool try_persist) const;
+    // Try write the committed kvs into cache of columnar DeltaMergeStore.
+    // Flush the cache if try_persist is set to true.
+    // The original name for this method is flushRegion.
```
JaySon-Huang, line 221:
```suggestion
```

dbms/src/Storages/Transaction/RegionTable.h
```diff
@@ -159,9 +159,15 @@ class RegionTable : private boost::noncopyable
...
 
+    // Find all regions with data, call writeBlockByRegionAndFlush with try_persist = true.
+    // This function is only for debug.
     bool tryFlushRegions();
-    RegionDataReadInfoList tryFlushRegion(RegionID region_id, bool try_persist = false);
-    RegionDataReadInfoList tryFlushRegion(const RegionPtrWithBlock & region, bool try_persist);
+
+    // Protects writeBlockByRegionAndFlush and ensures it's executed by only one thread at the smae time.
+    // Only one thread can do this at the same time.
+    // The original name for this function is tryFlushRegion.
```
JaySon-Huang, line 168:
```suggestion
```

dbms/src/Storages/Transaction/KVStore.cpp
```diff
@@ -333,12 +330,27 @@ void KVStore::setRegionCompactLogConfig(UInt64 sec, UInt64 rows, UInt64 bytes)
...
+        region_persister->persist(region);
+    }
+    if (region_task_lock.has_value())
+    {
+        LOG_DEBUG(log, "Persist {} done", region.toString(false));
+    }
+    else
+    {
+        LOG_INFO(log, "After persisted {}, cache {} bytes", region.toString(false), region.dataSize());
+    }
```
JaySon-Huang, line 353: the statements inside this if branch can be merged into the previous if branch?

dbms/src/Storages/Transaction/RegionTable.h
CalvinNeo: Why?

dbms/src/Storages/Transaction/KVStore.cpp
CalvinNeo: fixed

dbms/src/Storages/Transaction/RegionTable.h
JaySon-Huang: I think we don't need to keep this message... it doesn't help

JaySon-Huang (approved):
LGTM

PR #7418, opened 2023-05-05
Donot reserve memory for non-fixed type columns in decodeAndSquash

### LINKED ISSUE #7416: String column decode reserved much larger memory than needed
## Bug Report

When decode columns, we use IColumn's reserve(size_t rows) interface to avoid reallocating memories each time. However, for string columns, the element size is not fixed, and we used 64 bytes as default value. This will lead to much larger memory usage when actual element size is much smaller than 64 bytes.
This problem will be worsen for Join operator, since build side's all blocks will be in memory for a while.

yibin87:

Issue Number: close #7416 

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the query might use much larger memories than actual needed when Join build side data is very large and contains many small-size string type columns. 
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/CHBlockChunkCodecV1.cpp
+++ b/dbms/src/Flash/Coprocessor/CHBlockChunkCodecV1.cpp
@@ -15,6 +15,7 @@
 // limitations under the License.
 
 #include <DataTypes/DataTypeFactory.h>
+#include <DataTypes/DataTypeNullable.h>
 #include <Flash/Coprocessor/CHBlockChunkCodecV1.h>
 #include <IO/ReadBufferFromString.h>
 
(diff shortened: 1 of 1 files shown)


dbms/src/Flash/Coprocessor/CHBlockChunkCodecV1.cpp
```diff
@@ -96,12 +97,20 @@ static inline void decodeColumnsByBlock(ReadBuffer & istr, Block & res, size_t r
...
-        else
-            column->reserve(rows_to_read + column->size());
+    const auto & name_and_type_list = res.getColumnsWithTypeAndName();
+    size_t column_size = mutable_columns.size();
+    for (size_t i = 0; i < column_size; ++i)
+    {
+        auto && column = mutable_columns[i];
+        /// For non-fixed size type, reserve function might cause too much memory usage, i.e. string type column reserves 64 bytes
+        /// size for each element.
+        if (removeNullable(name_and_type_list[i].type)->isValueUnambiguouslyRepresentedInFixedSizeContiguousMemoryRegion())
```
windtalker, line 107: maybe add a new function? `isValueUnambiguouslyRepresentedInFixedSizeContiguousMemoryRegion` return `false` for `decimal64` based type

dbms/src/Flash/Coprocessor/CHBlockChunkCodecV1.cpp
```diff
@@ -96,12 +97,20 @@ static inline void decodeColumnsByBlock(ReadBuffer & istr, Block & res, size_t r
...
-        if (reserve_size > 0)
-            column->reserve(std::max(rows_to_read, reserve_size));
-        else
-            column->reserve(rows_to_read + column->size());
+    const auto & name_and_type_list = res.getColumnsWithTypeAndName();
+    size_t column_size = mutable_columns.size();
+    for (size_t i = 0; i < column_size; ++i)
+    {
+        auto && column = mutable_columns[i];
+        /// For non-fixed size type, reserve function might cause too much memory usage, i.e. string type column reserves 64 bytes
```
solotzg, line 105: Maybe we can change the codec format and add field about data size for the columns with non-fixed size type. But such way need to change codec & mpp version.

yibin87: Yes, nice suggestion, I plan to introduce some mechanism to "guess" the every data size, since we need to reserve the column size before knowning how many blocks will be squashed.

solotzg (approved):

windtalker (approved):
LGTM

Issue #7400, opened 2023-04-27
TiFlash with s3 crash in S3GCManager


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

ch worload and tiflash disaggreted with s3 using minio
scale a write node and tiflash crash
```
[2023/04/27 14:58:34.506 +08:00] [ERROR] [BaseDaemon.cpp:368] ["(from thread 18) Terminate called without an active exception\n"] [source=BaseDaemon] [thread_id=392]
[2023/04/27 14:58:34.506 +08:00] [ERROR] [BaseDaemon.cpp:376] [########################################] [source=BaseDaemon] [thread_id=392]
[2023/04/27 14:58:34.506 +08:00] [ERROR] [BaseDaemon.cpp:377] ["(from thread 18) Received signal Aborted(6)."] [source=BaseDaemon] [thread_id=392]
[...]

PR #7445, opened 2023-05-09
Restore checkpoint location after restart

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/7400

Problem Summary:

The checkpoint info could possible lag behind the S3 manifest after restart. If we only rely on the local persisted WAL to restore the checkpoint info into memory, it will cause data inconsistency problem that a tiflash wn referring a deleted S3 CheckpointDataFile.

* For a bootstrapped write node, wait for fetching checkpoint info from S3 before receiving GRPC requests
* After restarting, try to fetch the checkpoint info from the latest manifest on S3, for each page_id in the manifest:
  * If a local entry has the same sequence as in the S3 manifest, then copy the checkpoint info
  * If a local entry has newer sequence to the S3 manifest, then the page is updated, just ignore
  * If the page_id does not exist in local after restart, it means the page_id is deleted, just ignore
* Add an HTTP API "http://127.0.0.1:20170/tiflash/sync-remote-store" to re-sync all local page data to S3

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
  - Deploy a disaggregated cluster with 1 wn, 1 cn
  - Enable `profiles.default.remote_gc_verify_consistency: 1`
  - Run chbenchmark and check whether all check passed
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lidezhu (approved):
LGTM

PR #7489, opened 2023-05-17
Print MPPTask info when it's alive for a long time

### LINKED ISSUE #7351: Add a way to detect and log suspicious hang query/task in TiFlash compute layer
## Enhancement
Sometimes a query/task may hang due to some implementation bugs, we can observe this hang issue in metrics like a long non-zero `MPPTaskCount` when there is no active query, but the information in metrics is not enough, since we don't which query/task is hang, we need add a way to detect and log suspicious hang query/task in TiFlash compute layer.

An possible way to do this is adding a periodic detection thread to check and print the suspicious hang query/task.

### LINKED ISSUE #7488: It's hard to find hung MPPTask

xzhangxian1008:

- Print MPPTask info when it's alive for a long time with a background thread

Issue Number: close #7488 close #7351

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/MPPTask.h
```diff
@@ -47,6 +47,16 @@ enum class AbortType
     ONERROR,
 };
 
+// This struct only print logs when MPPTask is destructed
+struct MPPTaskDestructedHinter
+{
+    explicit MPPTaskDestructedHinter(const LoggerPtr log_)
+        : log(log_)
+    {}
+    ~MPPTaskDestructedHinter() { LOG_INFO(log, "MPPTask is completely destructed"); }
```
SeaRise, line 56: Maybe `LOG_DEBUG`?

SeaRise (approved):
Others LGTM

dbms/src/Flash/Mpp/MPPTask.h
xzhangxian1008: > Maybe `LOG_DEBUG`?

okk

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -309,4 +348,51 @@ void MPPTaskManager::releaseThreadsFromScheduler(const int needed_threads)
...
+    monitored_tasks.insert(std::make_pair(task_unique_id, Stopwatch()));
+}
+
+void MPPTaskManager::removeMonitoredTask(const String & task_unique_id)
+{
+    std::lock_guard lock(mu);
+    auto iter = monitored_tasks.find(task_unique_id);
+    if (iter == monitored_tasks.end())
+    {
+        LOG_ERROR(log, "Unexpected behavior! task {} is not found in monitored_task.");
```
yibin87, line 391: Do we consider change to Info level? It would allow removeMonitorerTask been called multiple times for the same task.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -31,11 +34,47 @@ extern const char random_task_manager_find_task_failure_failpoint[];
...
+    String log_info;
+    for (const auto & iter : monitored_tasks)
+    {
+        auto alive_time = iter.second.elapsedSeconds();
+        if (alive_time >= 300)
+            log_info = fmt::format("{} <MPPTask is alive for {} secs, {}>", log_info, alive_time, iter.first);
+    }
+
+    if (!log_info.empty())
+        LOG_INFO(log, log_info);
```
yibin87, line 51: Consider just collect overtimed task info, and do format and logging work outside the lock protection?

xzhangxian1008: > Consider just collect overtimed task info, and do format and logging work outside the lock protection?

okk, I will log info outside of the lock

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -31,11 +34,46 @@ extern const char random_task_manager_find_task_failure_failpoint[];
...
+{
+// Print log for MPPTask which hasn't been removed for over 5 minutes.
+String getLongLiveMPPTasks(const std::unordered_map<String, Stopwatch> & monitored_tasks)
+{
+    String log_info;
+    for (const auto & iter : monitored_tasks)
+    {
+        auto alive_time = iter.second.elapsedSeconds();
+        if (alive_time >= 300)
+            log_info = fmt::format("{} <MPPTask is alive for {} secs, {}>", log_info, alive_time, iter.first);
```
yibin87, line 47: how about move a step forward, just return vector<pair<string, time>>, and do formating outside too?

dbms/src/Flash/Mpp/MPPTask.h
```diff
@@ -113,10 +129,14 @@ class MPPTask : public std::enable_shared_from_this<MPPTask>
...
 
 private:
+    // We must ensure this member variable is put at this place to be destructed at proper time
+    MPPTaskMonitorHelper mpp_task_monitor_helper;
+
     // To make sure dag_req is not destroyed before the mpp task ends.
     tipb::DAGRequest dag_req;
     mpp::TaskMeta meta;
     MPPTaskId id;
+    const LoggerPtr log;
```
windtalker, line 139: Why change this

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -31,11 +34,46 @@ extern const char random_task_manager_find_task_failure_failpoint[];
 extern const char pause_before_register_non_root_mpp_task[];
 } // namespace FailPoints
 
+namespace
+{
+// Print log for MPPTask which hasn't been removed for over 5 minutes.
```
windtalker, line 39: 5 minutes is too short, I think it should be at least 30 minutes.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
xzhangxian1008: > how about move a step forward, just return vector<pair<string, time>>, and do formating outside too?

okk

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -31,11 +34,46 @@ extern const char random_task_manager_find_task_failure_failpoint[];
...
     : scheduler(std::move(scheduler_))
     , log(Logger::get())
+    , is_shutdown(false)
 {}
 
+MPPTaskManager::~MPPTaskManager()
+{
+    std::lock_guard lock(mu);
+    is_shutdown = true;
+    monitor_cv.notify_all();
```
windtalker, line 64: I think this does not work because `MonitorMPPTaskThread` holds a shared ptr of `MPPTaskManager`, the destruct will never be called.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -31,11 +34,46 @@ extern const char random_task_manager_find_task_failure_failpoint[];
...
+}
+
+void MPPTaskManager::startMonitorMPPTaskThread()
```
windtalker, line 67: I suggest to put this monitor thread to `BackgroundSerive` in TMTContext, and add some sync method to make sure the monitor thread always visit a valid mpp task manager.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -309,4 +347,58 @@ void MPPTaskManager::releaseThreadsFromScheduler(const int needed_threads)
     scheduler->releaseThreadsThenSchedule(needed_threads, *this);
 }
 
+void MPPTaskManager::monitorMPPTasks()
+{
+    std::unique_lock lock(mu, std::defer_lock);
+    while (true)
+    {
+        lock.lock();
```
windtalker, line 355: I think just hold the lock and get a snapshot of current monitored_tasks is enough. We can do the monitor things based on the snapshot.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -309,4 +347,58 @@ void MPPTaskManager::releaseThreadsFromScheduler(const int needed_threads)
...
+    }
+}
+
+void MPPTaskManager::addMonitoredTask(const String & task_unique_id)
+{
+    std::lock_guard lock(mu);
+    auto iter = monitored_tasks.find(task_unique_id);
+    if (iter != monitored_tasks.end())
+    {
+        LOG_INFO(log, "task {} is repeatedly added to be monitored which is not an expected behavior!");
```
windtalker, line 384: Maybe use warning log?

dbms/src/Flash/Mpp/MPPTaskManager.cpp
```diff
@@ -309,4 +347,58 @@ void MPPTaskManager::releaseThreadsFromScheduler(const int needed_threads)
...
+    monitored_tasks.insert(std::make_pair(task_unique_id, Stopwatch()));
+}
+
+void MPPTaskManager::removeMonitoredTask(const String & task_unique_id)
+{
+    std::lock_guard lock(mu);
+    auto iter = monitored_tasks.find(task_unique_id);
+    if (iter == monitored_tasks.end())
+    {
+        LOG_INFO(log, "Unexpected behavior! task {} is not found in monitored_task.");
```
windtalker, line 397: Ditto

dbms/src/Flash/Mpp/MPPTask.h
xzhangxian1008: > Why change this

This change is affected by the previous aborted implement, I will move it back.

dbms/src/Flash/Mpp/MPPTaskManager.cpp
xzhangxian1008: > 5 minutes is too short, I think it should be at least 30 minutes.

okk

yibin87 (approved):
LGTM

PR #7493, opened 2023-05-18
storage: Add AddExtraTableIDColumn action and stream

breezewish:

Problem Summary:

This PR is extracted from https://github.com/pingcap/tiflash/pull/7457

**Change1:**

Previously AddExtraTableIDColumn is a single action. Now we also provide a AddExtraTableIDColumnInputStream.

**Change2:**

Previously we perform AddExtraTableIDColumn **inside** UnorderedInputStream and DMSegmentThreadInputStream. Now:

- When ReadThread is not enabled: Apply AddExtraTableIDColumnInputStream after DMSegmentThreadInputStream

- When ReadThread is enabled: Apply AddExtraTableIDColumnInputStream after SegmentInputStream. This means, UnorderedInputStream will receive blocks with extra table ID column and do nothing but just throwing them out.

This change helps us to support **single UnorderedInputStream** over **multiple partitions**, which is a pattern used in disaggregated mode. This pattern is helpful to balance reads from multiple partitions.

TODO: DMSegmentSourceOp is not changed. It still perform AddExtraTableIDColumn **inside** the operator. After https://github.com/pingcap/tiflash/pull/7480 is merged we could provide AddExtraTableIDColumnOp.

### Check List

Tests 

- [x] Unit test - Existing tests
- [x] Integration test - Existing tests

Executed the following SQL when read_thread is on (default) and off:

```sql
create table t1(id int, age int, key(id)) partition by range(id) (
  partition p0 values less than (100),
  partition p1 values less than (200),
  partition p2 values less than (300),
  partition p3 values less than (400));

insert into t1 values (1, 10);
insert into t1 values (10, 10);

alter table t1 set tiflash replica 1;

set @@session.tidb_isolation_read_engines='tiflash';
set @@session.tidb_partition_prune_mode = 'dynamic';

analyze table t1;

select * from information_schema.tiflash_replica;

explain select count(*) from t1 where id < 150;

select count(*) from t1 where id < 150;
```

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


SeaRise (approved):
lgtm

Lloyd-Pottiger:
How about applying AddExtraTableIDColumnInputStream after `StorageDeltaMerge::read`? 

in `DAGStorageInterpreter::executeImpl` just like `executeGeneratedColumnPlaceholder`

breezewish:
> How about applying AddExtraTableIDColumnInputStream after `StorageDeltaMerge::read`?
> 
> in `DAGStorageInterpreter::executeImpl` just like `executeGeneratedColumnPlaceholder`

@Lloyd-Pottiger Maybe not a good idea, because `StorageDisaggregated::read` will accept multiple partitions and produce a stream for these partition. In this case, we must ensure the table id column exists to distinguish different partitions. To unify the behavior for disaggregated mode and non-disaggregated mode, we will produce the table id in the ReadThread pool.

Lloyd-Pottiger (approved):

PR #7509, opened 2023-05-19
*: refine appendExtraCastsAfterTS

Lloyd-Pottiger:

Issue Number: ref #6233 

Problem Summary:

The logic of appendExtraCastsAfterTS is kind of mess now, which means it is hard to understand.

1. remove `ExtraCastAfterTSMode`
2. use a `std::vector<bool>` to record the columns may need to add a cast.
3. judge which really need to add cast in `DAGExpressionAnalyzer::buildExtraCastsAfterTS` uniformly

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
+++ b/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
@@ -180,7 +180,7 @@ bool hasRegionToRead(const DAGContext & dag_context, const TiDBTableScan & table
 // <has_cast, extra_cast, project_for_remote_read>
 std::pair<bool, ExpressionActionsPtr> addExtraCastsAfterTs(
     DAGExpressionAnalyzer & analyzer,
-    const std::vector<bool> & may_need_add_cast_column,
+    const std::vector<UInt8> & may_need_add_cast_column,
     const TiDBTableScan & table_scan)
 {
     // if no column need to add cast, return directly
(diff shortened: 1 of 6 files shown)


SeaRise (approved):
good job! LGTM

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -180,18 +180,16 @@ bool hasRegionToRead(const DAGContext & dag_context, const TiDBTableScan & table
 // <has_cast, extra_cast, project_for_remote_read>
 std::pair<bool, ExpressionActionsPtr> addExtraCastsAfterTs(
     DAGExpressionAnalyzer & analyzer,
-    const std::vector<ExtraCastAfterTSMode> & need_cast_column,
+    const std::vector<bool> & may_need_add_cast_column,
```
JaySon-Huang, line 183: `vector<bool>` is a special specialization for `vector`, it could brings unexpected behavior sometimes. Maybe use `vector<UInt8>` instead?

Lloyd-Pottiger: The special case is
```
# vec[i] = true;
auto b = vec[i];
b = false.
# vec[i] become false now.
```

We will not do this way, so it is safe.

ywqzzy (approved):

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
JaySon-Huang: We won't do it now and we know this problem. But it could bring trouble in further code changes who don't familiar with `vector<bool>`.
Compare to `vector<UInt8>`, I think `vector<bool>` here does not bring any special benefit.

JaySon-Huang (approved):
LGTM

Issue #7539, opened 2023-05-24
A more reasonable capacity metrics reported to PD under disagg arch


JaySon-Huang:
## Enhancement

Under S3 disagg arch. Now each TiFlash write node will report these messages to PD:

* capacity size: always 1 PiB
* used size: total used size of local disk and owned size of S3 data of this write node

Now we used 1 PiB as capacity size to avoid PD from not adding peer to a specified tiflash write node when its S3 data is more than the local disk capacity. But the "1 PiB" could bring unexpected effect on the whole cluster capacity.

JaySon-Huang:
Change it to a moderate bug to backport it to release-7.1

PR #7544, opened 2023-05-24
Report local disk usage to PD

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/7539

Problem Summary:
Now we used 1 PiB as capacity size to avoid PD from not adding peer to a specified tiflash write node when its S3 data is more than the local disk capacity. But the "1 PiB" could bring unexpected effect on the whole cluster capacity.

* Report the local disk used size/capacity size/available to PD, excluding the data size stored on S3
  * The data size stored on S3 only shown at Grafana as "remote size", we have no way to report this size to PD. And also the PD does not take it into consideration
* Show the Region count of each TiFlash instance in TiFlash-summary panel
* Fix the duplicate panel id 166 between 'Data size in send and receive queue' and 'Stale Read OPS' that causes display error

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/PathCapacityMetrics.cpp
+++ b/dbms/src/Storages/PathCapacityMetrics.cpp
@@ -262,15 +262,9 @@ FsStats PathCapacityMetrics::getFsStats(bool finalize_capacity)
             for (const auto & [keyspace_id, used_bytes] : keyspace_id_to_used_bytes)
             {
                 UNUSED(keyspace_id);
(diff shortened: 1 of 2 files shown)


lidezhu (approved):
LGTM

Issue #7531, opened 2023-05-23
The buffer in executor should support memory bound


windtalker:
## Enhancement
When running MPP query, some non-memory intensive executors will still buffer data(like `ExchangeReceiver/UnionBlockInputStream/SharedQueryBlockInputStream/ExchangeSender`). Currently the buffer inside these executors are usually size bounded, since the memory usage varies a lot between different elements in buffer, it will be more memory friendly if the buffer inside the executors are memory bounded.

PR #7554, opened 2023-05-26
Refine ExchangeReceiver's receiver queue for fine grained shuffle

windtalker:

Issue Number: ref #7531

Problem Summary:

This pr refines the exchange receiver's queue for fine grained shuffle.

For fine grained shuffle, one received packet contains multiple data that aims at different receiver streams, in current implementation, assuming there is N receiver streams, ExchangeReceiver will maintain N receiver queues for fine grained shuffle. For each received packet, it need to be written to all the N queues(except that the packet does not contain any data for i-th receiver stream, which is a very rare situation). All the N receiver queues are size bounded, that is, a received packet's written could be blocked by any one of the receiver queues. This differs a lot from non fine grainded shuffle's receiver: in non fine grained shuffle's receiver, there is only one receiver queue, a received packet is either blocked by that queue or written into that queue. Obviously, this difference between fine graind and non fine grained makes the ExchangeReceiver's code complex and hard to maintain.

This pr tries to unify the code between fine grained and non fine grained receiver queue, the basic idea is
1. There is only one blocking queue in ExchangeReciver
2. For fine grained shuffle, besides the blocking queue, there are N non-blocking queues

For non fine grained shuffle:
- all the read/write to ExchangeReciver is based on the blocking queue directly.

For fine grained shuffle:
- write is divided into 2 stages
  - stage 1: the writer writes the received packet to the blocking queue
  - stage 2: if stage 1 success, write received packet to N's non blocking queues, since write to non blocking queue never blocks, so the write could only be blocked by stage 1, which is the same as non fine grained shuffle
- read is also divided into 2 stage
  - stage 1: each reader reads packet from its own received queue(one of the non blocking queue), and update `remaining_consumer` in the packet by 1
  - stage 2: if `remaining_consumer` is 0, then pop the packet from the blocking queue

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,160 @@
...
+        RUNTIME_ASSERT(false, "Illegal ReceiverMode");
+    }
+}
+} // namespace
+
+template <bool fine_grained_shuffle, bool need_wait>
+std::pair<MPMCQueueResult, ReceivedMessagePtr> ReceivedMessageQueue::pop(size_t stream_id)
+{
+    MPMCQueueResult res;
+    ReceivedMessagePtr recv_msg;
```
SeaRise, line 51: It seems unnecessary to use `shared_ptr`, how about using `ReceivedMessage` directly?

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,160 @@
...
+        {
+            if (recv_msg->remaining_consumers->fetch_sub(1) == 1)
+            {
+                ReceivedMessagePtr original_msg;
+                auto pop_result [[maybe_unused]] = grpc_recv_queue->tryPop(original_msg);
+                /// if there is no remaining consumer, then pop it from original queue, the message must stay in the queue before the pop
+                /// so even use tryPop, the result must not be empty
+                assert(pop_result != MPMCQueueResult::EMPTY);
+                if (original_msg != nullptr)
+                    RUNTIME_CHECK_MSG(*original_msg->remaining_consumers == 0, "Fine grained receiver pop a message that is not full consumed, remaining consumer: {}", *original_msg->remaining_consumers);
```
SeaRise, line 72:
```suggestion
                if (original_msg != nullptr)
                    RUNTIME_CHECK_MSG(original_msg != nullptr && *original_msg->remaining_consumers == 0, "Fine grained receiver pop a message that is not full consumed, remaining consumer: {}", *original_msg->remaining_consumers);
```

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,160 @@
...
+    std::function<MPMCQueueResult(ReceivedMessagePtr &)> write_func;
+    if constexpr (is_force)
+        write_func = [&](ReceivedMessagePtr & recv_msg) {
+            return msg_channel->forcePush(recv_msg);
+        };
+    else
+        write_func = [&](ReceivedMessagePtr & recv_msg) {
+            return msg_channel->push(recv_msg);
+        };
+    bool success = write_func(received_message) == MPMCQueueResult::OK;
```
SeaRise, line 102:
```suggestion
    bool success = false;
    if constexpr (is_force)
        success = msg_channel->forcePush(recv_msg) == MPMCQueueResult::OK;
    else
        success = msg_channel->push(recv_msg) == MPMCQueueResult::OK;
```

dbms/src/Common/LooseBoundedMPMCQueue.h
```diff
@@ -230,11 +235,16 @@ class LooseBoundedMPMCQueue
...
-    ALWAYS_INLINE void pushFront(U && data)
+    ALWAYS_INLINE MPMCQueueResult pushFront(U && data)
     {
         Int64 memory_usage = get_auxiliary_memory_usage(data);
         queue.emplace_front(std::forward<U>(data), memory_usage);
         current_auxiliary_memory_usage += memory_usage;
+        if (push_callback)
+        {
+            if (!push_callback(queue.front().data))
+                return MPMCQueueResult::CANCELLED;
```
SeaRise, line 246: Is there a need to call `cancel()` here?

dbms/src/Common/LooseBoundedMPMCQueue.h
```diff
@@ -230,11 +235,16 @@ class LooseBoundedMPMCQueue
...
     template <typename U>
-    ALWAYS_INLINE void pushFront(U && data)
+    ALWAYS_INLINE MPMCQueueResult pushFront(U && data)
     {
         Int64 memory_usage = get_auxiliary_memory_usage(data);
         queue.emplace_front(std::forward<U>(data), memory_usage);
         current_auxiliary_memory_usage += memory_usage;
+        if (push_callback)
+        {
+            if (!push_callback(queue.front().data))
```
SeaRise, line 245:
```suggestion
            if unlikely (!push_callback(queue.front().data))
```

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: Because the queue stores the shared_ptr, if we don't hold the ptr, it will be released.

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    injectFailPointReceiverPushFail(success, mode);
+    return success;
+}
+
+template <bool enable_fine_grained_shuffle>
```
xzhangxian1008, line 108: Why we need `enable_fine_grained_shuffle` here? It seems that it's not used.

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+        else
+        {
+            res = grpc_recv_queue->tryPop(recv_msg);
+        }
+    }
+    return {res, recv_msg};
+}
+
+template <bool enable_fine_grained_shuffle, bool is_force>
```
xzhangxian1008, line 90: ditto

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: The result could be `Cancelled`, so `original_msg ` is not always not null

SeaRise: my mistake...
```suggestion
                RUNTIME_CHECK_MSG(original_msg == nullptr || *original_msg->remaining_consumers == 0, "Fine grained receiver pop a message that is not full consumed, remaining consumer: {}", *original_msg->remaining_consumers);
```

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
SeaRise: ok, maybe `unique_ptr` can be used here?

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: Yes, you are right, after move fine grained write to LooseBoundedQueue, `enable_fine_grained_shuffle ` is not needed.

dbms/src/Common/LooseBoundedMPMCQueue.h
windtalker: Done

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: But it need to be shared among the read streams for fine grained shuffle.

SeaRise: ok

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,150 @@
...
+        for (size_t i = 0; i < fine_grained_channel_size; ++i)
+            /// these are unbounded queues
+            msg_channels_for_fine_grained_shuffle.push_back(std::make_shared<LooseBoundedMPMCQueue<ReceivedMessagePtr>>(std::numeric_limits<size_t>::max()));
+        /// use pushcallback to make sure that the order of messages in msg_channels_for_fine_grained_shuffle is exactly the same as it in msg_channel,
+        /// because pop from msg_channel rely on this assumption. An alternative is to make msg_channel a set/map of messages for fine grained shuffle, but
+        /// it need many more changes
+        msg_channel = std::make_shared<LooseBoundedMPMCQueue<ReceivedMessagePtr>>(max_buffer_size, [this](const ReceivedMessagePtr & element) {
+            for (size_t i = 0; i < fine_grained_channel_size; ++i)
+            {
+                if (msg_channels_for_fine_grained_shuffle[i]->tryPush(element) != MPMCQueueResult::OK)
```
SeaRise, line 130: use `forcePush`?

dbms/src/Common/LooseBoundedMPMCQueue.h
```diff
@@ -230,11 +241,22 @@ class LooseBoundedMPMCQueue
...
     template <typename U>
-    ALWAYS_INLINE void pushFront(U && data)
+    ALWAYS_INLINE MPMCQueueResult pushFront(U && data)
     {
         Int64 memory_usage = get_auxiliary_memory_usage(data);
         queue.emplace_front(std::forward<U>(data), memory_usage);
         current_auxiliary_memory_usage += memory_usage;
+        if (push_callback)
+        {
+            if unlikely (!push_callback(queue.front().data))
```
SeaRise, line 251: If `push_callback` returns false because `msg_channels_for_fine_grained_shuffle` was called `finish()`, it should not be possible to call `cancel` here?
Is it necessary for A to return the `MPMCQueueResult`?

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,150 @@
...
+                    return false;
+            }
+            return true;
+        });
```
SeaRise, line 134: why not moving the `push_callback` to `pushToMessageChannel`?
Adding a `push_callback` to `LooseBoundMPMCQueue` is a bit weird

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: Since the queue is unbounded, I think both `forcePush` and `tryPush` is ok.

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+template <bool is_force>
+bool ReceivedMessageQueue::pushToMessageChannel(ReceivedMessagePtr & received_message, ReceiverMode mode)
+{
+    bool success = false;
+    if constexpr (is_force)
+        success = msg_channel->forcePush(received_message) == MPMCQueueResult::OK;
+    else
+        success = msg_channel->push(received_message) == MPMCQueueResult::OK;
```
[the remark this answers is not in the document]
windtalker: I think `forcePush` is used in pipeline model.

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
windtalker: Moving `push_callback` to `pushToMessageChannel` is not enough under currently implemantion. The main idea of current implementation is to use `msg_channel` as a bounded queue, it only handle block `push`, the consumer streams does not `pop` messages from `msg_channel` directly, instead, they `pop` messages from `fine_grained_msg_channels`,  and only if a message is poped by all the consumers(remaining_consumers == 0), it will be poped from `msg_channel`. Since the queue is a FIFO queue, there is actually no way to pop a specific message from a queue, the hack thing is we keep the order of messages the same in all the queues, then we can ensure that if a message has no remaining consumers, it must be the earliest message in `msg_channel`, so it can be safely poped from `msg_channel`. If the order of all the messages is not guaranteed, `msg_channel` need to be a message set instead of a message queue, and every message need a unique key, once a message is consumed by all the consumers, it can be moved from the message set by its unique key.

dbms/src/Common/LooseBoundedMPMCQueue.h
windtalker: I think just return `MPMCQueueStatus::CANCELLED` and not change the queue status is ok? Because the caller should check the push result, and do something if the result is not `MPMCQueueStatus::OK`?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -945,13 +911,24 @@ ExchangeReceiverResult ExchangeReceiverBase<RPCContext>::handleUnnormalChannel(
...
+    bool handle_resp = false;
+    if (recv_msg->remaining_consumers != nullptr)
+    {
+        /// fine grained shuffle
+        handle_resp = stream_id == 0 && recv_msg->resp_ptr != nullptr;
+    }
+    else
+    {
+        handle_resp = recv_msg->resp_ptr != nullptr;
+    }
```
SeaRise, line 930: Looks like it can be wrapped into a function.
```
ReceivedMessage::respPtr(size_t stream_id) { return stream_id  == 0 ? resp_ptr : nullptr;  }
```
Because the stream_id of non-fine-grained shuffle must be 0, so there is no need to judge remaining_consumers here?

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
```diff
@@ -772,14 +735,16 @@ void ExchangeReceiverBase<RPCContext>::readLoop(const Request & req)
...
+    size_t stream_id,
+    const ReceivedMessagePtr & recv_msg,
     std::queue<Block> & block_queue,
     std::unique_ptr<CHBlockChunkDecodeAndSquash> & decoder_ptr)
 {
     assert(recv_msg != nullptr);
     DecodeDetail detail;
 
-    if (recv_msg->chunks.empty())
+    auto & chunks = recv_msg->remaining_consumers == nullptr ? recv_msg->chunks : recv_msg->fine_grained_chunks[stream_id];
```
SeaRise, line 746: Looks like it can be wrapped into a function.

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
```diff
@@ -0,0 +1,148 @@
...
+        else
+        {
+            res = msg_channels_for_fine_grained_shuffle[stream_id]->tryPop(recv_msg);
+        }
+        if (recv_msg != nullptr)
+        {
+            if (recv_msg->remaining_consumers->fetch_sub(1) == 1)
+            {
+                ReceivedMessagePtr original_msg;
+                auto pop_result [[maybe_unused]] = grpc_recv_queue->tryPop(original_msg);
```
SeaRise, line 67: So if the consumer of the fine grained queue has data skew, grpc_recv_queue will never be popped, and grpc_recv_queue will still be blocked, right?

dbms/src/Flash/Mpp/ReceivedMessageQueue.h
```diff
@@ -0,0 +1,127 @@
...
+    size_t source_index;
+    String req_info;
+    // shared_ptr<const MPPDataPacket> is copied to make sure error_ptr, resp_ptr and chunks are valid.
+    const std::shared_ptr<DB::TrackedMppDataPacket> packet;
+    const mpp::Error * error_ptr;
+    const String * resp_ptr;
+    std::vector<const String *> chunks;
+    /// used for fine grained shuffle
+    std::vector<std::vector<const String *>> fine_grained_chunks;
+    std::shared_ptr<std::atomic<size_t>> remaining_consumers;
```
SeaRise, line 44: Add some comments here?
When non-fine-grained shuffle, remaining_consumers is null ...

dbms/src/Flash/Mpp/ReceiverChannelBase.cpp
```diff
@@ -19,47 +19,66 @@
...
+            // indicating which stream_id this chunk belongs to.
+            RUNTIME_CHECK_MSG(packet.chunks_size() == packet.stream_ids_size(), "Packet's chunk size({}) not equal to its size of streams({})", packet.chunks_size(), packet.stream_ids_size());
+
+            for (int i = 0; i < packet.stream_ids_size(); ++i)
+            {
+                UInt64 stream_id = packet.stream_ids(i) % fine_grained_consumer_size;
+                ret->fine_grained_chunks[stream_id].push_back(&packet.chunks(i));
+            }
+        }
+    }
```
SeaRise, line 81: Looks like it could be wrapped in a ReceivedMessage

dbms/src/Flash/Mpp/ReceivedMessageQueue.cpp
SeaRise: ok

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
windtalker: Ok, I've made `ReceivedMessage` as a class, and added some member functions.

dbms/src/Flash/Mpp/ExchangeReceiver.cpp
windtalker: Done

dbms/src/Flash/Mpp/ReceivedMessageQueue.h
windtalker: Done

dbms/src/Flash/Mpp/ReceiverChannelBase.cpp
windtalker: Done

SeaRise (approved):
lgtm

xzhangxian1008 (approved):
LGTM

Issue #7587, opened 2023-06-02
TiFlash read snapshot can't be released for long time in the cancel mpp scenario


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Approximately 10 concurrent executions of the "SELECT COUNT(*) FROM t" statement were used to access tiflash. When the query reported an error, the next round of queries immediately commenced. When injecting tiflash restart chaos to the cluster, the mpp query count increased suddenly, and the tiflash snapshot was not released for a long time.
![image](https://github.com/pingcap/tiflash/assets/7677963/fa62177d-cec8-45c1-997b-2c48c8a39343)
![image](https://github.com/pingcap/tiflash/assets/7677963/311744f0-d98d-4d47-bb72-cd53edbbefbc)
![image](https://github.com/pingcap/tiflash/assets/7677963/973a7ec0-6236-473a-ab3f-f0b57aa079a4)

### 4. What is your TiFlash version? (Required)
master

windtalker:
Two things can be improved
1. In TiDB side, if there is any region does not have an alive TiFlash, TiDB will just send the tasks to all the existing TiFlash, this will almost certainly cause the query fail because it will send task to TiFlash nodes that is not alive. I think it can be improved by 
- Just throw error if some regions don't have an alive replica
- Or still construct mpp tasks based only on the alive TiFlash nodes
2. In TiFlash, `FlashService::CancelMPPTask` is an async method, when TiDB meet error, it will call `FlashService::CancelMPPTask` to cancel current query, but since `FlashService::CancelMPPTask` is an async method, return from TiFlash does not means the cancel is done. And for this issue, the test's code will keep retry immediately when the query meet error. This immediate retry makes TiFlash's handing mpp task number increased so rapidly, and the concurrent query count is out of control. An improvement is to `FlashService::CancelMPPTask` wait a little time before it return to TiDB.

SeaRise:
Maybe having tidb mppgather and mpptask heartbeat would solve the problem?
If mpptask fails to ping tidb mppgather for a long time, it will be released automatically.

zanmato1984:
Should be fixed by https://github.com/pingcap/tiflash/pull/7589.

Please reopen if it occurs again.

PR #7589, opened 2023-06-02
Fix "lost cancel" for mpp query

### LINKED ISSUE #6727: Cache the CancelMPPTask request if task does not exist
## Enhancement

It's an optimization for a corner case.
Suppose a query has failed on some TiFlash nodes very quickly or TiDB receives a cancel request from a MySQL client very quickly. 
In that case, it's possible that some TiFlash nodes can receive a `CancelMPPTask` request before `EstablishMPPConnection`.
Then these tasks on TiFlash nodes will be canceled after running for sometime.

In normal cases, it's not a problem.
However, if the client keeps sending this query many times, many tasks will continue to accumulate to a large amount, which may affect other normal queries.
We can introduce an LRU cache to alleviate this problem.

windtalker:

Issue Number: ref #7587, close #6727

Problem Summary:

When TiDB cancel a mpp query, there is a chance that the `cancel` is lost:
| time | TiDB | TiFlash |
|-----|------|---------|
| t1 | Send Disptach MPPTask to TiFlash| Dispatch request is not received yet |
| t2 | Send CancelMPPTask request to TiFlash| Received CancelMPPTask, and found nothing to cancel |
| t3 |   | Receive Dispatch request of t1, then the cancel is lost | 

This pr add a `AbortedMPPGatherCache` to avoid this cancel lost
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Mpp/MPPTaskId.h
+++ b/dbms/src/Flash/Mpp/MPPTaskId.h
@@ -55,6 +55,22 @@ struct MPPQueryIdHash
     size_t operator()(MPPQueryId const & mpp_query_id) const noexcept;
 };
 
+struct MPPGatherId
+{
+    UInt64 gather_id;
+    MPPQueryId query_id;
+    MPPGatherId(Int64 gather_id_, const MPPQueryId & query_id_)
+        : gather_id(gather_id_)
+        , query_id(query_id_)
+    {}
+    bool operator==(const MPPGatherId & rid) const;
+};
+
+struct MPPGatherIdHash
+{
+    size_t operator()(MPPGatherId const & mpp_gather_id) const noexcept;
+};
+
 // Identify a mpp task.
 struct MPPTaskId
 {
--- a/dbms/src/Flash/Mpp/MPPTaskId.h
+++ b/dbms/src/Flash/Mpp/MPPTaskId.h
@@ -55,6 +55,11 @@ struct MPPQueryIdHash
     size_t operator()(MPPQueryId const & mpp_query_id) const noexcept;
 };
 
+/// A MPP query has one or more MPPGathers, each mpp gather has one or more MPPTasks. The mpp tasks in different mpp gathers are independent
+/// to each other, so theoretically speaking, the smallest cancel/retry unit of MPP query could be MPPGather. However, MPPGathers belong to
+/// the same MPP query could have dependence to each other(e.g. for query A join B, if the join is not supported in TiFlash, TiDB will generate
+/// two mpp gathers, one is reading from A and the other is reading from B, the probe side's mpp gather depends on the build side's mpp gather),
+/// so the smallest scheduling unit in TiFlash is still the MPP query
 struct MPPGatherId
 {
     UInt64 gather_id;
--- a/dbms/src/Flash/Mpp/MPPTaskId.h
+++ b/dbms/src/Flash/Mpp/MPPTaskId.h
@@ -60,6 +60,8 @@ struct MPPQueryIdHash
 /// the same MPP query could have dependence to each other(e.g. for query A join B, if the join is not supported in TiFlash, TiDB will generate
 /// two mpp gathers, one is reading from A and the other is reading from B, the probe side's mpp gather depends on the build side's mpp gather),
 /// so the smallest scheduling unit in TiFlash is still the MPP query
+/// Note currently, TiDB does not fill `gather_id` in mpp::TaskMeta, TiFlash hard coded 0 as the gather id for all MPPGatherId, so a mpp query now
+/// only has one MPPGather, and the schedule/cancel/retry unit in TiFlash is MPP query, we may implement mpp gather level's cancel/retry in the future.
 struct MPPGatherId
 {
     UInt64 gather_id;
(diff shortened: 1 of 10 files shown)


SeaRise:
And close https://github.com/pingcap/tiflash/issues/6727?

dbms/src/Flash/Mpp/MPPTaskManager.h
```diff
@@ -54,6 +54,35 @@ struct MPPQueryTaskSet
...
+    AbortedMPPGatherCache(size_t capacity_)
+        : capacity(capacity_)
+    {}
+    bool exists(const MPPGatherId & id)
+    {
+        return gather_ids_set.find(id) != gather_ids_set.end();
+    }
+    void add(const MPPGatherId & id)
+    {
+        if (gather_ids_set.size() >= capacity && gather_ids_set.find(id) == gather_ids_set.end())
```
SeaRise, line 75: It looks like L81 also needs to check `gather_ids_set.find(id) == gather_ids_set.end()` first
```suggestion
        if (gather_ids_set.find(id) == gather_ids_set.end())
            return;
        if (gather_ids_set.size() >= capacity)
```

SeaRise: Or it is only written in this way, the duplicate id will also enter the cache, but the existing entry will not be eliminated?

dbms/src/Flash/Mpp/MPPTaskId.h
```diff
@@ -55,6 +55,27 @@ struct MPPQueryIdHash
     size_t operator()(MPPQueryId const & mpp_query_id) const noexcept;
 };
 
+/// A MPP query has one or more MPPGathers, each mpp gather has one or more MPPTasks. The mpp tasks in different mpp gathers are independent
+/// to each other, so theoretically speaking, the smallest cancel/retry unit of MPP query could be MPPGather. However, MPPGathers belong to
+/// the same MPP query could have dependence to each other(e.g. for query A join B, if the join is not supported in TiFlash, TiDB will generate
+/// two mpp gathers, one is reading from A and the other is reading from B, the probe side's mpp gather depends on the build side's mpp gather),
+/// so the smallest scheduling unit in TiFlash is still the MPP query
+struct MPPGatherId
```
SeaRise, line 63: I'm a little confused, isn't the current MPPQueryId at the gather level?

SeaRise: As mentioned in the case, if there are gather1 join gather2, will gather2 be cancelled when gather1 is cancelled due to limit? Because the cancellation is based on MPPQueryId.

windtalker: Currently, the cancel unit is MPPQuery, but we can implement mpp gather level's cancel. In this pr, all the `gather_id` is hard coded as 0, which mean even for `A join B`, TiFlash still get only one MPPGather

windtalker: Add more comments.

SeaRise: ok

dbms/src/Flash/Mpp/MPPTaskManager.h
yibin87: Seems duplicated gather id will cause duplicate ids in gather_ids, a little strange

dbms/src/Flash/Mpp/tests/gtest_aborted_mpp_gather_cache.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <Flash/Mpp/MPPTaskManager.h>
+#include <TestUtils/TiFlashTestBasic.h>
+#include <gtest/gtest.h>
+
+
+namespace DB
+{
+namespace tests
+{
+class TestAbortedMPPGatherCache : public testing::Test
```
SeaRise, line 24: It looks like we can construct the case described in pr in `gtest_compute_server.cpp`.
```
TEST_F(ComputeServerRunner, xxx)
try
{
    WRAP_FOR_SERVER_TEST_BEGIN
    startServers(4);
    {
        auto [query_id, holder] = prepareMPPStreams(context
                                                     .scan("test_db", "test_table_1")
                                                     .aggregation({Max(col("s1"))}, {col("s2"), col("s3")})
                                                     .project({"max(s1)"}));
        MockComputeServerManager::instance().cancelQuery(query_id);
       EXPECT_TRUE(!assertQueryActive(query_id)); 
       holder.runMPPTasks();
        EXPECT_TRUE(assertQueryCancelled(query_id));
    }
    WRAP_FOR_SERVER_TEST_END
}
CATCH
```

SeaRise: just an option

windtalker: OK, I've added a test in `gtest_compute_server.cpp`

SeaRise (approved):
lgtm

yibin87 (approved):
LGTM

yibin87 (approved):
LGTM

SeaRise (approved):

SeaRise (approved):

yibin87 (approved):
LGTM

PR #7616, opened 2023-06-08
avoid unnecessary fsync in ps

lidezhu:

Issue Number: ref #6827 

Problem Summary:
1. Avoid unnecessary sync when writing to wal, after optimization, the write performance is about 14MB/s(previous is about 7MB/s)
image

2. Add some metrics
image
image

3. Add some unit tests for pipeline write

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -1484,6 +1491,8 @@ std::unordered_set<String> PageDirectory<Trait>::apply(PageEntriesEdit && edit,
     watch.restart();
 
     writers.push_back(&w);
+    SYNC_FOR("after_PageDirectory::enter_write_group");
+    CurrentMetrics::set(CurrentMetrics::PSWriterQueueSize, writers.size());
```
JaySon-Huang, line 1495: when there are multiple PS instances(deployed under non-disagg mode), this will randomly show the queue size of one of the PS instance. 🤔️

dbms/src/Storages/Page/V3/LogFile/LogWriter.cpp
```diff
@@ -158,9 +141,10 @@ void LogWriter::addRecord(ReadBuffer & payload, const size_t payload_size, const
         begin = false;
     } while (payload.hasPendingData());
 
-    if (!manual_flush)
+    flush(write_limiter, background);
```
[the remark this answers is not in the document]
lidezhu: Actually, `manual_flush` is almost always false. So this `flush` is always called here originally.
And flush without sync is a cheap operation. So we can always do it even `manual_sync` is true.

dbms/src/Storages/Page/V3/PageDirectory.cpp
lidezhu: Yes, it also has other problems. Change to `CurrentMetrics::Increment pending_writer_size{CurrentMetrics::PSPendingWriterSize};`

JaySon-Huang (approved):
LGTM

Lloyd-Pottiger (approved):

Issue #7376, opened 2023-04-25
Window functions


xzhangxian1008:
## Feature Request

- [x] Support first_value function
- [x] Support last_value function
- [x] Support offset boundry type
- [x] Support range frame type
- [ ] Support aggregation function

PR #7629, opened 2023-06-08
Support range frame type

xzhangxian1008:

Issue Number: ref #7376

Problem Summary:

Following steps are the precedure that how range frame type works:
1. Compilation stage: as range frame needs the arithmetic calculation, we build a plus/minux auxiliary function at the compilation stage, and record the result column's index and get the column types of order by column and result columns for further use. How to build this function? tipb::Expr sent from tidb will answer this question.
2. Runtime stage: main operation of this stage is the comparison between order by column and result column that produced by the plux/minux function. Comparison is easy, however, we have to handle different types. So, most of codes in this stage are to process the data type, they are unrelated to the main operation.

Core files changed for compilation stage:
- DAGExpressionAnalyzer.cpp

Core files changed for runtime stage:
- WindowBlockInputStream.cpp

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,9 +36,132 @@ extern const int BAD_ARGUMENTS;
...
+template <typename T>
+T getValue(const ColumnPtr & col_ptr, size_t idx)
+{
+    if constexpr (checkIfSimpleNumericType<T>())
+    {
+        const auto * col = static_cast<const ColumnVector<T> *>(&(*col_ptr));
+        const typename ColumnVector<T>::Container & data = col->getData();
+        return data[idx];
+    }
+    else if (checkIfDecimalType<T>())
```
SeaRise, line 107:
```suggestion
    else if constexpr (checkIfDecimalType<T>())
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -360,6 +495,142 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToFrameEnd(const RowNumbe
     return std::make_tuple(frame_end_row, true);
 }
 
+RowNumber WindowTransformAction::stepToFrameStartForRange(const RowNumber & current_row)
+{
+    if (col_type == Window::ColumnType::UnInitialized)
+        // order_column_indices must contain value as tidb should ensure this.
+        col_type = getColumnType(inputAt(current_row)[order_column_indices[0]]);
```
SeaRise, line 502: Can this be obtained from metadata when compiling WindowFunction?

xzhangxian1008: > Can this be obtained from metadata when compiling WindowFunction?

okk, I have changed to this behaviour.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,146 @@ extern const int BAD_ARGUMENTS;
...
+template <typename T>
+constexpr bool checkIfSimpleNumericType()
+{
+    return checkIfIntType<T>() || checkIfFloatType<T>();
+}
+
+template <typename T>
+constexpr bool checkIfDecimalType()
+{
+    return std::is_same_v<T, Decimal32> || std::is_same_v<T, Decimal64> || std::is_same_v<T, Decimal128> || std::is_same_v<T, Decimal256>;
```
SeaRise, line 63:
```suggestion
    return IsDecimal<T>;
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,146 @@ extern const int BAD_ARGUMENTS;
...
+template <typename T>
+constexpr bool checkIfIntType()
+{
+    return std::is_same_v<T, UInt8> || std::is_same_v<T, UInt16> || std::is_same_v<T, UInt32> || std::is_same_v<T, UInt64> || std::is_same_v<T, Int8> || std::is_same_v<T, Int16> || std::is_same_v<T, Int32> || std::is_same_v<T, Int64>;
+}
+
+template <typename T>
+constexpr bool checkIfFloatType()
+{
+    return std::is_same_v<T, Float32> || std::is_same_v<T, Float64>;
```
SeaRise, line 51:
```suggestion
    return std::is_floating_point_v<T>;
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,146 @@ extern const int BAD_ARGUMENTS;
 extern const int NOT_IMPLEMENTED;
 } // namespace ErrorCodes
 
+namespace
+{
+template <typename T>
+constexpr bool checkIfIntType()
+{
+    return std::is_same_v<T, UInt8> || std::is_same_v<T, UInt16> || std::is_same_v<T, UInt32> || std::is_same_v<T, UInt64> || std::is_same_v<T, Int8> || std::is_same_v<T, Int16> || std::is_same_v<T, Int32> || std::is_same_v<T, Int64>;
```
SeaRise, line 45:
```suggestion
    return std::is_integral_v<T>;
```

xzhangxian1008: > std::is_integral_v<T>

done

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > 

done

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > 

done

dbms/src/Interpreters/WindowDescription.h
```diff
@@ -109,6 +135,24 @@ struct WindowDescription
...
+    Window::ColumnType order_by_col_type = Window::ColumnType::UnInitialized;
+
+    // Sometimes, we may cast order by column or const column to the target type.
+    // When the casted column is nullable, we need to check if there is null in
+    // the casted column as range frame type forbids the occurrence of null value.
+    // Tip: only used for range frame type
+    bool is_aux_begin_col_nullable = false;
+    bool is_aux_end_col_nullable = false;
+
+    Window::ColumnType begin_aux_col_type = Window::ColumnType::UnInitialized;
```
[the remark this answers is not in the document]
xzhangxian1008: > 

I think it's wasteful to hold two more pointers that only used to mark the data type.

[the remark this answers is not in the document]
SeaRise: Then use `enum class TypeIndex` directly?
`TypeIndex` and `DataType` are used for data type in Function-related code.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -1422,22 +1592,39 @@ String DAGExpressionAnalyzer::alignReturnType(
...
+template <bool need_col_ptr>
 void DAGExpressionAnalyzer::initChain(ExpressionActionsChain & chain) const
 {
     if (chain.steps.empty())
     {
         const auto & columns = getCurrentInputColumns();
-        NamesAndTypesList column_list;
         std::unordered_set<String> column_name_set;
-        for (const auto & col : columns)
+        if constexpr (need_col_ptr)
```
SeaRise, line 1602: Some hack...
In TiFlash, the sample block does not have a column pointer, while the header does have the block.
And here, we are setting a sample block for ExpressionAction.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,11 +198,200 @@ void appendWindowDescription(
     window_columns.emplace_back(func_string, result_type);
     source_columns.emplace_back(func_string, result_type);
 }
+
+Window::ColumnType getColumnType(const ColumnPtr & col_ptr)
```
SeaRise, line 202: If use TypeIndex, the code can be simplify like `col_ptr->getDataTypePtr()->getTypeIndex()`

dbms/src/Interpreters/WindowDescription.h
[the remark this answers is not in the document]
xzhangxian1008: > Then use `enum class TypeIndex` directly? `TypeIndex` and `DataType` are used for data type in Function-related code.

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
xzhangxian1008: > 

However, in order to get the data type at compilation stage we need this column pointer.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
xzhangxian1008: > getTypeIndex

okk, I will use `DataType`

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
SeaRise: DataTypePtr is enough, all type-related judgments can be done with DataTypePtr

SeaRise: `need_col_ptr` is still needed now?

xzhangxian1008: > `need_col_ptr` is still needed now?

okk, I have deleted this now.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -31,14 +32,21 @@
...
 #include <Interpreters/Set.h>
 #include <Interpreters/Settings.h>
 #include <Interpreters/convertFieldToType.h>
 #include <Parsers/ASTIdentifier.h>
 #include <Storages/Transaction/TypeMapping.h>
 #include <WindowFunctions/WindowFunctionFactory.h>
+#include <tipb/executor.pb.h>
+#include <tipb/expression.pb.h>
+
+#include "Core/Types.h"
```
SeaRise, line 49: please check all `#include "xxx.h"`

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,162 @@ void appendWindowDescription(
...
+    if (!aux_col_name.empty())
+    {
+        // Prepare something
+        Int32 & range_auxiliary_column_index = is_begin ? window_desc.frame.begin_range_auxiliary_column_index : window_desc.frame.end_range_auxiliary_column_index;
+        TypeIndex & aux_col_type = is_begin ? window_desc.begin_aux_col_type : window_desc.end_aux_col_type;
+        bool & is_aux_col_nullable = is_begin ? window_desc.is_aux_begin_col_nullable : window_desc.is_aux_end_col_nullable;
+
+        // Set auxiliary columns' indexes
+        size_t aux_col_idx = tmp_block.getPositionByName(aux_col_name);
+        range_auxiliary_column_index = aux_col_idx; // Set here
```
SeaRise, line 219: `Set here` Seems meaningless comments

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,162 @@ void appendWindowDescription(
...
+        // Set auxiliary columns' types
+        const auto & col_and_name = tmp_block.getByName(aux_col_name);
+        auto data_type = col_and_name.type;
+        bool is_nullable = data_type->isNullable();
+        if (is_nullable)
+        {
+            if (const auto * const data_type_nullable = typeid_cast<const DataTypeNullable *>(data_type.get()))
+                aux_col_type = data_type_nullable->getNestedType()->getTypeId(); // Set here
+            else
+                throw Exception("Invalid data type");
```
SeaRise, line 230:
```suggestion
            aux_col_type = removeNullable(data_type)->getTypeId();
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -360,6 +508,252 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToFrameEnd(const RowNumbe
...
+    ColumnPtr cur_row_aux_column;
+    if (is_aux_col_nullable)
+    {
+        const auto & nullable_col = static_cast<const ColumnNullable &>(*(inputAt(current_row)[cur_row_aux_col_idx]));
+        cur_row_aux_column = nullable_col.getNestedColumnPtr();
+    }
+    else
+        cur_row_aux_column = inputAt(current_row)[cur_row_aux_col_idx];
+    typename ActualCmpDataType<T>::Type current_row_aux_value = getValue<T>(cur_row_aux_column, current_row.row);
+    return moveCursorAndFindFrameBoundary<typename ActualCmpDataType<T>::Type, is_begin, is_desc>(cursor, current_row_aux_value);
```
SeaRise, line 668: If it is the function code style of tiflash/clickhouse, the code should be
```
auto column = inputAt(current_row)[cur_row_aux_col_idx];
if (!column->isNullAt(current_row.row))
{
    ...
}
```
Please follow the code style of tiflash.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -360,6 +508,252 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToFrameEnd(const RowNumbe
...
+    if constexpr (is_begin)
+        is_aux_col_nullable = window_description.is_aux_begin_col_nullable;
+    else
+        is_aux_col_nullable = window_description.is_aux_end_col_nullable;
+
+    ColumnPtr cur_row_aux_column;
+    if (is_aux_col_nullable)
+    {
+        const auto & nullable_col = static_cast<const ColumnNullable &>(*(inputAt(current_row)[cur_row_aux_col_idx]));
+        cur_row_aux_column = nullable_col.getNestedColumnPtr();
```
SeaRise, line 663: Will it be read even if the value is null?

xzhangxian1008: I have removed all the handling of nullable column

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,131 @@ namespace ErrorCodes
...
+{
+    if constexpr (checkIfSimpleNumericType<T>())
+    {
+        const auto * col = static_cast<const ColumnVector<T> *>(&(*col_ptr));
+        const typename ColumnVector<T>::Container & data = col->getData();
+        return data[idx];
+    }
+    else if (IsDecimal<T>)
+    {
+        const auto * col = static_cast<const ColumnDecimal<T> *>(&(*col_ptr));
```
SeaRise, line 95: Could it be ColumnConst?
Does Decimal appear here?

dbms/src/Debug/MockExecutor/WindowBinder.cpp
```diff
@@ -149,7 +149,7 @@ bool WindowBinder::toTiPBExecutor(tipb::Executor * tipb_executor, int32_t collat
     return children[0]->toTiPBExecutor(children_executor, collator_id, mpp_info, context);
 }
 
-ExecutorBinderPtr compileWindow(ExecutorBinderPtr input, size_t & executor_index, ASTPtr func_desc_list, ASTPtr partition_by_expr_list, ASTPtr order_by_expr_list, mock::MockWindowFrame frame, uint64_t fine_grained_shuffle_stream_count)
+ExecutorBinderPtr compileWindow(ExecutorBinderPtr input, size_t & executor_index, ASTPtr func_desc_list, ASTPtr partition_by_expr_list, ASTPtr order_by_expr_list, mock::MockWindowFrame & frame, uint64_t fine_grained_shuffle_stream_count)
```
SeaRise, line 152: Why modify here?

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,131 @@ namespace ErrorCodes
...
+{
+    using Type = std::conditional_t<checkIfSimpleNumericType<T>(), T, DecimalWithScale<T>>;
+};
+
+template <typename T>
+typename ActualCmpDataType<T>::Type getValue(const ColumnPtr & col_ptr, size_t idx)
+{
+    if constexpr (checkIfSimpleNumericType<T>())
+    {
+        const auto * col = static_cast<const ColumnVector<T> *>(&(*col_ptr));
```
SeaRise, line 89: Can Field be used like https://github.com/pingcap/tiflash/blob/892c2cac41d95d3399bd46278a4eba44f1caa25e/dbms/src/WindowFunctions/IWindowFunction.cpp#L336-L361 ? Although it will introduce the overhead of constructing Field

dbms/src/Debug/MockExecutor/WindowBinder.cpp
xzhangxian1008: > Why modify here?

The previous refactor pr is from this pr, I forget to delete it. I will delete it.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Could it be ColumnConst? Does Decimal appear here?

1. It's impossible to be const column here
2. This if branch is especially used for handling Decimal.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Can Field be used like
> 
> https://github.com/pingcap/tiflash/blob/892c2cac41d95d3399bd46278a4eba44f1caa25e/dbms/src/WindowFunctions/IWindowFunction.cpp#L336-L361
> 
> ? Although it will introduce the overhead of constructing Field

I think it's better to avoid `Field` as possible as we can.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > > Could it be ColumnConst? Does Decimal appear here?
> 
> 1. It's impossible to be const column here
> 2. This if branch is especially used for handling Decimal, so the Decimal will appear here.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+void setOrderByColumnTypeAndDirection(WindowDescription & window_desc, const ExpressionActionsPtr & actions, const tipb::Window & window)
+{
+    // Execute this function only when the frame type is Range
+    if (window.frame().type() != tipb::WindowFrameType::Ranges)
+        return;
+
+    if (!window_desc.order_by.empty())
+    {
+        const Block & sample_block = actions->getSampleBlock();
+        const String & order_by_col_name = window_desc.order_by[0].column_name;
```
SeaRise, line 258: So order by must have only one column?
If so please add a comment and `runtime_check`

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+        return;
+
+    const Block & tmp_block = actions->getSampleBlock();
+    setAuxiliaryColumnInfoImpl(window_desc, begin_aux_col_name, tmp_block, true);
+    setAuxiliaryColumnInfoImpl(window_desc, end_aux_col_name, tmp_block, false);
+}
+
+void setOrderByColumnTypeAndDirection(WindowDescription & window_desc, const ExpressionActionsPtr & actions, const tipb::Window & window)
+{
+    // Execute this function only when the frame type is Range
```
SeaRise, line 251: why?
Can some explaining notes be added?
And it looks like the function name could be changed to `setOrderByColumnTypeAndDirectionForRangeFrame`

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+    for (const auto & window_function_description : window_desc.window_functions_descriptions)
+    {
+        for (const auto & argument_name : window_function_description.argument_names)
+            step.required_output.push_back(argument_name);
+    }
+
+    window_desc.before_window = chain.getLastActions();
+    setAuxiliaryColumnInfo(step.actions, window_desc, aux_col_names.first, aux_col_names.second, window);
+
+    // We should not call finalize here as some useful columns will be unexpected removed
```
SeaRise, line 323: The handling here is a bit weird.
why not setting require_output?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+
+    analyzer->appendWindowColumns(window_desc, window, step.actions);
+    // set required output for window funcs's arguments.
+    for (const auto & window_function_description : window_desc.window_functions_descriptions)
+    {
+        for (const auto & argument_name : window_function_description.argument_names)
+            step.required_output.push_back(argument_name);
+    }
+
+    window_desc.before_window = chain.getLastActions();
```
SeaRise, line 320: Both chain.getLastActions and step.getActions use only one,

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+
+        // Set auxiliary columns' indexes
+        size_t aux_col_idx = tmp_block.getPositionByName(aux_col_name);
+        range_auxiliary_column_index = aux_col_idx;
+
+        // Set auxiliary columns' types
+        const auto & col_and_name = tmp_block.getByName(aux_col_name);
+        auto data_type = col_and_name.type;
+        if (data_type->isNullable())
+            throw Exception("Get an unexpected nullable column when the frame type is range");
```
SeaRise, line 224:
```suggestion
        RUNTIME_CHECK(!data_type->isNullable(), "Get an unexpected nullable column when the frame type is range");
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
SeaRise: Why must it not be ColumnConst, doesn't this depend on the input?
What if the operator below window outputs ColumnConst?

dbms/src/WindowFunctions/tests/gtest_first_value.cpp
```diff
@@ -25,6 +25,7 @@
 #include <memory>
 #include <optional>
 
+
```
SeaRise, line 28: `// TODO support unsigned int.`
https://github.com/pingcap/tiflash/pull/7755#discussion_r1254210104
Support unsigned or improve this comment?

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Why must it not be ColumnConst, doesn't this depend on the input? What if the operator below window outputs ColumnConst?

Yes, it depend on the input, but tidb will ensure that the input column that user designate is not const.
```sql
select *, first_value(v) over (partition by p order by 1 range between 1 preceding and 1 following) over a from a; // Error, order by 1, 1 is a const column.
```

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+    auto & after_window_step = analyze->initAndGetLastStep(chain);
+    analyze->appendCastAfterWindow(after_window_step.actions, window, source_size);
+    window_desc.after_window_columns = analyze->getCurrentInputColumns();
+    analyze->appendSourceColumnsToRequireOutput(after_window_step);
+    window_desc.after_window = chain.getLastActions();
+    chain.finalize();
+    chain.clear();
+}
+
+ExpressionActionsChain::Step & createStepForBuildingWindowDescription(DAGExpressionAnalyzer * const analyzer, ExpressionActionsChain & chain)
```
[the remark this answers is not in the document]
xzhangxian1008: > The function name is a bit strange, because there is no logic related to window here. Is it necessary to wrap a new function here?

I will change this function's name.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
xzhangxian1008: > So order by must have only one column? If so please add a comment and `runtime_check`

okk

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > what if `select *, first_value(v) over (partition by p order by a range between b preceding and c following) over a from (select 1 as a, 2 as b, 3 as c, v from table); `?
I will add comment and test for it.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
xzhangxian1008: > The handling here is a bit weird. why not setting require_output?

Because these auxiliary column don't appear in sample_block, and it will raise error.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,132 @@ namespace ErrorCodes
...
+
+// When T is Decimal, we should convert it to DecimalWithScale type
+// as we need scale value when executing the comparison operation.
+template <typename T>
+struct ActualCmpDataType
+{
+    using Type = std::conditional_t<checkIfSimpleNumericType<T>(), T, DecimalWithScale<T>>;
+};
+
+// ColumnConst should not appear at here as MySQL does not allow the order by column is const, so we do not handle it.
```
SeaRise, line 84: xsMloCmevd
In this plan, if Projection_11 is on ExchangeReceiver_14, order by column const.
The fact that Projection_11 is on ExchangeReceiver_14 does not affect the semantics of the query.

SeaRise: OgS3VJLnOY
Here comes `order by ColumnConst`, see Sort_21

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
SeaRise: https://github.com/pingcap/tiflash/pull/7629/files#r1270426007
But here getting aux col from sample block.

dbms/src/Parsers/ASTWithAlias.h
```diff
@@ -34,13 +34,13 @@ class ASTWithAlias : public IAST
...
 
-    String getColumnName() const override final { return prefer_alias_to_column_name && !alias.empty() ? alias : getColumnNameImpl(); }
+    String getColumnName() const final { return prefer_alias_to_column_name && !alias.empty() ? alias : getColumnNameImpl(); }
     String getAliasOrColumnName() const override { return alias.empty() ? getColumnNameImpl() : alias; }
     String tryGetAlias() const override { return alias; }
     void setAlias(const String & to) override { alias = to; }
 
     /// Calls formatImplWithoutAlias, and also outputs an alias. If necessary, encloses the entire expression in brackets.
-    void formatImpl(const FormatSettings & settings, FormatState & state, FormatStateStacked frame) const override final;
+    void formatImpl(const FormatSettings & settings, FormatState & state, FormatStateStacked frame) const final;
```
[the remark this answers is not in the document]
xzhangxian1008: > why delete `override`?

'override' is redundant since the function is already declared 'final'

dbms/src/Parsers/ASTWithAlias.h
```diff
@@ -34,13 +34,13 @@ class ASTWithAlias : public IAST
 
     using IAST::IAST;
 
-    String getColumnName() const override final { return prefer_alias_to_column_name && !alias.empty() ? alias : getColumnNameImpl(); }
+    String getColumnName() const final { return prefer_alias_to_column_name && !alias.empty() ? alias : getColumnNameImpl(); }
```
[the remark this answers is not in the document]
xzhangxian1008: > why delete `override`?

'override' is redundant since the function is already declared 'final'

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
xzhangxian1008: > https://github.com/pingcap/tiflash/pull/7629/files#r1270426007 But here getting aux col from sample block.

done

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: Now, we get value with Field.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -627,38 +782,13 @@ void DAGExpressionAnalyzer::appendWindowColumns(WindowDescription & window_descr
 WindowDescription DAGExpressionAnalyzer::buildWindowDescription(const tipb::Window & window)
 {
     ExpressionActionsChain chain;
-    ExpressionActionsChain::Step & step = initAndGetLastStep(chain);
-    appendSourceColumnsToRequireOutput(step);
+    ExpressionActionsChain::Step & step = createStep(this, chain);
```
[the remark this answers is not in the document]
xzhangxian1008: > It doesn't feel necessary to wrap a new function `createStep`.

Wrapping a function is just to make reader could quickly get the meaning of these codes, as they can know the meaning by the function name instead of comments.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,153 @@ void appendWindowDescription(
...
+
+void initBeforeWindow(
+    DAGExpressionAnalyzer * analyzer,
+    WindowDescription & window_desc,
+    ExpressionActionsChain & chain,
+    const tipb::Window & window,
+    ExpressionActionsChain::Step & step)
+{
+    // Prepare auxiliary function for range frame type
+    auto aux_col_names = addRangeFrameAuxiliaryFunctionAction(analyzer, step.actions, window);
```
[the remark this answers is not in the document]
xzhangxian1008: > auto actions = chain.getLastActions();

done

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
SeaRise: It looks like there is no difference in readability between
```
    auto & step = initAndGetLastStep(chain);
    appendSourceColumnsToRequireOutput(step);
```
and
```
    ExpressionActionsChain::Step & step = createStep(this, chain);
```
...
Instead, the latter hides the logic of `appendSourceColumnsToRequireOutput` on the function name. .

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,105 @@ namespace ErrorCodes
...
+    if constexpr (checkIfSimpleNumericType<T>())
+    {
+        return (*col_ptr)[idx].get<T>();
+    }
+    else if (IsDecimal<T>)
+    {
+        return (*col_ptr)[idx].get<DecimalField<T>>();
+    }
+    else
+        throw Exception("Unexpected column type!");
```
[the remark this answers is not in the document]
xzhangxian1008: done

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
xzhangxian1008: > createStep

changed

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,95 @@ namespace ErrorCodes
...
+template <typename T>
+typename ActualCmpDataType<T>::Type getValue(const ColumnPtr & col_ptr, size_t idx)
+{
+    return (*col_ptr)[idx].get<typename ActualCmpDataType<T>::Type>();
+}
+
+template <typename T, bool is_start, bool is_desc>
+bool isInRangeCommonImpl(T current_row_aux_value, T cursor_value)
+{
+    if constexpr ((is_start && is_desc) || (!is_start && !is_desc))
```
windtalker, line 73: how about use 
`if constexpr !(is_start ^ is_desc)`

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,95 @@ namespace ErrorCodes
...
+        return cursor_value <= current_row_aux_value;
+    else
+        return cursor_value >= current_row_aux_value;
+}
+
+template <bool is_start, bool is_desc>
+bool isInRangeIntImpl(Int64 current_row_aux_value, Int64 cursor_value)
```
windtalker, line 80: Is it still correct if the column type is `UInt64`?

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,95 @@ namespace ErrorCodes
...
+}
+
+template <typename AuxColType, typename OrderByColType, bool is_start, bool is_desc>
+bool isInRangeNotIntImpl(AuxColType current_row_aux_value, OrderByColType cursor_value)
+{
+    Float64 current_row_aux_value_float64;
+    Float64 cursor_value_float64;
+
+    if constexpr (checkIfDecimalFieldType<AuxColType>())
+        current_row_aux_value_float64 = current_row_aux_value.getValue().template toFloat<Float64>(current_row_aux_value.getScale());
```
windtalker, line 92: Why decimal value need to be converted to float?

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+    // Used for calculating the frame end for rows frame type
+    std::tuple<RowNumber, bool> stepToFrameEndForRows();
+
+    // Used for calculating the frame start for range frame type
+    RowNumber stepToFrameStartForRange();
+    // Used for calculating the frame end for range frame type
+    std::tuple<RowNumber, bool> stepToFrameEndForRange();
+
+    template <bool is_desc>
+    RowNumber stepToFrameStartForRange();
```
windtalker, line 183: Maybe rename it to `stepToFrameStartForRangeFrameOrderCase`

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+
+    template <bool is_desc>
+    std::tuple<RowNumber, bool> stepToFrameEndForRange();
```
windtalker, line 186: Ditto

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
 
     void appendInfo(FmtBuffer & buffer) const;
 
+private:
+    void stepToFrameStartWithOffsetBoundry();
+    void stepToFrameEndWithOffsetBoundary();
+
+    // Used for calculating the frame start for rows frame type
+    RowNumber stepToFrameStartForRows();
```
windtalker, line 173: maybe rename it to `stepToFrameStartForRowsFrame`

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+    // Used for calculating the frame end for rows frame type
+    std::tuple<RowNumber, bool> stepToFrameEndForRows();
```
windtalker, line 175: Ditto

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+
+    // Used for calculating the frame start for range frame type
+    RowNumber stepToFrameStartForRange();
```
windtalker, line 178: Maybe rename it to `stepToFrameStartForRangeFrame`

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+    // Used for calculating the frame end for range frame type
+    std::tuple<RowNumber, bool> stepToFrameEndForRange();
```
windtalker, line 180: Ditto

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+    RowNumber stepToFrameEndForRangeImpl();
+
+    template <typename T, bool is_begin, bool is_desc>
+    RowNumber stepToFrameForRangeImpl();
+
+    // distance is left - right.
+    UInt64 distance(RowNumber left, RowNumber right);
+
+    template <typename AuxColType, bool is_begin, bool is_desc>
+    RowNumber moveCursorAndFindFrameBoundary(RowNumber cursor, AuxColType current_row_aux_value);
```
windtalker, line 201: Maybe rename it to `moveCursorAndFindRangeFrameBoundary`

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+
+    template <typename AuxColType, typename OrderByColType, bool is_begin, bool is_desc>
+    RowNumber moveCursorAndFindFrameBoundary(RowNumber cursor, AuxColType current_row_aux_value);
```
windtalker, line 204: ditto

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -171,6 +165,48 @@ struct WindowTransformAction
...
+
+    template <typename AuxColType, typename OrderByColType, int CmpDataType, bool is_begin, bool is_desc>
+    RowNumber moveCursorAndFindFrameBoundaryImpl(RowNumber cursor, AuxColType current_row_aux_value);
```
windtalker, line 207: ditto

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -509,13 +829,7 @@ void WindowTransformAction::advanceFrameEnd()
     }
     case WindowFrame::BoundaryType::Offset:
```
windtalker, line 830: In `WindowDescription::setWindowFrame`, for WindowBoundType of both `Following` and `Preceding` is converted to `WindowFrame::BoundaryType::Offset`, can you double confirm that if this information lost is as expected after we truely support frame?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -627,38 +779,17 @@ void DAGExpressionAnalyzer::appendWindowColumns(WindowDescription & window_descr
...
-    {
-        for (const auto & argument_name : window_function_description.argument_names)
-            step.required_output.push_back(argument_name);
-    }
+    initChain(chain);
 
-    window_description.before_window = chain.getLastActions();
-    chain.finalize();
-    chain.clear();
+    ExpressionActionsChain::Step & step = chain.getLastStep();
```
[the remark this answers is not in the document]
xzhangxian1008: > `initAndGetLastStep`

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,150 @@ void appendWindowDescription(
...
+        TypeIndex & aux_col_type = is_begin ? window_desc.begin_aux_col_type : window_desc.end_aux_col_type;
+
+        // Set auxiliary columns' indexes
+        size_t aux_col_idx = tmp_block.getPositionByName(aux_col_name);
+        range_auxiliary_column_index = aux_col_idx;
+
+        // Set auxiliary columns' types
+        const auto & col_and_name = tmp_block.getByName(aux_col_name);
+        auto data_type = col_and_name.type;
+        RUNTIME_CHECK_MSG(!data_type->isNullable(), "Get an unexpected nullable column when the frame type is range");
```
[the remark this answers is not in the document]
xzhangxian1008: > Why nullable type is not allowed?

Because MySQL forbids to input a nullable order by column.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > how about use `if constexpr !(is_start ^ is_desc)`

I think it's easier to understand the origin code and it is only calculated in the compile time and will not slow the runtime performance.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Why decimal value need to be converted to float?

When these two data type are all decimal, it's no need to convert then to float, I will fix it. However, when one of the data type is float, converting decimal to float is necessary.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Is it still correct if the column type is `UInt64`?

I will fix it.

dbms/src/DataStreams/WindowBlockInputStream.h
xzhangxian1008: > Maybe rename it to `stepToFrameStartForRangeFrameOrderCase`

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
SeaRise:
```
 explain select *, first_value(v) over (partition by p order by (case 0 is null when true then o else null end) range between 3 preceding and 0 following) as a from test.first_range;
+----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
| id                               | estRows | task      | access object     | operator info                                                                                                                               |
+----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
| Projection_6                     | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, Column#7                                                                        |
| └─Shuffle_13                     | 17.00   | root      |                   | execution info: concurrency:5, data sources:[Projection_9]                                                                                  |
|   └─Window_7                     | 17.00   | root      |                   | first_value(test.first_range.v)->Column#7 over(partition by test.first_range.p order by Column#6 range between 3 preceding and 0 following) |
|     └─Sort_12                    | 17.00   | root      |                   | test.first_range.p, Column#6                                                                                                                |
|       └─Projection_9             | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, <nil>->Column#6                                                                 |
|         └─TableReader_11         | 17.00   | root      |                   | data:TableFullScan_10                                                                                                                       |
|           └─TableFullScan_10     | 17.00   | cop[tikv] | table:first_range | keep order:false, stats:pseudo                                                                                                              |
+----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
```
It looks like `Column#6` is an only null column const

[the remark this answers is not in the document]
xzhangxian1008: > ```
>  explain select *, first_value(v) over (partition by p order by (case 0 is null when true then o else null end) range between 3 preceding and 0 following) as a from test.first_range;
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> | id                               | estRows | task      | access object     | operator info                                                                                                                               |
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> | Projection_6                     | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, Column#7                                                                        |
> | └─Shuffle_13                     | 17.00   | root      |                   | execution info: concurrency:5, data sources:[Projection_9]                                                                                  |
> |   └─Window_7                     | 17.00   | root      |                   | first_value(test.first_range.v)->Column#7 over(partition by test.first_range.p order by Column#6 range between 3 preceding and 0 following) |
> |     └─Sort_12                    | 17.00   | root      |                   | test.first_range.p, Column#6                                                                                                                |
> |       └─Projection_9             | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, <nil>->Column#6                                                                 |
> |         └─TableReader_11         | 17.00   | root      |                   | data:TableFullScan_10                                                                                                                       |
> |           └─TableFullScan_10     | 17.00   | cop[tikv] | table:first_range | keep order:false, stats:pseudo                                                                                                              |
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> ```
> 
> It looks like `Column#6` is an only null column const

okk, we will raise error when the column is nullable.

dbms/src/WindowFunctions/tests/gtest_lead_lag.cpp
```diff
@@ -368,8 +367,6 @@ CATCH
 TEST_F(LeadLag, Int)
 try
 {
-    // TODO support unsigned int.
```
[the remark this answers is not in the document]
xzhangxian1008: > Looks like no tests are added, why delete these comments

We have move this comment to L25

dbms/src/WindowFunctions/tests/gtest_first_value.cpp
```diff
@@ -145,4 +368,17 @@ try
...
+TEST_F(FirstValue, firstValueWithRangeFrameType)
+try
+{
+    testIntOrderByColForRangeFrame();
+    testFloatOrderByColForRangeFrame();
+
+    // TODO This test is disabled as we can not assign decimal field's flen now
+    // in MockStorage.cpp::mockColumnInfosToTiDBColumnInfos().
+    // However, we will test this data type in fullstack tests.
+    // testDecimalOrderByColForRangeFrame();
```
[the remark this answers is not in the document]
xzhangxian1008: > I don't get your point here, looks like the test only need to construct decimal column, which I think is already support, you can find examples in `gtest_arithmetic_functions.cpp`

Because gtest_arithmetic_functions.cpp directly executes the function without using `addMockTable` which need to construct decimal column infos.

dbms/src/WindowFunctions/tests/gtest_last_value.cpp
```diff
@@ -82,6 +83,181 @@ class LastValue : public DB::tests::WindowTest
              toNullableVec<FloatType>(/*value*/ {{}, 2, 3, 4, {}, 6, 7, 8, 9, {}, 11, 12, {}})},
             unbounded_type_frame);
     }
+
+    void testIntOrderByColForRangeFrame()
```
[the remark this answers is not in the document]
xzhangxian1008: > Range frame with time type is not supported?

We will test it in ft.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,150 @@ void appendWindowDescription(
...
+    WindowDescription & window_desc,
+    const String & aux_col_name,
+    const Block & tmp_block,
+    bool is_begin)
+{
+    if (!aux_col_name.empty())
+    {
+        // Prepare something
+        Int32 & range_auxiliary_column_index = is_begin ? window_desc.frame.begin_range_auxiliary_column_index : window_desc.frame.end_range_auxiliary_column_index;
+        TypeIndex & aux_col_type = is_begin ? window_desc.begin_aux_col_type : window_desc.end_aux_col_type;
```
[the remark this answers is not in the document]
xzhangxian1008: > I think we can make `range_auxiliary_column_index` and `aux_col_type` as the function's input argument, and don't need `is_begin` anymore

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -627,38 +779,17 @@ void DAGExpressionAnalyzer::appendWindowColumns(WindowDescription & window_descr
...
-    appendCastAfterWindow(after_window_step.actions, window, source_size);
-    window_description.after_window_columns = getCurrentInputColumns();
-    appendSourceColumnsToRequireOutput(after_window_step);
-    window_description.after_window = chain.getLastActions();
-    chain.finalize();
-    chain.clear();
+    WindowDescription window_description = createAndInitWindowDesc(this, window);
+    setOrderByColumnTypeAndDirectionForRangeFrame(window_description, step.actions, window);
+    initBeforeWindow(this, window_description, chain, window);
+    initAfterWindow(this, window_description, chain, window, source_size);
```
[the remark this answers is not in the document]
xzhangxian1008: > 

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,150 @@ void appendWindowDescription(
...
+    setAuxiliaryColumnInfoImpl(window_desc, end_aux_col_name, tmp_block, false);
+}
+
+void setOrderByColumnTypeAndDirectionForRangeFrame(WindowDescription & window_desc, const ExpressionActionsPtr & actions, const tipb::Window & window)
+{
+    // Execute this function only when the frame type is Range
+    if (window.frame().type() != tipb::WindowFrameType::Ranges)
+        return;
+
+    if (!window_desc.order_by.empty())
```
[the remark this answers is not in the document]
xzhangxian1008: > If `window_desc.order_by.empty()` is true, should throw error?

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -192,6 +200,150 @@ void appendWindowDescription(
...
+        return std::make_pair("", "");
+
+    String begin_aux_col_name;
+    String end_aux_col_name;
+
+    // For begin frame
+    if (window.frame().start().has_frame_range())
+        begin_aux_col_name = DAGExpressionAnalyzerHelper::buildFunction(analyzer, window.frame().start().frame_range(), actions);
+    // For end frame
+    if (window.frame().end().has_frame_range())
```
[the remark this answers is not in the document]
xzhangxian1008: > 

I think they must be set and I will check this.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
SeaRise: > okk, we will raise error when the column is nullable.
So is it expected that an error will be reported when executing this query?

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > In `WindowDescription::setWindowFrame`, for WindowBoundType of both `Following` and `Preceding` is converted to `WindowFrame::BoundaryType::Offset`, can you double confirm that if this information lost is as expected after we truely support frame?

According to tidb, `Following` and `Preceding` is useless when frame type is range.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
xzhangxian1008: > ```
>  explain select *, first_value(v) over (partition by p order by (case 0 is null when true then o else null end) range between 3 preceding and 0 following) as a from test.first_range;
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> | id                               | estRows | task      | access object     | operator info                                                                                                                               |
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> | Projection_6                     | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, Column#7                                                                        |
> | └─Shuffle_13                     | 17.00   | root      |                   | execution info: concurrency:5, data sources:[Projection_9]                                                                                  |
> |   └─Window_7                     | 17.00   | root      |                   | first_value(test.first_range.v)->Column#7 over(partition by test.first_range.p order by Column#6 range between 3 preceding and 0 following) |
> |     └─Sort_12                    | 17.00   | root      |                   | test.first_range.p, Column#6                                                                                                                |
> |       └─Projection_9             | 17.00   | root      |                   | test.first_range.p, test.first_range.o, test.first_range.v, <nil>->Column#6                                                                 |
> |         └─TableReader_11         | 17.00   | root      |                   | data:TableFullScan_10                                                                                                                       |
> |           └─TableFullScan_10     | 17.00   | cop[tikv] | table:first_range | keep order:false, stats:pseudo                                                                                                              |
> +----------------------------------+---------+-----------+-------------------+---------------------------------------------------------------------------------------------------------------------------------------------+
> ```
> 
> It looks like `Column#6` is an only null column const

I think we should raise error when the column is nullable as it's meaningless to calculate range frame with a nullable column.

tests/fullstack-test/mpp/window_range_frame.test
```diff
@@ -0,0 +1,471 @@
...
+mysql> insert into test.first_range (p, o, v) values (0, 0, 0), (1, 1, 1), (1, 2, 2), (1, 4, 4), (1, 8, 8), (2, 0, 0), (2, 3, 3), (2, 10, 10), (2, 13, 13), (2, 15, 15), (3, 1, 1), (3, 3, 3), (3, 5, 5), (3, 9, 9), (3, 15, 15), (3, 20, 20), (3, 31, 31);
+mysql> alter table first_range set tiflash replica 1;
+
+mysql> drop table if exists test.first_range_f;
+mysql> create table test.first_range_f(p int not null, o double not null, v int not null);
+mysql> insert into test.first_range_f (p, o, v) values (0, 0.1, 0), (1, 1.0, 1), (1, 2.1, 2), (1, 4.1, 4), (1, 8.1, 8), (2, 0.0, 0), (2, 3.1, 3), (2, 10.0, 10), (2, 13.1, 13), (2, 15.1, 15), (3, 1.1, 1), (3, 2.9, 3), (3, 5.1, 5), (3, 9.1, 9), (3, 15.0, 15), (3, 20.1, 20), (3, 31.1, 31);
+mysql> alter table first_range_f set tiflash replica 1;
+
+mysql> drop table if exists test.first_range_d;
+mysql> create table test.first_range_d(p int not null, o decimal(5,1) not null, v int not null);
```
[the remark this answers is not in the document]
xzhangxian1008: > Should cover all 4 sub-types of Decimal in TiFlash

okk

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
SeaRise: Will tidb report an error when executing such SQL?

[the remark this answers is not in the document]
SeaRise:
```
mysql> select *, first_value(v) over (partition by p order by (case 0 is null when true then o else null end) range between 3 preceding and 0 following) as a from test.first_range;
+---+----+----+---+
| p | o  | v  | a |
+---+----+----+---+
| 0 |  0 |  0 | 0 |
| 1 |  1 |  1 | 1 |
| 1 |  2 |  2 | 1 |
| 1 |  4 |  4 | 1 |
| 1 |  8 |  8 | 1 |
| 3 |  1 |  1 | 1 |
| 3 |  3 |  3 | 1 |
| 3 |  5 |  5 | 1 |
| 3 |  9 |  9 | 1 |
| 3 | 15 | 15 | 1 |
| 3 | 20 | 20 | 1 |
| 3 | 31 | 31 | 1 |
| 2 |  0 |  0 | 0 |
| 2 |  3 |  3 | 0 |
| 2 | 10 | 10 | 0 |
| 2 | 13 | 13 | 0 |
| 2 | 15 | 15 | 0 |
+---+----+----+---+
```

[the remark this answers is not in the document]
SeaRise: If you are sure not to consider the nullable column, please do not push such sql down to tiflash, and inform the user in warnings

dbms/src/Interpreters/WindowDescription.cpp
```diff
@@ -58,9 +60,11 @@ void WindowDescription::setWindowFrame(const tipb::WindowFrame & frame_)
     frame.begin_offset = frame_.start().offset();
     frame.begin_type = getBoundaryTypeFromTipb(frame_.start());
     frame.begin_preceding = (frame_.start().type() == tipb::WindowBoundType::Preceding);
+    frame.begin_cmp_data_type = frame_.start().cmp_data_type();
```
SeaRise, line 63: What is the relationship between `begin_cmp_data_type`, `begin_aux_col_type` and `order_by_col_type`?

dbms/src/DataStreams/WindowBlockInputStream.cpp
SeaRise: I am a little confused, decimal should be the parent-type of float, will converting decimal to float not lose precision?

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -360,6 +468,248 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToFrameEnd(const RowNumbe
     return std::make_tuple(frame_end_row, true);
 }
 
+void WindowTransformAction::stepToFrameStartWithOffsetBoundry()
+{
+    if (window_description.frame.type == WindowFrame::FrameType::Rows)
```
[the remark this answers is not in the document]
xzhangxian1008: > And `WindowFrame::FrameType::Groups`

I think WindowFrame::FrameType::Groups is only used in ClickHouse.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -25,8 +36,102 @@ namespace ErrorCodes
 {
 extern const int BAD_ARGUMENTS;
 extern const int NOT_IMPLEMENTED;
+extern const int CANNOT_INSERT_NULL_IN_ORDINARY_COLUMN;
```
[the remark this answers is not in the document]
xzhangxian1008: > why need this?

It seems useless, I will delete it.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -12,8 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <Columns/ColumnDecimal.h>
+#include <Columns/ColumnNullable.h>
+#include <Columns/ColumnVector.h>
+#include <Columns/ColumnsNumber.h>
```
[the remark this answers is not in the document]
xzhangxian1008: > Useless include

I will delete it.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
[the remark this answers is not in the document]
windtalker: Looks like mysql does not throw error if the order by column is nullable or even contains null value, since both TiDB and MySQL does not throw error here, I think TiFlash should not throw error, we can either support it or as @SeaRise suggests, "do not push to tiflash and inform the user in warnings"

dbms/src/DataStreams/WindowBlockInputStream.cpp
windtalker: I think it should depends on MySQL/TiDB's behavior?

SeaRise (approved):

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,119 @@ extern const int BAD_ARGUMENTS;
...
+        // Two operand must be integer
+        if constexpr (std::is_integral_v<OrderByColType> && std::is_integral_v<AuxColType>)
+        {
+            if constexpr (std::is_unsigned_v<OrderByColType> && std::is_unsigned_v<AuxColType>)
+                return isInRangeIntImpl<UInt64, UInt64, is_start, is_desc>(current_row_aux_value, cursor_value);
+            return isInRangeIntImpl<Int64, Int64, is_start, is_desc>(current_row_aux_value, cursor_value);
+        }
+        throw Exception("Unexpected Data Type!");
+    }
+    else if (CmpDataType == tipb::RangeCmpDataType::Float)
```
SeaRise, line 139:
```suggestion
    else if constexpr (CmpDataType == tipb::RangeCmpDataType::Float)
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,119 @@ extern const int BAD_ARGUMENTS;
...
+template <typename T>
+struct ActualCmpDataType
+{
+    using Type = std::conditional_t<checkIfSimpleNumericType<T>(), T, DecimalField<T>>;
+};
+
+template <typename T>
+typename ActualCmpDataType<T>::Type getValue(const ColumnPtr & col_ptr, size_t idx)
+{
+    return (*col_ptr)[idx].get<typename ActualCmpDataType<T>::Type>();
```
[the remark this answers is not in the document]
xzhangxian1008: > Since all the possible column is actually using `PaddedPODArray` to store the data, can we avoid using `Field` for row level calculation?

Yes, but it will introduce many complexity for the current codes, and after discussion with SeaRise we think it's ok to use `Field` so far.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,119 @@ extern const int BAD_ARGUMENTS;
...
+}
+
+template <typename AuxColType, typename OrderByColType, bool is_start, bool is_desc>
+bool isInRangeDecimalImpl(AuxColType current_row_aux_value, OrderByColType cursor_value)
+{
+    if constexpr (checkIfDecimalFieldType<AuxColType>() && checkIfDecimalFieldType<OrderByColType>())
+        return isInRangeCommonImpl<AuxColType, OrderByColType, is_start, is_desc>(current_row_aux_value, cursor_value);
+
+    // When comparing Decimal and Int, we convert them to the Float first.
+    Float64 current_row_aux_value_float64;
```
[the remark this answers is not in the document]
xzhangxian1008: > Both `Decimal` and `Int` are acurate type, why convert them to in-acurate type before compare? Is MySQL/TiDB use the same compare stragety?

Because we need to reserve the fraction part of decimal before comparing and the accuracy loss of `Decimal` will not affect the result I think.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -27,6 +37,119 @@ extern const int BAD_ARGUMENTS;
...
+        if constexpr (std::is_integral_v<OrderByColType> && std::is_integral_v<AuxColType>)
+        {
+            if constexpr (std::is_unsigned_v<OrderByColType> && std::is_unsigned_v<AuxColType>)
+                return isInRangeIntImpl<UInt64, UInt64, is_start, is_desc>(current_row_aux_value, cursor_value);
+            return isInRangeIntImpl<Int64, Int64, is_start, is_desc>(current_row_aux_value, cursor_value);
+        }
+        throw Exception("Unexpected Data Type!");
+    }
+    else if (CmpDataType == tipb::RangeCmpDataType::Float)
+    {
```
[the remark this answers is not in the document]
xzhangxian1008: > Why not check if both type is floating point type here?

okk

tests/fullstack-test/mpp/window_range_frame.test
```diff
@@ -0,0 +1,644 @@
...
+| 3 | 20.1 | 20 | 20 |
+| 3 | 31.1 | 31 | 31 |
+| 2 |  0.0 |  0 |  0 |
+| 2 |  3.1 |  3 |  3 |
+| 2 | 10.0 | 10 | 10 |
+| 2 | 13.1 | 13 | 15 |
+| 2 | 15.1 | 15 | 15 |
++---+------+----+----+
+
+mysql> use test; set tidb_enforce_mpp=1; select *, last_value(v) over (partition by p order by o range between 0 preceding and 2.3 following) as a from test.last_range_d256;
```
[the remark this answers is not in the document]
xzhangxian1008: > Add some tests to cover interval literal based range frame

okk

dbms/src/DataStreams/WindowBlockInputStream.cpp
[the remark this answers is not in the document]
windtalker: I think you should first do some investigations on how TiFlash compare decimal value and int value, the accuracy loss is possible because double type only has 15 to 17 significant decimal digits precision, for example:
```
mysql> select value, cast(value as double), cast(value as double) + 1, cast(value as double) > cast(value as double) + 1, cast(value as double) < cast(value as double) + 1, cast(value as double) = cast(value as double) + 1 from t;
+---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
| value               | cast(value as double) | cast(value as double) + 1 | cast(value as double) > cast(value as double) + 1 | cast(value as double) < cast(value as double) + 1 | cast(value as double) = cast(value as double) + 1 |
+---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
| 9223372036854775801 |  9.223372036854776e18 |      9.223372036854776e18 |                                                 0 |                                                 0 |                                                 1 |
+---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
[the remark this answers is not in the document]
xzhangxian1008: > Why not check if both type is floating point type here?

Only one of the type is float is permitted here

dbms/src/DataStreams/WindowBlockInputStream.cpp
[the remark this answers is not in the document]
xzhangxian1008: > I think you should first do some investigations on how TiFlash compare decimal value and int value, the accuracy loss is possible because double type only has 15 to 17 significant decimal digits precision, for example:
> 
> ```
> mysql> select value, cast(value as double), cast(value as double) + 1, cast(value as double) > cast(value as double) + 1, cast(value as double) < cast(value as double) + 1, cast(value as double) = cast(value as double) + 1 from t;
> +---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
> | value               | cast(value as double) | cast(value as double) + 1 | cast(value as double) > cast(value as double) + 1 | cast(value as double) < cast(value as double) + 1 | cast(value as double) = cast(value as double) + 1 |
> +---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
> | 9223372036854775801 |  9.223372036854776e18 |      9.223372036854776e18 |                                                 0 |                                                 0 |                                                 1 |
> +---------------------+-----------------------+---------------------------+---------------------------------------------------+---------------------------------------------------+---------------------------------------------------+
> ```

fixed

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -28,6 +40,232 @@ extern const int BAD_ARGUMENTS;
...
+        else if constexpr (checkIfDecimalFieldType<RightType>())
+        {
+            right_scale = right.getScale();
+            return DecimalComparison<LeftType, typename RightType::DecimalType, LessOrEqualsOp>::compare(
+                left,
+                right.getValue(),
+                left_scale,
+                right_scale);
+        }
+    }
```
SeaRise, line 93:
```suggestion
    else if constexpr (checkIfDecimalFieldType<LeftType>())
    {
        return DecimalComparison<typename LeftType::DecimalType, RightType, LessOrEqualsOp>::compare(
            left.getValue(),
            right,
            left.getScale(),
            0);
    }
    else if constexpr (checkIfDecimalFieldType<RightType>())
    {
        return DecimalComparison<LeftType, typename RightType::DecimalType, LessOrEqualsOp>::compare(
            left,
            right.getValue(),
            0,
            right.getScale());
    }
```

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -28,6 +40,232 @@ extern const int BAD_ARGUMENTS;
...
+                   Decimal32>> || std::is_same_v<T, DecimalField<Decimal64>> || std::is_same_v<T, DecimalField<Decimal128>> || std::is_same_v<T, DecimalField<Decimal256>>;
+}
+
+template <typename LeftType, typename RightType>
+bool lessEqual(LeftType left, RightType right)
+{
+    if constexpr (checkIfDecimalFieldType<LeftType>() && checkIfDecimalFieldType<RightType>())
+    {
+        // This `if` branch is necessary or the next branch's codes will be generated
+        // when these two types are all DecimalField type which will cause compilation errors.
```
SeaRise, line 66: seems useless comments

SeaRise (approved):
Others LGTM

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -28,6 +40,232 @@ extern const int BAD_ARGUMENTS;
...
+        return left <= right;
+    }
+    else if constexpr (checkIfDecimalFieldType<LeftType>() || checkIfDecimalFieldType<RightType>())
+    {
+        // Only one type will be Decimal as Decimal vs Decimal can be handled in the `else` branch
```
windtalker, line 71: The comment should be `as Decimal vs Decimal is handled before`?

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -28,6 +40,232 @@ extern const int BAD_ARGUMENTS;
...
+{
+    return (*col_ptr)[idx].get<typename ActualCmpDataType<T>::Type>();
+}
+
+template <typename T, typename U, bool is_preceding, bool is_desc, bool is_begin>
+bool isInRangeCommonImpl(T current_row_aux_value, U cursor_value)
+{
+    if constexpr (is_begin)
+    {
+        if constexpr ((is_preceding && is_desc) || (!is_preceding && is_desc))
```
windtalker, line 159: I don't get why `is_preceding` is needed here? Looks like `((is_preceding && is_desc) || (!is_preceding && is_desc))` is the same as `is_desc`?

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -120,6 +109,63 @@ struct WindowTransformAction
...
+    std::tuple<RowNumber, bool> stepToEndForRangeFrameOrderCase();
+
+    template <typename T, bool is_desc>
+    RowNumber stepToStartForRangeImpl();
+
+    template <typename T, bool is_desc>
+    RowNumber stepToEndForRangeImpl();
+
+    template <typename T, bool is_begin, bool is_desc>
+    RowNumber stepForRangeImpl();
```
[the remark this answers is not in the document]
xzhangxian1008: > change `Range` to `RangeFrame` for the 3 functions above

okk

dbms/src/DataStreams/WindowBlockInputStream.h
```diff
@@ -120,6 +109,63 @@ struct WindowTransformAction
...
+
+    // We should use this function when the current auxiliary column row is null.
```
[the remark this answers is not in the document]
xzhangxian1008: > when `current auxiliary column row` is null or `current order column row` is null?

order column, I will change it.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > 

The code hasn't been updated when reviewing, and the comment here has been deleted.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > I don't get why `is_preceding` is needed here? Looks like `((is_preceding && is_desc) || (!is_preceding && is_desc))` is the same as `is_desc`?

From a logical perspective, which if-branch we choose to execute is decided by `is_preceding` and `is_desc`, so it's easier to understand the purpose of this if-condition if we use both `is_preceding` and `is_desc`

xzhangxian1008: changed

windtalker (approved):
LGTM

Issue #7568, opened 2023-05-30
[dynamic regions] tiflash panic due to data error


dbsid:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
tiup bench ch -H 172.16.5.140 -P 4000 -D tpcc prepare
```

during data preparation, enable tiflash replicas
```
ALTER DATABASE tpcc SET TIFLASH REPLICA 2;
```

### 2. What did you expect to see? (Required)
no tiflash crash

### 3. What did you see instead (Required)
tiflash crash

```
[...]

PR #7657, opened 2023-06-14
Handle orphan write cf key from a raftstore v2 tablet snapshot

CalvinNeo:

Issue Number: ref #7568

Problem Summary:

Raftstore v2 snapshot may contain extra write key which is from later raft log index than the snapshot has claimed. We decide to suppress this error rather than throwing an exception.

We will keep track of these orphan key, and throw an exception later if we are sure these keys are orphan because of data loss rather than an eager-replicated data.

The suppress part:
1. Deduce TiKV cluster's raftstore version by the first `/engine_type` response from TiKV.
    1. Note that TiKV later supports coexistence of v1 and v2 node in a cluster. After that, they can provide us a mechanism to fix this.
3. Store all orphan write key in `OrphanKeyInfo` struct when prehandling snapshot.
4. Remove a key from `OrphanKeyInfo` when insert into write cf.
5. When meet orphan key in a row2col job which is not triggered by prehandling snapshot, check `OrphanKeyInfo` first. If the key is registered in `OrphanKeyInfo`, consider it as a soft error.

The async checking part:
1. Log a `deadline_index` sent from TiKV's snapshot which must GE than `snapshot_index` in in `OrphanKeyInfo`.
2. If the `applied_index` is proceeded exceed the `deadline_index`, check if the `OrphanKeyInfo` is cleared. Otherwise raise a hard error.

What is not included:
1. `PreHandleSnapshot` requires `deadline_index` being sent from TiKV. So we can have async check only after TiKV support this.
2. Although, we have this PR merged first to avoid blocking test.

**NOTE**
Orphan keys will not be iterated by region read scanner, so it will not be presented in `data_list_read`, so it will not be removed as a committed entry. We don't eagerly do that, since it brings extra cost. These keys will either be removed when its default counterpart arrives in the following raft log, or being reported as an error when exceeds deadline_index.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
    + Tested the cluster version part
    ```
    [2023/06/12 12:41:25.170 +08:00] [INFO] [run.rs:147] ["start probing cluster's raftstore version"]
    [2023/06/12 12:41:25.267 +08:00] [INFO] [run.rs:150] ["cluster's raftstore version is V2"]
    [2023/06/12 14:14:17.974 +08:00] [INFO] [run.rs:147] ["start probing cluster's raftstore version"]
    [2023/06/12 14:14:18.072 +08:00] [INFO] [run.rs:150] ["cluster's raftstore version is V1"]
    ```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Handle orphan write cf key from a raftstore v2 tablet snapshot
```


Lloyd-Pottiger:
How about converting to draft before ready to avoid seeds messages to reviewers?

dbms/src/Storages/Transaction/RegionData.cpp
```diff
@@ -133,13 +132,39 @@ RegionDataReadInfo RegionData::readDataByWriteIt(const ConstWriteCFIter & write_
...
+                }
+                else
+                {
+                    // We can't delete this orphan key here, since it can be triggered from `onSnapshot`.
+                    if (orphan_keys_info.containsExtraKey(*key))
+                    {
+                        return std::nullopt;
+                    }
+                    // Otherwise, this is still a hard error.
+                    // TODO We still need to check if there are remained orphan keys after we have applied after peer's flushed_index.
```
CalvinNeo, line 154: This will be done when tikv support a deadline_index.

dbms/src/Storages/Transaction/tests/gtest_kvstore.cpp
```diff
@@ -336,17 +335,13 @@ void RegionKVStoreTest::testRaftSplit(KVStore & kvs, TMTContext & tmt)
         ASSERT_EQ(kvs.getRegion(1)->dataInfo(), "[write 1 lock 1 default 1 ]");
         ASSERT_EQ(kvs.getRegion(7)->dataInfo(), "[lock 1 ]");
     }
+    return;
```
JaySon-Huang, line 338: the following test code are skipped?

dbms/src/Storages/Transaction/ApplySnapshot.cpp
```diff
@@ -286,18 +287,22 @@ std::vector<DM::ExternalDTFileInfo> KVStore::preHandleSnapshotToFiles(
     const SSTViewVec snaps,
     uint64_t index,
     uint64_t term,
+    std::optional<uint64_t> deadline_index,
```
JaySon-Huang, line 290: Add some comments about the `deadline_index` before the function?

dbms/src/Storages/Transaction/SSTReader.cpp
```diff
@@ -62,7 +64,7 @@ MonoSSTReader::MonoSSTReader(const TiFlashRaftProxyHelper * proxy_helper_, SSTVi
     {
         auto && r = range->comparableKeys();
         auto start = r.first.key.toString();
-        LOG_DEBUG(log, "Seek cf {} to {}", static_cast<std::underlying_type<decltype(type)>::type>(type), Redact::keyToDebugString(start.data(), start.size()));
+        LOG_DEBUG(log, "Seek cf {} to {}", magic_enum::enum_name(type), Redact::keyToDebugString(start.data(), start.size()));
```
JaySon-Huang, line 67: also clean this
https://github.com/pingcap/tiflash/blob/b9899e9e302b4ad9b6aa4828f166fb9ec111530f/dbms/src/Storages/Transaction/SSTReader.h#L50-L54

dbms/src/Storages/Transaction/tests/gtest_kvstore.cpp
CalvinNeo: fixed

dbms/src/Storages/Transaction/SSTReader.cpp
CalvinNeo: fixed

dbms/src/Storages/Transaction/RegionScanner.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+bool Region::CommittedScanner::hasNext()
+{
+    if (peeked)
+    {
+        return true;
+    }
+    else
+    {
+        if (!write_map_size)
```
JaySon-Huang, line 48:
```suggestion
        if (write_map_size == 0)
```

dbms/src/Storages/Transaction/RegionScanner.cpp
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+        return false;
+    }
+}
+
+bool Region::CommittedScanner::tryNext()
+{
+    RUNTIME_CHECK_MSG(!peeked.has_value(), "Can't call CommittedScanner::tryNext() twice");
+    RUNTIME_CHECK(write_map_it != write_map_it_end);
```
JaySon-Huang, line 64: As `tryNext` is only called by `hasNext`, I think we can use `assert` to avoid the overhead in release mode
```suggestion
    assert(write_map_it != write_map_it_end);
```

dbms/src/Storages/Transaction/Region.cpp
```diff
@@ -497,6 +509,40 @@ ImutRegionRangePtr Region::getRange() const
...
+    }
+    return RaftstoreVer::Uncertain;
+}
+
+void Region::beforePrehandleSnapshot(uint64_t region_id, std::optional<uint64_t> deadline_index)
+{
+    if (getClusterRaftstoreVer() == RaftstoreVer::V2)
+    {
+        data.orphan_keys_info.snapshot_index = appliedIndex();
+        data.orphan_keys_info.pre_handling = true;
```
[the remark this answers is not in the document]
CalvinNeo: No, it marks that this region is prehandling a snapshot.

dbms/src/Storages/Transaction/TiKVKeyValue.h
```diff
@@ -46,6 +46,11 @@ struct StringObject : std::string
         : Base(str)
     {}
     static StringObject copyFrom(const Base & str) { return StringObject(str); }
+    template <bool B>
+    static StringObject<B> copyFrom(const StringObject<B> & other)
+    {
+        return StringObject(other.data(), other.dataSize());
+    }
```
JaySon-Huang, line 53: Do we really need to add this function? It accepts generating TiKVValue by copyFrom TiKVKey, which could be dangerous.

We can define it without the extra `bool B`. Or just calling `copyFrom(obj.getStr())`
```
    static StringObject copyFrom(const StringObject & other)
    {
        return StringObject(other.data(), other.dataSize());
    }
```

JaySon-Huang: Oh, I tried that even define it without the extra `bool B`, we can still copyFrom TiKVKey to TiKVValue :(

CalvinNeo: I get your point, let me try some sfinae magic

dbms/src/Storages/Transaction/Region.h
```diff
@@ -227,11 +230,10 @@ class Region : public std::enable_shared_from_this<Region>
 
 private:
     RegionData data;
-    mutable std::shared_mutex mutex;
-
     RegionMeta meta;
 
     LoggerPtr log;
+    mutable std::shared_mutex mutex;
```
JaySon-Huang, line 236: Does this mutex used to protect `RegionData`?

CalvinNeo: Yes, it protects RegionData and RegionMeta.

JaySon-Huang: then I think it should be defined next to those members. Why move it down to here?

CalvinNeo: I will move it back

CalvinNeo: fixed

lidezhu (approved):

Issue #6358, opened 2022-11-24
MPP Support warnings


SeaRise:
## Enhancement

- cop
![PdmLRbdSxQ](https://user-images.githubusercontent.com/24507159/203705853-2f9d2fc6-338f-46b3-b043-407067474fe8.jpg)
- mpp
![47trf3AiKo](https://user-images.githubusercontent.com/24507159/203705883-56be4c27-fb8f-4020-a50a-61075da6c819.jpg)

PR #7658, opened 2023-06-14
Add ReportMPPTaskStatus support in tiflash side

### LINKED ISSUE #7659: Allow TiFlash to report error and execution summaries directly to coordinator
## Enhancement
Related tidb issue: https://github.com/pingcap/tidb/issues/44032

yibin87:

Issue Number: close #7659 ref [tidb_issue](https://github.com/pingcap/tidb/issues/44032)

Problem Summary:

This PR aims to let mpp task reports to coordinator directly to refine existing error/execution_summary reporting mechanism:

1. When MppVersion >= 2 and coordinator_address in TaskMeta is not empty, execution summaries are not passed within mppDataPacket, they will be reported to coordinator directly. Thus execution summaries have little chance to loose in this way.
2. When MPPTask met exception during run, it will report the exception directly to coordinator, and avoid secondary error messages during exception delivering. 

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Use playground, test Compilation Failure case, wrong coordinator address case, and normal success case.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


yibin87:
hold until tidb aligned with tipb and kvproto version.

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -538,6 +547,47 @@ void MPPTask::runImpl()
...
+    if (!ReportStatusToCoordinator(meta.mpp_version(), meta.coordinator_address()))
+        return;
+
+    try
+    {
+        std::shared_ptr<mpp::ReportTaskStatusRequest> req = std::make_shared<mpp::ReportTaskStatusRequest>();
+        mpp::TaskMeta * req_meta = req->mutable_meta();
+        req_meta->CopyFrom(meta);
+
+        tipb::TiFlashExecutionInfo execution_info = mpp_task_statistics.genTiFlashExecutionInfo();
```
SeaRise, line 561: add `if (dag_context->collect_execution_summaries)` here?

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -516,14 +523,16 @@ void MPPTask::runImpl()
...
+        reportStatus(trimmed_err_msg);
         if (status == RUNNING)
         {
             LOG_ERROR(log, "task running meets error: {}", err_msg);
-            /// trim the stack trace to avoid too many useless information in log
-            trimStackTrace(err_msg);
             try
             {
-                handleError(err_msg);
+                handleError(trimmed_err_msg);
```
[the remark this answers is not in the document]
yibin87: Yes, the final error message is set in [handleError](https://github.com/pingcap/tiflash/blob/73731739ea00b6a5bc03872acc49c2eea8ff5b8e/dbms/src/Flash/Mpp/MPPTask.cpp#L591), which prefix the message by "From %id ". However, this process will trigger TaskManager to abort other  mpp tasks with the same query id. And personally, I think users may not case about "From %id" info.

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: okk

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -538,6 +547,50 @@ void MPPTask::runImpl()
     unregisterTask();
 }
 
+void MPPTask::reportStatus(const String & err_msg)
```
SeaRise, line 550: It seems that warnings can also be passed through coodinator, ref https://github.com/pingcap/tiflash/issues/6358

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -538,6 +547,50 @@ void MPPTask::runImpl()
...
+        mpp::TaskMeta * req_meta = req->mutable_meta();
+        req_meta->CopyFrom(meta);
+
+        if (dag_context->collect_execution_summaries)
+        {
+            tipb::TiFlashExecutionInfo execution_info = mpp_task_statistics.genTiFlashExecutionInfo();
+            if unlikely (!execution_info.SerializeToString(req->mutable_data()))
+            {
+                LOG_ERROR(log, "Failed to serialize TiFlash execution info");
+                return;
```
SeaRise, line 567: Maybe it shouldn't return here?
Because the following steps such as err_msg may not fail

dbms/src/Flash/Mpp/MPPTask.cpp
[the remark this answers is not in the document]
SeaRise: err_msg is usually an error caused by an upstream/downstream mpptask failure. So the err_msg passed to the coodinator here is usually meaningless.
getErrString is usually the error thrown by the earliest failed mpptask in the current tiflash instance, and the probability is not a secondary error.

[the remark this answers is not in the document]
SeaRise: If the coodinator gets a lot of meaningless error reports, how to filter out the correct error reports and pass them on to the user?
Does it depend on time?

dbms/src/Flash/Mpp/MPPTask.cpp
SeaRise: Or adding todo here?

yibin87: It's ok to pass warnings to coordinator, since the err_msg will take effect only when the root task returned error messages in tidb side.

dbms/src/Flash/Mpp/MPPTask.cpp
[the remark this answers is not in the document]
yibin87: Yes, tidb side only takes the first error message when any root task returned error.

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: The err_msg path is a better-do mechnism, so when it failed, something unexpected happened, no err msg returned seems ok.

SeaRise: One question: Will the tidb side wait for all tiflash mpptask report status?

dbms/src/Flash/Mpp/MPPTask.cpp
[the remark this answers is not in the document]
SeaRise: Does it mean that the err_msg of the non-root task will be discarded directly?.

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: ReportMPPTaskStatus is introduced to solve the error and cancel-by-limit case, and it is not guaranteed tidb side will received all reports before mpp_gather closed. I prefer to add warnings in MppDataPacket, since warning message was not that important when error happened.

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: Not for all queries, for query with limit operator, it will.

dbms/src/Flash/Mpp/MPPTask.cpp
SeaRise: I'm talking about the warnings in show warnings. I feel like adding a TODO or processing it in the same way as the summary.

dbms/src/Flash/Mpp/MPPTask.cpp
SeaRise: If directly return here, will it cause tidb to wait forever for the limit task to report the status?

SeaRise (commented):
Others LGTM

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: Yeah, I can add a TODO like "when MppDataPacket include warning messages info, support it in report status also"

dbms/src/Flash/Mpp/MPPTask.cpp
yibin87: No, tidb will wait no more than a statically set thres time.

SeaRise: ok

dbms/src/Flash/Mpp/MPPTask.cpp
SeaRise: ok

SeaRise (approved):

dbms/src/Flash/Statistics/ExecutorStatisticsCollector.cpp
```diff
@@ -142,6 +142,15 @@ tipb::SelectResponse ExecutorStatisticsCollector::genExecutionSummaryResponse()
     return response;
 }
 
+tipb::TiFlashExecutionInfo ExecutorStatisticsCollector::genTiFlashExecutionInfo()
+{
+    tipb::SelectResponse response = genExecutionSummaryResponse();
```
windtalker, line 147: `genExecutionSummaryResponse` will also generate remote execution summary, which in this case will be all zero, is that expected, or do we need these zero execution summaries?

yibin87: Yes, since remote executions won't use DispatchMPPTask interface, its execution summary needs to be included here now if has.

windtalker (approved):
LGTM

PR #7668, opened 2023-06-16
improve ps gc speed by dumping snapshot from memory directly

lidezhu:

Issue Number: ref #6827 

Problem Summary: Previously, when dumping snapshot for `PageDirectory`, we load all target wal files and create a temp PageDirectory for dumping. And after dumping the snapshot, we can reclaim all the target wal files.
But under heavy write scenario, the dump snapshot speed is a little slow. This will make the number of wal files keep increasing and make the process of dumping snapshot cost more and more memory.

Change the type of `being_ref_count` from `Int64` to `MultiVersionRefCount *`;
1. If `being_ref_count` is nullptr, it means `1`;
2. We maintain a map `PageVersion -> RefCount` inside `MultiVersionRefCount`;
3. When creating a ref page id, increasing the ref by calling `getLatestRefCount` and `appendRefCount`;
4. When delete a ref in gc process, deleting the ref by calling `decrLatestRefCountInSnap`. And it will also reclaim older version of ref count;

When dumping snapshot, ref count increasing caused by new ref page will just be appended to `MultiVersionRefCount`, so it won't affect the ref count value that the snapshot can see.
GC process can still change ref count value for a snapshot, but it doesn't matter because gc won't run concurrently with dumping snapshot.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,48 +82,170 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+            return 1;
+        }
+        RUNTIME_CHECK(!container->ref_map.empty());
+        return container->ref_map.rbegin()->second;
+    }
+
+#ifndef DBMS_PUBLIC_GTEST
+private:
+#endif
+    std::map<PageVersion, Int64> ref_map;
```
JaySon-Huang, line 189: using a `std::unique_ptr<std::map<PageVersion, Int64>>` maybe more exception safe

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,48 +82,170 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+    static MultiVersionRefCount * createFrom(const PageVersion & ver, Int64 ref_count)
+    {
+        if (ref_count == 1)
+        {
+            return nullptr;
+        }
+        return appendRefCount(nullptr, ver, ref_count);
+    }
+
+    static MultiVersionRefCount * copy(MultiVersionRefCount * container)
```
JaySon-Huang, line 99: why define all functions as static instead of using member functions?

lidezhu: Because the object will just created when ref happend. In other cases it is just a nullptr.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: It is just a member, why use unique_ptr can make it more exception safe? Or do you mean to be careful with the exception when creating raw pointer of `MultiVersionRefCount`?

dbms/src/Storages/Page/V3/PageDirectory.h
flowbehappy: How about we make `MultiVersionRefCount` an object which contains `{}` of `std::unique_ptr<std::map<PageVersion, Int64>>` by default?  So that we don't need to use static method.

flowbehappy: BTW `std::unique_ptr` has no size overhead.

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,48 +82,172 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+        }
+        return new_container;
+    }
+
+    static void destroy(MultiVersionRefCount * container)
+    {
+        delete container;
+    }
+
+    static void decrLatestRefCountInSnap(MultiVersionRefCount * container, UInt64 snap_seq, Int64 deref_count)
```
flowbehappy, line 119: I wonder do we really need to literate all the versions and decrease ref_count? How about make `ref_map` an update history instead of update result. I.e. We postpone the final ref_count of each version to the actual call of `getRefCountInSnap` and `getLatestRefCount`. So that `decrLatestRefCountInSnap` will be much simpler.

flowbehappy: And how about making `ref_map`  an `std::vector` instead of `std::map`. Because `vector` is faster and size efficient than `map`. And for the rare case that we need to do random update and truncate smaller versions, just create an new `vector` and do copy.

dbms/src/Storages/Page/V3/PageDirectory.h
JaySon-Huang: As all methods are static, we don't have a deconstructor for `MultiVersionRefCount `. We were possibly not calling `MultiVersionRefCount::destroy` when the exception happened. It will cause a memory leak.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: To maintain an update history will just make the code a lot complicated:
Consider the following operations:
1. put 1 at version 10
2. ref 2 -> 1 at version 11
3. ref 3 -> 1 at version 12
And now the version history becomes
```
version -> ref count
10 -> 1
11 -> 2
12 -> 3
```
Operations continue:
1. del 3 at version 13
2. del 2 at version 14
Then gc happens:
1. deref 2
2. deref 3
After the first deref op happens, the version history becomes
```
10 -> 1
11 -> 2
12 -> 3
14 -> 2
```
Then after the second deref op, the version history should become:
```
10 -> 1
11 -> 2
12 -> 3
13 -> 2
14 -> 1
```
Note the ref count of **version 14 changes from 2 to 1** after the second deref which make it essentially not an update history.
The reason why this problem happens is that if we deref at gc time, the delete operations of ref id doesn't come at version order.

And then a natural idea is to deref at write time. But because when do full gc, we may rewrite a ref id to a normal page id, so we will still meet a similar situation like the above at gc time.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: Avoid static method now.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,134 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+            }
+            RUNTIME_CHECK(ref_count_delta_in_snap > 0);
+            new_versioned_ref_counts->emplace_back(PageVersion(snap_seq, 0), ref_count_delta_in_snap);
+            versioned_ref_counts.swap(new_versioned_ref_counts);
+        }
+    }
+
+    Int64 getRefCountInSnap(UInt64 snap_seq) const
+    {
+        if (versioned_ref_counts == nullptr)
```
flowbehappy, line 163: Minor issue.
```suggestion
        if (!versioned_ref_counts)
        // or if (!versioned_ref_counts == {})
```
So that the compiler don't need to transform `nullptr` to `std:unique_ptr` instance and do comperation.

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,134 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+        {
+            versioned_ref_counts = std::make_unique<std::vector<std::pair<PageVersion, Int64>>>();
+        }
+        versioned_ref_counts->emplace_back(ver, ref_count_delta);
+    }
+
+    void decrRefCountInSnap(UInt64 snap_seq, Int64 deref_count_delta)
+    {
+        if (snap_seq == 0)
```
flowbehappy, line 129: Please add document to explain when will `snap_seq` be `0`;

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,134 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+                    new_versioned_ref_counts->emplace_back(ver, ref_count_delta);
+                }
+            }
+            if (ref_count_delta_in_snap == 0 && new_versioned_ref_counts->empty())
+            {
+                versioned_ref_counts = nullptr;
+                return;
+            }
+            RUNTIME_CHECK(ref_count_delta_in_snap > 0);
+            new_versioned_ref_counts->emplace_back(PageVersion(snap_seq, 0), ref_count_delta_in_snap);
```
flowbehappy, line 156: I don't understand why `emplace_back`?  I think it will cause the versions in `versioned_ref_counts` not sorted by increment.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: Added.

dbms/src/Storages/Page/V3/PageDirectory.h
lidezhu: `versioned_ref_counts` doesn't need to be sorted because we almost alway traverse all the elements in it.

flowbehappy: ok

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,135 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+                {
+                    new_versioned_ref_counts->emplace_back(ver, ref_count_delta);
+                }
+            }
+            if (ref_count_delta_in_snap == 0 && new_versioned_ref_counts->empty())
+            {
+                versioned_ref_counts = nullptr;
+                return;
+            }
+            RUNTIME_CHECK(ref_count_delta_in_snap > 0);
```
JaySon-Huang, line 156:
```suggestion
            RUNTIME_CHECK(ref_count_delta_in_snap > 0, deref_count_delta, ref_count_delta_in_snap);
```

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,135 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+            new_versioned_ref_counts->emplace_back(PageVersion(snap_seq, 0), ref_count_delta_in_snap);
```
JaySon-Huang, line 157: This append `{ver=<snap_seq,0>, count=ref_count_delta_in_snap}` to the end. But there could be ver.sequnce > snap_seq appended into `new_versioned_ref_counts` in line 148.
So the pairs in `versioned_ref_counts` are not sorted by versions ?

lidezhu: Yes, `versioned_ref_counts` doesn't need to be sorted because we almost alway traverse all the elements in it.

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -82,25 +82,135 @@ class PageDirectorySnapshot : public DB::PageStorageSnapshot
...
+    MultiVersionRefCount(const MultiVersionRefCount & other)
+    {
+        if (other.versioned_ref_counts)
+        {
+            versioned_ref_counts = std::make_unique<std::vector<std::pair<PageVersion, Int64>>>();
+            for (const auto & [ver, ref_count] : *(other.versioned_ref_counts))
+            {
+                versioned_ref_counts->emplace_back(ver, ref_count);
+            }
+        }
```
JaySon-Huang, line 101: there should be a else branch

```
else
{
    versioned_ref_counts = nullptr
}
```

lidezhu: Fixed.

JaySon-Huang (approved):
LGTM

PR #7693, opened 2023-06-25
fix active_set bug in min tso scheduler

### LINKED ISSUE #7692: `active_set_soft_limit` in MinTso scheduler does not work as expected.
## Bug Report

`active_set_soft_limit` is expected to limit the active query count even if `thread_soft_limit` is not reached.
For example, if `active_set_soft_limit` is set to 10, and currently, the there is 10 running queries, then any new comming query should not be scheduled except that the new comming query is a query with `min_tso`.
But in current implementation:
https://github.com/pingcap/tiflash/blob/4ecf3cd7649f0597c540cdbfe4ea609f26d7d169/dbms/src/Flash/Mpp/MinTSOScheduler.cpp#L193
Even if `active_set_soft_limit` is already reached, the schedule will still allow new task to running if the tso of current task is less than the `max_tso` in current active set(`active_set.size() < active_set_soft_limit || query_id <= *active_set.rbegin()`).

windtalker:

Issue Number: close #7692

Problem Summary:

If `active_set_soft_limit`, only schdule the non min-tso task if the task belongs to an active query.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Mpp/MinTSOScheduler.cpp
+++ b/dbms/src/Flash/Mpp/MinTSOScheduler.cpp
@@ -190,7 +190,7 @@ bool MinTSOScheduler::scheduleImp(const MPPQueryId & query_id, const MPPQueryTas
 {
     auto needed_threads = schedule_entry.getNeededThreads();
     auto check_for_new_min_tso = query_id <= min_query_id && estimated_thread_usage + needed_threads <= thread_hard_limit;
-    auto check_for_not_min_tso = (active_set.size() < active_set_soft_limit || query_id <= *active_set.rbegin()) && (estimated_thread_usage + needed_threads <= thread_soft_limit);
+    auto check_for_not_min_tso = (active_set.size() < active_set_soft_limit || active_set.find(query_id) != active_set.end()) && (estimated_thread_usage + needed_threads <= thread_soft_limit);
     if (check_for_new_min_tso || check_for_not_min_tso)
     {
         updateMinQueryId(query_id, false, isWaiting ? "from the waiting set" : "when directly schedule it");


SeaRise (approved):
Maybe add unit tests for MinTsoScheduler, in ComputeServerTest.
Add it or not, I think it's fine

yibin87 (approved):
LGTM

windtalker:
> Maybe add unit tests for MinTsoScheduler, in ComputeServerTest. Add it or not, I think it's fine

I've added a ut in ComputeServerTest.

dbms/src/Flash/tests/gtest_compute_server.cpp
```diff
@@ -849,6 +872,67 @@ try
...
+    context.context->setSetting("task_scheduler_active_set_soft_limit", active_set_soft_limit);
+    startServers(1);
+    setCancelTest();
+    ASSERT_TRUE(TiFlashMetrics::instance().tiflash_task_scheduler.get(tiflash_task_scheduler_metrics::type_active_queries_count).Value() == 0);
+    std::vector<std::thread> running_queries;
+    std::vector<MPPQueryId> query_ids;
+    try
+    {
+        /// case 1, min tso can be added
+        for (size_t i = 0; i < active_set_soft_limit; i++)
```
SeaRise, line 888:
```suggestion
        for (size_t i = 0; i < active_set_soft_limit; ++i)
```

dbms/src/Flash/tests/gtest_compute_server.cpp
```diff
@@ -849,6 +872,67 @@ try
...
+        std::this_thread::sleep_for(2s);
+        ASSERT_TRUE(TiFlashMetrics::instance().tiflash_task_scheduler.get(tiflash_task_scheduler_metrics::type_active_queries_count).Value() == 3);
+        for (const auto & query_id : query_ids)
+            MockComputeServerManager::instance().cancelQuery(query_id);
+        for (auto & t : running_queries)
+            t.join();
+        running_queries.clear();
+        query_ids.clear();
+        /// case 2, non-min tso can't be added
+        for (size_t i = 0; i < active_set_soft_limit; i++)
```
SeaRise, line 905: ditto

dbms/src/Flash/tests/gtest_compute_server.cpp
```diff
@@ -849,6 +872,67 @@ try
...
+        for (size_t i = 1; i < running_queries.size(); i++)
+            running_queries[i].join();
+    }
+    catch (...)
+    {
+        for (const auto & query_id : query_ids)
+            MockComputeServerManager::instance().cancelQuery(query_id);
+        for (auto & t : running_queries)
+            if (t.joinable())
+                t.join();
```
SeaRise, line 931: Is catch exception expected?

dbms/src/Flash/tests/gtest_compute_server.cpp
```diff
@@ -849,6 +872,67 @@ try
...
+        running_queries.clear();
+        query_ids.clear();
+        /// case 2, non-min tso can't be added
+        for (size_t i = 0; i < active_set_soft_limit; i++)
+        {
+            addOneQuery((i + 1) * 20, running_queries, query_ids);
+        }
+        using namespace std::literals::chrono_literals;
+        std::this_thread::sleep_for(2s);
+        ASSERT_TRUE(TiFlashMetrics::instance().tiflash_task_scheduler.get(tiflash_task_scheduler_metrics::type_active_queries_count).Value() == 2);
```
SeaRise, line 911: And check waiting query count?

dbms/src/Flash/tests/gtest_compute_server.cpp
windtalker: Catch exception here is to avoid destruct of non-joined thread crash the process, I will rethrow the exception after all threads are joined.

dbms/src/Flash/tests/gtest_compute_server.cpp
windtalker: Done

SeaRise (approved):

PR #7719, opened 2023-06-29
store raft data in dedicated blob file

lidezhu:

Issue Number: ref #6827 

Problem Summary: Currently we mix raft related data and other data in the same set of blob files. But raft related data have the characteristic of writing and deleting in a short time. So it is not suitable for full gc. And we need to separate them in a dedicated set of blob files.

Introduce `PageType` to distinguish different kinds of workload data and we can also specify different config for them. Different kinds of `PageType` will be stored in blob files with different ids. It can support at most `PageType::TypeCountLimit` kinds of workloads.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/Blob/BlobStat.h
```diff
@@ -118,6 +119,39 @@ class BlobStats
...
+#else
+    public:
+#endif
+        constexpr static BlobFileId FIRST_RAFT_FILE_ID = 4611686018427387904ULL; // 2^62
+
+#ifndef DBMS_PUBLIC_GTEST
+    private:
+#else
+    public:
+#endif
```
Lloyd-Pottiger, line 149: redundant?

dbms/src/Storages/Page/V3/Blob/BlobStat.h
```diff
@@ -118,6 +119,39 @@ class BlobStats
...
+
+        BlobFileId nextFileId(PageType page_type, const std::lock_guard<std::mutex> &);
+
+        static PageType getBlobFileType(BlobFileId file_id);
+
+    private:
+        static inline bool isRaftFileId(BlobFileId file_id)
+        {
+            return file_id >= FIRST_RAFT_FILE_ID;
+        }
```
Lloyd-Pottiger, line 136: `inline` is useless here?

Lloyd-Pottiger: `static bool isRaftFileId(BlobFileId file_id) const`

dbms/src/Storages/Page/Page.h
```diff
@@ -193,4 +193,11 @@ struct PageEntry
 };
 using PageIdU64AndEntry = std::pair<PageIdU64, PageEntry>;
 using PageIdU64AndEntries = std::vector<PageIdU64AndEntry>;
+
+enum PageType
```
Lloyd-Pottiger, line 197: `enum class PageType` is better.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
lidezhu: Removed.

dbms/src/Storages/Page/Page.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
lidezhu: It is a static function. So `const` is not applicable here. And `inline` may be useful here and even it is useless it should do no harm.

Lloyd-Pottiger: ok

Lloyd-Pottiger (approved):

dbms/src/Storages/Page/V3/Blob/BlobStat.h
```diff
@@ -118,6 +119,34 @@ class BlobStats
...
```
flowbehappy, line 136: I would recommend use an extendable structure for `PageType`. For example:

```
enum class PageType : UInt8
{
    Normal = 0;
    PageData = 1;

    // Support at most 10 types
    TypeCountLimit = 10;
}
```

And then 

```
        static inline bool isPageType(PageType t)
        {
            return file_id % PageType::TypeCountLimit == static_cast<UInt8>(t);
        }
```

flowbehappy: And we advance page_id by `TypeCountLimit`. Also I think `PageType` should not put inside PageStorage module. It should be an outside feature. And we can control the GC logical by configurations based on `PageType`.

flowbehappy: It means `PageStorage` should support configuration by passed-in `PageTypes`.

dbms/src/Storages/Page/V3/BlobStore.h
```diff
@@ -64,14 +64,23 @@ class BlobStore : private Allocator<false>
...
+        const WriteLimiterPtr & write_limiter = nullptr,
+        const ReadLimiterPtr & read_limiter = nullptr);
+
+    using PageTypeAndGcInfo = std::vector<std::tuple<PageType, GcEntriesMap, PageSize>>;
+    PageEntriesEdit gc(const PageTypeAndGcInfo & page_type_and_gc_info,
                        const WriteLimiterPtr & write_limiter = nullptr,
                        const ReadLimiterPtr & read_limiter = nullptr);
 
-    PageEntriesEdit write(typename Trait::WriteBatch && wb, const WriteLimiterPtr & write_limiter = nullptr);
+    PageEntriesEdit write(typename Trait::WriteBatch && wb, const WriteLimiterPtr & write_limiter = nullptr, PageType page_type = PageType::Normal);
```
flowbehappy, line 83: This should be better.

```suggestion
    PageEntriesEdit write(typename Trait::WriteBatch && wb, PageType page_type = PageType::Normal, const WriteLimiterPtr & write_limiter = nullptr);
```

dbms/src/Storages/Page/V3/BlobStore.cpp
```diff
@@ -1132,10 +1132,22 @@ std::vector<BlobFileId> BlobStore<Trait>::getGCStats()
             }
 
             // Check if GC is required
-            if (stat->sm_valid_rate <= config.heavy_gc_valid_rate)
+            // Raft related data contains raft log and other meta data.
+            // Raft log is written and deleted in a faster pace which is not suitable for full gc.
```
flowbehappy, line 1136: The GC behavior should be able to controlled by configurations, instead of directly put here.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
lidezhu: I don't see what currently makes it extendable and why it is not extendable previously. `PageType` is just an in memory type, so its value doesn't matter. There is nothing to prevent add new types in the future.

dbms/src/Storages/Page/V3/BlobStore.cpp
lidezhu: what kind of `GC behavior` do you mean, and what kind of configurations do you want to add?

lidezhu: Fixed.

dbms/src/Storages/Page/V3/BlobStore.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
lidezhu: Fixed.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
```diff
@@ -118,8 +124,30 @@ class BlobStats
...
+        void addFileId(BlobFileId file_id);
+
+        BlobFileId nextFileId(PageType page_type, const std::lock_guard<std::mutex> &);
+
+#ifndef DBMS_PUBLIC_GTEST
+    private:
+#else
+    public:
+#endif
+        std::unordered_map<PageType, BlobFileId> next_file_id_map;
```
flowbehappy, line 142: I think we can make it simpler, and do not need to store a `next_file_id_map`. I.e. 
```
BlobFileId nextFileId(PageType page_type)
{
    auto step = PageType::TypeCountLimit;
    auto id = cur_max_id + step;
    id = id - (id % step) + page_type;
    cur_max_id = id;
    return id;
}
```

flowbehappy:
Please update the PR's description.

dbms/src/Storages/Page/V3/Blob/BlobStat.h
lidezhu: Fixed.

lidezhu:
> Please update the PR's description.

Fixed.

flowbehappy (approved):
/unhold

PR #7742, opened 2023-07-04
.*: Pre refine for supporting join spill in pipeline

SeaRise:

Issue Number: ref #6518 

Problem Summary:

- support  spilling blocks containing only constant columns.
- support setting init task status for pipeline task, and the io event task can be submitted to io thread pool directly.
- some refine for `Interpreters::Join`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Core/Spiller.cpp
```diff
@@ -260,8 +258,16 @@ BlockInputStreams Spiller::restoreBlocks(UInt64 partition_id, UInt64 max_stream_
         /// clear the spilled_files so we can safely assume that the element in spilled_files is always not nullptr
         partition_spilled_files.clear();
     }
+    if (all_constant_block_rows[partition_id] > 0)
```
windtalker, line 261: Use `if else` to skip restore from file for a spiller that only spill all constant block?

dbms/src/Core/tests/gtest_spiller.cpp
```diff
@@ -502,14 +502,54 @@ try
...
+        Blocks expected_blocks;
+        Block block1 = constant_header;
+        for (auto & col : block1)
+            col.column = col.column->cloneResized(DEFAULT_BLOCK_SIZE);
+        expected_blocks.push_back(std::move(block1));
+        Block block2 = constant_header;
+        for (auto & col : block2)
+            col.column = col.column->cloneResized(1);
+        expected_blocks.push_back(std::move(block2));
+        verifyRestoreBlocks(spiller, 0, 100, 1, expected_blocks);
```
windtalker, line 549: Do you double check that for all cosntant block spiller, `Spiller::hasSpilledData()` will return the right value?

dbms/src/Core/Spiller.h
```diff
@@ -127,6 +135,12 @@ class Spiller
     std::atomic<bool> has_spilled_data{false};
     static std::atomic<Int64> tmp_file_index;
     std::vector<std::unique_ptr<SpilledFiles>> spilled_files;
+
+    // Used for the case that spilled blocks containing only constant columns.
+    // Record the rows of these blocks.
+    std::mutex all_constant_mutex;
+    std::vector<UInt64> all_constant_block_rows;
```
windtalker, line 142: use `std::vector<std::atomic<UInt64>>` is enough?

dbms/src/DataStreams/ConstantsBlockInputStream.h
```diff
@@ -0,0 +1,60 @@
...
+    ConstantsBlockInputStream(const Block & header, UInt64 rows_, UInt64 max_block_size_)
+        : header(header)
+        , remaining_rows(rows_)
+        , max_block_size(std::max(1, max_block_size_))
+    {
+        for (const auto & col : header)
+        {
+            RUNTIME_CHECK(col.column != nullptr && col.column->isColumnConst());
+        }
+        if unlikely (header.columns() == 0)
```
windtalker, line 34: I prefer to throw error there because just set `remaining_rows` to 0 will return different result as the spill input. In fact, I think we can need disable spill if the spill input schema is empty block.

dbms/src/Flash/Pipeline/Schedule/Tasks/Task.h
```diff
@@ -104,7 +104,7 @@ class Task
     // To reduce the overheads of `mem_tracker_holder.get()`
     MemoryTracker * mem_tracker_ptr;
 
-    ExecTaskStatus task_status{ExecTaskStatus::INIT};
```
windtalker, line 107: Is this change related to the join spill?

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -285,18 +285,12 @@ Aggregator::Aggregator(const Params & params_, const String & req_id)
...
         auto header = getHeader(false);
         bool is_convertible_to_two_level = AggregatedDataVariants::isConvertibleToTwoLevel(method_chosen);
-        bool is_input_support_spill = Spiller::supportSpill(header);
-        if (is_convertible_to_two_level && is_input_support_spill)
+        if (is_convertible_to_two_level)
         {
             /// for aggregation, the input block is sorted by bucket number
             /// so it can work with MergingAggregatedMemoryEfficientBlockInputStream
             spiller = std::make_unique<Spiller>(params.spill_config, true, 1, header, log);
         }
```
windtalker, line 293: still need to keep the `else` branch here?

dbms/src/Core/Spiller.h
SeaRise: Compilation errors will be reported because `std::atomic_unint64` is not copyable.

dbms/src/Flash/Pipeline/Schedule/Tasks/Task.h
SeaRise: Because the join spill/restore of the pipeline model will submit the io task asynchronously for execution.
The changes here allow the io task to be submitted directly to the io thread pool instead of being submitted to the cpu thread pool first and then to the io thread pool

dbms/src/DataStreams/ConstantsBlockInputStream.h
SeaRise: Ok, I feel that an error can be reported directly here. An empty header is abnormal even if it is not a spill.

dbms/src/Core/tests/gtest_spiller.cpp
SeaRise: It was indeed buggy, fixed and added check in ut

dbms/src/Interpreters/Aggregator.cpp
SeaRise: ok, fixed.

dbms/src/Core/Spiller.cpp
SeaRise: ok, update the code like
```
if (is_all_constant)
{
    xxxxx
}
else
{
    the origin codes.
}
```

dbms/src/Core/SpillHandler.cpp
```diff
@@ -174,5 +192,10 @@ void SpillHandler::finish()
         current_spilled_file_index = INVALID_CURRENT_SPILLED_FILE_INDEX;
         RUNTIME_CHECK_MSG(spiller->isSpillFinished() == false, "{}: spill after the spiller is finished.", spiller->config.spill_id);
     }
+    else if unlikely (spiller->isAllConstant())
+    {
+        std::lock_guard lock(spiller->all_constant_mutex);
```
windtalker, line 197: I suggest to keep total_rows in local and commit it to spiller in `SpillHandler::finish()`, just like the normal block spill.

dbms/src/Core/Spiller.cpp
```diff
@@ -201,67 +202,84 @@ BlockInputStreams Spiller::restoreBlocks(UInt64 partition_id, UInt64 max_stream_
...
-            RUNTIME_CHECK_MSG(file->exists(), "Spill file {} does not exists", file->path());
-            details.merge(file->getSpillDetails());
-            std::vector<SpilledFileInfo> file_infos;
-            file_infos.emplace_back(file->path());
-            restore_stream_read_rows.push_back(file->getSpillDetails().rows);
-            if (release_spilled_file_on_restore)
-                file_infos.back().file = std::move(file);
-            ret.push_back(std::make_shared<SpilledFilesInputStream>(std::move(file_infos), input_schema, header_without_constants, const_column_indexes, config.file_provider, spill_version));
+            // TODO use settings.max_block_size instead of DEFAULT_BLOCK_SIZE.
+            ret.push_back(std::make_shared<ConstantsBlockInputStream>(input_schema, all_constant_block_rows[partition_id], DEFAULT_BLOCK_SIZE));
```
windtalker, line 213: The output concurrency is always 1?

SeaRise: No, if append_dummy_read_stream is true, L283 will append NullBlockInputStream

dbms/src/Core/SpillHandler.cpp
SeaRise: ok, updated.

dbms/src/Core/Spiller.cpp
windtalker: I mean only one BlockInputStream will output non-empty data?

SeaRise: Yes, do we need to output the constants block in multiple streams?

windtalker: I think so, in both agg and join, it will convert constant column to normal column first.

dbms/src/Core/Spiller.cpp
```diff
@@ -201,67 +202,84 @@ BlockInputStreams Spiller::restoreBlocks(UInt64 partition_id, UInt64 max_stream_
...
+            all_constant_block_rows[partition_id] = 0;
```
windtalker, line 214: only set `all_constant_block_rows[partition_id]` to 0 if `release_spilled_file_on_restore ==true`?

SeaRise: ok, updated.

dbms/src/Core/Spiller.cpp
SeaRise: ok, updated

dbms/src/Core/Spiller.cpp
```diff
@@ -201,67 +202,105 @@ BlockInputStreams Spiller::restoreBlocks(UInt64 partition_id, UInt64 max_stream_
...
+                all_constant_block_rows[partition_id] = 0;
+        }
+        if (total_rows > 0)
+        {
+            if (max_stream_size == 0)
+                max_stream_size = config.for_all_constant_max_streams;
+            std::vector<UInt64> stream_rows;
+            stream_rows.resize(max_stream_size, 0);
+            size_t index = 0;
+            while (total_rows > 0)
```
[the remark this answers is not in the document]
SeaRise: To consider the case that total_rows is very small, and avoid stream output too many small block.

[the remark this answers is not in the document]
windtalker: OK

windtalker (approved):
LGTM

yibin87 (approved):
LGTM

PR #7779, opened 2023-07-10
PageStorage: remove wal files according to its max sequence when dump snapshot

lidezhu:

Issue Number: close https://github.com/pingcap/tiflash/issues/7780

Problem Summary: After dump snapshot from memory for ps, it will remove old wal files. But new page data is always write into wal before applying to memory. So the data in memory may be less than the data in wal files. So using the snapshot dumped from memory to replace old wal files may cause some data lost.

Add a cache to memorize the max sequence of wal files. 
When dump snapshot, traverse from the latest wal files and find the one whose max sequence is smaller than the snapshot's sequence. And remove the wal file and older ones.
We avoid maintaining each wal file's max sequence when writing, because it will make write process acquire one more lock which may cause performance penalty.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Page/V3/WALStore.cpp
```diff
@@ -270,4 +278,48 @@ bool WALStore::saveSnapshot(
...
+}
+
+UInt64 WALStore::getLogFileMaxSequence(const LogFilename & log_filename, std::function<UInt64(const String & record)> max_sequence_getter)
+{
+    std::unique_lock<std::mutex> lock(log_file_max_sequences_cache_mutex);
+    auto iter = log_file_max_sequences_cache.find(log_filename);
+    if (iter != log_file_max_sequences_cache.end())
+        return iter->second;
+    else
+    {
```
JaySon-Huang, line 304: As we directly return in the previous `if (...)` statement, we can remove this `else` indent layer to be 

```
if (...)
    return ...;

...
```

dbms/src/Storages/Page/V3/WALStore.cpp
```diff
@@ -270,4 +278,48 @@ bool WALStore::saveSnapshot(
...
+        while (true)
+        {
+            auto [ok, record] = log_reader->readRecord();
+            if (!ok)
+                break;
+
+            last_record = std::move(record);
+        }
+        if (last_record.empty())
+            return 0; // empty log file
```
JaySon-Huang, line 316: ditto

dbms/src/Storages/Page/V3/WALStore.cpp
```diff
@@ -270,4 +278,48 @@ bool WALStore::saveSnapshot(
...
+
+UInt64 WALStore::getLogFileMaxSequence(const LogFilename & log_filename, std::function<UInt64(const String & record)> max_sequence_getter)
+{
+    std::unique_lock<std::mutex> lock(log_file_max_sequences_cache_mutex);
+    auto iter = log_file_max_sequences_cache.find(log_filename);
+    if (iter != log_file_max_sequences_cache.end())
+        return iter->second;
+    else
+    {
+        auto log_reader = wal_store_reader->createLogReader(log_filename);
```
JaySon-Huang, line 305: Can we make `WALStoreReader::createLogReader` to be a static method? It is weird to me that we need to keep a `wal_store_reader` as long as `WALStore`

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -1825,12 +1833,9 @@ bool PageDirectory<Trait>::tryDumpSnapshot(const WriteLimiterPtr & write_limiter
     // The main reason write affect dumping snapshot is that we can not get a read-only
     // `being_ref_count` by the function `createSnapshot()`.
     assert(!files_snap.persisted_log_files.empty()); // should not be empty
-    auto log_num = files_snap.persisted_log_files.rbegin()->log_num;
-    auto identifier = fmt::format("{}.dump_{}", wal->name(), log_num);
 
     Stopwatch watch;
     // The records persisted in `files_snap` is older than or equal to all records in `edit`
```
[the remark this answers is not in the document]
lidezhu: Fixed

dbms/src/Storages/Page/V3/WALStore.cpp
lidezhu: Fixed

dbms/src/Storages/Page/V3/WALStore.cpp
lidezhu: Fixed.

Lloyd-Pottiger (approved):

dbms/src/Storages/Page/V3/PageDirectory.h
```diff
@@ -152,9 +152,16 @@ class MultiVersionRefCount
...
             {
-                versioned_ref_counts = nullptr;
+                if (new_versioned_ref_counts->empty())
+                {
+                    versioned_ref_counts = nullptr;
+                }
+                else
+                {
+                    versioned_ref_counts.swap(new_versioned_ref_counts);
+                }
```
JaySon-Huang, line 164:
```suggestion
                {
                    // There could be some new ref count created after `snap_seq`, we need to
                    // keep the newly added ref counts
                    versioned_ref_counts.swap(new_versioned_ref_counts);
                }
```

JaySon-Huang (approved):

Issue #7721, opened 2023-06-29
Select * from T where a in (x, y, z) makes TiFlash OOM


flowbehappy:
## Enhancement

The table `T` has 200 columns, most of them are `varchar` or `decimal`. And `T` has 100 list partitions.

PR #7806, opened 2023-07-14
Limiting max block bytes

JinheLin:

Issue Number: ref #7721

Problem Summary:

By default, the number of row of a block is `DEFAULT_BLOCK_SIZE`(2^16). If a query read a lot of columns, the reading bytes of each row can be large, for example 3KB, and a block can be about 200MB. This is extremely large and can cause OOM since there are many blocks processing concurrently.

- Limiting max block bytes in query request. 
- The number of row of a block of background tasks, such as DeltaMerge, is `DEFAULT_MERGE_BLOCK_SIZE`, so we don't limit it. 
- Add configuration `profiles.default.max_block_bytes`, default value is `DEFAULT_BLOCK_BYTES`(16MB).
- `clipBlockRows` try to limit the block size not exceed `max_block_bytes` and it also make the clipped block rows align with `dt_segment_stable_pack_rows` since at least read one pack from DMFile each time.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2816,13 +2826,14 @@ BlockInputStreamPtr Segment::getLateMaterializationStream(BitmapFilterPtr && bit
     }
 
     // construct rest column stream
+    auto clipped_rest_block_rows = clipBlockRows(dm_context.db_context, expected_block_size, *rest_columns_to_read);
```
Lloyd-Pottiger, line 2829: `clipped_rest_block_rows` should equal to `clipped_filter_block_rows`

dbms/src/Core/Defines.h
```diff
@@ -64,6 +64,8 @@ static constexpr UInt64 DEFAULT_DISAGG_TASK_TIMEOUT_SEC = 5 * 60;
   */
 #define DEFAULT_BLOCK_SIZE 65536
 
+constexpr size_t DEFAULT_BLOCK_BYTES = DEFAULT_BLOCK_SIZE * 256; // 256B per row, total 16MB.
```
Lloyd-Pottiger, line 67: maybe too small...

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2900,5 +2911,27 @@ BlockInputStreamPtr Segment::getBitmapFilterInputStream(const DMContext & dm_con
...
+        return expected_block_rows;
+    }
+    else
+    {
+        auto row_bytes = std::max(1, stable->avgRowBytes(read_columns)); // Avoid row_bytes to be 0.
+        auto rows = max_block_bytes / row_bytes;
+        rows = std::max(rows / pack_rows * pack_rows, pack_rows); // Align with pack and at least read one pack.
+        return std::min(expected_block_rows, rows);
+    }
+}
```
Lloyd-Pottiger, line 2934: define as functions in anonymous namespace?

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2867,14 +2868,23 @@ BlockInputStreamPtr Segment::getBitmapFilterInputStream(const DMContext & dm_con
...
         return std::make_shared<EmptyBlockInputStream>(toEmptyBlock(columns_to_read));
     }
+
+    // buildBitmapFilter will only read handle, version and tag.
+    const ColumnDefines build_bitmap_columns{
+        getExtraHandleColumnDefine(is_common_handle),
+        getVersionColumnDefine(),
+        getTagColumnDefine(),
+    };
+    auto build_bitmap_block_rows = clipBlockRows(dm_context.db_context, expected_block_size, build_bitmap_columns);
```
Lloyd-Pottiger, line 2878: There is no need to clip, just pass `expected_block_size` is ok.

dbms/src/Storages/DeltaMerge/StableValueSpace.cpp
```diff
@@ -564,5 +564,53 @@ StableValueSpace::Snapshot::getAtLeastRowsAndBytes(const DMContext & context, co
...
+        return 8;
+    case Field::Types::UInt128:
+    case Field::Types::Int128:
+    case Field::Types::Decimal128:
+        return 16;
+    case Field::Types::Int256:
+    case Field::Types::Decimal256:
+        return 32;
+    case Field::Types::String:
+        return f.get<String>().size();
```
Lloyd-Pottiger, line 586: Usually default value of string is `""` or `NULL`, which may cause inaccuracy.

Can we get `255` for type `varchar(255)`?

Lloyd-Pottiger: If the type is `varchar(255)`, default `""`, but the length of the data in this col is larger than `200`?

dbms/src/Storages/DeltaMerge/Segment.cpp
Lloyd-Pottiger: The memory usage of get bitmap is small, so I think pass `expected_block_size` is acceptable.

Lloyd-Pottiger: And those data will not pass to compute layer.

dbms/src/Core/Defines.h
Lloyd-Pottiger: ok

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -714,15 +714,17 @@ BlockInputStreamPtr Segment::getInputStream(const ReadMode & read_mode,
...
+        return getInputStreamModeNormal(dm_context, columns_to_read, segment_snap, read_ranges, filter ? filter->rs_operator : EMPTY_RS_OPERATOR, max_version, clipped_block_rows);
     case ReadMode::Fast:
-        return getInputStreamModeFast(dm_context, columns_to_read, segment_snap, read_ranges, filter ? filter->rs_operator : EMPTY_RS_OPERATOR, expected_block_size);
+        return getInputStreamModeFast(dm_context, columns_to_read, segment_snap, read_ranges, filter ? filter->rs_operator : EMPTY_RS_OPERATOR, clipped_block_rows);
     case ReadMode::Raw:
-        return getInputStreamModeRaw(dm_context, columns_to_read, segment_snap, read_ranges, expected_block_size);
+        return getInputStreamModeRaw(dm_context, columns_to_read, segment_snap, read_ranges, clipped_block_rows);
     case ReadMode::Bitmap:
+        // getBitmapFilterInputStream will read columns by multi-stages, so we clip its block rows inside.
         return getBitmapFilterInputStream(dm_context, columns_to_read, segment_snap, read_ranges, filter, max_version, expected_block_size);
```
Lloyd-Pottiger, line 728:
```suggestion
        return getBitmapFilterInputStream(dm_context, columns_to_read, segment_snap, read_ranges, filter, max_version, expected_block_size, clipped_block_rows);
```

we can pass `expected_block_size` and  `clipped_block_rows` at the same time now.

Lloyd-Pottiger (approved):
Others LGTM

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -2897,7 +2899,29 @@ BlockInputStreamPtr Segment::getBitmapFilterInputStream(const DMContext & dm_con
...
+{
+    if (unlikely(max_block_bytes <= 0)) // Disable block bytes limit.
+    {
+        return expected_block_rows;
+    }
+    else
+    {
+        auto row_bytes = std::max(1, stable->avgRowBytes(read_columns)); // Avoid row_bytes to be 0.
+        auto rows = max_block_bytes / row_bytes;
+        rows = std::max(rows / pack_rows * pack_rows, pack_rows); // Align with pack and at least read one pack.
```
SeaRise, line 2922:
```suggestion
        rows = std::ceil(rows / pack_rows) * pack_rows; // Align with pack and at least read one pack.
```
just a little suggestion

SeaRise (approved):
Others LGTM

SeaRise:
/hold
@JinheLin You can comment `/unhold` when you want to merge this pr.

PR #7819, opened 2023-07-18
More refine for join spill

SeaRise:

Issue Number: ref #6518 

Problem Summary:

- refine the code of `HashJoinProbeTransformOp` as follows, in order to be more convenient to use the state machine to implement the switch of join spill in the io/cpu thread pool.
```
while (true)
{
    switch(status)
    {
    case Status1:
        ...
        status = Status2:
        break;
    case Status2:
        ...
        status = Status3:
        break;
    case Status4:
        ...
    }
}
```
- refine the code of build/probe spilling as follows, In order to avoid frequent switching when the pipeline model supports join spill
```
## build
Join->insertFrom(Block, op_index);
if (join->hasMarkNeedSpillData(op_index))
    join->spillMarkedData(op_index);

## build finish
join_ptr->finishOneBuild(op_index);
if (join->hasMarkNeedSpillData(op_index))
    join->spillMarkedData(op_index);

## probe
Block block = join->joinBlock(ProbeProcessInfo & info, op_index);
if (join->hasMarkNeedSpillData(op_index))
    join->spillMarkedData(op_index);

## probe finish
join->finishOneProbe(op_index);
if (join->hasMarkNeedSpillData(op_index))
    join->spillMarkedData(op_index);
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Join.cpp
```diff
@@ -1460,91 +1523,109 @@ void Join::finishOneBuild()
...
-        /// since shallow_copy_probe_threshold is at least 1, if strictness is any, it will never use SHALLOW_COPY_RIGHT_BLOCK
-        cross_probe_mode = right_rows_to_be_added_when_matched_for_cross_join > shallow_copy_cross_probe_threshold
-            ? CrossProbeMode::SHALLOW_COPY_RIGHT_BLOCK
-            : CrossProbeMode::DEEP_COPY_RIGHT_BLOCK;
-        LOG_DEBUG(log, "Cross join will use {} probe mode", magic_enum::enum_name(cross_probe_mode));
-    }
+        finalizeCrossJoinBuild();
+
+    // set rf is ready
+    finalizeRuntimeFilter();
```
SeaRise, line 1549: Moved here to avoid runtime filter getting stuck with spill

dbms/src/Interpreters/Join.h
```diff
@@ -34,6 +34,8 @@
 #include <memory>
 #include <shared_mutex>
 
+#include "Interpreters/Users.h"
```
windtalker, line 37: Format the code and why need to include this header file?

dbms/src/Interpreters/Join.cpp
windtalker: I think for now, spill should be disabled explicitly if runtime filter is enabled because runtime filter is not support spill, and for current implementation, for partitions that is spilled, runtime filter is not constructed, which will cause wrong result.

dbms/src/Interpreters/Join.cpp
```diff
@@ -1768,7 +1862,19 @@ void Join::spillProbeSideBlocks(UInt64 part_id, Blocks && blocks)
     probe_spiller->spillBlocks(std::move(blocks), part_id);
 }
 
-void Join::spillMostMemoryUsedPartitionIfNeed()
+void Join::markBuildSideSpillData(UInt64 part_id, Blocks && blocks, size_t stream_index)
+{
+    if (!blocks.empty())
+        getBuildSideMarkdeSpillData(stream_index).emplace_back(part_id, std::move(blocks));
```
windtalker, line 1868: assert that `getBuildSideMarkdeSpillData(stream_index).second` is empty before emplace_back?

dbms/src/Interpreters/Join.cpp
```diff
@@ -1768,7 +1862,19 @@ void Join::spillProbeSideBlocks(UInt64 part_id, Blocks && blocks)
...
+}
+
+void Join::markProbeSideSpillData(UInt64 part_id, Blocks && blocks, size_t stream_index)
+{
+    if (!blocks.empty())
+        getProbeSideMarkdeSpillData(stream_index).emplace_back(part_id, std::move(blocks));
```
windtalker, line 1874: Ditto

dbms/src/DataStreams/HashJoinBuildBlockInputStream.cpp
```diff
@@ -24,10 +24,17 @@ Block HashJoinBuildBlockInputStream::readImpl()
         Block block = children.back()->read();
         if (!block)
         {
-            join->finishOneBuild();
+            if (join->finishOneBuild(stream_index))
+            {
+                if (join->hasBuildSideMarkedSpillData(stream_index))
```
windtalker, line 29: Why the last thread only spill data for `stream_index`, what about other threads that has the remaining blocks that is expected to be spilled?

dbms/src/Interpreters/Join.cpp
SeaRise: ok, added.

dbms/src/Interpreters/Join.h
SeaRise: Sorry, it is automatically written by vscode and has been deleted.

dbms/src/Interpreters/Join.cpp
SeaRise: ok, added.

dbms/src/DataStreams/HashJoinBuildBlockInputStream.cpp
SeaRise: At this time, only stream_index has marked_spill_data, and each thread will call finishOneBuild/finishOneProbe after flushing its own marked_spill_data.

dbms/src/Interpreters/Join.cpp
SeaRise: ok, updated the code as
```
    if (isEnableSpill())
    {
        if (isSpilled())
        {
            // TODO support runtime filter with spill.
            cancelRuntimeFilter("Currently runtime filter is not compatible with join spill, so cancel runtime filter here.");
        }
        else
        {
            // set rf is ready
            finalizeRuntimeFilter();
        }
        ...
    else
    {
            // set rf is ready
            finalizeRuntimeFilter();
    }
```

dbms/src/Interpreters/Join.h
```diff
@@ -272,8 +271,18 @@ class Join
     void meetError(const String & error_message);
     void meetErrorImpl(const String & error_message, std::unique_lock<std::mutex> & lock);
 
-    void spillBuildSideBlocks(UInt64 part_id, Blocks && blocks);
-    void spillProbeSideBlocks(UInt64 part_id, Blocks && blocks);
+    // std::unordered_map<partition_index, Blocks>
+    using MarkdeSpillData = std::unordered_map<size_t, Blocks>;
```
windtalker, line 275: Should be `MarkedSpillData`?

windtalker (approved):
Others LGTM

dbms/src/Interpreters/Join.h
SeaRise: my mistake...

yibin87 (approved):
LGTM

Issue #7761, opened 2023-07-06
another tiflash_write node oom after inject one tiflash_write node failure for 50m and recover


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1、run ch
go-tpc ch run -D tpcc --host tc-tidb.endless-ha-test-tps-1815226-1-478 -P4000 --warehouses 2000 -T 32 --acThreads 1 --queries q16 --ignore-error '2013,1213,1105,1205,8022,8028,9004,9007,1062' --time 36000m --user root --password '' --interval '10s'
2、inject one of  tiflash_write node failure for 50m
3、after 48m，one tiflash_write node oom and workload always report error as below：
    err execute query q16 failed Error 1105: other error for mpp stream: Poco::Exception. Code: 1000, e.code() = 15, e.displayText() = Exception: no alive tiflash, cannot dispatch BatchCopTask, e.what() = Exception

### 2. What did you expect to see? (Required)
no error report after 2m when fault injection
no oom

### 3. What did you see instead (Required)
![2edf829b-0d39-4904-afa8-baf8f194b862](https://github.com/pingcap/tiflash/assets/84712107/ca879f5a-fbf1-4f6f-b0d5-d8a7dcc3c5c8)
workload report “err execute query q16 failed Error 1105: other error for mpp stream: Poco::Exception. Code: 1000, e.code() = 15, e.displayText() = Exception: no alive tiflash, cannot dispatch BatchCopTask, e.what() = Exception” after inject one tiflash_write node failure for 50m

### 4. What is your TiFlash version? (Required)
git hash：f08e4d66ec09b2fb6ea727a23bc7956c15caae40

JaySon-Huang:
Seems this bug is caused by
TiKV leader will **NOT** send `CompactLog` when one TiFlash write node is down. It leads to the number of RaftLog in the other TiFlash write node grows rapidly. It finally cause the TiFlash write node OOM cause by too many pages in UniPS.

PR #7858, opened 2023-07-28
Disagg: Eager remove applied RaftLog to fix OOM when down-peers exist

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/7761

Problem Summary:

* Add `flash.eager_gc_log_gap` with default value 512

#### Eager RaftLog GC tasks
When there are down peer(s), the TiKV CompactLog is quite slow. And the un-deleted RaftLog takes up lots of memory because each Page in `UniPS` takes about 500 bytes currently.
Because the RaftLog is useless after being applied, so the RaftLog with index in `[truncated_index, applied_index)` can be removed from TiFlash's UniPS. We name this as `Eager RaftLog GC`

#### Details
* Add `eager_truncated_index` to `Region` in TiFlash
  * When `CompactLog` is executed (`KVStore::forceFlushRegionDataImpl`), we will update `eager_truncated_index` to `applied_index`
  * `eager_truncated_index` will be persisted to disk. If we restore a `Region` from disk without `eager_truncated_index`, it will be set to the `truncated_index`.
* When using `UniPS`, which means RaftLog are stored in the global UniPageStorage instance, we
  * After a normal write cmd is handled in `KVStore::handleWriteRaftCmdInner`, we will check the Region's `eager_truncated_index`, `applied_index` and the `gap_threshold`
  * If the actual gap is larger than `gap_threshold`, then we generate a `task hint` for eager RaftLog GC
  * Add a background task to check the `task hint` and run eager RaftLog GC tasks

#### Others
* Remove useless `Region::last_compact_log_time`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

#### Workload
```
>  cat config
mysql-host=10.2.12.81
mysql-port=8020
mysql-user=root
mysql-password=
mysql-db=sbtest
time=900
threads=64
report-interval=10
db-driver=mysql

> sysbench --config-file=config oltp_update_index --tables=1 --table-size=100000000 --db-ps-mode=auto --rand-type=uniform run
```

1 TiDB, 1 PD, 1 TiKV
2 TiFlash wn

stop 1 TiFlash wn and run sysbench oltp_update_index

Before this PR
* the "number of Pages" increases from ~200K to more than 23M
* the memory usage reach about 19GB

After this PR
* the "number of Pages" increases from ~200K to 1M
* the memory usage reach about 7GB

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
+// Copyright 2023 PingCAP, Ltd.
```
CalvinNeo, line 1: Why not still put these files in Transaction?

JaySon-Huang: there are too many codes that mix Raft, RegionMeta, codec under the Transaction folder. I want to split it into different subfolders.
But sure we can do it later.

CalvinNeo: Totally agree for the first part, the Transaction folder is too messy. I think it is the time we divide the folder into several sub folders.
However, I think `KVStore/Transaction` and `KVStore/Raft` is better than `TiDB/Raft`, since raft and transaction is in TiKV(KVStore) layer, rather than in TiDB layer.

dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+        const auto region_task = getRegionTask(region_id, region_apply_state, hint, eager_gc_rows, logger);
+        if (region_task.skip)
+        {
+            num_skip += 1;
+            continue;
+        }
+
+        for (UInt64 log_index = region_task.eager_truncate_index; log_index < region_task.applied_index; ++log_index)
+        {
+            del_batch.delPage(UniversalPageIdFormat::toRaftLogKey(region_id, log_index));
```
CalvinNeo, line 150: Will the write batch be too large when there are too many entries to be removed? I am afraid it may happens when the `eager_truncate_index` is   not updated, may be after a restart.

dbms/src/Storages/Transaction/KVStore.h
```diff
@@ -228,6 +237,10 @@ class KVStore final : private boost::noncopyable
         UInt64 compact_term,
         bool is_background,
         bool lock_held = true);
+
+    RaftLogEagerGcTasks::Hints getRaftLogGcHints();
+    void applyRaftLogTaskRes(const RaftLogGcTasksRes & res) const;
```
CalvinNeo, line 242: It should be `applyRaftLogGcTaskRes`, there are other tasks related to raft, not in this module though.

dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+struct RegionGcTask
+{
+    bool skip;
+    UInt64 eager_truncate_index;
+    UInt64 applied_index;
+};
+
+RegionGcTask getRegionTask(
```
CalvinNeo, line 84:
```suggestion
static RegionGcTask getRegionTask(
```

dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+    const RegionID region_id,
+    const raft_serverpb::RaftApplyState & region_state,
+    const RaftLogEagerGcHint & hint,
+    UInt64 threshold,
+    const LoggerPtr & logger)
+{
+    // LOG_DEBUG(logger, "Load persisted region apply state, region_id={} state={}", region_id, region_state.ShortDebugString());
+    if (region_state.applied_index() <= 1)
```
CalvinNeo, line 92: Why using 1?

dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+    LoggerPtr logger = Logger::get();
+
+    Stopwatch watch;
+    RaftLogGcTasksRes eager_truncated_indexes;
+    size_t num_skip = 0;
+    size_t total_num_raft_log_removed = 0;
+    for (const auto & [region_id, hint] : hints)
+    {
+        // Read region apply_state from UniPS
+        auto region_apply_state_key = UniversalPageIdFormat::toRaftApplyStateKeyInKVEngine(region_id);
```
CalvinNeo, line 131: `Region::lastCompactLogApplied` will be updated to current `applied_index` every time we finished a flush in `persistRegion`.
And will be updated to what is read from `RegionPersister` after restart.
I think may be we can just use what `Region::lastCompactLogApplied` returns, instead of performing a extra read?
Still, we need to assert `Region::lastCompactLogApplied` to be not zero here.

dbms/src/TiDB/Raft/RaftLogManager.h
```diff
@@ -0,0 +1,48 @@
...
+    UInt64 applied_index;
+};
+class RaftLogEagerGcTasks
+{
+public:
+    // Check whether a new eager GC task for given region_id will be created.
+    bool updateHint(RegionID region_id, UInt64 eager_truncated_index, UInt64 applied_index, UInt64 threshold);
+
+    using Hints = std::map<RegionID, RaftLogEagerGcHint>;
+    Hints getAndClearHints();
```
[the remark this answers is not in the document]
JaySon-Huang: When there are no down-peers/pending-peers, the gap between applied index and truncated index usually won't reach log gap, so it will not generate a hints.
If the gap does happen, the hints will be taken by `KVStore::getRaftLogGcHints` in `eager_raft_log_gc_handle` and be removed from the memory.

dbms/src/TiDB/Raft/RaftLogManager.cpp
JaySon-Huang: it is a "private" function only used in this file, I'll limit its scope

dbms/src/TiDB/Raft/RaftLogManager.cpp
```diff
@@ -0,0 +1,184 @@
...
+            region_task.eager_truncate_index,
+            region_task.applied_index,
+            num_raft_log_removed,
+            watch.elapsedSecondsFromLastTime());
+    }
+
+    const double cost_seconds = watch.elapsedSeconds();
+    const size_t num_process_regions = hints.size() - num_skip;
+    GET_METRIC(tiflash_raft_eager_gc_duration_seconds, type_run).Observe(cost_seconds);
+    GET_METRIC(tiflash_raft_eager_gc_count, type_num_raft_logs).Increment(total_num_raft_log_removed);
```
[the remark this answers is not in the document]
JaySon-Huang: Is this what you expected? https://github.com/pingcap/tiflash/blob/b8fbc4507725a25041cb4f6d189bb00b979907d9/dbms/src/Storages/Transaction/RaftLogManager.cpp#L157-L164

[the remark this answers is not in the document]
CalvinNeo: Yes

dbms/src/TiDB/Raft/RaftLogManager.cpp
JaySon-Huang: 1. There is a time gap between we update `Region::lastCompactLogApplied` and tiflash-proxy real persist the region apply state.
2. We must ensure all needed RaftLog can be read after restart

So we have to read the region apply state from UniPS to ensure the correctness

I've updated the way to read RegionApplyState in UniPS

CalvinNeo: You are right, may be we can put these as comments for these codes?

dbms/src/Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h
```diff
@@ -109,6 +109,18 @@ struct UniversalPageIdFormat
         return buff.releaseStr();
     }
 
+    // RAFT_PREFIX LOCAL_PREFIX REGION_RAFT_PREFIX region_id APPLY_STATE_SUFFIX
+    static UniversalPageId toRaftApplyStateKeyInRaftEngine(UInt64 region_id)
```
CalvinNeo, line 113: What's this?

dbms/src/TiDB/Raft/RaftLogManager.cpp
JaySon-Huang: I've added some comments here. Or you think we should add comments about why not using the member of `Region` in TiFlash memory?
https://github.com/pingcap/tiflash/blob/2904e72374e1e36b454d8164d54114f626614cad/dbms/src/Storages/Transaction/RaftLogManager.cpp#L136-L143

dbms/src/Storages/Page/V3/Universal/UniversalPageIdFormatImpl.h
JaySon-Huang: We use it to resolve RaftApplyState when TiFlash-proxy transfer its data from KVEngine to RaftEngine later (for example, adapting to RaftStore v2)

https://github.com/pingcap/tiflash/pull/7858/files#diff-993429dba2f708aefb20bb3620fa3d03dd72561a21bf370eb708a1821b47ce1fR33-R42

CalvinNeo (approved):
lgtm

metrics/grafana/tiflash_summary.json
```diff
@@ -10677,7 +10677,7 @@
           "targets": [
             {
               "exemplar": true,
-              "expr": "sum(delta(tiflash_raft_raft_log_lag_count_bucket{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", type=~\"applied_index\"}[1m])) by (le, type)",
```
CalvinNeo, line 10680: Could you also help me to change this metrics into `tsbuckets`?
It will be like this
image

JaySon-Huang: Fixed

PR #7901, opened 2023-08-04
fix row key not in standard format

### LINKED ISSUE #3099: TiFlash is incompatible with the split keys with TiSpark
https://github.com/pingcap/tispark/pull/2101
TiSpark split Region with a key with '\00' suffix before direct write. After that, some rows are not able to be scanned.

### LINKED ISSUE #7762: TiFlash has less rows than TiKV
## Bug Report

A use report that TiFlash has less rows than TiKV.

The root cause is that  the range of some regions are in illegal format. For example:
table id: 283
start: 7480000000000001FF1B5F728000000000FF11BDC00000000000FA
end: 7480000000000001FF1B5F728000000000FF11BDC00000000000FB

img_v2_25807fb2-a231-4b30-927e-895e33d2adcg

We don't know how the data is ingested into TiFlash. But we need to make some checks on TiFlash side to prevent data lost.

lidezhu:

Issue Number: close #3099 close #7762

Problem Summary: For int handle, the standard row key format should be t{tidb_id}_r{row_id}, but the cluster may generate keys not strictly adhere to this format. For example, it may append a `0x00` to the encoded value.
When tiflash decode the key, it will ignore the suffix `0x00` and may get an invalid row key value.

For row key which size is larger than the standard format, we add 1 on the decoded int key value.
This is ok, because
   1) if the key is the start range, then [t100_r1000 + 0x00, xxx) has the same semantics with [t100_r1001, xxx)
  2) if the key is the end range, then [xxx, t100_r1000 + 0x00) also has the same semantics with [xxx, t100_r1001)

Note if the decoded int value is `Int64::max_value`, it is a value generated by tiflash itself which is RowKeyValue::INT_HANDLE_MAX_KEY. This is a special key to means +inf, so we can just ignore it.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the potential consistency problem when the int type row key value is not in standard encoding format
```


breezewish (approved):

dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -84,6 +84,32 @@ struct RowKeyValue
         {
             size_t cursor = 0;
             int_value = DB::DecodeInt64(cursor, *value);
+            if (unlikely(value->size() != sizeof(Int64)))
```
flowbehappy, line 87: I think we should make sure the subfix string is exactly "0x00".

lidezhu: Fixed.

flowbehappy (approved):

dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -84,6 +84,42 @@ struct RowKeyValue
...
+                // This is ok, because
+                //  1) if the key is the start range, then [t100_r1000 + 0x00, xxx) has the same semantics with [t100_r1001, xxx)
+                //  2) if the key is the end range, then [xxx, t100_r1000 + 0x00) also has the same semantics with [xxx, t100_r1001)
+                //
+                // Note if the `int_value` is Int64::max_value,
+                // it is a value generated by tiflash itself to means +inf()(which is RowKeyValue::INT_HANDLE_MAX_KEY).
+                // So we can just ignore it.
+                if (value->size() == sizeof(UInt64) + 1)
+                {
+                    if (int_value < std::numeric_limits<Int64>::max() && value->back() == 0x00)
```
flowbehappy, line 103: Why don't we need an  "else" here ?

lidezhu: Refactor the structure to make it more clear.

CalvinNeo:
For the reason of 0x00 suffix, see https://github.com/pingcap/tidb/blob/a50a2323f4ba020cec9fa04a291c60a3d212d365/pkg/kv/key.go#L34-L68 and
[components/raftstore/src/store/worker/split_controller.rs#L878-L920](https://github.com/tikv/tikv/blob/77e5c58afa0fd14892799eba4bd4f55b45da7d8d/components/raftstore/src/store/worker/split_controller.rs#L878-L920)

PR #7917, opened 2023-08-08
Add ResourceControlQueue to support tiflash resource control

### LINKED ISSUE #7916: Support ResourceControlQueue
## Feature Request

guo-shaoge:

Issue Number: close #7916

Problem Summary:

This is the first part of supporting tiflash resource control. 

1. Add a new TaskQueue type `ResourceControlQueue`.
    1. `ResourceControlQueue` will have a nested TaskQueue, basically it calls nested TaskQueue. Magic happens in the `priority_queue`, which is ordered by the priority of resource group.
    2. `resource_group_infos`: this is the `priority_queue`
        1. it records resource group priority, name, nested task queue and keyspace_id.
        2. it order by resource group priority(**less value means higher priority**)
    3. `pipeline_tasks`: record <resource_group_name, nested task queue>
    4. `cancel_query_ids`: just for cancel
5. Add `ResourceGroup`
    1. `getPriority()`: this is the core logic of computing the priority of a resource group, which is used by `ResourceControlQueue::take()` to decide which resource group to pop task.
    2. `consumeResource()`: record cpu usage of resource group, which is used to decided priority of a resource group
7. Add `TokenBucket`. Two mode
    1. static mode: fill_rate is 0.0. Token will fetch from GAC.(Default mode)
    2. dynamic mode: fill_rate > 0.0. Local token bucket will fill token itself. Happens when:
        1. degrade bucket mode: If a network down or GAC crashes, causing LAC to be unable to communicate with GAC for a period of time
        2. trickle bucket mode: If GAC has not enouth tokens for this resource group.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Pipeline/Schedule/Tasks/TaskHelper.h
```diff
@@ -23,7 +23,7 @@
 
 namespace DB
 {
-#define FINISH_STATUS \
+#define FINISH_STATUS                                      \
```
guo-shaoge, line 26: todo why not endTraceMemory for `FINALIZE_TASK`

dbms/src/Flash/Pipeline/Schedule/TaskQueues/tests/gtest_resource_control_queue.cpp
```diff
@@ -0,0 +1,360 @@
...
+// 1. Tasks of different resource groups is restricted by RU, and there is still available cpu resource.
+// 2. Change rg-1 as burstable, the available cpu resource is used by rg1.
+// 3. Change other rg as burstable, the cpu usage should be same of all resource groups.
+TEST_F(TestResourceControlQueue, TestBurstable)
+{
+}
+
+TEST_F(TestResourceControlQueue, TestAddAndDelResourceGroup)
+{
+}
```
guo-shaoge, line 358: todo add cancel unittest

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,259 @@
...
+    // name can be empty, it means resource control is disabled.
+    const std::string & name = task->getResourceGroupName();
+
+    auto iter = pipeline_tasks.find(name);
+    if (iter == pipeline_tasks.end())
+    {
+        auto task_queue = std::make_shared<NestedQueueType>();
+        task_queue->submit(std::move(task));
+        resource_group_infos.push({LocalAdmissionController::global_instance->getPriority(name), task_queue, name});
+        pipeline_tasks.insert({name, task_queue});
```
guo-shaoge, line 59: rename to resource_group_task_queues

dbms/src/Flash/Pipeline/Schedule/TaskQueues/tests/gtest_resource_control_queue.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+// 1. Tasks of different resource groups is restricted by RU, and there is still available cpu resource.
+// 2. Change rg-1 as burstable, the available cpu resource is used by rg1.
+// 3. Change other rg as burstable, the cpu usage should be same of all resource groups.
+TEST_F(TestResourceControlQueue, TestBurstable)
+{
+}
+
+TEST_F(TestResourceControlQueue, TestAddAndDelResourceGroup)
+{
```
guo-shaoge, line 357: todo test default resource group

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,107 @@
...
+    bool is_finished = false;
+
+    CompatorType compator = compareResourceInfo;
+    ResourceGroupInfoQueue resource_group_infos{compator};
+    PipelineTasks pipeline_tasks;
+
+    // <resource_group_name, acculumated_cpu_time>
+    // when acculumated_cpu_time >= YIELD_MAX_TIME_SPENT_NS, will update resource group.
+    // This is to prevent the resource group from being updated too frequently.
+    std::unordered_map<std::string, UInt64> resource_group_statistic;
```
guo-shaoge, line 103: todo support delete

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,259 @@
...
+            break;
+        }
+        // todo: remove item from cancel_query_ids when this query is cancelled successfully,
+        // otherwise we may need to iterate all nested task queue to check if cancel task is empty or not.
+    }
+    return task != nullptr;
+}
+
+template <typename NestedQueueType>
+void ResourceControlQueue<NestedQueueType>::updateStatistics(const TaskPtr & task, ExecTaskStatus, size_t inc_value)
```
guo-shaoge, line 157: todo maybe change TaskThreadPool code, so only call updateStatistics each 100ms

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,107 @@
...
+    static constexpr auto DEFAULT_WAIT_INTERVAL_WHEN_RUN_OUT_OF_RU = std::chrono::seconds(1);
+
+    // ResourceGroupInfoQueue compator.
+    static bool compareResourceInfo(const ResourceGroupInfo & info1, const ResourceGroupInfo & info2)
+    {
+        auto priority1 = std::get<InfoIndexPriority>(info1);
+        auto priority2 = std::get<InfoIndexPriority>(info2);
+
+        // Return true means lower priority.
+        // Here we want make negative priority to be lower priority than positive priority.
```
guo-shaoge, line 72: todo add gtest

dbms/src/Flash/Pipeline/Schedule/TaskQueues/TaskQueueType.h
```diff
@@ -22,5 +22,7 @@ enum class TaskQueueType
     DEFAULT, // Determined internally by the task thread pool.
     MLFQ, // multi-level feedback queue
     IO_PRIORITY, // io priority queue
+    RCQ_MLFQ, // resource control queue nesting MLFQ.
+    RCQ_IO_PRIORITY, // resource control queue nesting IO_PRIORITY.
```
SeaRise, line 26: It seems that IOThreadPool should not use ResourceGroupQueue, because TiDB's resource control does not currently include io, and the execution in IOThreadPool basically does not consume cpu

dbms/src/Flash/Pipeline/Schedule/TaskQueues/TaskQueue.h
```diff
@@ -44,7 +44,7 @@ class TaskQueue
     // And the tasks in the queue can still be taken normally.
     virtual void finish() = 0;
 
-    virtual void cancel(const String & query_id) = 0;
+    virtual void cancel(const String & query_id, const String & resource_group_name) = 0;
```
SeaRise, line 47: How about maintaining a cancel queue in ResourceGroupQueue?

dbms/src/Flash/Pipeline/Schedule/Tasks/TaskHelper.h
```diff
@@ -52,4 +52,10 @@ namespace DB
 
 static constexpr int64_t YIELD_MAX_TIME_SPENT_NS = 100'000'000L;
 
+// The executing task should yield if it takes more than `YIELD_MAX_TIME_SPENT_NS`.
+inline bool pipelineTaskTimeExceedYieldThreshold(UInt64 total_time_spent)
```
SeaRise, line 56:
```suggestion
inline bool timeExceedYieldThreshold(UInt64 total_time_spent)
```

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPoolImpl.cpp
```diff
@@ -27,6 +28,10 @@ TaskQueuePtr CPUImpl::newTaskQueue(TaskQueueType type)
     {
     // the default queue is mlfq queue.
     case TaskQueueType::DEFAULT:
+#ifdef DBMS_PUBLIC_GTEST
+    case TaskQueueType::RCQ_MLFQ:
+        return std::make_unique<ResourceControlQueue<CPUMultiLevelFeedbackQueue>>();
+#endif
```
SeaRise, line 34: If the ResourceGroupQueue is temporarily unavailable, it can be set to non-default?

guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,107 @@
...
+    CompatorType compator = compareResourceInfo;
+    ResourceGroupInfoQueue resource_group_infos{compator};
+    ResourceGroupTaskQueue resource_group_task_queues;
+
+    // <resource_group_name, acculumated_cpu_time>
+    // when acculumated_cpu_time >= YIELD_MAX_TIME_SPENT_NS, will update resource group.
+    // This is to prevent the resource group from being updated too frequently.
+    std::unordered_map<std::string, UInt64> resource_group_statistic;
+
+    std::unordered_map<std::string, std::shared_ptr<NestedQueueType>> cancel_query_ids;
```
SeaRise, line 105: Maybe it would be simpler to maintain a cancel_queue in `ResourceControlQueue` like MLFQ.
```
bool take(TaskPtr & task) override
{
    if (!cancel_queue.empty())
        return popTask(cancel_queue);
}

void cancel(const String & query_id) override
{
    for (const auto & queue : resource_group_task_queues)
    {
        queue.moveCancelledTasks(query_id, cancel_queue);
    }
}
```
The time complexity of cancel is O(n), same as MLFQ, n is the total number of tasks.
This greatly simplifies the code of ResourceControlQueue.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -0,0 +1,27 @@
...
+namespace DB
+{
+
+#ifndef DBMS_PUBLIC_GTEST
+std::unique_ptr<LocalAdmissionController> LocalAdmissionController::global_instance;
+#else
+auto LocalAdmissionController::global_instance = std::make_unique<MockLocalAdmissionController>();
+#endif
+
+const std::string ResourceGroup::DEFAULT_RESOURCE_GROUP_NAME = "default";
```
SeaRise, line 26: If resource-group-name in dag-request is empty, will it be assigned "default" in tiflash?

guo-shaoge: tiflash should can handle resource_group_name is empy.

guo-shaoge: if resource_group_name is empty, it means resource control is disabled, we should not ask GAC.
LAC should just ignore it(`getPriority() just return 1; computeResource() should do nothing and isResourceGroupThrottled() should return false`)

guo-shaoge: todo next LAC pr will handle this!

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+    const auto & query_id = task->getQueryId();
+    if unlikely (cancel_query_id_cache.contains(query_id))
+    {
+        cancel_task_queue.push_back(std::move(task));
+        return;
+    }
+
+    // name can be empty, it means resource control is disabled.
+    const std::string & name = task->getResourceGroupName();
```
SeaRise, line 57: Who has higher priority, "" or "default"?
If the resource control is turned on at runtime, both "" and default will appear.
Need to treat "" as "default"?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,101 @@
...
+
+#ifndef DBMS_PUBLIC_GTEST
+private:
+#endif
+    // <resource_group_name, resource_group_task_queues>
+    using ResourceGroupTaskQueue = std::unordered_map<std::string, std::shared_ptr<NestedQueueType>>;
+
+    // <priority, corresponding_task_queue, resource_group_name>
+    using ResourceGroupInfo = std::tuple<UInt64, std::shared_ptr<NestedQueueType>, std::string>;
+    using CompatorType = bool (*)(const ResourceGroupInfo &, const ResourceGroupInfo &);
```
SeaRise, line 57: Maybe wrapping a class would be better?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+}
+
+template <typename NestedQueueType>
+bool ResourceControlQueue<NestedQueueType>::take(TaskPtr & task)
+{
+    assert(task == nullptr);
+    std::unique_lock lock(mu);
+    while (true)
+    {
+        if (popTask(cancel_task_queue, task))
```
SeaRise, line 82: Should check `finished` before?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: "default" resource group will talk with GAC to fetch tokens as other resource group.
But ""(when resource control is disabled) should not.

guo-shaoge: In the next LAC pr, priority of "" is always `1.0`, which is higher than "default".

SeaRise: Does this mean that "" has the highest priority, even if the user has set a resource group with the highest priority?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: Maybe too many `is_finished` check? We already have two:
1. In `while (!resource_groups_info.empty())`
2. Before calling `cv.wait_for(lock, DEFAULT_WAIT_INTERVAL_WHEN_RUN_OUT_OF_RU)`

IMO, whether to check `is_finished` before `pop cancel task queue` has little impact.

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPoolImpl.cpp
```diff
@@ -25,8 +26,10 @@ TaskQueuePtr CPUImpl::newTaskQueue(TaskQueueType type)
 {
     switch (type)
     {
-    // the default queue is mlfq queue.
+    // the default queue is RC_MLFQ.
     case TaskQueueType::DEFAULT:
+    case TaskQueueType::RCQ_MLFQ:
```
guo-shaoge, line 31: todo maybe add a new config

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: Yes

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: But not checking "finished" contradicts the semantics of "take".
https://github.com/pingcap/tiflash/blob/6f7c6472e1b59abb7e101f3b708094e81d52c568/dbms/src/Flash/Pipeline/Schedule/TaskQueues/TaskQueue.h#L43-L44

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: OK, I have some doubts, does tidb/tikv handle it in the same way?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: Maybe checking is_finished in L81 is enough, like MLFQ
https://github.com/pingcap/tiflash/blob/6f7c6472e1b59abb7e101f3b708094e81d52c568/dbms/src/Flash/Pipeline/Schedule/TaskQueues/MultiLevelFeedbackQueue.cpp#L178-L180

SeaRise: Since the mutex is used in finish, it doesn't make sense to check is_finished in the second while.
There is condition_cv.wait in the first while, which can give up the control of mutex

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+                logger,
+                "trying to schedule task of resource group {}, priority: {}, is_finished: {}, task_queue.empty(): {}",
+                name,
+                priority,
+                is_finished,
+                task_queue->empty());
+
+            // When highest priority of resource group is less than zero, means RU of all resource groups are exhausted.
+            // Should not take any task from nested task queue for this situation.
+            if (LocalAdmissionController::isRUExhausted(priority))
```
SeaRise, line 109: Does burst resource group also included?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+                break;
+
+            if (task_queue->empty() || !task_queue->take(task))
```
SeaRise, line 112: When will task_queue->take return false?
it seems impossible that this is the case
Add a runtime check?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+                    "cannot erase corresponding TaskQueue for task of resource group {}",
+                    name);
+            }
+            else
+            {
+                LOG_TRACE(
+                    logger,
+                    "schedule task of resource group {} succeed, cur cpu time of MPPTask: {}",
+                    name,
+                    task->getQueryExecContext().getQueryProfileInfo().getCPUExecuteTimeNs());
```
SeaRise, line 134: Getting `task->getQueryExecContext().getQueryProfileInfo().getCPUExecuteTimeNs()` is meaningless, because this value will not be updated at runtime, only when the task ends

SeaRise: Or `task->profile_info.getCPUExecuteTimeNs()`?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+                break;
+            }
+        }
+
+        if (task != nullptr)
+            return true;
+
+        // Check is_finished again for situation when resource_group_infos never insert any resource group.
+        if unlikely (is_finished)
+            return false;
```
SeaRise, line 145: Checking again seems meaningless, because the mutex has not been released at this time, is_finished is still the original value

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,265 @@
...
+    std::lock_guard lock(mu);
+    auto iter = resource_group_statistic.find(name);
+    if (iter == resource_group_statistic.end())
+    {
+        UInt64 accumulated_cpu_time = inc_value;
+        if (accumulated_cpu_time >= YIELD_MAX_TIME_SPENT_NS)
+        {
+            updateResourceGroupStatisticWithoutLock(name, accumulated_cpu_time);
+            accumulated_cpu_time = 0;
+        }
```
SeaRise, line 170: The cache value is a bit weird...how about doing that todo directly `maybe change TaskThreadPool code, so only call updateStatistics each 100ms`

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,101 @@
...
+    using ResourceGroupInfoQueue = std::priority_queue<ResourceGroupInfo, std::vector<ResourceGroupInfo>, CompatorType>;
+
+    // Index of ResourceGroupInfo.
+    static constexpr auto InfoIndexPriority = 0;
+    static constexpr auto InfoIndexPipelineTaskQueue = 1;
+    static constexpr auto InfoIndexResourceGroupName = 2;
```
SeaRise, line 63: I tend to use struct instead of tuple

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
```diff
@@ -0,0 +1,101 @@
...
+    static constexpr auto DEFAULT_WAIT_INTERVAL_WHEN_RUN_OUT_OF_RU = std::chrono::seconds(1);
```
SeaRise, line 64: Why is this needed?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: ok

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: Maybe cache the inc_value in the LAC?
In fact, the LAC itself will cache the local token. Is it meaningful to cache inc_value in the Resource Group queue?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Flash/ResourceControl/LocalAdmissionController.h>
+
+namespace DB
+{
+
+#ifndef DBMS_PUBLIC_GTEST
+std::unique_ptr<LocalAdmissionController> LocalAdmissionController::global_instance;
```
SeaRise, line 21: So global_instance is null for prod environment?

dbms/src/Flash/Executor/toRU.h
```diff
@@ -19,6 +19,10 @@
 namespace DB
 {
 using RU = UInt64;
+
+// Convert cpu time nanoseconds to cpu time millisecond, and round up.
+UInt64 toCPUTimeMillisecond(UInt64 cpu_time_ns);
+
 // Convert cpu time nanoseconds to Request Unit.
 RU toRU(UInt64 cpu_time_ns);
```
SeaRise, line 27: Is RU actually a double?
Was it a bug to use UInt64 before? Do we need to fix it in this pr?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: For bustable rg, `ResourceGroup::getPriority()` will not return special priority(which is `uint64_max`, which means ru is exhausted). 
So in theoty, `isRUExhausted()` always return false for burstable rg.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: Yes. It make sense to add a runtime check

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: ok

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: Next pr will update this value. For this pr, `MockLocalAdmissionController` is used

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: ok

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
guo-shaoge: When all RG runs out of RU, wait for `DEFAULT_WAIT_INTERVAL_WHEN_RUN_OUT_OF_RU` to let LAC fetch tokens from GAC.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: ok

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
SeaRise: https://github.com/pingcap/tiflash/pull/7917/files#diff-6e66cd26a4a06ad28fe982176727d5cf64f7b9c6ed0b167a64dcf68d14642882R30-R32
ResourceGroupQueue seems to be turned on by default, will it cause CI/daily to fail

dbms/src/Flash/Executor/toRU.h
guo-shaoge: I think this pr use UInt64 is still ok.
Next pr need to update to double, because next pr will report RU usage to PD, and the pb between LAC and PD use double type for RU consumption.

SeaRise: ok

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: good idea

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
SeaRise: How about proactively notifying `ResourceGroupQueue` after `LAC` gets the token?

SeaRise: If ResourceGroupQueue waits exactly at 4.9s, then ResourceGroupQueue will wait 0.9s longer.
LAC   ----------
RCQ              -------

SeaRise: Adding a callback to LAC, in order to notify RCQ after fetching the token?
```
LAC.register([&]() {
    condition_cv.notify();
})
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: CI is ok, because LocalAdmissionControl::getPriority is simple. So will not coredump even if it's not init.
But it's not reasonable. So i changed it.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: 1. `call updateStatistics each 100ms` this may affects mlfq
2. move `inc_value` to LAC means every time `updateStaticstics` is called, LAC need lock to add `inc_value` to `accmulated_cpu_time`. 
After discuss, leave current implementation unchanged.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
guo-shaoge: changed

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.h
guo-shaoge: Should register callback to LAC, so we can delete when RG is deleted.

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef DBMS_PUBLIC_GTEST
+private:
+#endif
+    // Priority of resource group set by user.
+    // This is specified by tidb: parser/model/model.go
+    static constexpr int32_t LowPriorityValue = 1;
+    static constexpr int32_t MediumPriorityValue = 8;
+    static constexpr int32_t HighPriorityValue = 16;
+
+    // Minus 1 because uint64 max is used as special flag.
```
windtalker, line 78: `uint64 max` as special flag or `uint64 max`>>4 as a special flag?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -0,0 +1,173 @@
...
+
+    mutable std::mutex mu;
+
+    // Local token bucket.
+    TokenBucketPtr bucket;
+
+    // Total used cpu_time_in_ns of this ResourceGroup.
+    uint64_t cpu_time_in_ns = 0;
+
+    double ru_consumption_delta = 0.0;
```
windtalker, line 159: Where is this used?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Flash/ResourceControl/LocalAdmissionController.h>
+
+namespace DB
+{
+
+auto LocalAdmissionController::global_instance = std::make_unique<MockLocalAdmissionController>();
```
windtalker, line 20: So this pr only implement `MockLocalAdmissionController`?

guo-shaoge: Yes, next pr will implement real `LocalAdmissionController`, which will communicate to GAC to fetch GAC.

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: For now tiflash report RU usage to prometheus directly. Later we need to report to PD, just like tidb does.

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.cpp
```diff
@@ -0,0 +1,66 @@
...
+            if (cv.wait_for(lock, std::chrono::seconds(1), [&]() { return stopped; }))
+                return;
+
+            resource_group_empty = resource_groups.empty();
+            for (auto & ele : resource_groups)
+            {
+                auto & rg = ele.second;
+                if (rg->bucket->isStatic())
+                {
+                    std::cout << "refill token bucket of rg " << rg->name << std::endl;
```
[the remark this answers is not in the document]
guo-shaoge: Only for debug. It's ok to delete this line.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    std::lock_guard lock(mu);
+    auto ru = toRU(inc_value);
+    LOG_TRACE(logger, "resource group {} will consume {} RU(or {} cpu time in ns)", name, ru, inc_value);
+    LocalAdmissionController::global_instance->consumeResource(name, ru, inc_value);
+    updateResourceGroupInfosWithoutLock();
+}
+
+template <typename NestedTaskQueueType>
+void ResourceControlQueue<NestedTaskQueueType>::updateResourceGroupInfosWithoutLock()
```
SeaRise, line 158: It seems that `updateResourceGroupInfosWithoutLock` just needs to be called before take.
No need to call in both `take` and `updateStatistics`

SeaRise:
```
while (true)
{
    ...
    updateResourceGroupInfosWithoutLock();
    while (!resource_group_infos.empty())
    {
        ...
    }
    cv.wait(lock);
}
```

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+    std::priority_queue<ResourceGroupInfo> new_resource_group_infos;
+    while (!resource_group_infos.empty())
+    {
+        const ResourceGroupInfo & group_info = resource_group_infos.top();
+        auto new_priority = LocalAdmissionController::global_instance->getPriority(group_info.name);
+        new_resource_group_infos.push({group_info.name, new_priority, group_info.task_queue});
+        resource_group_infos.pop();
+    }
+    resource_group_infos = new_resource_group_infos;
```
SeaRise, line 168:
```suggestion
    resource_group_infos = std::move(new_resource_group_infos);
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: Looks `uint64_max >> 4` is more reasonable, because user_priority cannot be zero. Changed

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: Maybe it would be simpler to use `std::vector<ResourceGroupInfo>` and `std::sort`

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,223 @@
...
+void ResourceControlQueue<NestedTaskQueueType>::submit(std::vector<TaskPtr> & tasks)
+{
+    {
+        std::lock_guard lock(mu);
+        for (auto & task : tasks)
+        {
+            submitWithoutLock(std::move(task));
+        }
+    }
+    cv.notify_one();
```
SeaRise, line 44: Please call `notify_one` multiple times in `for` like `MLFQ`, or call `notify_all` here.
Because there may be multiple threads waiting at this time.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,223 @@
...
+                    erase_num);
+                resource_group_infos.pop();
+            }
+            else
+            {
+                // Take task from nested task queue, and should always take succeed.
+                // Because this task queue should not be finished inside lock_guard.
+                RUNTIME_CHECK(group_info.task_queue->take(task));
+                assert(task != nullptr);
+                break;
```
SeaRise, line 129: return true directly?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: And then resource_group_infos  can be sorted directly, no need to create new container every time.

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.cpp
```diff
@@ -0,0 +1,63 @@
...
+                    const auto bucket_config = rg->bucket->getCurrentConfig();
+                    const double new_fill_rate = std::get<1>(bucket_config);
+                    const double new_capacity = std::get<2>(bucket_config);
+                    // Use capacity as new token, because for static token bucket, its fill rate is zero.
+                    const double new_token = rg->bucket->peek() + std::get<2>(bucket_config);
```
SeaRise, line 38:
```suggestion
                    std::tie(new_fill_rate, new_capacity, new_token) = rg->bucket->getCurrentConfig();
                    // Use capacity as new token, because for static token bucket, its fill rate is zero.
                    new_token += rg->bucket->peek()
```

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: If put `updateResourceGroupInfosWithoutLock()` only in `take()`, it means every time we call `take()`, need to get priority of each rg and reorder priority queue.

For current implement, only needs to reorder priority queue when RU is exhausted.

guo-shaoge: Current impl:
1. take()
2. call updateResourceGroupInfosWithLock() in updateStatistics()
3. submit()

After change:
1. updateResourceGroupInfosWithLock()
2. take()
4. submit()

So in the current implementation, submit() basically use the newest priority info to push priority queue.

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: But using `uint64 max` is more convenient to implement the comparator of priority_queue. No need to add special if branch to check special flag in the comparator.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: Using a vector can cause the deletion operation to be more cumbersome

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.h
```diff
@@ -0,0 +1,104 @@
...
+class ResourceGroup;
+inline void nopConsumeResource(const std::string &, double, uint64_t) {}
+inline uint64_t nopGetPriority(const std::string &)
+{
+    return 1.0;
+}
+inline bool nopIsResourceGroupThrottled(const std::string &)
+{
+    return false;
+}
```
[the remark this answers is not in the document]
guo-shaoge: "" resource group use this to get its priority.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: ok

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.h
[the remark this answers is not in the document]
SeaRise: ok

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.cpp
guo-shaoge: changed to structured binding

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.cpp
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        refill_token_callback();
+    }
+}
+
+std::string MockLocalAdmissionController::dump() const
+{
+    std::string info_str;
```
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -98,14 +98,13 @@ void TaskThreadPool<Impl>::handleTask(TaskPtr & task)
...
-        auto inc_time_spent = task->profile_info.elapsedFromPrev();
-        task_queue->updateStatistics(task, cur_status, inc_time_spent);
+        total_time_spent += task->profile_info.elapsedFromPrev();
         cur_status = after_status;
-        total_time_spent += inc_time_spent;
         // The executing task should yield if it takes more than `YIELD_MAX_TIME_SPENT_NS`.
         if (!Impl::isTargetStatus(cur_status) || total_time_spent >= YIELD_MAX_TIME_SPENT_NS)
             break;
     }
+    task_queue->updateStatistics(task, cur_status, total_time_spent);
```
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/ResourceControl/TokenBucket.cpp
```diff
@@ -0,0 +1,95 @@
...
+    RUNTIME_CHECK(new_fill_rate >= 0.0);
+    RUNTIME_CHECK(new_capacity >= 0.0);
+
+    auto now = std::chrono::steady_clock::now();
+    tokens = new_tokens;
+    fill_rate = new_fill_rate;
+    capacity = new_capacity;
+
+    compact(now);
+    // Update because token number may increase, which may cause token_changed be nigative.
```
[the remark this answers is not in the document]
guo-shaoge: done

SeaRise (commented):
Others LGTM

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -0,0 +1,223 @@
...
+                // Take task from nested task queue, and should always take succeed.
+                // Because this task queue should not be finished inside lock_guard.
+                RUNTIME_CHECK(group_info.task_queue->take(task));
+                assert(task != nullptr);
+                return true;
+            }
+        }
+
+        if (task != nullptr)
+            return true;
```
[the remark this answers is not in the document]
guo-shaoge: If `resource_group_infos` is empty, task is nullptr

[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Common/tests/gtest_threaded_worker.cpp
```diff
@@ -296,23 +294,10 @@ TEST(ThreadedWorker, ErrorInWorkerWithNonEmptyQueue)
...
-    result = w.result_queue->pop(r);
-    ASSERT_EQ(result, MPMCQueueResult::OK);
-
-    result = w.result_queue->pop(r);
-    ASSERT_EQ(result, MPMCQueueResult::OK);
-
-    // Because the concurrency is 2, there could be a chance that some new result
-    // is pushed into the result_queue before exception cancel all the wokers.
-    // As long as the queue is cancelled before all elements are handled, consider
-    // this test passed.
```
SeaRise, line 311: /cc @JaySon-Huang

SeaRise (approved):
LGTM

windtalker (approved):
LGTM

PR #7927, opened 2023-08-10
Support LocalAdmissionController for tiflash resource control

### LINKED ISSUE #7918: Support LocalAdmissionControl
## Feature Request

guo-shaoge:

Issue Number: close #7918

Problem Summary:

1. LocalAdmissionController:
    1. Main member function:
        1. consumeResource()
        2. getPriority()
        3. checkResourceGroupValid()
    2. LAC manages all ResourceGroups in tiflash:
        1. Creation: When MPPTask of specific RG comes, will create new resource group info
        2. Deletion: Will watch etcd Delete event and remove resource group info
        3. Update: Will watch etcd Put event and update config of resource group
    3. Two background thread:
        1. bg-1:
            1. call fetchTokensFromGAC() periodically or low token threshold
            2. call checkDegradeMode() when the network between PD is down
            3. call callbacks(refill_token) periodically
         2. bg-2: watchGAC()
2. changes of min tso is in next pr.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test: setup tpch1G dataset and test two RG's isolation
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -128,16 +141,109 @@ class ResourceGroup final : private boost::noncopyable
         return priority;
     }
 
-    void initStaticTokenBucket(uint64_t capacity = std::numeric_limits<uint64_t>::max())
+    bool lowToken() const
```
guo-shaoge, line 144: todo 
1. maybe reorder function
2. split ResourceGroup into another file

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,18 +263,195 @@ class ResourceGroup final : private boost::noncopyable
...
+#endif
+
+    // gjt todo private?
+    // Interval of fetch from GAC periodically.
+    static constexpr uint64_t DEFAULT_FETCH_GAC_INTERVAL = 5;
+    // DEFAULT_TOKEN_FETCH_ESAPSED * token_avg_consumption_speed as token num to fetch from GAC.
+    static constexpr uint64_t DEFAULT_TOKEN_FETCH_ESAPSED = 5;
+    // If we cannot get GAC resp for DEGRADE_MODE_DURATION seconds, enter degrade mode.
+    static constexpr auto DEGRADE_MODE_DURATION = 120;
+    static constexpr auto TARGET_REQUEST_PERIOD_MS = 5000;
```
guo-shaoge, line 355: todo check GAC of this meaing

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -578,6 +579,14 @@ void MPPTask::runImpl()
             runtime_statistics.blocks,
             runtime_statistics.bytes);
 
+        // Only report RU consumption to GAC for now.
+        // Will not affect the execution process of MPPTask,
+        // as the consult operation will only be performed after MPPTask is completed.
+        // NOTE: May throw if resource group is deleted.
```
SeaRise, line 585: `May throw if resource group is deleted.`
catch and ignore the exception?

dbms/src/Flash/Mpp/MPPTask.cpp
```diff
@@ -578,6 +579,14 @@ void MPPTask::runImpl()
...
+        LocalAdmissionController::global_instance->consumeResource(
+            dag_context->getResourceGroupName(),
+            toRU(read_ru),
+            0);
```
SeaRise, line 589: cop/batch_cop/streaming_cop also needs this

dbms/src/Flash/Mpp/MPPTaskId.h
```diff
@@ -27,17 +27,28 @@ struct MPPQueryId
     UInt64 local_query_id;
     UInt64 server_id;
     UInt64 start_ts;
-    MPPQueryId(UInt64 query_ts, UInt64 local_query_id, UInt64 server_id, UInt64 start_ts)
+    // 1. For toString(), so we can check the resource group in log.
+    // 2. When release min tso threads, it's easy to find corresponding MinTSOScheduler by query_id.resource_group_name.
```
SeaRise, line 31: Comments look useless

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -104,9 +121,16 @@ void TaskThreadPool<Impl>::handleTask(TaskPtr & task)
         if (!Impl::isTargetStatus(status_after_exec) || total_time_spent >= YIELD_MAX_TIME_SPENT_NS)
             break;
     }
-    task_queue->updateStatistics(task, status_before_exec, total_time_spent);
     metrics.addExecuteTime(task, total_time_spent);
     metrics.decExecutingTask();
+    // updateStatistics may throw exception, so need to update metrics first.
```
SeaRise, line 126: Why does `updateStatistics` throw an error? This will cause tiflash crash

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -74,18 +74,35 @@ void TaskThreadPool<Impl>::doLoop(size_t thread_no)
...
     TaskPtr task;
-    while (likely(task_queue->take(task)))
+    ExecTaskStatus status_after_exec;
+    while (true)
     {
-        metrics.decPendingTask();
-        handleTask(task);
+        try
+        {
+            if unlikely (!task_queue->take(task))
```
SeaRise, line 82: The function of taskQueue should not throw errors

SeaRise: All methods called in TaskThreadPool should not throw errors

dbms/src/Server/Server.cpp
```diff
@@ -1545,6 +1544,21 @@ int Server::main(const std::vector<std::string> & /*args*/)
...
+    LOG_WARNING(
+        log,
+        "Compiled with ENABLE_TESTS enabled, will use MockLocalAdmissionController, which means tiflash "
+        "ResourceControl will be disabled.");
+#else
+    auto mpp_task_manager = tmt_context.getMPPTaskManager();
+    LocalAdmissionController::global_instance
+        = std::make_unique<LocalAdmissionController>(tmt_context.getKVCluster(), tmt_context.getEtcdClient());
+    LocalAdmissionController::global_instance->registerDeleteResourceGroupCallback(
+        [&mpp_task_manager]() { mpp_task_manager->deleteEmptyScheduler(); });
```
SeaRise, line 1559: Perhaps passing TMTContext as an argument to the TaskScheduler constructor would eliminate the need for code modification.
` TaskScheduler::instance = std::make_unique<TaskScheduler>(config, tmt_context);`

guo-shaoge:
need to fix ci first

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
guo-shaoge: During the execution of Pipeline Tasks, if resource group is deleted, we have the following options:
1. report error(resource group not found)
    1. throw exception in TaskQueue::take()/submit()/updatestatistics9)
    2. set task a flag, so when exec task, report error
2. let tasks of this query finish without consuming any RU.

Current implement choose 1.1

dbms/src/Server/Server.cpp
guo-shaoge: `LAC` should be destructed after `MPPTaskManager`(so it can `unregisterDeleteResourceGroup()` successfully).
So it will make things complicated if bind the lifetime of `LAC` with `TaskScheduler`.

dbms/src/Flash/Executor/PipelineExecutor.cpp
```diff
@@ -137,7 +137,7 @@ RU PipelineExecutor::collectRequestUnit()
     // It may be necessary to obtain CPU time using a more accurate method, such as using system call `clock_gettime`.
     const auto & query_profile_info = exec_context.getQueryProfileInfo();
     auto cpu_time_ns = query_profile_info.getCPUExecuteTimeNs();
-    return toRU(ceil(cpu_time_ns));
+    return toRU(cpu_time_ns);
```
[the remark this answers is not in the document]
guo-shaoge: maybe no need to delete

dbms/src/Flash/Mpp/MPPTask.cpp
guo-shaoge: done

dbms/src/Flash/Mpp/MPPTask.cpp
guo-shaoge: done

dbms/src/Flash/Mpp/MPPTaskId.h
guo-shaoge: done

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
guo-shaoge: take()/updateStatistics()/submit() will handle exception inside.

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
guo-shaoge: take()/updateStatistics()/submit() will handle exception inside.

dbms/src/Server/Server.cpp
```diff
@@ -1546,6 +1545,18 @@ int Server::main(const std::vector<std::string> & /*args*/)
         }
     });
 
+    auto & tmt_context = global_context->getTMTContext();
+    // Resource Control.
+#ifdef DBMS_PUBLIC_GTEST
+    LOG_WARNING(
+        log,
+        "Compiled with ENABLE_TESTS enabled, will use MockLocalAdmissionController, which means tiflash "
+        "ResourceControl will be disabled.");
```
SeaRise, line 1554: Why is this needed?
What's the problem with getting rid of it?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,245 @@ class ResourceGroup final : private boost::noncopyable
...
+
+        // todo: refine: ResourceControlQueue can record the statics of running pipeline task.
+        // If the proportion of scheduled pipeline task is less than 10%, we say this resource group is throttled.
+        ResourceGroupPtr group = findResourceGroup(name);
+        if (group == nullptr)
+            return false;
+        return group->getRU() <= 0.0;
+    }
+
+    std::string isResourceGroupValid(const std::string & name)
```
SeaRise, line 417: Where is this function used?
And it's a bit strange that isXXX returns a string.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,491 @@
...
+void LocalAdmissionController::startBackgroudJob()
+{
+    try
+    {
+        unique_client_id = etcd_client->acquireServerIDFromPD();
+    }
+    catch (...)
+    {
+        auto err_msg = getCurrentExceptionMessage(false);
+        LOG_FATAL(log, err_msg);
```
SeaRise, line 58: Need to crash tiflash?
What happens if tiflash continues to run?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,491 @@
...
+        // 3. watch is cancel.
+        // Will sleep and try again.
+        std::unique_lock lock(mu);
+        if (cv.wait_for(lock, std::chrono::seconds(30), [this]() { return stopped.load(); }))
+            return;
+    }
 }
 
-LocalAdmissionController::~LocalAdmissionController()
+void LocalAdmissionController::doWatch()
```
SeaRise, line 346: Why is watch gac needed here?

SeaRise: Isn’t the fetch resource group and refill bucket token pinging the GAC?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -106,28 +132,24 @@ bool ResourceControlQueue<NestedTaskQueueType>::take(TaskPtr & task)
...
-                    group_info.name,
-                    erase_num);
-                resource_group_infos.pop();
+                auto iter = error_resource_groups.find(group_info.name);
+                if likely (iter != error_resource_groups.end())
+                {
+                    mustTakeTask(group_info.task_queue, task);
+                    lock.unlock();
+                    task->onErrorOccurred(iter->second);
+                    lock.lock();
```
SeaRise, line 143: It doesn't feel necessary...lock and unlock also have overhead

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -24,62 +24,88 @@ namespace DB
...
+                NestedTaskQueuePtr task_queue = getSubmitTaskQueuWithoutLock(task);
+                if likely (task_queue)
+                    task_queue->submit(std::move(task));
+            }
+            catch (...)
+            {
+                assert(task);
+                lock.unlock();
+                task->onErrorOccurred(std::current_exception());
+                lock.lock();
```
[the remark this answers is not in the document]
guo-shaoge: Will dead lock if dont unlock.
1. task->onErrorOccurred()
2. exec_context.onErrorOccurred()
3. task_scheduler.cancel()
4. RCQ::cancel(): will use the same lock as submit()

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: We must have correct `unique_client_id` for GAC(GAC use this id to distinguish different LAC and balance token).

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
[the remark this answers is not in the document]
SeaRise: how about
```
bool xxx;
{
      yyy
}
if (xxx)
    task->onErrorOccurred(zzz);
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: Or maybe we can deadloop retry(sleep X second before next retry) and LOG_ERROR when got exception?

SeaRise: So it looks like `std::terminate()` need to call here

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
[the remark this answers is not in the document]
guo-shaoge: You mean like the one task version(`TaskQueue::submit(TaskPtr && task)`)?

How about:
```
for (task : tasks)
{
    submit(std::move(task));
}
```

dbms/src/Server/Server.cpp
guo-shaoge: If you compile tiflash server using `cmake .. -DENABLE_TESTS=ON`, `DBMS_PUBLIC_GTESTS` will be defined. So `MockLocalAdmissionController` will be used by `RCQ` instead of the real `LAC`.

`MockLAC` will not talk to `GAC` and it return same priority for all pipeline tasks, this is to remaind this behavior.

SeaRise: Does `DB::LocalAdmissionController::global_instance = std::make_unique<DB::MockLocalAdmissionController>();` need to be call here?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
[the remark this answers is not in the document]
SeaRise: Just saying not to use lock, unlock. You can call onError after the {}.
```
std::vector<TaskPtr> error_tasks
{}
for (auto & task : error_tasks)
    task->onError
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: `watchGAC` is to watch the `deletion` and `config change` of resource group.

We can fetch these info from GAC periodically, but I think it's no accurate. And tidb/tikv also watch etcd for resource group creation/deletion/change

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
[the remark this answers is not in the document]
guo-shaoge: get

dbms/src/Server/Server.cpp
guo-shaoge: It's called [here](https://github.com/pingcap/tiflash/pull/7927/files/25cb1a0ccf60667dd2e2e284a8ba75eabc1661c0#diff-7688037759fb8f57bcf85cdcb13657390b20b0cf711f8c15ebba7773e529f4d3R494). Because gtest will not compile `Server.cpp`.

SeaRise: If Server will not be compiled on DBMS_PUBLIC_GTEST, is it possible to remove this code?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
SeaRise: ok

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: ok. I thought `LOG_FATAL` would automatically crash for  looks I'm wrong.

dbms/src/Server/Server.cpp
guo-shaoge: Yes. It's ok to remove this log if Server will not be compiled on DBMS_PUBLIC_GTEST.
But how can we disable this behavior? static_assert?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: dup

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: `isResourceGroupThrottled` and `isResourceGroupValid` removed. It's used for the former code of min tso.

dbms/src/Server/Server.cpp
SeaRise: Maybe it's not necessary to consider.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -24,62 +24,92 @@ namespace DB
...
         for (auto & task : tasks)
         {
-            submitWithoutLock(std::move(task));
-            cv.notify_one();
+            try
+            {
+                // May throw resource group not found exception.
+                NestedTaskQueuePtr task_queue = getSubmitTaskQueuWithoutLock(task);
+                if likely (task_queue)
+                    task_queue->submit(std::move(task));
```
SeaRise, line 61: You can use `// NOLINT(bugprone-use-after-move)`, so you don't need to modify the code

SeaRise:
```suggestion
                submitWithoutLock(std::move(task));  // NOLINT(bugprone-use-after-move)
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,491 @@
...
+        return group;
+
+    resource_manager::GetResourceGroupRequest req;
+    req.set_resource_group_name(name);
+    resource_manager::GetResourceGroupResponse resp;
+
+    try
+    {
+        resp = cluster->pd_client->getResourceGroup(req);
+    }
```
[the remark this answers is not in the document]
guo-shaoge: Yes, it may happens.

dbms/src/Server/Server.cpp
```diff
@@ -1546,6 +1545,18 @@ int Server::main(const std::vector<std::string> & /*args*/)
         }
     });
 
+    auto & tmt_context = global_context->getTMTContext();
+    // Resource Control.
+#ifdef DBMS_PUBLIC_GTEST
```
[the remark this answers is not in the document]
guo-shaoge: If you compile tiflash binary using `cmake ../ -DENABLE_TESTS=ON && make tiflash`, `DBMS_PUBLIC_GTEST` will be defined in Server.cpp.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
guo-shaoge: I think it's dangerous to do this?
After std::move, the state of Task is unspecified.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -106,28 +136,24 @@ bool ResourceControlQueue<NestedTaskQueueType>::take(TaskPtr & task)
...
-                    group_info.name,
-                    erase_num);
-                resource_group_infos.pop();
+                auto iter = error_resource_groups.find(group_info.name);
+                if likely (iter != error_resource_groups.end())
+                {
+                    mustTakeTask(group_info.task_queue, task);
+                    lock.unlock();
+                    task->onErrorOccurred(iter->second);
+                    lock.lock();
```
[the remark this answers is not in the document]
guo-shaoge: how about:
```
{
    lock_guard lock;
    while (true)
    {
        ...
        if has error task: set exception_ptr;
            break;
    }
}

if unlikely (exception_ptr)
    task->onErrorOccurred(exception_ptr);
return true;
```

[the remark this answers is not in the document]
SeaRise:
```
auto exception_ptr;
SCOPE_EXIT({
        if unlikely (exception_ptr)
            task->onErrorOccurred(exception_ptr);
    });

origin code.
```
Maybe it would be simpler

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
SeaRise: Before `other = moved_obj` is called, the task remains unchanged, so it is safe

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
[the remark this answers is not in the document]
guo-shaoge: ok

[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Server/Server.cpp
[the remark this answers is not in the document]
guo-shaoge: done. Defined in LAC.cpp, assigned in Server.cpp and unittests

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Debug/dbgFuncMisc.cpp
```diff
@@ -141,7 +141,10 @@ void dbgFuncWaitUntilNoTempActiveThreadsInDynamicThreadPool(Context &, const AST
         auto wait_time = safeGet<UInt64>(typeid_cast<const ASTLiteral &>(*args[0]).value);
         wait_time = std::min(wait_time, MAX_WAIT_TIME);
         /// should update the value when there is long running threads using dynamic thread pool
-        static const int expected_value = 0;
+        /// Long running thread:
+        /// 1. LocalAdmissionController::fetchGAC
+        /// 2. LocalAdmissionController::watchGAC
+        static const int expected_value = 2;
```
SeaRise, line 147: `wetchGAC` and `watchGAC` can use `std::thread` directly, so there is no need to modify the value here.
Otherwise, `random failpoint` daily run test also needs to be modified.

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,228 @@ class ResourceGroup final : private boost::noncopyable
...
+    // So we can avoid dead lock.
+    ResourceGroupPtr getOrFetchResourceGroup(const std::string & name);
+    ResourceGroupPtr findResourceGroup(const std::string & name)
+    {
+        std::lock_guard lock(mu);
+        for (const auto & group : resource_groups)
+        {
+            if (group.first == name)
+                return group.second;
+        }
```
SeaRise, line 462: why not
```
auto it = resource_groups.find(name);
return it == resource_groups.end() ? nullptr : it.second;
```
?

guo-shaoge: done

dbms/src/Debug/dbgFuncMisc.cpp
guo-shaoge: done

dbms/src/Common/Exception.h
```diff
@@ -349,4 +349,14 @@ constexpr auto maybeStringLiteralExpr(const std::string_view sv)
         }                                                              \
     } while (false)
 
+#define CATCH_AND_LOG(stmt, log)                              \
```
SeaRise, line 352: If `CATCH_AND_LOG` is only used by `consumeResource`, how about letting `consumeResource` not throw an error?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,504 @@
...
+                    handleBackgroundError(WATCH_GAC_ERR_PREFIX + err_msg);
+            }
+            else if (event.type() == mvccpb::Event_EventType_PUT)
+            {
+                if (!handlePutEvent(kv, err_msg))
+                    handleBackgroundError(WATCH_GAC_ERR_PREFIX + err_msg);
+            }
+            else
+            {
+                __builtin_unreachable();
```
SeaRise, line 407:
```suggestion
                RUNTIME_ASSERT(false, log, "unexpect event type {}", magic_enum::enum_to_name(event.type()));
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,504 @@
...
+        if (resp.canceled())
+        {
+            handleBackgroundError(WATCH_GAC_ERR_PREFIX + "watch is canceled");
+            break;
+        }
+        for (const auto & event : resp.events())
+        {
+            std::string err_msg;
+            const mvccpb::KeyValue & kv = event.kv();
+            if (event.type() == mvccpb::Event_EventType_DELETE)
```
SeaRise, line 395: maybe switch case is clearer

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,504 @@
...
+    LOG_ERROR(log, err_msg);
+}
+
+void LocalAdmissionController::checkGACRespValid(const resource_manager::ResourceGroup & new_group_pb)
+{
+    if unlikely (new_group_pb.name().empty())
+        throw Exception("resource group name from GAC is empty");
+
+    if unlikely (new_group_pb.mode() != resource_manager::GroupMode::RUMode)
+        throw Exception("resource group is not RUMode");
```
SeaRise, line 503: Better to use `RUNTIME_CHECK_MSG`

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,220 @@ class ResourceGroup final : private boost::noncopyable
...
+
+    ~LocalAdmissionController() { stop(); }
+
+    void consumeResource(const std::string & name, double ru, uint64_t cpu_time_in_ns)
+    {
+        // When tidb_enable_resource_control is disabled, resource group name is empty.
+        if (name.empty())
+            return;
+
+        ResourceGroupPtr group = findResourceGroupWithException(name);
```
SeaRise, line 379:
```suggestion
        ResourceGroupPtr group = findResourceGroup(name);
        if unlikely (group.empty())
            return;
```
Throwing an exception here and then catch and log outside doesn't seem to make sense.

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,220 @@ class ResourceGroup final : private boost::noncopyable
...
+    }
+
+    uint64_t getPriority(const std::string & name)
+    {
+        if (name.empty())
+            return HIGHEST_RESOURCE_GROUP_PRIORITY;
+
+        ResourceGroupPtr group = findResourceGroupWithException(name);
+        return group->getPriority(max_ru_per_sec.load());
+    }
```
SeaRise, line 398:
```suggestion
    std::optional<uint64_t> getPriority(const std::string & name)
    {
        if (name.empty())
            return HIGHEST_RESOURCE_GROUP_PRIORITY;

        ResourceGroupPtr group = findResourceGroup(name);
        if unlikely (group == nullptr)
            return {};
        return group->getPriority(max_ru_per_sec.load());
    }
```
I think it would be better to return optional, so that there is no need to catch exception in ResourceGroupQueue, but to do different processing according to the return value.
@windtalker What do you think about it?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,220 @@ class ResourceGroup final : private boost::noncopyable
...
+    void registerRefillTokenCallback(const std::function<void()> & cb)
+    {
+        std::lock_guard lock(mu);
+        RUNTIME_CHECK_MSG(refill_token_callback == nullptr, "callback cannot be registered multiple times");
+        refill_token_callback = cb;
+    }
+    void unregisterRefillTokenCallback()
+    {
+        std::lock_guard lock(mu);
+        RUNTIME_CHECK_MSG(refill_token_callback != nullptr, "callback cannot be nullptr before unregistering");
```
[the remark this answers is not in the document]
SeaRise: I think it is still necessary to throw an exception, because if refill_token_callback is not registered correctly, it will cause the tiflash to hang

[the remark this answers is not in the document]
SeaRise: So here you need to ensure that `unregisterRefillTokenCallback` will only be called once

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
SeaRise: On the other hand, the cost of throwing exception is much more expensive than returning value.

SeaRise:
```
# run sanitizer test on PR
# check the workflow in
# https://ci.pingcap.net/blue/organizations/jenkins/tiflash-sanitizer-daily/activity
/run-sanitizer-test asan
/run-sanitizer-test tsan
```
You can pre-run tsan/asan to check for data races and memory leaks.
Can run tsan/asan on CI or locally.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -24,62 +24,97 @@ namespace DB
...
 
 template <typename NestedTaskQueueType>
 bool ResourceControlQueue<NestedTaskQueueType>::take(TaskPtr & task)
 {
-    assert(task == nullptr);
+    assert(!task);
+    std::exception_ptr exception_ptr;
+    SCOPE_EXIT({
+        if unlikely (exception_ptr)
+            task->onErrorOccurred(exception_ptr);
```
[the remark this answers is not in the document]
guo-shaoge: task cannot be null here.
Maybe add a assert is ok?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,220 @@ class ResourceGroup final : private boost::noncopyable
...
+            thread.join();
+    }
+
+    // Interval of fetch from GAC periodically.
+    static constexpr uint64_t DEFAULT_FETCH_GAC_INTERVAL = 5;
+    // DEFAULT_TOKEN_FETCH_ESAPSED * token_avg_consumption_speed as token num to fetch from GAC.
+    static constexpr uint64_t DEFAULT_TOKEN_FETCH_ESAPSED = 5;
+    // If we cannot get GAC resp for DEGRADE_MODE_DURATION seconds, enter degrade mode.
+    static constexpr auto DEGRADE_MODE_DURATION = 120;
+    static constexpr auto TARGET_REQUEST_PERIOD_MS = 5000;
```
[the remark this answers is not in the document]
guo-shaoge: use std::chrono::seconds or std::chrono::milliseconds instead

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -61,16 +64,27 @@ class ResourceGroup final : private boost::noncopyable
...
     ~ResourceGroup() = default;
 
+#ifndef DBMS_PUBLIC_GTEST
+private:
+#endif
     enum TokenBucketMode
     {
         normal_mode,
         degrade_mode,
         trickle_mode,
```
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +336,220 @@ class ResourceGroup final : private boost::noncopyable
...
+        watch_gac_grpc_context.TryCancel();
+        cv.notify_all();
+        for (auto & thread : background_threads)
+            thread.join();
+    }
+
+    // Interval of fetch from GAC periodically.
+    static constexpr uint64_t DEFAULT_FETCH_GAC_INTERVAL = 5;
+    // DEFAULT_TOKEN_FETCH_ESAPSED * token_avg_consumption_speed as token num to fetch from GAC.
+    static constexpr uint64_t DEFAULT_TOKEN_FETCH_ESAPSED = 5;
```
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,501 @@
...
+
+            double added_tokens = granted_token_bucket.granted_tokens().tokens();
+            RUNTIME_CHECK(added_tokens >= 0);
+
+            int64_t capacity = granted_token_bucket.granted_tokens().settings().burst_limit();
+
+            // Check GAC code, fill_rate is never setted.
+            RUNTIME_CHECK(granted_token_bucket.granted_tokens().settings().fill_rate() == 0);
+
+            auto resource_group = findResourceGroup(one_resp.resource_group_name());
```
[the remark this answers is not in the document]
guo-shaoge: done. remote for loop for granted_r_u_tokens.

dbms/src/Flash/ResourceControl/TokenBucket.cpp
```diff
@@ -41,15 +41,15 @@ double TokenBucket::peek(const TokenBucket::TimePoint & timepoint) const
...
+    RUNTIME_CHECK(config.fill_rate >= 0.0);
+    RUNTIME_CHECK(config.capacity >= 0.0);
 
     auto now = std::chrono::steady_clock::now();
-    tokens = new_tokens;
-    fill_rate = new_fill_rate;
-    capacity = new_capacity;
+    tokens = config.tokens;
+    fill_rate = config.fill_rate;
+    capacity = config.capacity;
```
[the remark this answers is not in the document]
guo-shaoge: In TokenBucket.cpp, It's a bit cumbersome to access the config variable every time to update the TokenBucket status.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,501 @@
...
+                    return;
+            }
+        }
+
+        try
+        {
+            if (fetch_token_periodically)
+            {
+                fetchTokensForAllResourceGroups();
+                checkDegradeMode();
```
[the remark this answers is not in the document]
guo-shaoge: check degrade mode should also be periodically

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -128,26 +143,189 @@ class ResourceGroup final : private boost::noncopyable
...
-        // If token bucket is normal mode, it's static, so fill_rate is zero.
-        const double init_fill_rate = 0.0;
-        const double init_tokens = user_ru_per_sec;
-        const double init_cap = capacity;
-        bucket = std::make_unique<TokenBucket>(init_fill_rate, init_tokens, init_cap);
+        group_pb = group_pb_;
+        user_priority = group_pb_.priority();
+        const auto & setting = group_pb.r_u_settings().r_u().settings();
+        user_ru_per_sec = setting.fill_rate();
+        burstable = (setting.burst_limit() < 0);
```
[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,504 @@
...
+    {
+        // For each resource group.
+        if unlikely (!one_resp.granted_resource_tokens().empty())
+        {
+            handleBackgroundError("GAC return RAW granted tokens, but LAC expect RU tokens");
+            continue;
+        }
+
+        handled_resource_group_names.emplace_back(one_resp.resource_group_name());
+        for (const resource_manager::GrantedRUTokenBucket & granted_token_bucket : one_resp.granted_r_u_tokens())
```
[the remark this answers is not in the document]
guo-shaoge: for loop remove and check size.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: done

dbms/src/Common/Exception.h
guo-shaoge: done. it will return bool to indicate succ or not

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: done

dbms/src/Flash/Executor/toRU.h
```diff
@@ -18,7 +18,7 @@
 
 namespace DB
 {
-using RU = UInt64;
+using RU = double;
```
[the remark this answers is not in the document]
guo-shaoge: Only one place to interact with other component: tiflash report RU usage to prometheus/GAC

SeaRise:
https://github.com/pingcap/tiflash/blob/569b5672dfcdee6da25e6e73d89675010b4339f8/dbms/src/Flash/Pipeline/Schedule/TaskQueues/TaskQueueType.h#L19
This comment should also be removed

dbms/src/Flash/Executor/toRU.h
SeaRise, line 21: https://github.com/pingcap/tiflash/blob/569b5672dfcdee6da25e6e73d89675010b4339f8/dbms/src/Flash/Coprocessor/DAGContext.cpp#L440
`ceil` here also should be removed.

guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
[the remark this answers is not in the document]
windtalker: Is it possible that some of the resource groups are always in low token mode, so there is no chance to do periodically fetch at all?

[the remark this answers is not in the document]
guo-shaoge: Yes, make sense!

dbms/src/Flash/executeQuery.cpp
```diff
@@ -200,6 +202,7 @@ std::optional<QueryExecutorPtr> executeAsPipeline(Context & context, bool intern
     if (likely(!internal))
         LOG_INFO(logger, fmt::format("Query pipeline:\n{}", executor->toString()));
     dag_context.switchToPipelineMode();
+    LocalAdmissionController::global_instance->warmupResourceGroupInfoCache(dag_context.getResourceGroupName());
```
SeaRise, line 205: how about moving to the constructor of `PipelineExecutor` and moving `consumeResource` to `PipelineExecutor::execute`.

SeaRise: Because resource control should only be enabled when enable_pipeline.
In the previous logic, even if the pipeline is not enabled, ru will be consumed by read_bytes

dbms/src/Flash/Coprocessor/DAGDriver.cpp
```diff
@@ -212,6 +213,14 @@ try
 
     auto cpu_ru = query_executor->collectRequestUnit();
     auto read_ru = dag_context.getReadRU();
+    if unlikely (!LocalAdmissionController::global_instance
+                      ->consumeResource(dag_context.getResourceGroupName(), toRU(read_ru), 0))
+    {
+        LOG_WARNING(
+            log,
+            "cannot consume read_ru for {}, maybe has been deleted. Just ignore",
+            dag_context.getResourceGroupName());
```
SeaRise, line 222: maybe the warning log can be moved into `consumeResource`.

dbms/src/Flash/Executor/toRU.cpp
```diff
@@ -24,9 +24,7 @@ UInt64 toCPUTimeMillisecond(UInt64 cpu_time_ns)
     if (unlikely(cpu_time_ns == 0))
         return 0;
 
-    double cpu_time_millisecond = static_cast<double>(cpu_time_ns) / 1'000'000L;
-    auto ceil_cpu_time_millisecond = ceil(cpu_time_millisecond);
-    return ceil_cpu_time_millisecond;
+    return static_cast<double>(cpu_time_ns) / 1'000'000L;
```
SeaRise, line 27: Why modify here, because you want to round down?

dbms/src/Flash/Executor/toRU.cpp
```diff
@@ -36,8 +34,6 @@ RU toRU(UInt64 cpu_time_ns)
         return 0;
 
     auto cpu_time_millisecond = toCPUTimeMillisecond(cpu_time_ns);
-    auto ru = static_cast<double>(cpu_time_millisecond) / 3;
-    auto ceil_ru = ceil(ru);
-    return ceil_ru;
+    return static_cast<double>(cpu_time_millisecond) / 3;
```
SeaRise, line 37: why not `cpu_time_millisecond / 3.0` ?

dbms/src/Flash/Executor/toRU.h
SeaRise, line 21: https://github.com/pingcap/tiflash/blob/569b5672dfcdee6da25e6e73d89675010b4339f8/dbms/src/Flash/Coprocessor/DAGContext.cpp#L440
If RU is changed to double, can the ceil here be removed? Because there is no need to round up

dbms/src/Flash/ResourceControl/MockLocalAdmissionController.h
```diff
@@ -30,22 +30,18 @@ inline uint64_t nopGetPriority(const std::string &)
...
-    return false;
-}
-
 
 // This is only for ResourceControlQueue gtest.
 class MockLocalAdmissionController final : private boost::noncopyable
 {
 public:
-    MockLocalAdmissionController()
+    explicit MockLocalAdmissionController(bool enable = true)
```
SeaRise, line 38: If the overhead of enable by default is not large, maybe this bool variable may not be needed.

dbms/src/Flash/Executor/toRU.cpp
guo-shaoge: https://en.cppreference.com/w/c/language/conversion#Usual_arithmetic_conversions
1'000'000L will convert to double implicitly.

SeaRise: But the return type is UInt64 of `toCPUTimeMillisecond`, and then double value will be rounded down

dbms/src/Flash/Executor/PipelineExecutor.cpp
```diff
@@ -42,6 +43,8 @@ PipelineExecutor::PipelineExecutor(
     physical_plan.build(context.getDAGContext()->dag_request());
     physical_plan.outputAndOptimize();
     root_pipeline = physical_plan.toPipeline(exec_context, context);
+    LocalAdmissionController::global_instance->warmupResourceGroupInfoCache(
+        context.getDAGContext()->getResourceGroupName());
```
SeaRise, line 47:
```suggestion
        dagContext().getResourceGroupName());
```

dbms/src/Flash/Executor/PipelineExecutor.cpp
```diff
@@ -107,6 +110,12 @@ ExecutionResult PipelineExecutor::execute(ResultHandler && result_handler)
         wait();
     }
     LOG_TRACE(log, "query finish with {}", exec_context.getQueryProfileInfo().toJson());
+
+    // For read_ru, only report it to GAC for now.
+    LocalAdmissionController::global_instance->consumeResource(
+        exec_context.getResourceGroupName(),
+        context.getDAGContext()->getReadRU(),
```
SeaRise, line 117:
```suggestion
        dagContext().getReadRU(),
```

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -52,34 +55,47 @@ void ResourceControlQueue<NestedTaskQueueType>::submitWithoutLock(TaskPtr && tas
...
-    assert(task == nullptr);
+    assert(!task);
+    bool is_error_task = false;
+    SCOPE_EXIT({
+        if unlikely (is_error_task)
+        {
+            assert(task);
+            task->onErrorOccurred(fmt::format(error_template, task->getResourceGroupName()));
+        }
+    });
```
SeaRise, line 98:
```suggestion
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -158,22 +345,227 @@ class ResourceGroup final : private boost::noncopyable
...
+        : cluster(cluster_)
+        , etcd_client(etcd_client_)
+    {
+        background_threads.emplace_back([this] { this->startBackgroudJob(); });
+        background_threads.emplace_back([this] { this->watchGAC(); });
+    }
+
+    ~LocalAdmissionController() { stop(); }
+
+    bool consumeResource(const std::string & name, double ru, uint64_t cpu_time_in_ns)
```
SeaRise, line 382: It seems that there is no need to return bool

dbms/src/Flash/Pipeline/Schedule/TaskQueues/ResourceControlQueue.cpp
```diff
@@ -144,26 +150,49 @@ template <typename NestedTaskQueueType>
...
     assert(task);
-    const String & name = task->getResourceGroupName();
-
-    std::lock_guard lock(mu);
     auto ru = toRU(inc_value);
+    const String & name = task->getResourceGroupName();
     LOG_TRACE(logger, "resource group {} will consume {} RU(or {} cpu time in ns)", name, ru, inc_value);
-    LocalAdmissionController::global_instance->consumeResource(name, ru, inc_value);
+    if unlikely (!LocalAdmissionController::global_instance->consumeResource(name, ru, inc_value))
+        LOG_WARNING(logger, "cannot consume ru for {}, maybe has been deleted. Just ignore", name);
```
SeaRise, line 157:
```suggestion
    LocalAdmissionController::global_instance->consumeResource(name, ru, inc_value);
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,496 @@
...
+    return handled_resource_group_names;
+}
+
+void LocalAdmissionController::watchGAC()
 {
-    if (!global_instance)
-        global_instance = std::make_unique<MockLocalAdmissionController>();
+    while (true)
+    {
+        doWatch();
```
SeaRise, line 343: Is a catch exception needed here, could the thread exit due to a network error?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,496 @@
...
+    std::string name;
+    if (!parseResourceGroupNameFromWatchKey(kv.key(), name, err_msg))
+        return false;
+
+    size_t erase_num = 0;
+    {
+        std::lock_guard lock(mu);
+        erase_num = resource_groups.erase(name);
+    }
+    LOG_INFO(log, "delete resource group {}, erase_num: {}", name, erase_num);
```
SeaRise, line 428: Maybe it shouldn't be info level?

SeaRise: Or debug？

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,496 @@
...
+        err_msg = "parse pb from etcd value failed";
+        return false;
+    }
+    {
+        std::lock_guard lock(mu);
+        auto iter = resource_groups.find(name);
+        if (iter == resource_groups.end())
+        {
+            // It happens when query of this resource group has not came.
+            LOG_INFO(
```
SeaRise, line 450: It seems that this log should not be info level

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -128,26 +151,190 @@ class ResourceGroup final : private boost::noncopyable
...
+            return;
+
+        bucket_mode = TokenBucketMode::degrade_mode;
+        double avg_speed = bucket->getAvgSpeedPerSec();
+        auto config = bucket->getConfig();
+        std::string ori_bucket_info = bucket->toString();
+
+        config.fill_rate = avg_speed;
+        bucket->reConfig(config);
+        LOG_INFO(
```
SeaRise, line 295: There are too many info logs here, and too many info logs will cause the log of tidb cloud to be lost.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -13,19 +13,496 @@
...
+                fetchTokensForLowTokenResourceGroups();
+
+            if (local_refill_token_callback)
+                local_refill_token_callback();
+
+            checkDegradeMode();
+        }
+        catch (...)
+        {
+            handleBackgroundError(getCurrentExceptionMessage(true));
```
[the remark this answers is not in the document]
guo-shaoge: do we have log_evenry_n in glog?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: I think grpc use `status` as error handling. But it's ok to catch and log here.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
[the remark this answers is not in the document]
SeaRise: There might not be one, perhaps use trace level?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
guo-shaoge: The frequency of calling `toDegradeMode()` is not high, it only occurs when there are network issues. I think it ok to do `LOG_INFO` for `toDegradeMode()`.

And I have changed most `LOG_INFO` to `LOG_DEBUG` for LAC.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: done

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
guo-shaoge: done

SeaRise (approved):
lgtm

windtalker (approved):
LGTM

Issue #7915, opened 2023-08-08
TiFlash crash after restart for Trying to add ref to non-exist page


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

consistency test and chaos tiflash failure
```
[...]

PR #7920, opened 2023-08-09
ensure every record in wal is only applied once when restart

lidezhu:
### What problem does this PR solve?

Issue Number: close #7915 

Problem Summary:
After dumping snapshot to checkpoint file, we only remove log files which max sequence is smaller than the sequence used to dump snapshot.
So the remaining wal files may have some overlap with the checkpoint file.
And at restart, we will apply all records in the checkpoint file and other wal files which will cause some records to be applied more than once.
This may cause some problem like the following picture.

image

Because page 88 was already deleted, it is not shown in the checkpoint file `log_100_1`. But the operation `ref 90 -> 88` is still in later log files. So at restart, when it try to apply this record, it will fail to complete the ref operation.

### What is changed and how it works?
Record the snap sequence used to dump checkpoint in the filename of checkpoint file. Specifically, change it from `log_100_1` to `log_100_1_400`.
When restart, after apply all records in checkpoint file, we will skip later record which sequence is smaller than the snap sequence in the checkpoint file name.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [x] Breaking backward compatibility
  - Can not downgrade to a version earlier than v7.4.0

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
Fix the problem that tiflash may fail to restart under some cases.
```

PR #7931, opened 2023-08-11
PageStorage: handle ref and full gc after restart

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/7915

Problem Summary:
In the previous PR https://github.com/pingcap/tiflash/pull/7920, we can not handle the full GC properly. It will ignore all 
`UPSERT` entries that is created by `blobstore full gc`. Because `full gc` will use a version=`<old_version, old_epoch+1>` for the `UPSERT` entries in WAL log files.

As only the REF is not an idempotent operation, when loading edit from disk to restore the PageDirectory, it could re-apply a REF to an non-existing page_id that is already deleted in the dumped snapshot.
All other operation is idempotent operations. So we filter the REF record which is less than or equal to the `filter_seq`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


lidezhu (approved):
LGTM

dbms/src/Storages/Page/V3/PageDirectory.cpp
```diff
@@ -134,11 +134,22 @@ void VersionedPageEntries<Trait>::createNewEntry(const PageVersion & ver, const
...
 typename VersionedPageEntries<Trait>::PageId VersionedPageEntries<Trait>::createUpsertEntry(
     const PageVersion & ver,
-    const PageEntryV3 & entry)
+    const PageEntryV3 & entry,
+    bool strict_check)
 {
     auto page_lock = acquireLock();
 
     // For applying upsert entry, only `VAR_ENTRY`/`VAR_REF` is valid state.
+    // But when `strict_check == true`, we will create a new entry when it is
```
JaySon-Huang, line 143:
```suggestion
    // But when `strict_check == false`, we will create a new entry when it is
```

JaySon-Huang, line 143:
```suggestion
    // But when `strict_check == false`, we will create a new entry when it is
```

Lloyd-Pottiger (approved):

PR #7939, opened 2023-08-15
Make window function return null when frame is valid

### LINKED ISSUE #7930: Window function should return null when frame is valid
## Bug Report

Frame is valid when start with following is greater or equal to the partition end or end with preceding is less than the partition start.

For example:
`select first_value(v) over (partition by p order by o rows between 2 following and 3 following) as a from agg;`, 2 following may make the frame_start greater than the partition end.

`select last_value(v) over (partition by p order by o rows between 3 preceding and 2 preceding) as a from agg;`, 2 preceding may frame_end less then the partition start.

```sql
drop table if exists agg;
create table agg(
    p int not null,
    o int not null,
    v int not null
);

insert into agg (p, o, v) values (0, 0, 1);
insert into agg (p, o, v) values (1, 1, 2);
insert into agg (p, o, v) values (1, 2, 3);
insert into agg (p, o, v) values (1, 3, 4);
insert into agg (p, o, v) values (1, 4, 5);
insert into agg (p, o, v) values (2, 5, 6);
insert into agg (p, o, v) values (2, 6, 7);
insert into agg (p, o, v) values (2, 7, 8);
insert into agg (p, o, v) values (2, 8, 9);
insert into agg (p, o, v) values (2, 9, 10);
insert into agg (p, o, v) values (3, 10, 11);
insert into agg (p, o, v) values (3, 11, 12);
insert into agg (p, o, v) values (3, 12, 13);

alter table agg set tiflash replica 1;
set tidb_enforce_mpp=1;
```

```sql
mysql> select first_value(v) over (partition by p order by o rows between 2 following and 2 following) as a from agg;
+------+
| a    |
+------+
| NULL |
|    4 |
|    5 |
| NULL |
| NULL |
|    8 |
|    9 |
|   10 |
| NULL |
| NULL |
|   13 |
| NULL |
| NULL |
+------+

mysql> select last_value(v) over (partition by p order by o rows between 2 preceding and 2 preceding) as a from agg;
+------+
| a    |
+------+
| NULL |
| NULL |
| NULL |
|    2 |
|    3 |
| NULL |
| NULL |
|    6 |
|    7 |
|    8 |
| NULL |
| NULL |
|   11 |
+------+
```

```sql
mysql> select first_value(v) over (partition by p order by o rows between 2 following and 2 following) as a from agg;
+----+
| a  |
+----+
|  2 |
|  2 |
|  2 |
|  3 |
| 11 |
| 11 |
| 11 |
|  6 |
|  6 |
|  6 |
|  7 |
|  8 |
|  1 |
+----+

mysql> select last_value(v) over (partition by p order by o rows between 2 preceding and 2 preceding) as a from agg;
+----+
| a  |
+----+
|  4 |
|  5 |
|  5 |
|  5 |
|  8 |
|  9 |
| 10 |
| 10 |
| 10 |
|  1 |
| 13 |
| 13 |
| 13 |
+----+
```

nightly: 28591c4b18305172fd4c1f12572542ac6ad454f1

xzhangxian1008:

Issue Number: close #7930

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/WindowFunctions/IWindowFunction.h
```diff
@@ -45,8 +45,25 @@ class IWindowFunction
...
 using WindowFunctionPtr = std::shared_ptr<IWindowFunction>;
 
+// Runtime data for computing one window function.
+struct WindowFunctionWorkspace
+{
+    // TODO add aggregation function
+    WindowFunctionPtr window_function = nullptr;
+
+    ColumnNumbers arguments;
+};
```
SeaRise, line 68: why not moving to `WindowUtils.h`?

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -31,6 +32,29 @@ namespace ErrorCodes
...
+bool defaultCheckFrameValidAndHandle(WindowTransformAction & action, size_t function_index)
+{
+    if (isFrameValid(action.frame_start, action.frame_end, action.partition_start, action.partition_end))
+        return true;
+
+    IColumn & to = *action.blockAt(action.current_row).output_columns[function_index];
+    static Field null_field;
+    to.insert(null_field);
+    return false;
+}
```
SeaRise, line 51: Does this only work with lead/lag? Why not as the default implementation of `checkFrameValidAndHandle`?

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -31,6 +32,29 @@ namespace ErrorCodes
...
+} // namespace
+
+// In window function, some of the functions must return nullable column.
+// Return nullable column functions: FIRST_VALUE, LAST_VALUE, LAG, LEAD, NTH_VALUE, NTILE and all aggregation functions
```
windtalker, line 55: This comment is contradictory with the comment here https://github.com/pingcap/tiflash/pull/7939/files#diff-0f7c99237fd95a3f57b8eedb9e90e2f3b48f44d9abc8af6b0dcef0960d5ea37aR29 for LEAD/LAG

dbms/src/WindowFunctions/IWindowFunction.cpp
xzhangxian1008: > Does this only work with lead/lag? Why not as the default implementation of `checkFrameValidAndHandle`?

No, lead/lag don't need this function. So far, only first_value and last_value need this.

Because the class `WindowTransformAction` can not be used in the IWindowFunction.h, so we have to define the default function in the .cpp file.

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -180,25 +186,23 @@ struct WindowFunctionLastValue final : public IWindowFunction
         : IWindowFunction(argument_types_)
     {
         RUNTIME_CHECK(argument_types_.size() == 1);
-        return_type = argument_types_[0];
+        if (argument_types_[0]->isNullable())
+            return_type = argument_types_[0];
+        else
+            return_type = std::make_shared<DataTypeNullable>(argument_types_[0]);
```
windtalker, line 192: I think you can use `return_type = makeNullable(argument_types_[0])`

dbms/src/WindowFunctions/IWindowFunction.h
xzhangxian1008: > why not moving to `WindowUtils.h`?

It relies on `WindowFunctionPtr` and the circle include will happen if we move this to `WindowUtils.h`

dbms/src/WindowFunctions/IWindowFunction.h
```diff
@@ -45,8 +45,25 @@ class IWindowFunction
...
     DataTypes argument_types;
+
+private:
+    // Return true when the frame is valid
+    //
+    // Check if the frame is invalid, or we will insert null value into result column
+    // Frame is invalid when:
+    //   1. partition_end <= frame_start
+    //   2. frame_end < partition_start
+    virtual bool checkFrameValidAndHandle(WindowTransformAction &, size_t) { return true; }
```
windtalker, line 56: Will this function be called for all impl of  IWindowFunction, if not it seems not neccessary to add a top level virtual function here

dbms/src/WindowFunctions/IWindowFunction.cpp
SeaRise: Is the return type of first/last must be nullable?
What if it's not nullable?

SeaRise: Doesn't the frame vaild exist for other window functions?
For lead/lag, what if frame vaild?
What about other window functions in the future?

SeaRise: Is there any side effect if defaultCheckFrameValidAndHandle is made the default implementation of checkFrameValidAndHandle

dbms/src/WindowFunctions/IWindowFunction.h
SeaRise: ok

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -273,91 +280,120 @@ Int64 WindowTransformAction::getPartitionEndRow(size_t block_rows)
     return left;
 }
 
-RowNumber WindowTransformAction::stepToFrameStart(const RowNumber & current_row, const WindowFrame & frame)
+std::tuple<RowNumber, bool> WindowTransformAction::stepToFrameStart(
```
SeaRise, line 283: Please add some explanation why this is changed, comment or pr description

dbms/src/WindowFunctions/IWindowFunction.cpp
xzhangxian1008: > Is the return type of first/last must be nullable? What if it's not nullable?

Yes, the return type of first/last must be nullable. If it returns non-nullable, the result will be different from mysql's

xzhangxian1008: > Doesn't the frame vaild exist for other window functions? For lead/lag, what if frame vaild? What about other window functions in the future?

Some other window functions don't rely on frame such as lead/lag, so it's needless to consider frame valid for these window functions. Other window functions may have different handle method for invalid frame(maybe default handling is enough, I haven't research it yet).

xzhangxian1008: > Is there any side effect if defaultCheckFrameValidAndHandle is made the default implementation of checkFrameValidAndHandle

I think there is no side affect.

dbms/src/Debug/MockExecutor/WindowBinder.cpp
```diff
@@ -22,98 +22,80 @@
...
-            break;
-        }
-        default:
-            ft->set_tp(TiDB::TypeLongLong);
-            ft->set_flag(TiDB::ColumnFlagBinary);
-            ft->set_collate(collator_id);
-            ft->set_flen(21);
-            ft->set_decimal(-1);
+            const auto third_arg_type = window_expr->children(2).field_type();
+            assert(first_arg_type.tp() == third_arg_type.tp());
```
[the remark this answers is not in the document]
xzhangxian1008: > I think there is no such restrict for lead/lag, so it is actually the restriction of test framework?

This code is copied, so it's the restriction of test framework.

dbms/src/Debug/MockExecutor/WindowBinder.cpp
```diff
@@ -22,98 +22,80 @@
...
-        by->set_desc(elem->direction < 0);
-        tipb::Expr * expr = by->mutable_expr();
-        astToPB(children[0]->output_schema, elem->children[0], expr, collator_id, context);
+        auto field_type = TiDB::fieldTypeToColumnInfo(arg_type);
+        field_type.clearNotNullFlag();
+        window_field_type->set_flag(field_type.flag);
+        break;
     }
+    default:
+        window_field_type->set_tp(TiDB::TypeLongLong);
```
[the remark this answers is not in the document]
xzhangxian1008: > default type is nullable(int64)?

I don't know, it the code copied from origin test framework.

dbms/src/Debug/MockExecutor/WindowBinder.cpp
```diff
@@ -234,6 +274,7 @@ ExecutorBinderPtr compileWindow(ExecutorBinderPtr input, size_t & executor_index
             case tipb::ExprType::LastValue:
             {
                 ci = children_ci[0];
+                ci.clearNotNullFlag();
```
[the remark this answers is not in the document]
xzhangxian1008: > Looks like these code are very much the same as `setWindowFieldType`, so why need to derive the return type twice?

I don't know, it's not my codes.

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -31,6 +32,29 @@ namespace ErrorCodes
 extern const int NUMBER_OF_ARGUMENTS_DOESNT_MATCH;
 } // namespace ErrorCodes
 
+namespace
+{
+bool isFrameValid(RowNumber frame_start, RowNumber frame_end, RowNumber partition_start, RowNumber partition_end)
+{
+    return !((partition_end <= frame_start) || (frame_end <= partition_start));
```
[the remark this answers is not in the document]
xzhangxian1008: > Is `frame_start >= frame_end` valid?

No, but we have checked it before entering this function.

dbms/src/WindowFunctions/IWindowFunction.cpp
xzhangxian1008: > I think you can use `return_type = makeNullable(argument_types_[0])`

okk

dbms/src/WindowFunctions/IWindowFunction.cpp
xzhangxian1008: > This comment is contradictory with the comment here https://github.com/pingcap/tiflash/pull/7939/files#diff-0f7c99237fd95a3f57b8eedb9e90e2f3b48f44d9abc8af6b0dcef0960d5ea37aR29 for LEAD/LAG

okk, I will refine it.

dbms/src/WindowFunctions/IWindowFunction.h
xzhangxian1008: > Will this function be called for all impl of IWindowFunction, if not it seems not neccessary to add a top level virtual function here

okk, I will make it as a static function in .cpp file.

dbms/src/DataStreams/WindowBlockInputStream.cpp
xzhangxian1008: > Please add some explanation why this is changed, comment or pr description

okk

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -404,13 +444,18 @@ void WindowTransformAction::advanceFrameStart()
     }
     case WindowFrame::BoundaryType::Offset:
         if (window_description.frame.type == WindowFrame::FrameType::Rows)
-            frame_start = stepToFrameStart(current_row, window_description.frame);
+        {
+            RowNumber frame_start_tmp;
+            std::tie(frame_start_tmp, frame_started) = stepToFrameStart(current_row, window_description.frame);
+            if (frame_started)
+                frame_start = frame_start_tmp;
```
SeaRise, line 451: why not
```suggestion
            std::tie(frame_start, frame_started) = stepToFrameStart(current_row, window_description.frame);
```
?
Does the value of `frame_start` make sense when `frame_started` is false? If it doesn't make sense, why not assign it directly?

xzhangxian1008: > why not
> 
> ? Does the value of `frame_start` make sense when `frame_started` is false? If it doesn't make sense, why not assign it directly?

When `frame_started` is false, frame_start should not be changed, or it will be less than `prev_frame_start`, this is invalid.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -273,91 +280,124 @@ Int64 WindowTransformAction::getPartitionEndRow(size_t block_rows)
     return left;
 }
 
-RowNumber WindowTransformAction::stepToFrameStart(const RowNumber & current_row, const WindowFrame & frame)
+// When finding frame start with Following attribute, partition end
+// may haven't appeared and we can't find frame start in this case.
+// Returning false in the tuple's second parameter means the failuer
```
windtalker, line 285:
```suggestion
// Returning false in the tuple's second parameter means the failure
```

SeaRise (approved):

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -31,6 +32,40 @@ namespace ErrorCodes
 extern const int NUMBER_OF_ARGUMENTS_DOESNT_MATCH;
 } // namespace ErrorCodes
 
+namespace
+{
+bool isFrameValid(RowNumber frame_start, RowNumber frame_end, RowNumber partition_start, RowNumber partition_end)
+{
+    return !((partition_end <= frame_start) || (frame_end <= partition_start) || (frame_end <= frame_start));
```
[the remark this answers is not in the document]
xzhangxian1008: > I prefer to use `return partition_end > frame_start && frame_end > partition_start && frame_end > frame_start`

okk

dbms/src/WindowFunctions/IWindowFunction.cpp
```diff
@@ -31,6 +32,40 @@ namespace ErrorCodes
...
+}
+
+// Return true when the frame is valid
+//
+// Check if the frame is invalid, or we will insert null value into result column
+// Frame is invalid when:
+//   1. partition_end <= frame_start
+//   2. frame_end < partition_start
+//   3. frame_end <= frame_start
+bool checkFrameValidAndHandle(WindowTransformAction & action, size_t function_index)
```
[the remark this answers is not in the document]
xzhangxian1008: okk

windtalker (approved):
LGTM

Issue #7177, opened 2023-03-29
MPPTask's earier exit may cause false positive error


windtalker:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
Consider the case that 
{MPPTask1, MPPTask2} =======> {MPPTask3,MPPTask4}

Assuming that MPPTask3/MPPTask4 contains executor that may cause earier exit(for example, `limit` and `filter with always false filter condition`), then MPPTask3/MPPTask4 may exit earier before consumming all the data produced by MPPTask1/MPPTask2, and once MPPTask3/MPPTask4 finish executing, it will close the exchange reciever, so if MPPTask1/MPPTask2 is sending data to MPPTask3/MPPTask4 again, it will meet `closed tunnel` error, if this error happens before TiDB get all the result from the query, then there is a chance that the error message will be sent to TiDB, and TiDB think the query fails, which is actually a false positive error.

yibin87:
Add a specific SQL and explain case here:
`
select  asceding.item_sk
from
    (select *
           from (select ss_item_sk item_sk,avg(ss_net_profit) rank_col
                 from store_sales ss1
                 where ss_store_sk = 47
                 group by ss_item_sk
                 having avg(ss_net_profit) > null)V1) asceding,
    (select *
           from (select ss_item_sk item_sk,avg(ss_net_profit) rank_col
                 from store_sales ss1
                 group by ss_item_sk)V2) descending
where asceding.item_sk = descending.item_sk;
`
The query plan:
![yC22WAZlg1](https://github.com/pingcap/tiflash/assets/12403562/a41d906a-98d0-4b67-99bb-b305373a17b4)

PR #7969, opened 2023-08-18
Not call abortMPPGather when mpp task met error

yibin87:

Issue Number: ref #7177 

Problem Summary: When MPPTask has 'always false' filter, it will close receiver before receiving all data. It will cause 'false positive' errors. And when there are other MPPTasks which consumes data from 'always false' MPPTasks in the same 
 tiflash node, abortMPPGather may convert 'false positive' errors to 'real' errors. Here is a specific example: https://github.com/pingcap/tiflash/issues/7177#issuecomment-1683272593

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Manually test tpcds q44, which contains 'always false' filter.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


windtalker (approved):
LGTM

Issue #7738, opened 2023-07-04
Support query level automatical spilling when query memory usage exceeds threshold


windtalker:
## Feature Request
The goal is user can simply set a memory quota for a query(per TiFlash node), and when the query memory usage exceeds the memory quota, TiFlash will trigger spill automatically if the query contains memory intensive operators.
**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #8033, opened 2023-08-28
Auto spill enhance

windtalker:

Issue Number: ref #7738

Problem Summary:

- `AutoSpillTrigger` support spill check without cooldown time if memory usage is really high
- Avoid too many logs in `QueryOperatorSpillContexts::triggerAutoSpill`
- Support trigger auto spill check inside `Join::InsertFromBlock`
- In `HashTable` and `Arena`, support check if current data is already marked as spilled when before trying to allocate more memory, currently, used in `Join` spill
- Fix a bug that the `ReceivedMessageQueue` is actually not memory bounded even if `max_buffered_bytes_in_executor` is set
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Common/Arena.h
```diff
@@ -93,6 +97,11 @@ class Arena : private boost::noncopyable
     /// Add next contiguous chunk of memory with size not less than specified.
     void NO_INLINE addChunk(size_t min_size)
     {
+        if (resize_callback != nullptr)
+        {
+            if (!resize_callback())
+                throw ResizeException("Error in arena resize");
+        }
```
SeaRise, line 104: Will there be a performance regression?

SeaRise (approved):
lgtm

dbms/src/Common/Arena.h
yibin87: Add unlikely for exception throw path? And add failpoint here?

yibin87 (approved):
Others LGTM

dbms/src/Common/Arena.h
windtalker: OK, I will add unlikely for it, but for failpoint, actually it is not a standalone behavior, because if this function throws error, the related partition must be marked as to spill, otherwise, it may cause wrong result. I will add failpoint to test this case in test pr.

PR #8045, opened 2023-09-01
Refine auto spill for fine grained executor

windtalker:

Issue Number: ref #7738

Problem Summary:
When fine grained shuffle is enabled, due to implicit depentency of all the operator belongs to the same TiDB executor, auto spill will meet a big performance regression if auto spill treated all the operator spill contexts from the same TiDB executor independently. This pr add `FineGrainedOperatorSpillContext` and try to align the spill behavior for fine grained executor.
- Add `FineGrainedOperatorSpillContext`
- For operator that comes from the same TiDB executor, auto spill will be triggered at TiDB executor level, instead of triggering each operator independently

Test on TPCH-100 dataset, with 2 TiFlash node
For fine grained agg, test sql is
```
select /*+ mpp_1phase_agg() */ l_orderkey, max(L_COMMENT), max(L_SHIPMODE), max(L_SHIPINSTRUCT), max(L_SHIPDATE), max(L_EXTENDEDPRICE) from lineitem group by l_orderkey having sum(l_quantity) > 314
``` 
per node memory usage limit to 12G:

| without this pr | with this pr| 
|-------------|--------------|
| ~45s            | ~33s |

per node memory usage limit to 7G:

| without this pr | with this pr| 
|-------------|--------------|
| ~37s            | ~30s |

For fine grained sort, test sql is
```
select * from (select L_ORDERKEY,L_COMMENT,L_SHIPINSTRUCT, row_number() over(partition by l_orderkey order by l_orderkey) as rn from lineitem) d where l_orderkey + rn < 50;
```
per node memory usage limit to 12G:

| without this pr | with this pr| 
|-------------|--------------|
| ~45s            | ~24s |

per node memory usage limit to 7G:

| without this pr | with this pr| 
|-------------|--------------|
| ~45s            | ~25s |
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Core/FineGrainedOperatorSpillContext.cpp
+++ b/dbms/src/Core/FineGrainedOperatorSpillContext.cpp
@@ -26,19 +26,14 @@ bool FineGrainedOperatorSpillContext::supportFurtherSpill() const
     return false;
 }
 
-bool FineGrainedOperatorSpillContext::isSpillEnabled() const
-{
-    for (const auto & operator_spill_context : operator_spill_contexts)
-        if (operator_spill_context->isSpillEnabled())
-            return true;
-    return false;
-}
-
 void FineGrainedOperatorSpillContext::addOperatorSpillContext(const OperatorSpillContextPtr & operator_spill_context)
 {
     /// fine grained operator spill context only used in auto spill
-    operator_spill_context->setAutoSpillMode();
-    operator_spill_contexts.push_back(operator_spill_context);
+    if likely (operator_spill_context->supportSpill() && operator_spill_context->supportAutoTriggerSpill())
+    {
+        operator_spill_contexts.push_back(operator_spill_context);
+        operator_spill_context->setAutoSpillMode();
+    }
 }
 
 Int64 FineGrainedOperatorSpillContext::getTotalRevocableMemoryImpl()
--- a/dbms/src/Core/FineGrainedOperatorSpillContext.cpp
+++ b/dbms/src/Core/FineGrainedOperatorSpillContext.cpp
@@ -1,0 +1,64 @@
+// Copyright 2023 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Core/FineGrainedOperatorSpillContext.h>
+
+namespace DB
+{
+bool FineGrainedOperatorSpillContext::supportFurtherSpill() const
+{
+    for (const auto & operator_spill_context : operator_spill_contexts)
+    {
+        if (operator_spill_context->supportFurtherSpill())
+            return true;
+    }
+    return false;
+}
+
+bool FineGrainedOperatorSpillContext::isSpillEnabled() const
+{
+    for (const auto & operator_spill_context : operator_spill_contexts)
+        if (operator_spill_context->isSpillEnabled())
+            return true;
+    return false;
+}
+
+void FineGrainedOperatorSpillContext::addOperatorSpillContext(const OperatorSpillContextPtr & operator_spill_context)
+{
+    /// fine grained operator spill context only used in auto spill
+    operator_spill_context->setAutoSpillMode();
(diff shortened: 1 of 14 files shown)


dbms/src/Core/FineGrainedOperatorSpillContext.h
```diff
@@ -0,0 +1,51 @@
...
+/// with fine grained shuffle enable. When fine grained shuffle is enabled, executors like sort and aggregation
+/// will have n independent operators in TiFlash side, each of them have their own operator spill context. However
+/// these n operators are not really independent: they share the same source operator(exchange receiver), if any one
+/// of the operator stop consuming data from exchange receiver(for example, begin spill data), all the others will
+/// be stuck because the exchange receiver has a bounded queue for all the operators, any one of the operator stop
+/// consuming will make the queue full, and no other data can be pushed to exchange receiver. If all the n operator
+/// is triggered to spill serially, it will affects the overall performance seriously.
+/// FineGrainedOperatorSpillContext is used to make sure that all the operators belongs to the same executor
+/// will be triggered to spill almost at the same time
+class FineGrainedOperatorSpillContext : public OperatorSpillContext
```
yibin87, line 31: If we don't intend to inherit this class, better add final keyword.

dbms/src/Core/FineGrainedOperatorSpillContext.h
```diff
@@ -0,0 +1,51 @@
...
+{
+protected:
+    std::vector<OperatorSpillContextPtr> operator_spill_contexts;
```
yibin87, line 34: why not private field member?

dbms/src/Core/FineGrainedOperatorSpillContext.h
windtalker: Done

dbms/src/Core/FineGrainedOperatorSpillContext.h
windtalker: Done

yibin87 (approved):
LGTM

dbms/src/Core/FineGrainedOperatorSpillContext.cpp
```diff
@@ -0,0 +1,59 @@
...
+    return ret;
+}
+
+Int64 FineGrainedOperatorSpillContext::triggerSpillImpl(Int64 expected_released_memories)
+{
+    Int64 original_expected_released_memories = expected_released_memories;
+    for (auto & operator_spill_context : operator_spill_contexts)
+    {
+        if (expected_released_memories <= 0)
+            operator_spill_context->triggerSpill(original_expected_released_memories);
```
SeaRise, line 53: Why is this needed? Does this mean that more data than `expected_released_memories` will be spilled.

dbms/src/Flash/Planner/Plans/PhysicalTopN.cpp
```diff
@@ -79,7 +79,7 @@ void PhysicalTopN::buildPipelineExecGroupImpl(
     // TODO find a suitable threshold is necessary; 10000 is just a value picked without much consideration.
     if (group_builder.concurrency() * limit <= 10000)
     {
-        executeLocalSort(exec_context, group_builder, order_descr, limit, context, log);
+        executeLocalSort(exec_context, group_builder, order_descr, limit, false, context, log);
```
SeaRise, line 82: It looks like this is similar to fine grained.
How about using the same spilling strategy?

dbms/src/Core/FineGrainedOperatorSpillContext.cpp
windtalker: Yes, because the worst case is each operator is triggered to spill serially.

dbms/src/Flash/Planner/Plans/PhysicalTopN.cpp
windtalker: But for non fine grained case, one operator spill usually will not block others, that is why I only handle it for fine grained executor.

dbms/src/Core/FineGrainedOperatorSpillContext.cpp
SeaRise: ok.

dbms/src/Flash/Planner/Plans/PhysicalTopN.cpp
SeaRise: But topN is usually together with sender, so there are similar situations.
I think it's all right.

SeaRise (approved):
lgtm

SeaRise:
/hold
When you want to merge pr, comment `/hold cancel` @windtalker

Issue #8050, opened 2023-09-04
Enhance the memory usage of mark cache and minmax index cache


hongyunyan:
## Enhancement
Currently, in minmax_index_cache and mark_cache we use the long file name as the key, which occupy much memory, especially when the memory cost of value is small(with a little pack number).

Besides, we statistic memory overhead of mark cache size, but only counted about the memory cost of MarksInCompressedFile, but not consider the memory cost of keys, and the structual cost of LRUCache, which cause a misleading of memory usage. 

Thus, we should optimize the memory usage of these caches, and make a better memory statistical  for these memory usage of caches.

JaySon-Huang:
Change the type from enhancement to bug because it leads to much more memory consumption than expected under some senario

PR #8058, opened 2023-09-06
Update the weight function for Mark Cache and Min Max Index Cache

hongyunyan:

Issue Number: ref https://github.com/pingcap/tiflash/issues/8050

Problem Summary:
Update the weight function for Mark Cache and Min Max Index Cache

We only calculate the inner value weight as the cost of LRUCache now, while sometimes the memory cost of key and structural part also occupy a lot, especially when the pack number is quite small.

Thus, we update the weight function of mark cache and minmax index cache, to provider more accurate memory calculation(while it's still an approximate memory usage, but more accurate approximat).

Besides, we add the metrics for minmax cache index, and add their correponding panel in grafana.

In test we can see the whole memory gap is similar with the memory gap of mark cache and minmax index cache.

image

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.h
```diff
@@ -60,7 +60,8 @@ class MinMaxIndex
 
     size_t byteSize() const
     {
-        return sizeof(UInt8) * has_null_marks->size() + sizeof(UInt8) * has_value_marks->size() + minmaxes->byteSize();
+        return sizeof(UInt8) * has_null_marks->size() + sizeof(UInt8) * has_value_marks->size() + minmaxes->byteSize()
+            + 3 * sizeof(PaddedPODArray<UInt8>);
```
JaySon-Huang, line 64: Add some comments about where these values comes from

dbms/src/Storages/DeltaMerge/Remote/RNDeltaIndexCache.h
```diff
@@ -85,7 +85,7 @@ class RNDeltaIndexCache : private boost::noncopyable
 
     struct CacheValueWeight
     {
-        size_t operator()(const CacheValue & v) const { return v.bytes; }
+        size_t operator()(const CacheKey & /*key*/, const CacheValue & v) const { return v.bytes; }
```
JaySon-Huang, line 88: We'd better add `sizeof(CacheKey)` here cause the key is composed by plenty of members

dbms/src/Interpreters/AsynchronousMetrics.cpp
```diff
@@ -196,6 +196,14 @@ void AsynchronousMetrics::update()
...
     }
 
+    {
+        if (auto min_max_cache = context.getMinMaxIndexCache())
+        {
+            set("MinMaxIndexCacheBytes", min_max_cache->weight());
+            set("MinMaxIndexFiles", min_max_cache->count());
+        }
+    }
+
```
JaySon-Huang, line 206: We'd better add metrics about `RNDeltaIndexCache` when it is not empty

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.h
```diff
@@ -116,7 +117,24 @@ class MinMaxIndex
...
+        auto str_len = key.size(); // key_len
+        auto key_memory_usage = sizeof(String); // String struct memory cost
+
+        // 3. the memory cost of hash table
+        auto unordered_map_memory_usage = 28; // hash table struct approximate memory cost
+
+        // 4. the memory cost of LRUQueue
+        auto list_memory_usage = sizeof(std::list<String>); // list struct memory cost
+
+        return index_memory_usage + cells_memory_usage + str_len * 2 + key_memory_usage * 2 + unordered_map_memory_usage
```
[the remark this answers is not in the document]
hongyunyan: One for LRUCache::Queue and one for key of LRUCache::Cells

JaySon-Huang (approved):

Issue #4646, opened 2022-04-13
Breakdown the "dbms/src/Storages/Transaction" module


JaySon-Huang:
## Enhancement
The codes under "dbms/src/Storages/Transaction" are some codes we adapt TiFlash with TiDB. After years of development, it become kind of crowded and confusing.

Suggest breakdown "dbms/src/Storages/Transaction” into these modules:
* Some global context for TiFlash management
* Basic structure from TiDB (TableInfo/DatabaseInfo, etc)
* Syncing schema with TiDB
* Decoding data from row to column (Block)
* Raft layer
  * Region info
  * Handling admin command
  * Handling write command
  * Handling apply snapshot/ingest sst
  * Handling learner read

<hr/>

1. Suggest creating a `dbms/src/TiDB` as the module for basic components from TiDB. And
    * [ ] Rename TMTContext to TiFlashContext
    * [ ] Move TableInfo/DBInfo, etc from TiDB.h to `dbms/src/TiDB`
    * [ ] Move TiFlashContext/ManagedStorages to the `dbms/src/TiDB` module
    * [ ] Add a `dbms/src/TiDB/Codec` and move the logic of decoding data from row to column there
    * [x] Add a `dbms/src/TiDB/Schema` and move schema syncing logic there
2. Rename `dbms/src/Transaction` to `dbms/src/KVStore`. And maybe we should further break down the codes for different Raft processes under this module.
    * [x] Rename `dbms/src/Transaction` to `dbms/src/KVStore` https://github.com/pingcap/tiflash/pull/8073
    * [x] Break down `KVStore`. https://github.com/pingcap/tiflash/pull/8073

hzh0425:
Hi, can you assign this issue to me, I want to try it @JaySon-Huang

JaySon-Huang:
/assign @hzh0425 

Yes. And suggest beginning with a PR that does a part of the changes, cause it will change lots of files.

hzh0425:
> /assign @hzh0425
> 
> Yes. And suggest beginning with a PR that does a part of the changes, cause it will change lots of files.

Ok, thanks, I will try it

mischaZhang:
Hi, I think I can help with this refactor.

I am a user of tidb, sync data to another database with ticdc is slow which costs 1s at least, so I came up with an idea to change tiflash source code and impliment our own data storage.

Will this refactor support 5.4.0? We are using version 5.4.0 in production.

JaySon-Huang:
[...]

PR #8073, opened 2023-09-08
Refactor kvstore

CalvinNeo:

Issue Number: close https://github.com/pingcap/tiflash/issues/4646

Problem Summary:

The Transaction folder is renamed to KVStore, since it is originally a KVStore with a percolator executor to decode row data into column data built on raft kv.

The new KVStore folder will be made into a new `kvstore` library which will be linked to `dbms`.

The following files will be moved to TiDB:
1. Collator
2. TiDB row/column format. These codes are purely definitions of TiDB encoding, including Datum and RowCodec and TypeMapping. It does not relate to TiKV key's encoding.

The KVStore folder includes:
1. `Storages/KVStore/Decode` is a bridge between KVStore and DeltaMerge. It is about:
    - How we read keys and values from KVStore.
    - How we use utils from `TiDB` and decode the value in row format kvs. 
    - How we will transform row format into column format. It includes modules like RegionTable.
3. `Storages/KVStore/FFI` is FFI utils and interfaces with TiFlash Proxy.
4. `Storages/KVStore/MultiRaft` contains all modules that implements a raft kv. Including region, region manager, region range index.
5. `Storages/KVStore/Utils`
6. `Storages/KVStore/TiKVHelpers` contains some util functions about definitions in TiKV which can be extract as common util function, such as key format for percolator/keyspace, and pd client. The key format part is about how to parse TiKVKey, and uses some information in TiKVValue to perform percolator 2pc commit. Note some TiKV key utils are related to schema or range of tables, we will put them in `Storages/KVStore/Decode`.
7. `Storages/KVStore/Read` contains all about learner read with strong consistency, inclduing mvcc and region lock info and read index. **It does not include read from data KVStore and ingest into DeltaMerge**, which is in `Storages/KVStore/Decode`.

The whole classification is not accurate, since we still rely on some decoded data such as DecodedTiKVKey in MultiRaft.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Transaction/Decode/Datum.cpp
JaySon-Huang, line 1: I'd like to move these files about Decoding to `TiDB/Decode`

CalvinNeo: The code in the current `KVStore/Decode` folder includes:
1. TiKV key format. This part is most messy. Some key format is for percolator transaction protocol, some key format relates to TiDB schema
1. Row2Column transformation
1. Transaction

I don't think they totally belongs to TiDB, since most implementation of transaction is in TiKV.

dbms/src/Storages/KVStore/Utils/JsonBinary.cpp
JaySon-Huang, line 1: I think these `Utils/JsonXXX` (and their tests) should be moved to `TiDB/Decoding` or `Functions/`

dbms/src/Storages/KVStore/MultiRaft/Disagg/KeyspaceSnapshot.cpp
JaySon-Huang, line 1: `KeyspaceXXX` is not a part of `Disagg` arch. Move these files to `KVStore/TiKV` is better

dbms/src/Storages/KVStore/Utils/JsonBinary.cpp
CalvinNeo: I think the former is better

JaySon-Huang (approved):

dbms/src/TiDB/Collator/Collator.h
JaySon-Huang, line 1: "Collation" may be a better name for the folder "Collator"

flowbehappy (approved):

PR #8078, opened 2023-09-12
support agg resize callback

windtalker:

Issue Number: ref #7738

Problem Summary:

This pr set ResizeCallback for aggregation hash table, so when the hash table need resize in aggregation, it has the chance to check if current hash table is already maked to spill, if yes, it can simply give up resize and release the memory in time.
### Check List

Tests 

- [ ] Unit test
- [x] Integration test
   will add integration test in endless
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/Aggregator.cpp
+++ b/dbms/src/Interpreters/Aggregator.cpp
@@ -160,7 +160,7 @@ void AggregatedDataVariants::setResizeCallbackIfNeeded(size_t thread_num) const
         {
             auto resize_callback = [agg_spill_context, thread_num]() {
                 return !(
-                    agg_spill_context->supportAutoTriggerSpill()
+                    agg_spill_context->supportFurtherSpill()
                     && agg_spill_context->isThreadMarkedForAutoSpill(thread_num));
             };
 #define M(NAME)                                                                                         \
--- a/dbms/src/Interpreters/Aggregator.cpp
+++ b/dbms/src/Interpreters/Aggregator.cpp
@@ -159,7 +159,9 @@ void AggregatedDataVariants::setResizeCallbackIfNeeded(size_t thread_num) const
         if (agg_spill_context->isSpillEnabled() && agg_spill_context->isInAutoSpillMode())
         {
             auto resize_callback = [agg_spill_context, thread_num]() {
-                return !(agg_spill_context->supportAutoTriggerSpill() && agg_spill_context->isThreadMarkedForAutoSpill(thread_num));
+                return !(
+                    agg_spill_context->supportAutoTriggerSpill()
+                    && agg_spill_context->isThreadMarkedForAutoSpill(thread_num));
             };
 #define M(NAME)                                                                                         \
     case AggregationMethodType(NAME):                                                                   \
@@ -653,6 +655,24 @@ void NO_INLINE Aggregator::executeImpl(
 }
 
 template <typename Method>
+std::optional<typename Method::EmplaceResult> Aggregator::emplaceKey(
+    Method & method,
+    typename Method::State & state,
+    size_t index,
+    Arena & aggregates_pool,
+    std::vector<std::string> & sort_key_containers) const
+{
+    try
+    {
+        return state.emplaceKey(method.data, index, aggregates_pool, sort_key_containers);
+    }
+    catch (ResizeException &)
+    {
+        return {};
+    }
+}
+
+template <typename Method>
 ALWAYS_INLINE void Aggregator::executeImplBatch(
     Method & method,
     typename Method::State & state,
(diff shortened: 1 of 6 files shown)


dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -645,9 +672,21 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
...
+                state.emplaceKey(method.data, i, *aggregates_pool, sort_key_containers).setMapped(place);
+                processed_rows = i;
+            }
+        }
+        catch (ResizeException &)
+        {
+            LOG_INFO(log, "HashTable resize throw ResizeException since the data is already marked for spill");
+        }
+        if (processed_rows != std::numeric_limits<size_t>::max())
+            agg_process_info.start_row = processed_rows + 1;
```
SeaRise, line 689: how about
```suggestion
        try
        {
            for (size_t i = 0; i < agg_size; ++i)
            {
                state.emplaceKey(method.data, agg_process_info.start_row, *aggregates_pool, sort_key_containers).setMapped(place);
                ++agg_process_info.start_row;
            }
            ++agg_process_info.start_row;
        }
        catch (ResizeException &)
        {
            LOG_INFO(log, "HashTable resize throw ResizeException since the data is already marked for spill");
        }
```

SeaRise: yes, you are right.

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -678,43 +716,65 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
...
-            inst->state_offset,
-            inst->batch_arguments,
-            aggregates_pool);
+        /// Add values to the aggregate functions.
+        for (AggregateFunctionInstruction * inst = agg_process_info.aggregate_functions_instructions.data(); inst->that;
+             ++inst)
+        {
+            inst->batch_that->addBatch(
+                agg_process_info.start_row,
+                processed_rows - agg_process_info.start_row + 1,
```
SeaRise, line 770:
```suggestion
                agg_size,
```
If `processed_rows != std::numeric_limits<size_t>::max()`, it seems that `processed_rows` must be `agg_process_info.start_row + 1 + agg_size`

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -678,43 +716,65 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
     /// Generic case.
 
     std::unique_ptr<AggregateDataPtr[]> places(new AggregateDataPtr[agg_size]);
+    size_t processed_rows = std::numeric_limits<size_t>::max();
```
SeaRise, line 719: maybe a  bool var is enough.

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -678,43 +716,65 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
...
-    for (size_t i = agg_process_info.start_row; i < agg_process_info.start_row + agg_size; ++i)
+    try
     {
-        AggregateDataPtr aggregate_data = nullptr;
+        for (size_t i = agg_process_info.start_row; i < agg_process_info.start_row + agg_size; ++i)
+        {
+            AggregateDataPtr aggregate_data = nullptr;
 
-        auto emplace_result = state.emplaceKey(method.data, i, *aggregates_pool, sort_key_containers);
+            allow_exception = true;
```
SeaRise, line 728: What does allow_exception do?

dbms/src/Interpreters/Aggregator.cpp
windtalker: Must use `processed_rows - agg_process_info.start_row + 1` here, because the basic idea is:
- `processed_rows` is used to record how many rows are already insert its key to the hash table
- during insert, it may throw ResizeException, if exception happens, the processed_rows is pointing the last row that have their key in the hash table
- if `allow_exception` is true when `ResizeException` happens, the aggregator will aggregate data in [start_rows, processed_rows], then update `start_rows` to `processed_rows + 1`, so next time in `executeOnBlock`, it will not re-aggregate the ones that is already aggregated.
- `allow_exception` is used to make sure the valid exception only comes when insert key to hash table, for example, allocating memory in pool does not allow exception.

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -645,9 +692,19 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
...
+            {
+                state.emplaceKey(method.data, agg_process_info.start_row, *aggregates_pool, sort_key_containers)
+                    .setMapped(place);
+                ++agg_process_info.start_row;
+            }
+        }
+        catch (ResizeException &)
+        {
+            LOG_INFO(log, "HashTable resize throw ResizeException since the data is already marked for spill");
+        }
```
SeaRise, line 707:
```suggestion
            for (size_t i = 0; i < agg_size; ++i)
            {
                auto res = emplaceKey(state, method.data, agg_process_info.start_row, *aggregates_pool, sort_key_containers)
                    .setMapped(place);
                if (!state)
                {
                    LOG_INFO(log, "HashTable resize throw ResizeException since the data is already marked for spill");
                    break;
                }
                ++agg_process_info.start_row;
            }
```

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -678,12 +736,20 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
     /// Generic case.
 
     std::unique_ptr<AggregateDataPtr[]> places(new AggregateDataPtr[agg_size]);
+    size_t processed_rows = std::numeric_limits<size_t>::max();
```
SeaRise, line 739:
```suggestion
    std::optional<size_t> processed_rows;
```
It looks better to use `if (processed_rows) { *processed_rows - agg_process_info.start_row + 1, }`

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -657,6 +714,7 @@ ALWAYS_INLINE void Aggregator::executeImplBatch(
         for (AggregateFunctionInstruction * inst = agg_process_info.aggregate_functions_instructions.data(); inst->that;
              ++inst)
         {
+            /// no resize will happen for this kind of hash table, so don't catch resize exception
```
SeaRise, line 717:
```suggestion
```

dbms/src/Interpreters/Aggregator.cpp
windtalker: Done

dbms/src/Interpreters/Aggregator.cpp
windtalker: Good idea, refined.

SeaRise (approved):
lgtm

yibin87 (approved):
LGTM

PR #8064, opened 2023-09-07
Fully support aggregate on part of a input block

windtalker:
### What problem does this PR solve?

Issue Number: ref #7738

Problem Summary:

### What is changed and how it works?
This pr is after #8056, and in the pr, aggregate part of the input block in `Aggregator` fully supported. However, although supported, this feature can only be triggered by failpoint.
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8101, opened 2023-09-15
Fix MPPTask hang introduced by #8064

windtalker:

Issue Number: ref #7738

Problem Summary:

In #8064, it use a `while` loop to handle aggregate input block, like 
https://github.com/pingcap/tiflash/blob/f6e8ffa5b76768c307368e45f7e668f78c1a1e71/dbms/src/DataStreams/ParallelAggregatingBlockInputStream.cpp#L150-L155

while, unfortunately, both the while loop and `allBlockDataHandled()` does not aware that the agg is cancelled, so it may cause hang.
This pr makes `AggProcessInfo` aware the cancellation of current aggregator. 
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Operators/AggregateContext.h
+++ b/dbms/src/Operators/AggregateContext.h
@@ -27,7 +27,10 @@ struct ThreadData
     size_t src_rows = 0;
     size_t src_bytes = 0;
 
-    Aggregator::AggProcessInfo agg_process_info{};
+    Aggregator::AggProcessInfo agg_process_info;
+    ThreadData(Aggregator * aggregator)
+        : agg_process_info(aggregator)
+    {}
 };
 
 /// Aggregated data shared between AggBuild and AggConvergent Pipeline.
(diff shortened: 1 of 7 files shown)


yibin87 (approved):
LGTM

dbms/src/Operators/AggregateContext.h
```diff
@@ -27,7 +27,10 @@ struct ThreadData
     size_t src_rows = 0;
     size_t src_bytes = 0;
 
-    Aggregator::AggProcessInfo agg_process_info{};
+    Aggregator::AggProcessInfo agg_process_info;
+    ThreadData(Aggregator * aggregator)
```
SeaRise, line 31: how about just holding the `CancellationHoo`

windtalker: Because `CancellationHook` inside `Aggregator` maybe reset by `setCancellationHook`

SeaRise: ok

SeaRise (approved):

PR #8108, opened 2023-09-18
Handle snapshot in parallel if there is only one big region in raftstore v2

### LINKED ISSUE #8081: [Dynamic Region] TiFlash can't catch up with TiKV when there is only one big region pending prehandle
## Bug Report

CalvinNeo:

Issue Number: close #8081

Problem Summary:

In this PR, we also reorganized some codes in KVStore.cpp and Region.cpp.

Interface changes:
1. Add an interface `fn_get_config_json` to get **final** config on Proxy's side.
2. Add an interface `fn_approx_size` to get approximate size of different cf **only** for tablet sst reader.
3. Add an interface `fn_get_split_keys` to try to split a raft snapshot into `splits_count` parts, according to write cf. It may result in less or even zero parts if the size of the snapshot is small. The returned keys are not evenly distributed.

The feature only works when in raftstore v2 and there is few concurrent prehandling.

The whole process:
1. SSTReader calls proxy for approximate size, if it meets some threshold, we will start parallel prehandling.
2. SSTFilesToBlockInputStream compute how may splits we want to divide.
3. SSTReader ask proxy for all split keys, the whole process will fallback to single thread if proxy can't return enough keys.
4. For all `count` split parts, we will create another `count - 1` to handle in parallel.
5. If exception or other failure happens, the sub task will throw or return the error. If one sub task fails, all other prehandling subtask will be aborted.
6. Otherwise, it will return ingest files.
7. The caller thread will handle the head(leftmost) split, and wait for all other threads join. It will aggregate outputs of all subtasks. The outputs includes ingested files(in disk) and uncommitted key-value pairs(in memory)

How to split tasks by split keys?
Given a `split_key[i]`, a subtask who runs `split_id=i` will handle range in `(pk(split_key[i]), pk(split_key[i+1])]`. Which means it will skip all versions of pk of `split_key[i]`, and accepts all versions of pk of `split_key[i+1]`.

The order
```
region start -- head split(will also do final agg) -- split key 0 -- split 0 -- split key 1 -- split 1 -- split key n -- split n -- region end
```

#### Performance

| item | master | this pr |
|:-:| :-: |:-:|
| splits | 1(always) | 3 |
| lines | 11095216 | 10474970 |
| interval between two ddls | 2min | 1min20s |

| item | master | this pr |
|:-:| :-: |:-:|
| splits | 1(always) | 4 |
| lines | - | 20856771 |
| interval between two ddls | - | 1min21s |

img_v2_85a6a738-2024-42a5-b11f-da9e78b1da5g

### Check List

Tests 

- [x] Unit test
- [x] Integration test
      use sysbench to generate data for 300s, then add tiflash replica
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -41,29 +42,30 @@ SSTFilesToBlockInputStream::SSTFilesToBlockInputStream( //
...
     , opts(std::move(opts_))
 {
     log = Logger::get(opts.log_prefix);
-}
-
-SSTFilesToBlockInputStream::~SSTFilesToBlockInputStream() = default;
 
-void SSTFilesToBlockInputStream::readPrefix()
-{
+    // We have to initialize sst readers at an earlier stage,
```
CalvinNeo, line 57: that is before `readPrefix`

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -100,7 +103,8 @@ void SSTFilesToBlockInputStream::readPrefix()
             make_inner_func,
             ssts_write,
             log,
-            region->getRange());
+            region->getRange(),
+            soft_limit.has_value() ? soft_limit.value().split_id : DM::SSTScanSoftLimit::HEAD_SPLIT);
```
CalvinNeo, line 107: If soft_limit is not set, then it is the only split, which is the legacy case

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -130,12 +135,27 @@ void SSTFilesToBlockInputStream::readPrefix()
...
 {
     // There must be no data left when we write suffix
-    assert(!write_cf_reader || !write_cf_reader->remained());
-    assert(!default_cf_reader || !default_cf_reader->remained());
-    assert(!lock_cf_reader || !lock_cf_reader->remained());
+    if (!write_cf_reader)
+        return;
+    if (!write_cf_reader->remained())
+        return;
+    RUNTIME_CHECK(soft_limit.has_value());
```
CalvinNeo, line 149: If neither the previous two cases match, then this must be a parallel-handling-single-big-region-snapshot case, then the `cur` cursor is definetely beyond end_limit

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -146,33 +166,46 @@ void SSTFilesToBlockInputStream::readSuffix()
...
+            // or less that the last rowkey from the write column family.
             {
-                loaded_write_cf_key.assign(key.data, key.len);
-            }
-        } // Notice: `key`, `value` are string-view-like object, should never use after `next` called
-        write_cf_reader->next();
+                BaseBuffView key = write_cf_reader->keyView();
+                BaseBuffView value = write_cf_reader->valueView();
+                auto tikv_key = TiKVKey(key.data, key.len);
+                region->insert(ColumnFamilyType::Write, std::move(tikv_key), TiKVValue(value.data, value.len));
```
CalvinNeo, line 186: Different split uses different `RegionPtr`, so there will be no race

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -331,5 +380,112 @@ Block SSTFilesToBlockInputStream::readCommitedBlock()
...
+        return false;
+    const auto & start_limit = soft_limit.value().getStartLimit();
+    // If start is set to "", then there is no soft limit for start.
+    if (!start_limit)
+        return false;
+    // Safety `soft_limit` outlives returned base buff view.
+    reader->seek(cppStringAsBuff(soft_limit.value().raw_start));
+
+    // Skip other versions of the same PK.
+    // TODO(split) use seek to optimize if failed several iterations.
```
CalvinNeo, line 415: I prefer not do this optimation, since I am not too familiar with key pattern.

dbms/src/Storages/DeltaMerge/SSTFilesToBlockInputStream.cpp
```diff
@@ -331,5 +380,112 @@ Block SSTFilesToBlockInputStream::readCommitedBlock()
...
+    while (reader && reader->remained())
+    {
+        // Read until find the next pk.
+        auto key = reader->keyView();
+        // TODO the copy could be eliminated, but with many modifications.
+        auto tikv_key = TiKVKey(key.data, key.len);
+        auto current_truncated_ts = RecordKVFormat::getRawTiDBPK(RecordKVFormat::decodeTiKVKey(tikv_key));
```
CalvinNeo, line 422: I suppose this to be a raw pk without timestamp

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -240,21 +599,52 @@ PrehandleResult KVStore::preHandleSSTsToDTFiles(
...
+            auto split_keys = getSplitKey(log, this, new_region, sst_stream);
+
             ReadFromStreamResult result;
-            std::tie(result, prehandle_result)
-                = executeTransform(new_region, prehandle_task, job_type, storage, sst_stream, opt, tmt);
+            if (split_keys.empty())
+            {
+                LOG_INFO(
+                    log,
+                    "Single threaded prehandling for single big region, range={}, region_id={}",
```
JaySon-Huang, line 613:
```suggestion
                    "Single threaded prehandling for this region, range={}, region_id={}",
```

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+    RUNTIME_CHECK_MSG(
+        split_key_count >= 1,
+        "split_key_count should be more or equal than 1, actual {}",
+        split_key_count);
+    LOG_INFO(
+        log,
+        "Parallel prehandling for single big region, range={}, split keys={}, region_id={}",
+        new_region->getRange()->toDebugString(),
+        split_key_count,
+        new_region->id());
```
JaySon-Huang, line 411: Move this logging to line 620 is better for reading

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+        if (auto v = FailPointHelper::getFailPointVal(FailPoints::force_set_parallel_prehandle_threshold); v)
+            parallel_prehandle_threshold = std::any_cast<size_t>(v.value());
+    });
+    // If size is 0, do no parallel prehandle for this snapshot, which is legacy.
+    // If size is non-zero, use extra this many threads to prehandle.
+    std::vector<std::string> split_keys;
+    // Don't change the order of following checks, `getApproxBytes` involves some overhead,
+    // although it is optimized to bring about the minimum overhead.
+
+    if (new_region->getClusterRaftstoreVer() == RaftstoreVer::V2 && kvstore->getOngoingPrehandleTaskCount() < 2
```
JaySon-Huang, line 225:
```
    ... && kvstore->getOngoingPrehandleTaskCount() < 2
```

So when TiFlash is running pre-handle for one large Region snapshot, then the coming concurrent Region snapshot decoding will receive empty `split_keys` and run in single thread?

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+    // TODO(split) find solution if the snapshot has too many untrimmed data.
+    // TODO(split) recover this after integration test passes.
+    // constexpr size_t PARALLEL_PREHANDLE_THRESHOLD = 1 * 1024 * 1024 * 1024;
+    constexpr size_t PARALLEL_PREHANDLE_THRESHOLD = 1 * 1024 * 1024;
+    size_t parallel_prehandle_threshold = PARALLEL_PREHANDLE_THRESHOLD;
+    fiu_do_on(FailPoints::force_set_parallel_prehandle_threshold, {
+        if (auto v = FailPointHelper::getFailPointVal(FailPoints::force_set_parallel_prehandle_threshold); v)
+            parallel_prehandle_threshold = std::any_cast<size_t>(v.value());
+    });
+    // If size is 0, do no parallel prehandle for this snapshot, which is legacy.
```
JaySon-Huang, line 219:
```suggestion
    // If size is 0, do not parallel prehandle for this snapshot, which is legacy.
```

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+    LoggerPtr log,
+    KVStore * kvstore,
+    RegionPtr new_region,
+    std::shared_ptr<DM::SSTFilesToBlockInputStream> sst_stream)
+{
+    // We don't use this is the single snapshot is small, due to overhead in decoding.
+    // TODO(split) find solution if the snapshot has too many untrimmed data.
+    // TODO(split) recover this after integration test passes.
+    // constexpr size_t PARALLEL_PREHANDLE_THRESHOLD = 1 * 1024 * 1024 * 1024;
+    constexpr size_t PARALLEL_PREHANDLE_THRESHOLD = 1 * 1024 * 1024;
```
JaySon-Huang, line 213: does it pass?

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+            auto limit = DM::SSTScanSoftLimit(
+                extra_id,
+                std::string(split_keys[extra_id]),
+                extra_id + 1 == split_key_count ? std::string("") : std::string(split_keys[extra_id + 1]));
+            LOG_INFO(
+                log,
+                "Add extra parallel prehandle task split_id={}/total={} limit {}",
+                extra_id,
+                split_keys.size(),
+                limit.toDebugString());
```
JaySon-Huang, line 430:
```suggestion
            LOG_INFO(
                log,
                "Add extra parallel prehandle task split_id={} total={} limit={}",
                extra_id,
                split_keys.size(),
                limit.toDebugString());
```

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+                index,
+                extra_id,
+                ctx,
+                std::move(limit),
+                storage);
+            return true;
+        });
+        RUNTIME_CHECK_MSG(
+            add_result,
+            "Failed when add {}-th task for prehandling region_id={}",
```
JaySon-Huang, line 449:
```suggestion
            "Failed when adding {}-th task for prehandling region_id={}",
```

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +200,328 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+            new_region->id());
+    }
+    LOG_INFO(
+        log,
+        "Add extra parallel prehandle task split_id={}/total={} limit {}",
+        DM::SSTScanSoftLimit::HEAD_OR_ONLY_SPLIT,
+        split_keys.size(),
+        sst_stream->getSoftLimit().value().toDebugString());
+    auto [head_result, head_prehandle_result]
+        = executeTransform(new_region, prehandle_task, job_type, storage, sst_stream, opt, tmt);
```
JaySon-Huang, line 460: why we still need adding this?

dbms/src/Storages/KVStore/FFI/SSTReader.h
```diff
@@ -106,18 +150,25 @@ class MultiSSTReader : public SSTReader
...
         , current(0)
         , range(range_)
+        , split_id(split_id_)
     {
         assert(args.size() > 0);
-        LOG_INFO(log, "Open sst file first {} range {}", buffToStrView(args[current].path), range->toDebugString());
-        mono = initer(proxy_helper, args[current], range);
+        LOG_INFO(
+            log,
+            "Open sst file first {} range {} split_id={}",
```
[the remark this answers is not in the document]
CalvinNeo: `args[current].path` is not a path, but a unique identifier

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: There is another log says `Parallel prehandling for single big region ...`, the `single big` is just a specific string for identification

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: Yes, but it will make `KVStore::preHandleSSTsToDTFiles` even longer, it takes already ~200 lines with deep indention

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: `kvstore->getOngoingPrehandleTaskCount() < 2` ensures the coming snapshot is the only prehanding snapshot.
If another snapshot comes later before the current prehandling finishes, `kvstore->getOngoingPrehandleTaskCount() < 2` is false, so it will be handle in single thread.
This is because we tend to use all available cores to prehandle when there is only one snapshot. It will bring about many context switches if we do this to another snapshot in the mean time.

dbms/src/Storages/KVStore/FFI/SSTReader.h
```diff
@@ -95,7 +139,7 @@ class MultiSSTReader : public SSTReader
                 // We don't drop if mono is the last instance for safety,
                 // and it will be dropped as MultiSSTReader is dropped.
                 LOG_INFO(log, "Open sst file {}", buffToStrView(args[current].path));
```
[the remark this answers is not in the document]
CalvinNeo: ditto

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: yes, it passed the ut and manual test. however, i'd like to leave this until performance test is finished

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: I will move these logs into executeTransform

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: After discussion with @JaySon-Huang , this is a problem when there are a few but not a single big region snapshot. Meanwhile, if there are many cores, and a single big region, we will not use all cores to do prehandling, since it will result in too many splits.
These are two optimaztions, we will have another pr for this.

dbms/src/Storages/DeltaMerge/Decode/SSTFilesToBlockInputStream.cpp
```diff
@@ -146,33 +166,46 @@ void SSTFilesToBlockInputStream::readSuffix()
...
+                    loaded_write_cf_key.assign(key.data, key.len);
+                }
+            } // Notice: `key`, `value` are string-view-like object, should never use after `next` called
+            write_cf_reader->next();
+        }
+        else
+        {
+            // Load the last batch
+            break;
+        }
```
JaySon-Huang, line 200:
```
if (should_stop_advancing)
{
  // Load the last batch
  break;
}

// To decode committed rows from key-value pairs into block, we need to load
...
```

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -163,6 +206,318 @@ PrehandleResult KVStore::preHandleSnapshotToFiles(
...
+            return true;
+        });
+        RUNTIME_CHECK_MSG(
+            add_result,
+            "Failed when adding {}-th task for prehandling region_id={}",
+            extra_id,
+            new_region->id());
+    }
+    auto [head_result, head_prehandle_result]
+        = executeTransform(log, new_region, prehandle_task, job_type, storage, sst_stream, opt, tmt);
```
JaySon-Huang, line 455:
```suggestion
    // This will read the keys from the beginning to the first split key
    auto [head_result, head_prehandle_result]
        = executeTransform(log, new_region, prehandle_task, job_type, storage, sst_stream, opt, tmt);
```

JaySon-Huang: Is my understanding right?

JaySon-Huang (commented):
Rest LGTM

dbms/src/Debug/MockRaftStoreProxy.cpp
```diff
@@ -111,6 +111,13 @@ void fn_gc_rust_ptr(RawVoidPtr ptr, RawRustPtrType type_)
     case RawObjType::MockAsyncWaker:
         delete reinterpret_cast<MockAsyncWaker *>(ptr);
         break;
+    case RawObjType::MockString:
+        delete reinterpret_cast<std::string *>(ptr);
+        break;
+    case RawObjType::MockVecOfString:
+        auto * inner = reinterpret_cast<RustStrWithViewVecInner *>(ptr);
```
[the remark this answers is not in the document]
CalvinNeo: fixed

dbms/src/Debug/MockRaftStoreProxy.cpp
```diff
@@ -167,12 +174,21 @@ void fn_notify_compact_log(
...
-RaftstoreVer fn_get_cluster_raftstore_version(RaftStoreProxyPtr ptr, uint8_t, int64_t)
+static RaftstoreVer fn_get_cluster_raftstore_version(RaftStoreProxyPtr ptr, uint8_t, int64_t)
 {
     auto & x = as_ref(ptr);
     return x.cluster_ver;
 }
 
+static RustStrWithView fn_get_config_json(RaftStoreProxyPtr, uint64_t)
+{
+    auto * s = new std::string(R"({"raftstore":{"snap-handle-pool-size":4}})");
```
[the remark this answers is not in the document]
CalvinNeo: comment added

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -370,16 +756,17 @@ void Region::beforePrehandleSnapshot(uint64_t region_id, std::optional<uint64_t>
...
-void Region::afterPrehandleSnapshot()
+void Region::afterPrehandleSnapshot(int64_t ongoing)
 {
     if (getClusterRaftstoreVer() == RaftstoreVer::V2)
     {
         data.orphan_keys_info.pre_handling = false;
         LOG_INFO(
             log,
-            "After prehandle, remains orphan keys {} removed orphan keys {} [region_id={}]",
+            "After prehandle, remains orphan keys {} removed orphan keys {} ongoing {} [region_id={}]",
```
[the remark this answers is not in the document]
CalvinNeo: ongoing means ongoing prehandling snapshots

dbms/src/Storages/DeltaMerge/Decode/SSTFilesToBlockInputStream.cpp
CalvinNeo: fixed

dbms/src/Storages/DeltaMerge/PKSquashingBlockInputStream.h
```diff
@@ -147,8 +157,12 @@ class PKSquashingBlockInputStream final : public IBlockInputStream
                 {
                     if (unlikely(next_pk < last_curr_pk))
                         throw Exception(
-                            "InputStream is not sorted, pk in next block is smaller than current block: "
-                                + next_pk.toDebugString() + " < " + last_curr_pk.toDebugString(),
+                            fmt::format(
```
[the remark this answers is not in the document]
CalvinNeo: fixed

JaySon-Huang (approved):
lgtm

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: yes

PR #8115, opened 2023-09-20
Spill in pipeline model not driven by input block

windtalker:

Issue Number: ref #7738

Problem Summary:

Currently, spill is triggered by input block, that is to say for operator that support spill(`sort/agg/join`), even if it is marked for spill by auto spill trigger, it has to wait a new input block, and only after it handle the input block(which usually will use extra memory), then there is a chance for these operator to check if it has data to spill. This is required in pull model, since in pull model, the execution is actually driven by input block, the operator itself has no chance to spill without getting input block from its children, however, in pipeline model, this is not a must, for a operator that is marked to be spill by auto spill trigger, the operator can start spill without waiting for new input block, so this pr make spill in pipeline model not triggered by input block.
### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


SeaRise (approved):

yibin87 (approved):
LGTM

PR #8168, opened 2023-10-08
add some failpoints for remote read

SeaRise:

Issue Number: ref  #6233

Problem Summary:

- add some failpoint for remote read
- Use Limiter instead of thread pool for cop.
- move `analyzeDAGEncodeType` from `DAGUtils` to `DAGContext`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


SeaRise:
run-all-tests

SeaRise:
run-all-tests

SeaRise:
run-all-tests

dbms/src/Common/SimpleFixedThreadPool.cpp
```diff
@@ -0,0 +1,63 @@
...
+        std::unique_lock lock(mutex);
+        has_free_thread.wait(lock, [this] { return active_jobs < m_size; });
+        ++active_jobs;
+    }
+    thread_mgr->schedule(false, name, [&, job]() {
+        job();
+        {
+            std::lock_guard lock(mutex);
+            --active_jobs;
+            has_free_thread.notify_one();
```
JaySon-Huang, line 42: I think we'd better calling `notify_one` out of the scope of lock

SeaRise: https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one
```
The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock. However, some implementations (in particular many implementations of pthreads) recognize this situation and avoid this "hurry up and wait" scenario by transferring the waiting thread from the condition variable's queue directly to the queue of the mutex within the notify call, without waking it up.

Notifying while under the lock may nevertheless be necessary when precise scheduling of events is required, e.g. if the waiting thread would exit the program if the condition is satisfied, causing destruction of the notifying thread's condition variable. A spurious wakeup after mutex unlock but before notify would result in notify called on a destroyed object.
```
It doesn't seem to matter

windtalker (approved):
LGTM

dbms/src/Common/SimpleFixedThreadPool.cpp
```diff
@@ -0,0 +1,63 @@
...
+        }
```
SeaRise, line 43:
```suggestion
        }
        has_free_thread.notify_one();
```

dbms/src/Common/FailPoint.cpp
```diff
@@ -102,6 +102,8 @@ namespace DB
     M(exception_after_large_write_exceed)                    \
     M(proactive_flush_force_set_type)                        \
     M(exception_when_fetch_disagg_pages)                     \
+    M(force_agg_on_partial_block)                            \
```
SeaRise, line 105:
```suggestion
```

dbms/src/Common/SimpleFixedThreadPool.cpp
SeaRise: updated

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -131,6 +132,10 @@ std::tuple<std::optional<RegionRetryList>, RegionException::RegionReadStatus> Ma
                 if (batch_cop)
                     status = RegionException::RegionReadStatus::NOT_FOUND;
             });
+            fiu_do_on(FailPoints::random_trigger_remote_read, {
+                if (batch_cop)
+                    status = RegionException::RegionReadStatus::NOT_FOUND;
+            });
```
gengliqi, line 138: It seems these three failpoint can be merged.

dbms/src/Common/SimpleFixedThreadPool.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void SimpleFixedThreadPool::schedule(ThreadManager::Job job)
+{
+    {
+        std::unique_lock lock(mutex);
+        has_free_thread.wait(lock, [this] { return active_jobs < m_size; });
+        ++active_jobs;
+    }
+    thread_mgr->schedule(false, name, [&, job]() {
```
gengliqi, line 37: https://github.com/pingcap/tiflash/blob/f49779562c9ad3233e69ebca526ed98fc500f57a/dbms/src/Common/ThreadManager.cpp#L46-L70
The memory of `futures` vector will keep growing forever.

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
SeaRise: These three failpoints have their own uses.
The first failpoint will only be triggered once.
The second failpoint will always be triggered,
The third failpoint is triggered randomly according to probability

gengliqi: Hmmm. Is it necessary to use these three failpoints in one test at the same time? If not, keeping one would be enough.

dbms/src/Common/SimpleFixedThreadPool.cpp
SeaRise: ok, updated.

dbms/src/Common/Limiter.h
```diff
@@ -0,0 +1,66 @@
...
+    {
+        std::unique_lock lock(mu);
+        cv.wait(lock, [&]() { return active_count == 0; });
+    }
+
+    T execute(std::function<T()> func)
+    {
+        {
+            std::unique_lock lock(mu);
+            cv.wait(lock, [&]() { return active_count < limit; });
```
windtalker, line 45: Use `wait_for` here so we can return error if wait time execceded?

guo-shaoge (approved):

dbms/src/Common/Limiter.h
SeaRise: ok, updated.

windtalker (approved):
LGTM

PR #8176, opened 2023-10-10
refine LAC in disaggregated mode

### LINKED ISSUE #8178: Disaggregated storage tiflash should not start LAC
## Bug Report

For disaggregated storage tiflash node, will not use pipeline model(a.k.a. RCQ) to execute. So no need to start LAC

guo-shaoge:

Issue Number: close #8178

Problem Summary:
1. Do not start LAC and pipeline TaskScheduler for **WN**. Because WN will not run any queires. And for now WN doesn't support resource control.
2. Make low token threshold lower. 0.8 is too high. It may cause rpc to GAC too frequent.
3. Delete server id from GAC when LAC stopped. Because on Serverless Tier, **CN** will scale in/out frequently, so we can avoid too many server id allocated from GAC.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
1. start tidb cluster with one or more tiflash
2. search log(keyword: get unique_client_id succeed) to check the unique client id of LAC. Run `etcdctl get --prefix '/tidb/server_id/{client_id_of_LAC}'` to check it from GAC, should get successfully.
4. kill tiflash
5. run `etcdctl get --prefix '/tidb/server_id/{client_id_of_LAC}'` , the server id has been deleted.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/ResourceControl/LocalAdmissionController.h
+++ b/dbms/src/Flash/ResourceControl/LocalAdmissionController.h
@@ -441,7 +441,7 @@ public:
 
     // Fetch resource group info from GAC if necessary and store in local cache.
     // Throw exception if got error when fetching from GAC.
-    bool warmupResourceGroupInfoCache(const std::string & name);
+    void warmupResourceGroupInfoCache(const std::string & name);
 
     static bool isRUExhausted(uint64_t priority) { return priority == std::numeric_limits<uint64_t>::max(); }
 
@@ -529,7 +529,7 @@ private:
     }
 
     // Return true if new resource group is added.
-    bool addResourceGroup(const resource_manager::ResourceGroup & new_group_pb)
+    void addResourceGroup(const resource_manager::ResourceGroup & new_group_pb)
     {
         uint64_t user_ru_per_sec = new_group_pb.r_u_settings().r_u().settings().fill_rate();
         if (max_ru_per_sec.load() < user_ru_per_sec)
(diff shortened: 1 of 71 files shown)


SeaRise (approved):
lgtm

dbms/src/Flash/executeQuery.cpp
```diff
@@ -155,6 +155,7 @@ std::optional<QueryExecutorPtr> executeAsPipeline(Context & context, bool intern
     }
 
     prepareForExecute(context);
+    dag_context.enableResourceControl();
```
windtalker, line 158: Why change this

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -580,6 +603,7 @@ class LocalAdmissionController final : private boost::noncopyable
         = std::chrono::steady_clock::now();
 
     ::pingcap::kv::Cluster * cluster = nullptr;
+    std::atomic<bool> need_reset_unique_client_id{false};
```
windtalker, line 606: When will this be false? Looks like it is always true once `unique_client_id` is initialized?

guo-shaoge: This is to avoid `deleteServerIDFromGAC()` is called before `acquireServerIDFromGAC()` succeed.

dbms/src/Flash/executeQuery.cpp
guo-shaoge: We have to put it before PipelineExecutor is constructed.

windtalker (approved):
LGTM

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -479,6 +485,23 @@ class LocalAdmissionController final : private boost::noncopyable
             if (thread.joinable())
                 thread.join();
         }
+
+        if (need_reset_unique_client_id.load())
+        {
+            try
+            {
+                etcd_client->deleteServerIDFromGAC(unique_client_id);
```
SeaRise, line 493: It looks like `unique_client_id` has a data race, use `std::atomic_uint64` instead?

guo-shaoge: ok

PR #8244, opened 2023-10-25
keep toCase function the same as golang implementation

### LINKED ISSUE #7695: The result of `upper()` and `lower()` is different with TiDB
## Bug Report

```
mysql> drop table t;
Query OK, 0 rows affected (0.34 sec)

mysql> create table t(a char(20));
Query OK, 0 rows affected (0.15 sec)

mysql> insert into t values ('𞤀');
Query OK, 1 row affected (0.01 sec)

mysql> ALTER TABLE t set tiflash replica 1;
Query OK, 0 rows affected (0.14 sec)

mysql> set @@session.tidb_isolation_read_engines = 'tiflash';
Query OK, 0 rows affected (0.00 sec)

mysql> select * from t where lower(a) = '𞤀';
+------+
| a    |
+------+
| 𞤀     |
+------+
1 row in set (0.07 sec)

mysql> set @@session.tidb_isolation_read_engines = 'tikv';
Query OK, 0 rows affected (0.01 sec)

mysql> select * from t where lower(a) = '𞤀';
Empty set (0.00 sec)
```

```
Same as TiDB and TiKV
```

```
Result not same
```

2023-06-24 nightly

windtalker:

Issue Number: close #7695

Problem Summary:

The core code for TiFlash's `lower/upper` function is
https://github.com/pingcap/tiflash/blob/57c0c57e91a1e95f2ff4a13597a2898409ce11de/dbms/src/Functions/CharUtil.h#L235
And this function is actually a directly translation of golang's implementation:
https://github.com/golang/go/blob/go1.21.3/src/unicode/letter.go#L211

This pr update the code in TiFlash to keep it the same as latest golang's implemenation.
### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


tests/fullstack-test/expr/issue_7695.test
```diff
@@ -0,0 +1,30 @@
...
+mysql> alter table test.t set tiflash replica 1
+func> wait_table test t
+
+mysql> insert into test.t values ('𞤀');
+mysql> set tidb_enforce_mpp=1; select a, lower(a) from test.t;
++------+----------+
+| a    | lower(a) |
++------+----------+
+| 𞤀     | 𞤢         |
++------+----------+
```
windtalker, line 26: It should be a display issue, the real result looks like this
Screenshot 2023-10-25 at 17 53 19

SeaRise (approved):

tests/fullstack-test/expr/issue_7695.test
SeaRise: How about adding unit test instead of Integration  test?

yibin87 (approved):
LGTM

PR #8199, opened 2023-10-17
Make part of the ci work for raftstore-v2

CalvinNeo:
### What problem does this PR solve?

Issue Number: close #8089

Problem Summary:

We stablize the current support for raftstore-v2 into ci tests.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8245, opened 2023-10-25
Fix integration test of parallel prehandling

### LINKED ISSUE #8246: Integration test of parallel prehandling is not stable
## Bug Report

CalvinNeo:

Issue Number: close #8246

Problem Summary:

The problem is when a retryable exception is thrown in a thread, it will notify other thread to abort. As a result, other thread will abort. If the abort result is captured before the retryable result by head split, it will abort directly without retry.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -189,7 +181,10 @@ static inline std::tuple<ReadFromStreamResult, PrehandleResult> executeTransform
         if (stream->isAbort())
         {
             stream->cancel();
-            res = ReadFromStreamResult{.error = ReadFromStreamError::Aborted, .extra_msg = "", .region = new_region};
+            res = ReadFromStreamResult{
```
CalvinNeo, line 184: When abort, we will use abort_error, which is by default Aborted, but sometimes could be other value.

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -567,7 +563,6 @@ void executeParallelTransform(
         LOG_DEBUG(log, "Try fetch prehandle task split_id={}, region_id={}", extra_id, new_region->id());
         async_tasks.fetchResult(extra_id);
     }
-    if (head_result.error == ReadFromStreamError::Ok)
```
JaySon-Huang, line 570: Why must we run through the codes below if the "head split" is not OK now?

JaySon-Huang:
which unstable test case is caused by this issue?

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
```diff
@@ -448,6 +443,7 @@ static void runInParallel(
         if (part_result.error == ReadFromStreamError::ErrUpdateSchema)
         {
             prehandle_task->abort_flag.store(true);
+            prehandle_task->abort_error.store(ReadFromStreamError::ErrUpdateSchema);
```
JaySon-Huang, line 446: I think this fix could lead to the same issue again:

Thread A: run `prehandle_task->abort_flag.store(true)` and CPU yield before `prehandle_task->abort_error.store(ReadFromStreamError::ErrUpdateSchema);`
Thread B: read `prehandle_task->abort_error.load()` and see the reason is "abort"
Thread A: resume and set the reason is "update schema"

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/PrehandleSnapshot.cpp
CalvinNeo: fixed by using a single atomic variable

dbms/src/Storages/KVStore/MultiRaft/PreHandlingTrace.h
```diff
@@ -22,14 +22,35 @@
 
 namespace DB
 {
+enum class ReadFromStreamError
```
JaySon-Huang, line 25: I think `PrehandleError` or `PrehandleStatus` is a more reasonable name here

JaySon-Huang:
The fix of https://github.com/pingcap/tiflash/pull/8199#discussion_r1372552881 is not pushed?

dbms/src/Storages/KVStore/MultiRaft/PreHandlingTrace.h
CalvinNeo: Renamed to `PrehandleTransformStatus`, because there is a `PrehandleResult` for the whole prehandle process

JaySon-Huang (approved):
LGTM

PR #8297, opened 2023-11-02
Reduce redundant calculation in join probe, optimize `mergeNullAndFilterResult`

### LINKED ISSUE #8294: Avoid  redundant calculation of hash data during join probe stage
## Enhancement
In join probe stage, for each input block, it will be wrapped inside [`ProbeProcessInfo`](https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/ProbeProcessInfo.h#L32), this is because if the build side has many duplicated key entries, the probe stage may expand the input block greatly, and inorder to keep the max-row-size of a output block, `ProbeProcessInfo` is used to support probe only part of the input block. That is to say for an input block, it may be probed multiple times, each time of part of the data in the block is processed.

For each probe, it will call `probeBlockImplTypeCase` to do the probe, and inside that function, it will calculate the hash data of the whole block:
https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/JoinPartition.cpp#L1480-L1494

If a block is probed multiple times, the hash data will be calculated multiple times, which is meaningless and redundant.

windtalker:

Issue Number: close #8294

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
mysql> desc t;
+----------+
| count(*) |
+----------+
|   393216 |
+----------+
```
Before this pr(based on some other optimization)
```
mysql> select count(t1.value) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-----------------+                                  
| count(t1.value) |                                  
+-----------------+                                  
|     34359738368 |                                  
+-----------------+                                  
1 row in set (42.24 sec)
```
After this pr(based on the same optimization as above)
```
mysql> select count(t1.value) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-----------------+
| count(t1.value) |
+-----------------+
|     34359738368 |
+-----------------+
1 row in set (34.63 sec)
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/Join.cpp
+++ b/dbms/src/Interpreters/Join.cpp
@@ -1185,7 +1185,7 @@ void Join::handleOtherConditionsForOneProbeRow(Block & block, ProbeProcessInfo &
     throw Exception("Logical error: unknown combination of JOIN", ErrorCodes::LOGICAL_ERROR);
 }
 
-Block Join::doJoinBlockHash(ProbeProcessInfo & probe_process_info) const
+Block Join::doJoinBlockHash(ProbeProcessInfo & probe_process_info, const JoinBuildInfo & join_build_info) const
 {
     assert(probe_process_info.prepare_for_probe_done);
     probe_process_info.updateStartRow<false>();
@@ -1251,13 +1251,6 @@ Block Join::doJoinBlockHash(ProbeProcessInfo & probe_process_info) const
     auto & filter = probe_process_info.filter;
     auto & offsets_to_replicate = probe_process_info.offsets_to_replicate;
 
-    JoinBuildInfo join_build_info{
-        enable_fine_grained_shuffle,
-        fine_grained_shuffle_count,
-        isEnableSpill(),
-        hash_join_spill_context->isSpilled(),
-        build_concurrency,
-        restore_round};
     JoinPartition::probeBlock(
         partitions,
         rows,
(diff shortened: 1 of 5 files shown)


dbms/src/Interpreters/Join.cpp
```diff
@@ -793,20 +793,45 @@ void mergeNullAndFilterResult(
...
+            if (null_as_true)
+            {
+                for (size_t i = 0; i < nullmap_vec->size(); ++i)
+                    filter_column[i] = filter_column[i] || (*filter_vec)[i];
+            }
             else
-                filter_column[i] = filter_column[i] && (*filter_vec)[i];
+            {
+                for (size_t i = 0; i < nullmap_vec->size(); ++i)
+                    filter_column[i] = !filter_column[i] && (*filter_vec)[i];
```
SeaRise, line 807: Doesn't seem to be equivalent to https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/Join.cpp#L798-L803

SeaRise: sorry, my mistake..

dbms/src/Interpreters/Join.cpp
```diff
@@ -1365,7 +1387,16 @@ Block Join::joinBlockHash(ProbeProcessInfo & probe_process_info) const
 {
     std::vector<Block> result_blocks;
     size_t result_rows = 0;
-    probe_process_info.prepareForHashProbe(key_names_left, non_equal_conditions.left_filter_column, kind, strictness);
+    bool need_compute_hash = enable_fine_grained_shuffle
+        || (hash_join_spill_context->isSpillEnabled() && !hash_join_spill_context->isSpilled());
```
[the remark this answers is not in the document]
windtalker: Yes, updated.

SeaRise (approved):

gengliqi (approved):
LGTM

JaySon-Huang:
Seems like a pretty big performance boost to join. Can we backport it to the LTS branch? @windtalker

windtalker:
> Seems like a pretty big performance boost to join. Can we backport it to the LTS branch? @windtalker

It only gets performance benefit if the build side has many duplicate keys, for example, the average rows count for the same key is more than 10, which I think is kind of corner case, so there is no plan to backport, but if we really meet such performance issues, it can be backport easily.

PR #8320, opened 2023-11-06
Cache probe column when key has many entries

### LINKED ISSUE #8295: Optimize join probe stage if one hash table key has many entries
## Enhancement
In TiFlash, the entry of a join hash table is <key, RowRefList>, where `RowRefList` is a list of the rows that has the same key, and each element if the `RowRefList` is a pair of <block_index, row_index>. During hash join probe stage, if a probe row has an existing key in the hash table, it will combine all the matched rows in the `RowRefList` using [`AddFound`](https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/JoinPartition.cpp#L1296), inside `AddFound`, it just iterator over the `RowRefList` and insert the data row by row
https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/JoinPartition.cpp#L1321-L1329

If the list is very long (~10000), it will take a lot if time in `AddFound`. We need to find a way to optimize this implementation.

A straightforward optimization is to cache `AddFound` result of such key, so next time when a probe row which has the same key comes, we can just use `insertRange` to batch insert the data instead of iterator over `RowRefList` and insert the data row by row

windtalker:

Issue Number: close #8295

Problem Summary:

1. add `list_length` and `cache_column_info` in RefList of Join hash table
2. cache and reuse cached columns if a join key has many duplicated entries
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
mysql> desc t;
+----------+
| count(*) |
+----------+
|   393216 |
+----------+
```
Before this pr
```
mysql>  select count(*) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-------------+
| count(*)    |
+-------------+
| 34359738368 |
+-------------+
1 row in set (3 min 21.22 sec)
```
After this pr
```
mysql>  select count(*) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-------------+
| count(*)    |
+-------------+
| 34359738368 |
+-------------+
1 row in set (1 min 16.07 sec)
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/JoinHashMap.h
+++ b/dbms/src/Interpreters/JoinHashMap.h
@@ -45,7 +45,7 @@ struct CachedColumnInfo
     Columns columns;
     CachedColumnState state = CachedColumnState::NOT_CACHED;
     void * next;
-    CachedColumnInfo(void * next_)
+    explicit CachedColumnInfo(void * next_)
         : next(next_)
     {}
 };
--- a/dbms/src/Interpreters/JoinHashMap.h
+++ b/dbms/src/Interpreters/JoinHashMap.h
@@ -45,7 +45,6 @@ struct RowRefList : RowRef
 {
     RowRefList * next = nullptr;
     size_t list_length = 0;
-    /// todo use raw point and allocate memory in pool?
     std::unique_ptr<CachedColumnInfo> cached_column_info;
 
     RowRefList() = default;
@@ -66,7 +65,6 @@ struct RowRefListWithUsedFlag : RowRef
     mutable std::atomic<bool> used{};
     RowRefListWithUsedFlag * next = nullptr;
     size_t list_length = 0;
-    /// todo use raw point and allocate memory in pool?
     std::unique_ptr<CachedColumnInfo> cached_column_info;
 
     void setUsed() const
--- a/dbms/src/Interpreters/JoinHashMap.h
+++ b/dbms/src/Interpreters/JoinHashMap.h
@@ -37,6 +37,9 @@ struct RowRef
 struct RowRefList : RowRef
 {
     RowRefList * next = nullptr;
+    mutable std::mutex cached_columns_mu;
+    mutable Columns cached_columns;
+    mutable bool generate_cached_columns = false;
 
     RowRefList() = default;
     RowRefList(const Block * block_, size_t row_num_)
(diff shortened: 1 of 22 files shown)


dbms/src/Interpreters/JoinPartition.cpp
```diff
@@ -1239,6 +1282,65 @@ struct Adder<ASTTableJoin::Kind::LeftOuterSemi, ASTTableJoin::Strictness::Any, M
...
+        {
+            added_columns[j]->insertRangeFrom(*cached_column_info->columns[j], 0, rows_added);
+        }
+    }
+}
+
+Columns constructCachedColumns(MutableColumns & added_columns, size_t rows, size_t columns)
+{
+    Columns cached_columns;
+    size_t start_offset = added_columns[0]->size() - rows;
```
yibin87, line 1330: Better check size >= rows here.

dbms/src/Interpreters/JoinHashMap.h
```diff
@@ -33,15 +33,29 @@ struct RowRef
     {}
 };
 
+struct CachedColumnInfo
+{
+    std::mutex mu;
+    Columns columns;
+    bool generate_cached_columns = false;
```
yibin87, line 40: An enum value instead of boolean value may make it more clean.

windtalker: OK, add an enum here.

dbms/src/Interpreters/JoinPartition.cpp
windtalker: OK, add an assert

yibin87 (approved):
LGTM

dbms/src/Interpreters/JoinPartition.cpp
```diff
@@ -1239,6 +1294,80 @@ struct Adder<ASTTableJoin::Kind::LeftOuterSemi, ASTTableJoin::Strictness::Any, M
...
+        {
+            for (size_t j = 0; j < columns; ++j)
+                cached_columns.push_back(added_columns[j]->cut(start_offset, rows));
+        }
+    }
+    std::unique_lock lock(cached_column_info->mu);
+    if (!cached_columns.empty())
+    {
+        assert(cached_column_info->columns.empty());
+        cached_column_info->columns.insert(
```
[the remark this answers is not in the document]
windtalker: Done

PR #7064, opened 2023-03-14
disable join from tiflash client

windtalker:
### What problem does this PR solve?

Issue Number: close #7054

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8366, opened 2023-11-13
Support finalize for join

### LINKED ISSUE #8296: Join only construct joined columns that is needed by its parent operator
## Enhancement
In TiFlash, the output schema of join is [left_columns, right_columns], for example
```
select a.id from a join b on a.id = b.id and a.id1 = b.id1 and a.id2 = b.id2 and a.id3 = b.id3
```
The output schema of join is [a.id, a.id1, a.id2, a.id3, b.id, b.id1, b.id2, b.id3]

However, only `a.id` is actually needed. And inside join, the cost of combine the left columns with the right columns is actually very high: for left columns, it need to call [`replicateRange`](https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/Join.cpp#L1291), and for right columns, it need to be added row by row using [`AddFound`](https://github.com/pingcap/tiflash/blob/e597b78a6f5595ba50ffe839daca798ced1d1bc4/dbms/src/Interpreters/JoinPartition.cpp#L1296), if we can only return the column used by the parent operators, it will avoid some meaningless cost and increase the performance of join.

windtalker:

Issue Number: close #8296

Problem Summary:
Support `finalize` for join
- Support `finalize` for PhysicalJoin
- Make `FinalizeRule` works for the whole planner tree(before this pr, `FinalizeRule` will stop at the First PhyscialJoin node)
- Some refine in `ProbeProcessInfo` and `Join.cpp`
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
mysql> desc t;
+----------+
| count(*) |
+----------+
|   393216 |
+----------+
```
Before this pr
```
mysql>  select count(*) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-------------+
| count(*)    |
+-------------+
| 34359738368 |
+-------------+
1 row in set (1 min 5.94 sec)
```
After this pr
```
mysql>  select count(*) from t t1 join t t2 on t1.from_id = t2.to_id and t1.value > t2.value;
+-------------+
| count(*)    |
+-------------+
| 34359738368 |
+-------------+
1 row in set (52.86 sec)
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -68,11 +68,8 @@ PhysicalPlanNodePtr PhysicalJoin::build(
     RUNTIME_CHECK(left);
     RUNTIME_CHECK(right);
 
-    left->finalize();
-    right->finalize();
-
-    const Block & left_input_header = left->getSampleBlock();
-    const Block & right_input_header = right->getSampleBlock();
+    const Block & left_input_header = Block(left->getSchema());
+    const Block & right_input_header = Block(right->getSchema());
```
SeaRise, line 72: Do `probe/build_side_header` also need to be changed to `Block(probe/build_plan->getSchema())`?

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -324,14 +323,58 @@ void PhysicalJoin::buildPipeline(PipelineBuilder & builder, Context & context, P
...
+            probe_required.push_back(name);
+        else
+        {
+            has_missed_column = true;
+            missed_column_name = name;
+            break;
+        }
+    }
+    if (has_missed_column)
+        throw Exception("Meet unknown column: {}", missed_column_name);
```
SeaRise, line 359: The same exception will be thrown in https://github.com/pingcap/tiflash/blob/adf85696270f864c0317526919b37fbdfe5a6417/dbms/src/Interpreters/ExpressionActions.cpp#L570-L574.
No need to throw exception here.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -324,14 +323,58 @@ void PhysicalJoin::buildPipeline(PipelineBuilder & builder, Context & context, P
...
-    // schema.size() >= parent_require.size()
     FinalizeHelper::checkSchemaContainsParentRequire(schema, parent_require);
+    join_ptr->finalize(parent_require);
+    auto required_input_columns = join_ptr->getRequiredColumns();
+    std::unordered_set<String> build_schema_sets;
+    for (const auto & name : build_side_prepare_actions->getSampleBlock().getNames())
+    {
+        build_schema_sets.insert(name);
+    }
+    std::unordered_set<String> probe_schema_sets;
```
SeaRise, line 336: seems `set<string>` is useless, because there are no duplicate column names in the block...

dbms/src/Flash/Planner/PhysicalPlanNode.cpp
```diff
@@ -63,9 +63,34 @@ String PhysicalPlanNode::toSimpleString()
     return fmt::format("{}|{}", type.toString(), isTiDBOperator() ? executor_id : "NonTiDBOperator");
 }
 
-void PhysicalPlanNode::finalize()
+void PhysicalPlanNode::finalize(const Names & parent_require)
 {
-    finalize(DB::toNames(schema));
+    if (finalized)
```
SeaRise, line 68:
```suggestion
    if unlikely (finalized)
```

dbms/src/Flash/Planner/FinalizeHelper.cpp
```diff
@@ -69,6 +69,8 @@ void checkSampleBlockContainsParentRequire(const Block & sample_block, const Nam
 {
     for (const auto & parent_require_column : parent_require)
     {
+        if (!sample_block.has(parent_require_column))
+            throw Exception("test");
```
SeaRise, line 73:
```suggestion
```

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -324,14 +323,58 @@ void PhysicalJoin::buildPipeline(PipelineBuilder & builder, Context & context, P
...
+    {
+        child_required_columns.push_back(ExpressionActions::getSmallestColumn(build()->getSchema()));
+    }
+    build()->finalize(child_required_columns);
+    probe_side_prepare_actions->finalize(probe_required);
+    child_required_columns = probe_side_prepare_actions->getRequiredColumns();
+    if unlikely (child_required_columns.empty())
+    {
+        child_required_columns.push_back(ExpressionActions::getSmallestColumn(probe()->getSchema()));
+    }
```
SeaRise, line 372: Seems useless since this case is taken into account in ExpressionAction?
https://github.com/pingcap/tiflash/blob/adf85696270f864c0317526919b37fbdfe5a6417/dbms/src/Interpreters/ExpressionActions.cpp#L647-L653

dbms/src/Interpreters/ExpressionActions.cpp
```diff
@@ -562,6 +563,74 @@ std::string ExpressionActions::getSmallestColumn(const NamesAndTypesList & colum
     return res;
 }
 
+void ExpressionActions::pruneInputColumns(const Names & output_columns)
```
SeaRise, line 566: What is the difference from `finalize`?

dbms/src/DataStreams/ScanHashMapAfterProbeBlockInputStream.cpp
```diff
@@ -199,51 +201,50 @@ ScanHashMapAfterProbeBlockInputStream::ScanHashMapAfterProbeBlockInputStream(
...
     }
 
-    for (size_t i = 0; i < num_columns_right; ++i)
-        column_indices_right.push_back(num_columns_left + i);
-
-    for (size_t i = 0; i < num_columns_left; ++i)
+    for (size_t i = 0; i < column_indices_left.size(); ++i)
     {
-        const auto & column_with_type_and_name = result_sample_block.getByPosition(column_indices_left[i]);
+        const auto & column_with_type_and_name = result_sample_block.getByPosition(i);
```
SeaRise, line 233: why change here?

windtalker: Before this change, `column_indices_left[i]` is always `i`, and `i` is also the column pos in both `result_sample_block` and `left_sample_block`
After this change, `column_indices_left[i]` is not always `i`, and `column_indices_left[i]` is the column pos in `left_sample_block`, `i` is the column pos in `result_sample_block`

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
windtalker: OK.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
windtalker: Refined.

dbms/src/Interpreters/ExpressionActions.cpp
windtalker: It does not remove column, if a column is removed, it will cause error for nullaware join since it intends to reuse the columns: https://github.com/pingcap/tiflash/blob/adf85696270f864c0317526919b37fbdfe5a6417/dbms/src/Interpreters/NullAwareSemiJoinHelper.cpp#L322-L327

dbms/src/Flash/Planner/FinalizeHelper.cpp
windtalker: Done

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
windtalker: Yes, I've removed it.

dbms/src/Flash/Coprocessor/JoinInterpreterHelper.cpp
```diff
@@ -453,5 +450,20 @@ std::vector<RuntimeFilterPtr> TiFlashJoin::genRuntimeFilterList(
     }
     return result;
 }
+
+NamesAndTypes genDAGExpressionAnalyzerSourceColumns(Block block, const NamesAndTypes & tidb_schema)
```
SeaRise, line 454: The logic of this function makes me a little confused...
Can you add some comments?

dbms/src/Interpreters/ExpressionActions.cpp
SeaRise: If a column is required, why not set it to required_column?

windtalker: Because at the time of calling `pruneInputColumns`, we don't know which columns need to be kept.
For example, the original output columns are
`a, b, c, func(a), func(b), func(a) + func(b)`
And the output columns is `func(a) + func(b)`
If use `finalize`, then `finalize` will insert `REMOVE_COLUMN` to remove `func(a), func(b), a, b, c` to make sure the output column is `func(a) + func(b)`
While for nullaware join, it actually require the output block keeps `a, b` as the first/second column. So not easy to reuse `finalize` function.

dbms/src/Flash/Coprocessor/JoinInterpreterHelper.cpp
windtalker: OK, added.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -324,14 +327,33 @@ void PhysicalJoin::buildPipeline(PipelineBuilder & builder, Context & context, P
...
+        else
+            /// if name not exists in probe side, it will throw error when call `probe_size_prepare_actions->finalize(probe_required)`
+            probe_required.push_back(name);
+    }
+
+    build_side_prepare_actions->finalize(build_required);
+    build()->finalize(build_side_prepare_actions->getRequiredColumns());
+
+    probe_side_prepare_actions->finalize(probe_required);
+    probe()->finalize(probe_side_prepare_actions->getRequiredColumns());
```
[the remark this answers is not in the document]
windtalker: `Project` is not allowed due to nullaware join assums this
https://github.com/pingcap/tiflash/blob/72db40557a07c0148b885fb670ea3a05186666b3/dbms/src/Interpreters/NullAwareSemiJoinHelper.cpp#L322-L324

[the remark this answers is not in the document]
windtalker: I give a second thought on this, and I think it is ok to insert project to build/probe_side_prepare_actions. I will do some tests on it.

dbms/src/DataStreams/ScanHashMapAfterProbeBlockInputStream.cpp
SeaRise: ok

dbms/src/Interpreters/Join.cpp
```diff
@@ -2497,4 +2590,86 @@ void Join::wakeUpAllWaitingThreads()
...
+        updated_require.push_back(non_equal_conditions.other_cond_name);
+    /// nullaware/semi join will reuse the input columns so need to let finalize keep the input columns
+    if (non_equal_conditions.null_aware_eq_cond_expr != nullptr)
+    {
+        non_equal_conditions.null_aware_eq_cond_expr->finalize(updated_require, true);
+        updated_require = non_equal_conditions.null_aware_eq_cond_expr->getRequiredColumns();
+    }
+    if (non_equal_conditions.other_cond_expr != nullptr)
+    {
+        non_equal_conditions.other_cond_expr->finalize(updated_require, true);
```
SeaRise, line 2644: how about
```suggestion
        non_equal_conditions.other_cond_expr->finalize(updated_require, join_type == nullaware || semi);
```

dbms/src/Interpreters/Join.cpp
```diff
@@ -378,6 +379,15 @@ std::shared_ptr<Join> Join::createRestoreJoin(size_t max_bytes_before_external_j
         flag_mapped_entry_helper_name,
         probe_cache_column_threshold,
         is_test);
+    /// init output names after finalize, the restored join don't need to finalize
+    ret->output_columns_after_finalize = output_columns_after_finalize;
+    ret->output_column_names_set_after_finalize = output_column_names_set_after_finalize;
+    ret->output_columns_names_set_for_other_condition_after_finalize
+        = output_columns_names_set_for_other_condition_after_finalize;
+    ret->required_columns = required_columns;
+    ret->output_block_after_finalize = output_block_after_finalize;
```
SeaRise, line 388: why not using `output_block` directly?

windtalker: `output_block` is already inited in the constructor of `Join`, here we just assign some finalized releated field because the `finalize` method of the newly constructed `Join` is not called.

dbms/src/Interpreters/Join.cpp
windtalker: Done

dbms/src/Interpreters/Join.h
```diff
@@ -407,11 +410,18 @@ class Join
...
 
-    Names tidb_output_column_names;
+    NamesAndTypes output_columns;
+    Block output_block;
+    NamesAndTypes output_columns_after_finalize;
+    Block output_block_after_finalize;
+    NameSet output_column_names_set_after_finalize;
+    NameSet output_columns_names_set_for_other_condition_after_finalize;
+    Names required_columns;
+    bool finalized = false;
```
SeaRise, line 424: Can clickhouse-client execute join sql normally?

SeaRise (approved):
Others LGTM

dbms/src/Interpreters/Join.h
windtalker: No, join is not supported for sql from clickhouse-client. We already disable these sql since v7.0.0(https://github.com/pingcap/tiflash/pull/7064)

SeaRise: ok

gengliqi (approved):

Issue #8398, opened 2023-11-20
TiFlash data inconsistency caused by `ALTER TABLE ... EXCHANGE PARTITION`


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

This bug is introduced by "refactoring the DDL framework" https://github.com/pingcap/tiflash/issues/7630 that use TableID mapping instead of creating TiFlash storage instances all the time.

It's slightly different from https://github.com/pingcap/tiflash/issues/8372. 
The issue in #8372 is the partition table has been synced to tiflash, but the non-partition table is not synced to tiflash, then execute `ALTER TABLE ... EXCHANGE PARTITION` immediately.

This case is
1. The partition table table has been synced to tiflash's table id mapping
2. The non-partition table has been created and synced to tiflash's table id mapping
3. But no tiflash replica is added for partition table and non-partition table at this timepoint
4. Execute `ALTER TABLE ... EXCHANGE PARTITION` on tidb
5. Add tiflash replica for the partition table before tiflash's regular schema sync task is run.

This bug would result in the data on the partition is lost but not tiflash queries failed/crash. But when testing, I found the root cause may result in crashes under some cases too.

test case:
```
# disable schema sync service
>> DBGInvoke __enable_schema_sync_service('false')
>> DBGInvoke __refresh_schemas()
# cleanup
mysql> drop table if exists test.e;
mysql> drop table if exists test.e2;
mysql> drop table if exists test_new.e2;
mysql> drop database if exists test_new;

# case 11, create non-partition table and execute exchagne partition immediately
mysql> create table test.e(id INT NOT NULL,fname VARCHAR(30),lname VARCHAR(30)) PARTITION BY RANGE (id) ( PARTITION p0 VALUES LESS THAN (50),PARTITION p1 VALUES LESS THAN (100),PARTITION p2 VALUES LESS THAN (150), PARTITION p3 VALUES LESS THAN (MAXVALUE));
mysql> insert into test.e values (1, 'a', 'b'),(108, 'a', 'b');
# sync the partition table to tiflash
>> DBGInvoke __refresh_schemas()

mysql> create table test.e2(id int not null,fname varchar(30),lname varchar(30));
# sync the non-partition table to tiflash   <--- required on the master branch
>> DBGInvoke __refresh_schemas()
mysql> insert into test.e2 values (2, 'a', 'b');
[...]

PR #8403, opened 2023-11-21
ddl: Fix broken TableIDMapping lead to data loss or crashes (release-7.5)

JaySon-Huang:
manual cherry-pick of https://github.com/pingcap/tiflash/pull/8402

Issue Number: close https://github.com/pingcap/tiflash/issues/8398

Problem Summary:

The main problem is when we want to create a Storage instance

* `TiDBSchemaSyncer::syncTableSchema` first check whether we can find the TableID mapping in memory
* If we can find the TableID mapping, then `SchemaBuilder::applyTable` will fetch the table info from TiKV and try to create the table

https://github.com/pingcap/tiflash/blob/5ba01c04abc057f376eb71c8598cab65bc711460/dbms/src/TiDB/Schema/TiDBSchemaSyncer.cpp#L174-L215

https://github.com/pingcap/tiflash/blob/5ba01c04abc057f376eb71c8598cab65bc711460/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L1203-L1213

However, even if we can find the TableID from the TableID mapping, we can not ensure its freshness. There could be some DDLs has changed the TableID mapping when we really want to create a Storage instance, this could lead to incorrect TableInfo fetched from TiKV. For example:
* Rename table to a new database
* `ALTER TABLE ... EXCHANGE PARTITION` that will make a partition to be a non-partition table

commit: https://github.com/pingcap/tiflash/pull/8403/commits/76ee7091183730416bdd8b218b19c4da0349e289
- Iet `SchemaBuilder::applyTable` can detect the wrong TableID mapping and return false
  - Get the table info without `MVCC GET`, so that if a table has been renamed to a new database or dropped, we can return false
  - If we can not produce the table info by the logical_table_id, it is possible some partitions DDLs have been executed, return false
  - Only return true if we create the Storage instance by the latest table info or has applied the latest table info to the existing Storage instance
- Iet `TiDBSchemaSyncer:: syncTableSchema` try to sync the table schema to refresh the TableID mapping when need
  - First check the table id mapping and then try to call `SchemaBuilder::applyTable`
  - If something wrong happens in the first step, then try to sync the TiFlash instance schema using the same TiKV snapshot (same `SchemaGetter`). This will ensure the freshness of TableIDMapping
  - Then try to check the table id mapping and call `SchemaBuilder::applyTable` again with the same TiKV snapshot, it should OK for all cases now
- `SchemaBuilder:: applyRenameTable`
  - Will update the TableID -> DatabaseID mapping even if the Storage instance is not exist

commit: https://github.com/pingcap/tiflash/pull/8402/commits/babdf5588dbd7ac2cda9bee211905b859ae21f84
* Make `exchange partition` to be an atomic function `TableIDMap::exchangeTablePartition`
* `TableIDMap::emplaceTableID`/`TableIDMap::emplacePartitionTableID` check whether the old value is the same as new value before logging a warning message

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a data inconsistency issue after `ALTER TABLE ... EXCHANGE PARTITION`
```

### DIFF
--- a/dbms/src/Debug/MockSchemaGetter.h
+++ b/dbms/src/Debug/MockSchemaGetter.h
@@ -34,7 +34,7 @@ struct MockSchemaGetter
 
     static bool checkSchemaDiffExists(Int64 version) { return MockTiDB::instance().checkSchemaDiffExists(version); }
 
-    static TiDB::TableInfoPtr getTableInfo(DatabaseID, TableID table_id)
+    static TiDB::TableInfoPtr getTableInfo(DatabaseID, TableID table_id, [[maybe_unused]] bool try_mvcc = true)
     {
         return MockTiDB::instance().getTableInfoByID(table_id);
     }
--- a/dbms/src/Encryption/tests/gtest_rate_limiter.cpp
+++ b/dbms/src/Encryption/tests/gtest_rate_limiter.cpp
@@ -71,8 +71,10 @@ TEST(WriteLimiterTest, Rate)
         // hint: the range [0.8, 1.25] is copied from rocksdb,
         // if tests fail, try to enlarge this range.
         // enlarge the range to [0.75, 1.30]
-        EXPECT_GE(actual_rate / target, 0.75);
-        EXPECT_LE(actual_rate / target, 1.30);
+        EXPECT_GE(actual_rate / target, 0.75)
+            << fmt::format("actual_rate={} target={} elapsed={:.3f}s", actual_rate, target, elapsed);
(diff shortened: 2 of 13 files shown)


Lloyd-Pottiger (approved):

JaySon-Huang:
/hold
more local testing

hongyunyan (approved):

PR #8333, opened 2023-11-08
support `cast as json`

SeaRise:
### What problem does this PR solve?

Issue Number: close #8307 close #8371

Problem Summary:

### What is changed and how it works?
Because json and string are both DataTypeString in tiflash, cast as json cannot be implemented in `FunctionsTiDBConversion.h`, So here we implement cast as json separately.
support
- CastIntAsJson
- CastRealAsJson
- CastDecimalAsJson
- CastStringAsJson
- CastTimeAsJson
- CastDurationAsJson
- CastJsonAsJson

Introduce simdjson v3.6.0 https://github.com/simdjson/simdjson/releases/tag/v3.6.0 for cast string as json.
Because some functions require tipb::FieldType, pass tipb::FieldType in DAGExpressionAnalyzerHelper.

tidb pr: https://github.com/pingcap/tidb/pull/48550

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8407, opened 2023-11-22
Support Json unquote function

### LINKED ISSUE #8334: Support json_unquote function
## Enhancement
json_unquote only accepts string/json data type. When json type parameter is met, it will be cast as string before being fed into json_unquote like this:
json_unquote(cast(json as string))
TiFlash has already support basic functionality of json_unquote and cast_json_as_string in https://github.com/pingcap/tiflash/pull/6377

Two main work remain todo:

1. Check if [string is valid when quoted](https://dev.mysql.com/doc/refman/8.0/en/json-modification-functions.html#function_json-unquote:~:text=An%20error%20occurs%20if%20the%20value%20starts%20and%20ends%20with%20double%20quotes%20but%20is%20not%20a%20valid%20JSON%20string%20literal.).  
[tidb implementation](https://github.com/pingcap/tidb/blob/0cca59f2fdefc8d97b0870141df86c16edac126e/pkg/expression/builtin_json.go#L243C1-L245C3)
2. cast_json_as_string: already support cast json as unrestricted string.  However, the output string data type can be restricted, like char(20).
[tidb implementation](https://github.com/pingcap/tidb/blob/0cca59f2fdefc8d97b0870141df86c16edac126e/pkg/expression/builtin_cast.go#L1911)

yibin87:

Issue Number: close #8334

Problem Summary:

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


yibin87:
run-integration-test

dbms/src/TiDB/Decode/JsonScanner.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+namespace DB
+{
+bool JsonScanIsSpace(char c)
+{
+    return (static_cast<unsigned char>(c) <= static_cast<unsigned char>(' '))
+        && (c == ' ' || c == '\t' || c == '\r' || c == '\n');
+}
+
+bool checkJsonValid(const char * data, size_t length)
```
SeaRise, line 25: how about using `SIMDJson` directly?

dbms/src/Functions/tests/gtest_cast_json_as_string.cpp
```diff
@@ -27,6 +29,49 @@ namespace DB::tests
...
+            if (auto * function_cast_json_as_string = dynamic_cast<FunctionsCastJsonAsString *>(function_impl);
+                function_cast_json_as_string)
+            {
+                function_cast_json_as_string->setOutputTiDBFieldType(field_type);
+            }
+            else
+            {
+                throw TiFlashTestException(fmt::format("Function {} not found!", func_name));
+            }
+        }
```
SeaRise, line 67: Seems useless because `DAGExpressionAnalyerHelper` will be called when `raw_function_test` is false

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp
```diff
@@ -280,6 +281,102 @@ String DAGExpressionAnalyzerHelper::buildCastFunction(
...
+    }
+    else
+    {
+        throw Exception(fmt::format("Unexpected func {} in buildCastJsonAsStringFunction", func_name));
+    }
+
+    const ExpressionAction & apply_function
+        = ExpressionAction::applyFunction(function_builder, {arg}, result_name, nullptr);
+    actions->add(apply_function);
+    return result_name;
```
SeaRise, line 377: It looks like they can be combined into one function

dbms/src/TiDB/Decode/JsonScanner.cpp
yibin87: Not sure if SIMDJson behaves exactly the same with go-lib.

SeaRise: SIMDJson can take advantage of AVX2 and AVX512 to improve performance

yibin87: Yeah, it's not common case to do "valid check" here after introduce need_check flag. Thus I prefer use go-lib one currently, and I'll add TODOs to use simdJson in future.

SeaRise: ok.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp
yibin87: Do you mean extracting some common code here?

dbms/src/Functions/tests/gtest_cast_json_as_string.cpp
yibin87: Can't get your point here, just introduce this method for test to set tidb field type here

dbms/src/Functions/FunctionsJson.h
```diff
@@ -370,25 +416,60 @@ class FunctionsCastJsonAsString : public IFunction
...
+                    size_t orig_length = element_write_buffer.count();
+                    size_t byte_length = tidb_tp->flen();
+                    byte_length = charLengthToByteLengthFromUTF8(
+                        reinterpret_cast<char *>(container_per_element.data()),
+                        orig_length,
+                        byte_length);
+                    byte_length = std::min(byte_length, orig_length);
+                    if (byte_length < element_write_buffer.count())
+                        context.getDAGContext()->handleTruncateError("Data Too Long");
+                    write_buffer.write(reinterpret_cast<char *>(container_per_element.data()), byte_length);
```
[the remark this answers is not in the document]
yibin87: byte_length is expected to be equal or fewer than orig_length, thus shouldn't be byte_length > element_write_buffer.count() case.  
And it is not common to set char length here, thus use tmp result to make code more readable.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp
SeaRise: I mean that refine the code like
```
String DAGExpressionAnalyzerHelper::buildJsonFunction(
    DAGExpressionAnalyzer * analyzer,
    const tipb::Expr & expr,
    const ExpressionActionsPtr & actions)
{
    if unlikely (!exprHasValidFieldType(expr))
        throw TiFlashException("JsonUnquote function without valid field type", Errors::Coprocessor::BadRequest);

    const auto & input_expr = expr.children(0);
    auto func_name = getFunctionName(expr);

    String arg = analyzer->getActions(input_expr, actions);
    const auto & collator = getCollatorFromExpr(expr);
    String result_name = genFuncString(func_name, {arg}, {collator});
    if (actions->getSampleBlock().has(result_name))
        return result_name;

    const FunctionBuilderPtr & function_builder = FunctionFactory::instance().get(func_name, analyzer->getContext());
    auto * function_build_ptr = function_builder.get();
    if (auto * default_function_builder = dynamic_cast<DefaultFunctionBuilder *>(function_build_ptr);
        default_function_builder)
    {
        auto * function_impl = default_function_builder->getFunctionImpl().get();
        if (auto * function_cast_json_as_string = dynamic_cast<FunctionsCastJsonAsString *>(function_impl);
            function_cast_json_as_string)
        {
            function_cast_json_as_string->setOutputTiDBFieldType(expr.field_type());
        }
        else if (auto * function_json_unquote = dynamic_cast<FunctionsJsonUnquote *>(function_impl); function_json_unquote)
        {
const auto & input_expr = expr.children(0);
            bool valid_check = !(isScalarFunctionExpr(input_expr) && input_expr.sig() == tipb::ScalarFuncSig::CastJsonAsString);
            function_json_unquote->setNeedValidCheck(valid_check);
        }
        else
        {
            throw Exception(fmt::format("Unexpected func {}", func_name));
        }
    }
    else
    {
        throw Exception(fmt::format("Unexpected func {}", func_name));
    }

    const ExpressionAction & apply_function
        = ExpressionAction::applyFunction(function_builder, {arg}, result_name, nullptr);
    actions->add(apply_function);
    return result_name;
}
```

SeaRise: How about merging with https://github.com/pingcap/tiflash/pull/8333/files#diff-e78d988899b5d199db66439fd853d0c86526e77cbffcb20c2c9d10500bdb51eaR284

dbms/src/Functions/FunctionsJson.h
```diff
@@ -329,21 +328,68 @@ class FunctionsJsonUnquote : public IFunction
...
+        size_t current_offset = 0;
+        for (size_t i = 0; i < block.rows(); ++i)
+        {
+            size_t next_offset = offsets_from[i];
+            size_t data_length = next_offset - current_offset - 1;
+            if constexpr (validCheck)
+            {
+                // TODO(hyb): use SIMDJson to check when SIMDJson is proved in practice
+                if (data_length >= 2 && data_from[current_offset] == '"' && data_from[next_offset - 2] == '"'
+                    && !checkJsonValid(reinterpret_cast<const char *>(&data_from[current_offset]), data_length))
```
SeaRise, line 349:
```suggestion
                    && unlikely(!checkJsonValid(reinterpret_cast<const char *>(&data_from[current_offset]), data_length)))
```

dbms/src/Functions/tests/gtest_json_array.cpp
```diff
@@ -33,7 +33,10 @@ class TestJsonArray : public DB::tests::FunctionTest
         static auto json_array_return_type = std::make_shared<DataTypeString>();
         assert(json_array_return_type->equals(*json_column.type));
         // The `json_binary` should be cast as a string to improve readability.
-        return executeFunction("cast_json_as_string", {json_column});
+        tipb::FieldType field_type;
+        field_type.set_flen(-1);
+        field_type.set_tp(TiDB::TypeString);
+        return executeCastJsonAsStringFunction({json_column}, field_type);
```
windtalker, line 39: Why not construct `filed_type` inside `executeCastJsonAsStringFunction`?

dbms/src/Functions/FunctionsJson.h
```diff
@@ -386,25 +432,59 @@ class FunctionCastJsonAsString : public IFunction
...
+                        vec_null_map[i] = 1;
+                    else
+                    {
+                        JsonBinary json_binary(
+                            data_from[current_offset],
+                            StringRef(&data_from[current_offset + 1], json_length - 1));
+                        json_binary.toStringInBuffer(element_write_buffer);
+                    }
+
+                    size_t orig_length = element_write_buffer.count();
```
windtalker, line 475: L475-L483 should be inside the above else branch?

dbms/src/Functions/FunctionsJson.h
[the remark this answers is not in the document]
windtalker: But theoretical speaking, we still need to handle the case of `byte_length > element_write_buffer.count()`? Maybe we should throw Exception in `charLengthToByteLengthFromUTF8` if `ret >= length`?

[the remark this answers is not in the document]
yibin87: byte_length = std::min(byte_length, orig_length); is executed after charLengthToByteLengthFromUTF8, thus byte_length <= orig_length. Not sure if this answer your question.

dbms/src/Functions/tests/gtest_json_array.cpp
yibin87: FieldType has several attributes other than flen, like collations, thus pass it as a single parameter here.

dbms/src/Functions/FunctionsJson.h
yibin87: Yeah, it can reduce useless code for null case. I'll move it.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzerHelper.cpp
```diff
@@ -281,27 +281,31 @@ String DAGExpressionAnalyzerHelper::buildCastFunction(
...
     DAGExpressionAnalyzer * analyzer,
     const tipb::Expr & expr,
     const ExpressionActionsPtr & actions)
 {
-    if unlikely (expr.children_size() != 1)
-        throw TiFlashException("Cast function only support one argument", Errors::Coprocessor::BadRequest);
-    if unlikely (!exprHasValidFieldType(expr))
-        throw TiFlashException("CAST function without valid field type", Errors::Coprocessor::BadRequest);
-
     const auto & input_expr = expr.children(0);
```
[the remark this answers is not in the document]
yibin87: Good catch, will move it~

dbms/src/Functions/FunctionsJson.h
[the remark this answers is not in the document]
windtalker: But `charLengthToByteLengthFromUTF8` can not guarantee this if it is not a valid utf8 string, so I suggest to throw an exception in `charLengthToByteLengthFromUTF8` if `ret >= length`

dbms/src/Functions/tests/gtest_json_array.cpp
windtalker: But for json type, the collation and flen should always be binary and -1?

windtalker: OK.

dbms/src/Functions/FunctionsJson.h
```diff
@@ -386,25 +432,59 @@ class FunctionCastJsonAsString : public IFunction
...
+                        JsonBinary json_binary(
+                            data_from[current_offset],
+                            StringRef(&data_from[current_offset + 1], json_length - 1));
+                        json_binary.toStringInBuffer(element_write_buffer);
+                        size_t orig_length = element_write_buffer.count();
+                        auto byte_length = charLengthToByteLengthFromUTF8(
+                            reinterpret_cast<char *>(container_per_element.data()),
+                            orig_length,
+                            tidb_tp->flen());
+                        byte_length = std::min(byte_length, orig_length);
```
windtalker, line 478: Looks like this is not necessary since `charLengthToByteLengthFromUTF8` should ensure that the return value is less than `orig_length`?

dbms/src/Functions/FunctionsJson.h
```diff
@@ -386,25 +432,59 @@ class FunctionCastJsonAsString : public IFunction
...
+                        if (byte_length < element_write_buffer.count())
+                            context.getDAGContext()->handleTruncateError("Data Too Long");
+                        write_buffer.write(reinterpret_cast<char *>(container_per_element.data()), byte_length);
```
SeaRise, line 481: how about
```suggestion
                        auto start_pos = write_buffer.offset();
                        JsonBinary json_binary(
                            data_from[current_offset],
                            StringRef(&data_from[current_offset + 1], json_length - 1));
                        json_binary.toStringInBuffer(write_buffer);
                        auto end_pos = write_buffer.offset();
                        auto orig_length = end_pos - start_pos;
                        auto byte_length = charLengthToByteLengthFromUTF8(
                            reinterpret_cast<char *>(write_buffer.data() + start_offset),
                            orig_length,
                            tidb_tp->flen());
                        if (byte_length < orig_length)
                        {
                            context.getDAGContext()->handleTruncateError("Data Too Long");
                            write_buffer.setOffset(start_pos + byte_length);
                        }
```
?

SeaRise: To avoid one more memcpy.

yibin87: Yeah，I just think this code path is not common used(because cast json as fixed length char is a little strange), and even if it is used the performance won't drop significantly, thus choose to use the temporary buffer here to make code more easier.

windtalker (approved):
LGTM

Issue #8395, opened 2023-11-20
Some tables may not be physically dropped on TiFlash


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Apply the following three DDLs quickly
  1. Create a table on tidb
  2. Add tiflash for the table
  3. drop the table

There is a chance that the `drop table` is executed on TiFlash before the `IStorage` instance is actually created by `applying snapshot`.
Then the information that the table has been dropped is missed on the TiFlash instance and its related data can not be physically dropped.

This can be workaround by restarting TiFlash, which will re-sync all table info from TiDB. And it will mark the table as tombstones at that time.

logging from [testing pipeline](https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/15116/pipeline/) :  [table_id_571.log](https://github.com/pingcap/tiflash/files/13407289/table_id_571.log)

### 4. What is your TiFlash version? (Required)
https://github.com/pingcap/tiflash/tree/5ba01c04abc057f376eb71c8598cab65bc711460

Issue #3777, opened 2021-12-30
TiFlash may crash under very frequent truncate table, change column type and insert data


lidezhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
It may occurs with a small probability under the following scenario
1. change column type to a wider column type
2. insert some data into the new column that exceeds the range of the old column type
3. doing many truncate table
The reason may be that if a table is truncated, then it become a tombstone table in tiflash. And tiflash has two schema sync mode: incremental sync and full sync. If the schema version changed a lot, then it will trigger a full sync. But the tombstone table schema change is not showed in the full sync list, so it's schema change cannot be seen. And tiflash cannot use the old schema to decode the new data.

### 2. What did you expect to see? (Required)
tiflash works normally
### 3. What did you see instead (Required)
tiflash crash

### 4. What is your TiFlash version? (Required)
the problem may influence all version

flowbehappy:
Lower the severity since it is a very rare case. Only seen in testing env. And it is more like a design flaw.

Issue #1664, opened 2021-03-26
Cannot handle a rare case for flash back table


zanmato1984:
Current tombstone + flash back strategy cannot handle the following rare case:
- Alter table by widening a column
- Insert a row with the widened column's value to be out of the range of the type before widening
- Drop table
- TiFlash does a full schema sync and directly tombstones the table (with the alter diff missed)
- TiFlash flushes the new inserted row into storage

In such a case, TiFlash won't be able to decode this row and errors out "overflow detected".

But as you can see, this is a extremely rare case.

PR #8421, opened 2023-11-26
ddl: Refactor for fixing unstable drop table

JaySon-Huang:
### What problem does this PR solve?

Issue Number: ref https://github.com/pingcap/tiflash/issues/8395

Problem Summary:

### What is changed and how it works?

* Abstract class `DatabaseInfoCache` to wrap the logic about the database info cache in `TiDBSchemaSyncer` and `SchemaBuilder`
* Replace `TiDBSchemaSyncer::findDatabaseIDAndTableID` by `TableIDMap:: findDatabaseIDAndLogicalTableID`
* Simplify getting canonical name in `SchemaSyncService` to remove useless code
* Remove useless code `SchemaBuilder-internal.h`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8422, opened 2023-11-26
ddl: Fix unstable `DROP TABLE`/`FLASHBACK TABLE`/`RECOVER TABLE`

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8395, close https://github.com/pingcap/tiflash/issues/1664, close https://github.com/pingcap/tiflash/issues/3777

Problem Summary:

Problem 1:
There could be a chance that raft snapshot or raft command of a table comes after it has been dropped.
Because tiflash will ignore the previous `DROP TABLE`, because storage instance is not created yet. The storage instance is created as "non tombstone" after that when raft snapshot or raft command comes. And the storage instance will never be physically dropped until tiflash restarts. 

Problem 2:
In the second time when `syncTableSchema` calls `trySyncTableSchema` after table-id-mapping is up-to-date, it should use `mvcc get` to ensure it can use the table schema of tombstone table to create Storage instance or decode new raft logs, or some new columns data will not be decoded.
Previously `updateTiFlashReplica` will update the table info with the latest columns by accident. So tiflash pass the related tests sometimes.

```sql
CREATE TABLE t (a int, b int)
INSERT INTO t VALUES (1, 1);
ALTER TABLE t SET TIFLASH REPLICA 1;
# wait tiflash replica ready

ALTER TABLE t ADD COLUMN c int;
INSERT INTO t VALUES (1,2,3);
DROP TABLE t;

# the rows (1,2,3) comes after `drop table t` is executed in tiflash
RECOVER TABLE t; -- or FLASHBACK TABLE t

SELECT * FROM t;
# should return rows (1, 1, NULL), (1,2,3);
```

This is a PR following https://github.com/pingcap/tiflash/pull/8421
The logic changes is these commits: https://github.com/pingcap/tiflash/pull/8422/files/d88f6b6f4ae4c8026b969cd9c5ae50924b179529..1fe991997055f755aea89cd2e2bdb8ab26a848bf

* In the first time when `syncTableSchema` calls `trySyncTableSchema`, it will not use `mvcc get` so that we can detect whether we need to update the table-id-mapping
* In the second time when `syncTableSchema` calls `trySyncTableSchema` after table-id-mapping is up-to-date, it will use `mvcc get` to ensure it can use the table schema of tombstone table to create Storage instance or decode new raft logs.
* `SchemaGetter::getTableInfoImpl` does not check the existence of `db_key`, so that we can still get the table info after database is dropped. (get ready for `FLASHBACK DATABASE ... TO ...`)
* If a storage instance is created with the table info got by `client-c mvcc get`, then we will create the storage instance with a tombstone timestamp. So that it can be physically dropped after GC time.
* `applySetTiFlashReplica` should only update the tiflash replica info instead of replacing all the table info, or it will lead to some later DDLs (changing partitions, etc) is not executed
* Simplify `recover table` codes using early return

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a potential issue that some data can not be recovered by `RECOVER TABLE` and `FLASHBACK TABLE`
```


CalvinNeo:
Can this pr fix https://github.com/pingcap/tiflash/issues/1664??

CalvinNeo (commented):
Can this pr fix https://github.com/pingcap/tiflash/issues/1664 ? 
Seems the problem1 is https://github.com/pingcap/tiflash/issues/1664#issuecomment-808631344, and the problem2 is https://github.com/pingcap/tiflash/issues/1664#issue-842031800?

If so, please also close 1664.

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -86,16 +91,28 @@ void SchemaBuilder<Getter, NameMapper>::applyCreateTable(DatabaseID database_id,
     }
 
     // If table is partition table, we will create the logical table here.
-    // Because we get the table_info, so we can ensure new_db_info will not be nullptr.
     auto new_db_info = getter.getDatabase(database_id);
-    applyCreateStorageInstance(new_db_info, table_info);
+    if (new_db_info == nullptr)
```
Lloyd-Pottiger, line 95: Why is it called `new_db_info`?

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -86,16 +91,28 @@ void SchemaBuilder<Getter, NameMapper>::applyCreateTable(DatabaseID database_id,
...
+        // the database has been dropped
+        LOG_INFO(
+            log,
+            "database is not exist in TiKV, may have been dropped, applyCreateTable is ignored, database_id={} "
+            "table_id={}",
+            database_id,
+            table_id);
+        return;
+    }
+    applyCreateStorageInstance(new_db_info, table_info, get_by_mvcc);
```
Lloyd-Pottiger, line 106: add some comments to explain why creating logical table need to create storage instance.

JaySon-Huang:
> Can this pr fix #1664 ? Seems the problem1 is [#1664 (comment)](https://github.com/pingcap/tiflash/issues/1664#issuecomment-808631344), and the problem2 is [#1664 (comment)](https://github.com/pingcap/tiflash/issues/1664#issue-842031800)?
> 
> If so, please also close 1664.

Yes, it can. Thanks for reminding

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -373,68 +391,100 @@ void SchemaBuilder<Getter, NameMapper>::applySetTiFlashReplica(DatabaseID databa
...
-            }
-            else
-            {
-                applyRecoverTable(db_info->id, table_id);
-            }
+            applyCreateTable(db_info->id, table_id);
         }
-        else
+        return;
+    }
```
Lloyd-Pottiger, line 407:
```
1. create table t (I int);
2. alter table t partition by hash(id) partitions 8; <-- normal table to partition table
3. alter tiflash replica t set tiflash replica 1;
4. insert into t values (10001), (10002), ....; <-- insert data
```

What if the ordering come to tiflash is `3->1->2->4`?

JaySon-Huang: Does `3->1` mean the "Region snapshot" comes before the DDL `create table t` is executed in TiFlash?

Lloyd-Pottiger: yeah

1. tikv create normal table
2. tikv change the table to partition table
3. tikv send Region snapshot to tiflash
4. tiflash get `create table` DDL.

JaySon-Huang: When TiFlash try do decode the Region snapshot/raft command, it calls `AtomicGetStorageSchema`. Inside that function, if the storage instance is not yet created, it calls `syncTableSchema`. If `syncTableSchema` does not known the table id, it sync the DDL with tikv and apply `create table` in TiFlash.

https://github.com/pingcap/tiflash/blob/cf0e9affe7191b428a25142a91f240bce42823a1/dbms/src/Storages/KVStore/Decode/PartitionStreams.cpp#L465-L488

dbms/src/TiDB/Schema/SchemaBuilder.cpp
JaySon-Huang: not important, name is changed

dbms/src/TiDB/Schema/SchemaBuilder.cpp
JaySon-Huang: added

Lloyd-Pottiger (approved):
LGTM

tests/fullstack-test2/ddl/alter_drop_table.test
```diff
@@ -44,66 +59,29 @@ mysql> set session tidb_isolation_read_engines='tiflash';select * from test.t3;
...
-
 mysql> drop table test.t3;
-
 => DBGInvoke __refresh_schemas()
-
 >> select tidb_database,tidb_name from system.tables where tidb_database = 'test' and tidb_name = 't3' and is_tombstone = 0
 
-
-## drop table arrive tiflash before ddl and insert, and do recover, check the data is not lost
-## because we want to test we actually drop the table, so please not use the same name for this table
```
JaySon-Huang, line 56: This case is moved to `fullstack-test2/ddl/flashback`

hongyunyan (approved):
LGTM

PR #8458, opened 2023-12-04
FAP: Support cancel FAP tasks

### LINKED ISSUE #8382: Introduce two stages FAP

CalvinNeo:

Issue Number: close #8382

Problem Summary:

1. Support cancel FAP tasks
1. Fix empty segments
1. Fix the problem when restart with another regular snapshot after FAP has totally failed
1. Fix maybe throw when build fap snapshot
1. Fix test framework
1. Fix snapshot in proxy when there is not CachedRegionInfo

**RELATED PROXY PR**: https://github.com/pingcap/tidb-engine-ext/pull/363

FAP phase 1 can be canceled in:
1. `FastAddPeer` when timeout
    It this happens, `FastAddPeer` will block wait `FastAddPeerImpl` to finish all cleaning work, and then return to Proxy. Otherwise, a later legacy snapshot may meet a cleaning FAP phase1 task.

It can't be canceled elsewhere in TiFlash, since once the task is canceled, it is not accessible from `AsyncTasks`. Then a `FastAddPeer` will create a new task for this.

It can't be canceled from Proxy, since it involves many modifications in Proxy.

It can't be canceled by a legacy snapshot, since the only way a legacy snapshot is trigger is that the FAP is finished or fallback so it can continue to receive MsgSnapshot.

It can't be canceled by a destroy peer event, since we don't handle MsgAppend in FAP phase 1 either. If there are some corner cases, then it can:
1. Better be handled in proxy's side, in order to stop the FAP process.
2. Better not be handled in FastAddPeer, before polling FAP status, check if the region is still exists. Because the region is likely not exist at the FAP phase 1 stage.

FAP phase 1 result can be cleaned in:
1.  `FastAddPeerImpl` when received cancel signal from `FastAddPeer`
3. `handleDestroy`
4. `prehandleSnapshot`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/KVStore/MultiRaft/ApplySnapshot.cpp
```diff
@@ -121,17 +122,19 @@ void KVStore::checkAndApplyPreHandledSnapshot(const RegionPtrWrap & new_region,
...
         {
-            // TODO(fap): Better cancel fap process in first, however, there is no case currently where a legacy snapshot runs with fap phase1/phase2 in parallel.
-            // The only case is a fap failed after phase 1 and fallback and failed to clean its phase 1 result.
-            fap_ctx->cleanCheckpointIngestInfo(tmt, new_region->id());
-        }
-        // Another FAP will not take place if this stage is not finished.
-        if (fap_ctx->tasks_trace->discardTask(new_region->id()))
-        {
-            LOG_ERROR(log, "FastAddPeer: find old fap task, region_id={}", new_region->id());
+            // Legacy snapshot and FAP(both phase 1 and 2) for a region is exclusive for now.
```
CalvinNeo, line 129: When applying legacy snapshot, cancel all ongoing snapshot

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -185,9 +191,14 @@ std::variant<CheckpointRegionInfoAndData, FastAddPeerRes> FastAddPeerImplSelect(
...
     std::unordered_map<StoreID, UInt64> checked_seq_map;
     auto fap_ctx = tmt.getContext().getSharedContextDisagg()->fap_context;
+    auto cancel_handle = fap_ctx->tasks_trace->getCancelHandleFromExecutor(region_id);
     const auto & settings = tmt.getContext().getSettingsRef();
-    auto current_store_id = tmt.getKVStore()->getStoreMeta().id();
-    auto candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
+    auto current_store_id = tmt.getKVStore()->clonedStoreMeta().id();
+    std::vector<StoreID> candidate_store_ids;
+    fiu_do_on(FailPoints::force_set_fap_candidate_store_id, { candidate_store_ids = {1234}; });
+    if (candidate_store_ids.empty())
```
CalvinNeo, line 199: For failpoints

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -244,13 +260,20 @@ std::variant<CheckpointRegionInfoAndData, FastAddPeerRes> FastAddPeerImplSelect(
         {
             if (watch.elapsedSeconds() >= settings.fap_wait_checkpoint_timeout_seconds)
             {
-                // TODO(fap) Cancel and remove from AsyncTasks
                 // This could happen if there are too many pending tasks in queue,
                 LOG_INFO(log, "FastAddPeer timeout region_id={} new_peer_id={}", region_id, new_peer_id);
                 GET_METRIC(tiflash_fap_task_result, type_failed_timeout).Increment();
+                // We don't wait here.
```
CalvinNeo, line 266: Because the task will be canceled while cleaned

dbms/src/Storages/KVStore/KVStore.cpp
```diff
@@ -333,6 +333,12 @@ void KVStore::handleDestroy(UInt64 region_id, TMTContext & tmt)
 void KVStore::handleDestroy(UInt64 region_id, TMTContext & tmt, const KVStoreTaskLock & task_lock)
 {
     const auto region = getRegion(region_id);
+    if (tmt.getContext().getSharedContextDisagg()->isDisaggregatedStorageMode())
```
CalvinNeo, line 336: Ensure the FAP tasks being cleaned before apply legacy snapshot

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -294,7 +325,14 @@ FastAddPeerRes FastAddPeerImplWrite(
         storage->isCommonHandle(),
         storage->getRowKeyColumnSize());
 
+    if (cancel_handle->canceled())
+    {
+        LOG_INFO(log, "Cancel FAP before write, region_id={}", region_id);
+        fap_ctx->forciblyCleanTask(tmt, region_id);
```
CalvinNeo, line 331: Should we also clean written segments and raft log?

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
```diff
@@ -209,4 +209,37 @@ void FastAddPeerContext::insertCheckpointIngestInfo(
...
+        LOG_INFO(log, "FastAddPeer: find old finished fap task, region_id={}", region_id);
+        forciblyCleanTask(tmt, region_id);
+    }
+    else if likely (prev_state == FAPAsyncTasks::TaskState::NotScheduled)
+    {
+        // 1. There leaves some non-ingested data on disk after restart.
+        // 2. There has been no fap at all.
+        // 3. FAP is enabled before, but disabled for now.
+        LOG_DEBUG(log, "FastAddPeer: no find ongoing fap task, region_id={}", region_id);
+        forciblyCleanTask(tmt, region_id);
```
CalvinNeo, line 231: Should we also clean written segments and raft log?

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
```diff
@@ -209,4 +209,37 @@ void FastAddPeerContext::insertCheckpointIngestInfo(
...
+    }
+    else
+    {
+        // Currently, proxy will not actively cancel FAP. So it will not fallback if FAP phase 1 is still running.
+        // If we cancel the task asyncly, this will happen in some very corner cases where a legacy snapshot meets a cleaning FAP task.
+        throw Exception(
```
CalvinNeo, line 237: I still prefer throw here then `RUNTIME_CHECK_MSG(xxx, "yyy")`, because it makes the comment more clear

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -26,97 +30,278 @@ struct AsyncTasks
...
+    uint64_t queryStartTime(Key key, bool throw_if_not_found = true)
     {
         std::scoped_lock<std::mutex> l(mtx);
-        auto it2 = start_time.find(key);
-        RUNTIME_CHECK_MSG(it2 != start_time.end(), "queryElapsed meets empty key");
-        return it2->second;
+        auto it = tasks.find(key);
+        if unlikely (it == tasks.end())
+        {
+            if (throw_if_not_found)
```
CalvinNeo, line 267: I prefer throw here because we don't RUNTIME_CHECK throw_if_not_found but `unlikely (it == tasks.end())`. However, we first check unlikely (it == tasks.end())` because it is more unlikely.

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -125,9 +310,20 @@ struct AsyncTasks
...
+        return it->second.cancel;
+    }
+
+
+protected:
+    std::unordered_map<Key, Elem> tasks;
+    // TODO(fap) Use threadpool which supports purging from queue.
     std::unique_ptr<ThreadPool> thread_pool;
     mutable std::mutex mtx;
+    LoggerPtr log;
```
CalvinNeo, line 327: It is not used currently, however, leave it here, it could be useful when debugging some cases.

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -185,26 +191,44 @@ std::variant<CheckpointRegionInfoAndData, FastAddPeerRes> FastAddPeerImplSelect(
...
+
+    // Get candidate stores.
     const auto & settings = tmt.getContext().getSettingsRef();
-    auto current_store_id = tmt.getKVStore()->getStoreMeta().id();
-    auto candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
+    auto current_store_id = tmt.getKVStore()->clonedStoreMeta().id();
+    std::vector<StoreID> candidate_store_ids;
+    fiu_do_on(FailPoints::force_set_fap_candidate_store_id, { candidate_store_ids = {1234}; });
+    if (candidate_store_ids.empty())
+        candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
```
Lloyd-Pottiger, line 208:
```suggestion
    std::vector<StoreID> candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
    fiu_do_on(FailPoints::force_set_fap_candidate_store_id, { candidate_store_ids = {1234}; });
```

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -185,26 +191,44 @@ std::variant<CheckpointRegionInfoAndData, FastAddPeerRes> FastAddPeerImplSelect(
...
     {
         // Check all candidate stores in this loop.
         for (const auto store_id : candidate_store_ids)
         {
-            RUNTIME_CHECK(store_id != current_store_id);
+            RUNTIME_CHECK_MSG(
+                store_id != current_store_id,
+                "store_id {} != current_store_id {}",
+                store_id,
+                current_store_id);
```
Lloyd-Pottiger, line 227:
```suggestion
            RUNTIME_CHECK(
                store_id != current_store_id,
                store_id,
                current_store_id);
```

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
CalvinNeo: `getCandidateStoreIDsForRegion` could throw, this failpoint is to avoid this throw while provide a suitable value for `candidate_store_ids `

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -383,44 +435,56 @@ FastAddPeerRes FastAddPeerImpl(
     }
 }
 
-void ApplyFapSnapshotImpl(TMTContext & tmt, TiFlashRaftProxyHelper * proxy_helper, UInt64 region_id, UInt64 peer_id)
+uint8_t ApplyFapSnapshotImpl(TMTContext & tmt, TiFlashRaftProxyHelper * proxy_helper, UInt64 region_id, UInt64 peer_id)
```
Lloyd-Pottiger, line 438: why the return type is `uint8_t` rather than `bool`?

dbms/src/Storages/KVStore/tests/gtest_async_tasks.cpp
```diff
@@ -0,0 +1,216 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <Storages/KVStore/Utils/AsyncTasks.h>
+#include <common/logger_useful.h>
+
+#include "region_kvstore_test.h"
```
Lloyd-Pottiger, line 18: use full path

Lloyd-Pottiger (approved):
Others LGTM

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
CalvinNeo: In some previous Rust version, there is nothing which is designed to carry booleans through FFI

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -193,7 +209,32 @@ bool CheckpointIngestInfo::forciblyClean(TMTContext & tmt, UInt64 region_id, boo
...
+    UInt64 region_id)
+{
+    auto uni_ps = tmt.getContext().getWriteNodePageStorage();
+    auto snapshot = uni_ps->getSnapshot(fmt::format("read_fap_i_{}", region_id));
+    auto page_id
+        = UniversalPageIdFormat::toLocalKVPrefix(UniversalPageIdFormat::LocalKVKeyType::FAPIngestInfo, region_id);
+    // For most cases, ingest infos are deleted in `removeFromLocal`.
+    Page page = uni_ps->read(page_id, nullptr, snapshot, /*throw_on_not_exist*/ false);
+    auto maybe_checkpoint_info = CheckpointIngestInfo::restore(tmt, proxy_helper, region_id, 0);
+    if (maybe_checkpoint_info.has_value())
```
JaySon-Huang, line 230:
```suggestion
    // For most cases, ingest infos are deleted in `removeFromLocal`.
    auto maybe_checkpoint_info = CheckpointIngestInfo::restore(tmt, proxy_helper, region_id, 0);
    if (maybe_checkpoint_info.has_value())
```

Are these lines useless?

CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -176,12 +200,36 @@ void CheckpointIngestInfo::removeFromLocal(TMTContext & tmt, UInt64 region_id)
     uni_ps->write(std::move(del_batch), PageType::Local);
 }
 
-// Like removeFromLocal, but is static and with check.
-bool CheckpointIngestInfo::forciblyClean(TMTContext & tmt, UInt64 region_id, bool pre_check)
+void CheckpointIngestInfo::deleteWrittenData(TMTContext & tmt, RegionPtr region, const DM::Segments & segments)
+{
+    // auto & storages = tmt.getStorages();
```
CalvinNeo, line 205: This part will be reconsidered in another PR

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
```diff
@@ -110,7 +110,7 @@ std::pair<UInt64, ParsedCheckpointDataHolderPtr> FastAddPeerContext::getNewerChe
     return std::make_pair(cache_seq, checkpoint_data);
 }
 
-CheckpointIngestInfoPtr FastAddPeerContext::getOrRestoreCheckpointIngestInfo(
+std::optional<CheckpointIngestInfoPtr> FastAddPeerContext::getOrRestoreCheckpointIngestInfo(
```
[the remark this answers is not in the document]
CalvinNeo: Will it cost too much if I use `std::optional<CheckpointIngestInfoPtr>`? Because for safety issue, I prefer this one, since dereferencing a nullptr result in a panic which is hard to diagnose.

dbms/src/Storages/KVStore/tests/kvstore_helper.h
```diff
@@ -97,21 +97,27 @@ class KVStoreTestBase : public ::testing::Test
...
+    void resetKVStoreStorage()
     {
-        kvstore.reset();
         auto & global_ctx = TiFlashTestEnv::getGlobalContext();
         global_ctx.tryReleaseWriteNodePageStorageForTest();
         global_ctx.initializeWriteNodePageStorageIfNeed(*path_pool);
+    }
+    KVStore & reloadKVSFromDisk(bool with_reset = true)
+    {
+        auto & global_ctx = TiFlashTestEnv::getGlobalContext();
```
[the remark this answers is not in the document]
CalvinNeo: Yes, it is in `KVStore/tests` and it is in `namespace tests`

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -26,97 +30,281 @@ struct AsyncTasks
     // We use a big queue to cache, to reduce add task failures.
     explicit AsyncTasks(uint64_t pool_size, uint64_t free_pool_size, uint64_t queue_size)
```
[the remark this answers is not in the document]
CalvinNeo: Why?

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -26,97 +30,281 @@ struct AsyncTasks
...
+        void doCancel()
+        {
+            // Use lock here to prevent losing signal.
+            std::scoped_lock<std::mutex> lock(mut);
+            inner->store(true);
+            cv.notify_all();
+        }
+
+        friend struct AsyncTasks;
+        std::shared_ptr<std::atomic_bool> inner = std::make_shared<std::atomic_bool>(false);
```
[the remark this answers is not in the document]
CalvinNeo: I will check again if it is still necessary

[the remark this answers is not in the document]
CalvinNeo: It is not necessary, and been fixed

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -26,97 +30,281 @@ struct AsyncTasks
     // We use a big queue to cache, to reduce add task failures.
     explicit AsyncTasks(uint64_t pool_size, uint64_t free_pool_size, uint64_t queue_size)
         : thread_pool(std::make_unique<ThreadPool>(pool_size, free_pool_size, queue_size))
-    {}
+    {
+        log = DB::Logger::get("AsyncTasks");
```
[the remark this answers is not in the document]
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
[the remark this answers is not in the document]
CalvinNeo: > there is no difference with dereferenceing a `nullptr`

It actually does with common usage. Given the below example, if we call `.value` on `std::nullopt`, it will **throw** an `std::bad_optional_access` which [is not UB](https://en.cppreference.com/w/cpp/utility/optional/value). However, it we call `operator->` on `nullptr`, it will panic. In TiFlash, the `BaseDaemon` handles exceptions better than segment faults.
```
#include <optional>
#include <cstring>
#include <cstdio>

void f() {
    try {
        std::optional<int> x = std::nullopt;
        printf("%d\n", x.value());
    }
    catch (std::bad_optional_access & e) {
        printf("it throws in f\n");
    }
}

void g() {
    try {
        int * x = nullptr;
        printf("%d\n", *x);
    } catch (...){
        printf("it throws in g\n");
    }
}

int main() {
    f();
    g();
}
```

> I think `shared_ptr`/`unique_ptr` and `optional` are redundant

I think it depends. In rust, it is very common to use `Option<NotNull<T>>` to [ensure safety](https://stackoverflow.com/questions/54195517/should-we-use-option-or-ptrnull-to-represent-a-null-pointer-in-rust). In C++, there is no such thing like `NotNull`, however, I think it is also proper to use `std::option<std::shared_ptr<T>>` in API methods which explicitly claim the pointer could be null

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
CalvinNeo: The PR is https://github.com/pingcap/tiflash/pull/8498

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -193,10 +242,20 @@ bool CheckpointIngestInfo::forciblyClean(TMTContext & tmt, UInt64 region_id, boo
...
     }
     return false;
 }
 
+bool CheckpointIngestInfo::forciblyClean(
+    TMTContext & tmt,
+    const TiFlashRaftProxyHelper * proxy_helper,
+    UInt64 region_id)
+{
+    UNUSED(proxy_helper);
```
CalvinNeo, line 256: This part will be reconsidered in another PR. The commented lines are to be removed.

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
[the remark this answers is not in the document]
CalvinNeo: OK, I agree with you now. So how about we introduce `not_null` and some optional wrapper that totally prevents UB behaviors for these cases?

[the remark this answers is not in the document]
JaySon-Huang: > In TiFlash, the BaseDaemon handles exceptions better than segment faults.

I don't agree with this. The BaseDaemon only handles `DB::Exception` better than segment faults. But it handles the exceptions from C++ std lib badly. We can get the stack backtrace when `DB::Exception` is thrown or a null pointer is being accessed, but no stack backtrace when std exceptions are thrown.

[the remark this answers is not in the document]
CalvinNeo: it is changed back to std::shared_ptr

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -405,21 +405,46 @@ Segment::SegmentMetaInfos Segment::readAllSegmentsMetaInfoInRange( //
...
+    bool first_of_the_range = true;
     while (current_segment_id != 0)
     {
         Segment::SegmentMetaInfo segment_info;
         auto target_id = UniversalPageIdFormat::toFullPageId(
             UniversalPageIdFormat::toFullPrefix(context.keyspace_id, StorageType::Meta, context.physical_table_id),
             current_segment_id);
-        auto page = checkpoint_info->temp_ps->read(target_id);
+        auto page = checkpoint_info->temp_ps->read(target_id, nullptr, {}, false);
+        if unlikely (!page.isValid())
```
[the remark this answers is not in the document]
CalvinNeo: Do you mean the page exists, but is not visible to us because of it is filtered by snapshot version?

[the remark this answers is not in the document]
CalvinNeo: Yes, you are right. This optimization is redundant and unsafe. Changed it back

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -47,12 +49,7 @@ CheckpointIngestInfoPtr CheckpointIngestInfo::restore(
     Page page = uni_ps->read(page_id, nullptr, snapshot, /*throw_on_not_exist*/ false);
     if (!page.isValid())
     {
-        throw Exception(
-            ErrorCodes::LOGICAL_ERROR,
-            "Failed to restore CheckpointIngestInfo, region_id={} peer_id={} store_id={}",
-            region_id,
-            peer_id,
-            tmt.getKVStore()->getStoreID(std::memory_order_relaxed));
+        return nullptr;
```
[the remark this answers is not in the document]
CalvinNeo: addressed

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -14,109 +14,303 @@
...
+        {
+            result_dropper();
+        }
+        else
+        {
+            auto cancel_handle = getCancelHandleFromCaller(k);
+            cancel_handle->doCancel();
+            // Cancel logic should do clean itself
+        }
+        {
```
CalvinNeo, line 166: The tasks can still finished in the middle, leaving result not dropped

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
```diff
@@ -14,109 +14,303 @@
...
+    // 2. The cancel_handle is already set.
+    // NOTE: Consider a one producer thread one consumer thread scene, the first task is running,
+    // and the second task is in queue. If we block cancel the second task here, deadlock will happen.
+    // So we only allow block canceling running tasks, and users have to guaruantee all infinite loop having cancel checking.
+    [[nodiscard]] BlockCancelResult blockedCancelRunningTask(Key k)
+    {
+        auto cancel_handle = getCancelHandleFromCaller(k);
+        auto state = queryState(k);
+        RUNTIME_CHECK_MSG(state != TaskState::NotScheduled, "Can't block wait a non-scheduled task");
+        if (state == TaskState::InQueue)
```
CalvinNeo, line 197: If we allow cancel when start running, we can just discard here

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
```diff
@@ -209,4 +222,63 @@ void FastAddPeerContext::insertCheckpointIngestInfo(
...
+                ErrorCodes::LOGICAL_ERROR,
+                "FastAddPeer: find running scheduled fap task, region_id={} fap_state={}",
+                region_id,
+                magic_enum::enum_name(prev_state));
+        }
+        else
+        {
+            // If the region is sent to this store later, it will be with another peer_id.
+            LOG_INFO(log, "FastAddPeer: Cancel running/queueing FAP because region destroyed, region_id={}", region_id);
+            cleanTask(tmt, proxy_helper, region_id, false);
```
CalvinNeo, line 279: We'd to block wait the FAP to cancel to prevent potential.
There is no race for now, since either `handleDestroy` or `cleanTask` handles their own data. However, it is not a good way.

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
CalvinNeo: fixed

dbms/src/Storages/KVStore/Utils/AsyncTasks.h
CalvinNeo: fixed

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -1155,7 +1154,8 @@ Segments DeltaMergeStore::buildSegmentsFromCheckpointInfo(
         return {};
     }
     wbs.writeLogAndData();
-    LOG_INFO(log, "Finish write fap checkpoint, region_id={}", checkpoint_info->region_id);
+    wbs.writeRemoves();
```
CalvinNeo, line 1157: It is originally in tiflash/dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp at 821997be4c542acdacb7536c98016963f

Lloyd-Pottiger:
Is it possible to split this PR into several PRs? It is really hard to review for me, plus I am not completely familiar with those codes...

CalvinNeo:
> Is it possible to split this PR into several PRs? It is really hard to review for me, plus I am not completely familiar with those codes...

I can move `AsyncTasks` out to another PR.

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -406,21 +406,32 @@ Segment::SegmentMetaInfos Segment::readAllSegmentsMetaInfoInRange( //
...
+        {
+            // After #7642, DELTA_MERGE_FIRST_SEGMENT_ID may not exist, however, such checkpoint won't be selected.
+            // If it were to be selected, the FAP task could fallback to legacy.
```
JaySon-Huang, line 425:
```suggestion
            // If it were to be selected, the FAP task could fallback to normal snapshot.
```

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
JaySon-Huang: I didn't get it. This line is still in `DeltaMergeStore::ingestSegmentsFromCheckpointInfo` in your PR. Why we need to call `wbs.writeRemoves()` here and what pages tend to be removed?
https://github.com/pingcap/tiflash/blob/85e144c75266ee40dd0fc5e8720f1405e4344353/dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp#L1206-L1214

dbms/src/Debug/MockKVStore/MockRaftStoreProxy.cpp
```diff
@@ -621,20 +623,27 @@ std::tuple<RegionPtr, PrehandleResult> MockRaftStoreProxy::snapshot(
...
-    region->updateAppliedIndex(index, true);
-
-    // Region changes during applying snapshot, must re-get.
-    return std::make_tuple(kvs.getRegion(region_id), prehandle_result);
+    catch (const Exception & e)
+    {
+        LOG_ERROR(log, e.message());
+        e.rethrow();
+    }
+    exit(0);
```
JaySon-Huang, line 646: should use `exit(1)` for abnormal exit? Or this is just unreachable?

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -47,12 +46,8 @@ CheckpointIngestInfoPtr CheckpointIngestInfo::restore(
     Page page = uni_ps->read(page_id, nullptr, snapshot, /*throw_on_not_exist*/ false);
     if (!page.isValid())
     {
-        throw Exception(
-            ErrorCodes::LOGICAL_ERROR,
-            "Failed to restore CheckpointIngestInfo, region_id={} peer_id={} store_id={}",
-            region_id,
-            peer_id,
-            tmt.getKVStore()->getStoreID(std::memory_order_relaxed));
+        // The restore failed, we can safely return null here to make FAP fallback.
```
JaySon-Huang, line 49: Isn't `CheckpointIngestInfo::restore` used to continue the FAP phase 2 after restart? Why it is safe now

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
```diff
@@ -209,4 +219,39 @@ void FastAddPeerContext::insertCheckpointIngestInfo(
...
+    // 2. There has been no fap at all.
+    // 3. FAP is enabled before, but disabled for now.
+    LOG_DEBUG(
+        log,
+        "FastAddPeer: no find ongoing fap task, region_id={} is_legacy_snapshot={}",
+        region_id,
+        is_legacy_snapshot);
+    // Still need to clean because there could be data left.
+    cleanTask(tmt, proxy_helper, region_id, false);
+}
```
JaySon-Huang, line 255: I'd like to use "normal snapshot" for the classical apply snapshot from SSTFiles of a Region

CalvinNeo: In our previous document, we have the following terms:
1. normal snapshot vs incremental snapshot
2. legacy snapshot vs fap snapshot

dbms/src/Debug/MockKVStore/MockRaftStoreProxy.cpp
CalvinNeo: It's unreachable, but not worth a throw since it is only used in test codes.

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
CalvinNeo: The problem here is when received a snapshot, we don't know whether it is a fap snapshot or a real empty snapshot, especially after restarted.

dbms/src/Storages/KVStore/KVStore.cpp
```diff
@@ -627,11 +626,21 @@ KVStore::StoreMeta::Base KVStore::StoreMeta::getMeta() const
...
 
+const metapb::Store & KVStore::getStoreMeta() const
+{
+    return this->store.base;
+}
+
+metapb::Store & KVStore::debugMutStoreMeta()
+{
+    return this->store.base;
+}
```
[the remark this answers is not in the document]
CalvinNeo: `debugMutStoreMeta` is for debug to mutate store meta

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
CalvinNeo: will use regular for non-fap snapshot

dbms/src/Debug/MockKVStore/MockRaftStoreProxy.cpp
JaySon-Huang: I think the `exit` line can be removed. If there is no exception, it directly return. If there is exception, then it will rethrow the exception.

dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
CalvinNeo: Fixed

dbms/src/Debug/MockKVStore/MockRaftStoreProxy.cpp
CalvinNeo: Then what do we return here? Since the compiler forces us to return something here

JaySon-Huang, line 646: got it, it should be replaced by
```suggestion
    catch (...)
    {
        tryLogCurrentFatalException(...);
        exit(1);
    }
```

JaySon-Huang: got it, it should be replaced by
```suggestion
    catch (...)
    {
        tryLogCurrentFatalException(...);
        exit(1);
    }
```

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeerContext.cpp
CalvinNeo: fixed

dbms/src/Storages/DeltaMerge/Segment.cpp
CalvinNeo: Changed to regular snapshot

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
```diff
@@ -168,42 +167,115 @@ void CheckpointIngestInfo::persistToLocal() const
...
+    auto log = DB::Logger::get("CheckpointIngestInfo");
+    if (storage && storage->engineType() == TiDB::StorageEngine::DT)
     {
-        CheckpointIngestInfo::removeFromLocal(tmt, region_id);
-        return true;
+        auto dm_storage = std::dynamic_pointer_cast<StorageDeltaMerge>(storage);
+        auto dm_context = dm_storage->getStore()->newDMContext(tmt.getContext(), tmt.getContext().getSettingsRef());
+        for (const auto & segment_to_drop : segments)
+        {
+            DM::WriteBatches wbs(*dm_context->storage_pool, dm_context->getWriteLimiter());
```
JaySon-Huang, line 202: Suggest collecting one WriteBatches for all segments and then use one write call to remove all persisted page

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -185,26 +191,34 @@ std::variant<CheckpointRegionInfoAndData, FastAddPeerRes> FastAddPeerImplSelect(
...
+
+    // Get candidate stores.
     const auto & settings = tmt.getContext().getSettingsRef();
-    auto current_store_id = tmt.getKVStore()->getStoreMeta().id();
-    auto candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
+    auto current_store_id = tmt.getKVStore()->clonedStoreMeta().id();
+    std::vector<StoreID> candidate_store_ids;
+    fiu_do_on(FailPoints::force_set_fap_candidate_store_id, { candidate_store_ids = {1234}; });
+    if (candidate_store_ids.empty())
+        candidate_store_ids = getCandidateStoreIDsForRegion(tmt, region_id, current_store_id);
```
JaySon-Huang, line 202: maybe move the failpoint into `getCandidateStoreIDsForRegion`?

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -390,44 +446,63 @@ FastAddPeerRes FastAddPeerImpl(
...
+        GET_METRIC(tiflash_fap_task_duration_seconds, type_ingest_stage).Observe(watch_ingest.elapsedSeconds());
+        auto begin = checkpoint_ingest_info->beginTime();
+        auto current = FAPAsyncTasks::getCurrentMillis();
+        if (begin != 0)
+        {
+            GET_METRIC(tiflash_fap_task_duration_seconds, type_total).Observe((current - begin) / 1000.0);
+        }
+        LOG_INFO(
+            log,
+            "Finish apply fap snapshot, region_id={} peer_id={} begin_time={} current_time={}",
```
JaySon-Huang, line 487: log down the time cost is better than log down the begin_time and current_time

JaySon-Huang: when begin_time != 0, log down the time cost; when begin_time == 0, log down a "unknown"?

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -500,19 +590,62 @@ FastAddPeerRes FastAddPeer(EngineStoreServerWrap * server, uint64_t region_id, u
...
+                if (prev_state == FAPAsyncTasks::TaskState::Running)
+                {
+                    GET_METRIC(tiflash_fap_task_state, type_blocking_cancel_stage).Increment();
+                    [[maybe_unused]] auto result = fap_ctx->tasks_trace->blockedCancelRunningTask(region_id);
+                    GET_METRIC(tiflash_fap_task_state, type_blocking_cancel_stage).Decrement();
+                }
+                else
+                {
+                    [[maybe_unused]] auto s = fap_ctx->tasks_trace->blockedCancelRunningTask(region_id);
+                }
```
JaySon-Huang, line 622: why here do the same `blockedCancelRunningTask` but increase the `type_blocking_cancel_stage` in only one branch?

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -407,21 +407,32 @@ Segment::SegmentMetaInfos Segment::readAllSegmentsMetaInfoInRange( //
...
     {
         Segment::SegmentMetaInfo segment_info;
         auto target_id = UniversalPageIdFormat::toFullPageId(
             UniversalPageIdFormat::toFullPrefix(context.keyspace_id, StorageType::Meta, context.physical_table_id),
             current_segment_id);
-        auto page = checkpoint_info->temp_ps->read(target_id);
+        auto page = checkpoint_info->temp_ps->read(target_id, nullptr, {}, false);
+        if unlikely (!page.isValid())
+        {
+            // After #7642, DELTA_MERGE_FIRST_SEGMENT_ID may not exist, however, such checkpoint won't be selected.
```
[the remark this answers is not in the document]
CalvinNeo: fixed

dbms/src/Storages/DeltaMerge/Segment.cpp
```diff
@@ -407,21 +407,32 @@ Segment::SegmentMetaInfos Segment::readAllSegmentsMetaInfoInRange( //
...
+            // If it were to be selected, the FAP task could fallback to legacy.
```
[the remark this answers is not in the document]
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
CalvinNeo: What if a exception throws before we write the `wbs`? It will have all dmfiles deleted while themselves are still registered in pagestorage
```
void Segment::dropAsFAPTemp(const FileProviderPtr & file_provider, WriteBatches & wbs)
{
    delta->recordRemoveColumnFilesPages(wbs);
    stable->recordRemovePacksPages(wbs);
    wbs.writeAll();
    stable->drop(file_provider);
}
```

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
CalvinNeo: The begin time here is to find where the FAP snapshot is generated

CalvinNeo: I would like to keep the begin_time in the log of `Begin apply fap snapshot`

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
CalvinNeo: Because only cancel a Running task is blocking. Consider other 3 states:
1. Finished -- Cancel means immediate fetch and drop result
2. InQueue -- Cancel means setting the cancel flag, so the task will be canceled once it is executed
3. NotExisting -- Obviously

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
JaySon-Huang: Removing the stable files from disk but leaving the page_id to those dmfile on UniPS may make it fail to restore the StableVS (which tries to read the meta from dmfile) from removing next time. https://github.com/pingcap/tiflash/blob/8e90a26de37e039e37091af98681a3f3374f0a88/dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp#L87-L91

You can do it this way:
* collect the stable files
* collect the page_ids in StableVS and remove them in one write batch
* call `DMFile::remove` (or more lightweight `DMFile::enableGC` and the file will be GCed later)

JaySon-Huang: Making that logic to another PR is LGTM, because it is kind of complicated and not effect the correctness of this PR.

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
JaySon-Huang: Suggest keep it simple because
* There could be some bugs in `AsyncTasks` that make it not as expected
* Keep it simple to simplify the code as adding/subtracting atomic value is not performance critical here

CalvinNeo: The idea here is not to avoid adding/subtracting atomic value, but to record events that could really block cancel. However, point 1 is reasonable. If there is bug in AsyncTasks, log all events could help us to diagnose.

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.cpp
CalvinNeo: Consider it as a non-blocking issue, and use another pr to improve is LGTM.
I think we'd better first in all solve some compat issues for cse.

dbms/src/Storages/KVStore/MultiRaft/Disagg/FastAddPeer.cpp
```diff
@@ -500,19 +593,58 @@ FastAddPeerRes FastAddPeer(EngineStoreServerWrap * server, uint64_t region_id, u
...
+                    "Cancel FAP due to timeout region_id={} new_peer_id={} prev_state={}",
+                    region_id,
+                    new_peer_id,
+                    magic_enum::enum_name(prev_state));
+                GET_METRIC(tiflash_fap_task_state, type_blocking_cancel_stage).Increment();
+                if (prev_state == FAPAsyncTasks::TaskState::Running)
+                {
+                    [[maybe_unused]] auto s = fap_ctx->tasks_trace->blockedCancelRunningTask(region_id);
+                }
+                GET_METRIC(tiflash_fap_task_state, type_blocking_cancel_stage).Decrement();
```
JaySon-Huang, line 621: Now you only cal `blockedCancelRunningTask` when prev_state == FAPAsyncTasks::TaskState::Running, it is not the same logic as the commit before. Is it expected?

CalvinNeo: No, it is a typo, actually the test fires locally... I will have a push later

JaySon-Huang (approved):

CalvinNeo:
image
image

PR #8472, opened 2023-12-06
Enhance metrics for KVStore write

### LINKED ISSUE #8471: Enhance metrics for write from KVStore
## Enhancement

1. Provide a way to observe apply fsm's CPU usage(already fixed)
2. Provide a way to observe write flow of committed data to delta tree
3. Provide a way to observe size of each write
4. Provide a way to monitor events of write to delta merge

CalvinNeo:

Issue Number: close #8471

Problem Summary:
1. Add tiflash_raft_throughput_bytes, type_write for writes from raftstore, type_write_committed for writes to delta merge
    (Add a timeseries panel in Raft)
3. Add tiflash_raft_raft_frequent_events_count.type_write_commit for every write event to to delta merge
4. Add tiflash_raft_write_flow_bytes, type_ingest_uncommitted for uncommitted writes bytes after ingestsst, type_write_committed for committed write bytes, type_net_write for net write bytes(region's size after - region's size before)
    (Add two histogram panels in Raft)

The KVStore memory tracker test
```
tiup bench ycsb load tidb -p tidb.instances="127.0.0.1:4000" -p recordcount=10000
tiup bench ycsb run tidb -p tidb.instances="10.2.12.81:5711" -p operationcount=10000
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CalvinNeo:
y4lCYCEo6P
mjwdx8hkx5
GrhoaIfY3H

JaySon-Huang:
Can you commit the changes to tiflash_summary.json ? Or provide a preview URL of the changed grafana panels

CalvinNeo:
> `tiflash_raft_region_flush_bytes.type_unflushed` seems cannot tell the memory usage of region data. Because the flushed uncommited data can still use a lot of memory.

I think the name is ambiguous and misleading here. What I actually mean is how much memory does this region take when a "unflushing" event takes place, whereas the `type_flushed` is for how much memory this region takes when a "flushing" event takes place.

CalvinNeo:
> I know what you means.
> 
> The problems of `tiflash_raft_region_flush_bytes` are:
> 
> 1. Its reporting frequency is triggered by the compact log command, not particularly timely.
> 2. It it about memory usage of each region, not all.
> 
> I think we need a `MemoryTracker` object to track the memory usage of all regions' data.

Yes.
In TiKV, there is a module named `MemoryTrace` in https://github.com/tikv/tikv/blob/master/components/tikv_alloc/src/trace.rs. I think may be we can have something like this to observe all memory consuming objects.

CalvinNeo:
> MemoryTracker

Good idea, I will look into that

dbms/src/Common/TiFlashMetrics.h
```diff
@@ -484,12 +484,13 @@ namespace DB
...
+      F(type_write_commit, {{"type", "write_commit"}}),                                                                             \
       F(type_write, {{"type", "write"}}))                                                                                           \
     M(tiflash_raft_region_flush_bytes,                                                                                              \
       "Bucketed histogram of region flushed bytes",                                                                                 \
       Histogram,                                                                                                                    \
-      F(type_flushed, {{"type", "flushed"}}, ExpBuckets{32, 2, 21}),                                                                \
-      F(type_unflushed, {{"type", "unflushed"}}, ExpBuckets{32, 2, 21}))                                                            \
+      F(type_flushed, {{"type", "flushed"}}, ExpBucketsWithRange{32, 4, 32 * 1024 * 1024}),                                         \
```
CalvinNeo, line 492: Reduce reported size from 21 -> 11

CalvinNeo: It seldomly uses this much memory.
It used to be 16MB, and I changed it to 32MB.

dbms/src/Storages/KVStore/Decode/PartitionStreams.cpp
```diff
@@ -388,9 +388,16 @@ DM::WriteResult RegionTable::writeBlockByRegion(
...
     reportUpstreamLatency(*data_list_read);
     auto write_result = writeRegionDataToStorage(context, region, *data_list_read, log);
-
+    auto prev_region_size = region->dataSize();
     RemoveRegionCommitCache(region, *data_list_read, lock_region);
-
+    auto new_region_size = region->dataSize();
+    if likely (new_region_size <= prev_region_size)
+    {
+        auto net_write = prev_region_size - new_region_size;
```
CalvinNeo, line 396: These names here are strange and misleading, will reconsider

dbms/src/Common/TiFlashMetrics.h
```diff
@@ -501,6 +502,17 @@ namespace DB
...
+      "Raft handled bytes in global",                                                                                               \
+      Counter,                                                                                                                      \
+      F(type_write, {{"type", "write"}}),                                                                                           \
+      F(type_write_committed, {{"type", "write_committed"}}))                                                                       \
+    M(tiflash_raft_write_flow_bytes,                                                                                                \
+      "Bucketed histogram of bytes for each write",                                                                                 \
+      Histogram,                                                                                                                    \
+      F(type_ingest_uncommitted, {{"type", "ingest_uncommitted"}}, ExpBucketsWithRange{16, 4, 64 * 1024}),                          \
```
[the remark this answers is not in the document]
CalvinNeo: Increased to 1MB now. I think it is enough in all cases since `raft-entry-max-size` is 8MB at most.

dbms/src/Storages/KVStore/Decode/PartitionStreams.cpp
CalvinNeo: I will remove `type_net_write` here, I think the total size of KVStore do better than `type_net_write`

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -322,6 +326,10 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
         GET_METRIC(tiflash_raft_raft_frequent_events_count, type_write).Increment(1);
         GET_METRIC(tiflash_raft_process_keys, type_write_put).Increment(put_key_count);
         GET_METRIC(tiflash_raft_process_keys, type_write_del).Increment(del_key_count);
+        auto after_size = dataSize();
+        if (after_size > prev_size)
+            GET_METRIC(tiflash_raft_write_flow_bytes, type_net_write).Observe(after_size - prev_size);
```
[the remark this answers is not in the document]
CalvinNeo: No, it does not count those data which are written into delta merge

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -26,30 +26,50 @@ extern const int LOGICAL_ERROR;
...
-        cf_data_size += default_cf.insert(std::move(key), std::move(value), mode);
-        return;
+        auto delta = default_cf.insert(std::move(key), std::move(value), mode);
+        cf_data_size += delta;
+        reportAlloc(delta);
+        return delta;
     }
     case ColumnFamilyType::Lock:
     {
         lock_cf.insert(std::move(key), std::move(value), mode);
```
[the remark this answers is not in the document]
CalvinNeo: Lock cf contributes a little and will be deleted if the transaction is committed.
If the lock is not deleted, then there will be much more data in `Default`

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -238,6 +266,9 @@ void RegionData::splitInto(const RegionRange & range, RegionData & new_region_da
     size_t size_changed = 0;
     size_changed += default_cf.splitInto(range, new_region_data.default_cf);
     size_changed += write_cf.splitInto(range, new_region_data.write_cf);
+    // Put them before computing lock cf size change.
+    reportDealloc(size_changed);
```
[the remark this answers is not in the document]
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.cpp
```diff
@@ -94,6 +94,7 @@ RegionDataRes RegionCFDataBase<Trait>::insert(std::pair<Key, Value> && kv_pair,
                         + " new_val: " + prev_value.toDebugString(),
                     ErrorCodes::LOGICAL_ERROR);
             }
+            return 0;
```
JaySon-Huang, line 97:
```suggestion
            // duplicated key is ignored
            return 0;
```

dbms/src/Storages/KVStore/Region.h
```diff
@@ -235,15 +235,20 @@ class Region : public std::enable_shared_from_this<Region>
 
     Region() = delete;
 
+#ifdef DBMS_PUBLIC_GTEST
+public:
+#else
```
JaySon-Huang, line 240: what methods make us need to use this hack?

dbms/src/Storages/KVStore/MultiRaft/IngestSST.cpp
```diff
@@ -180,6 +180,8 @@ void Region::finishIngestSSTByDTFile(RegionPtr && temp_region, UInt64 index, UIn
     {
         std::unique_lock<std::shared_mutex> lock(mutex);
 
+        auto uncommitted_ingest = temp_region->dataSize();
+        GET_METRIC(tiflash_raft_write_flow_bytes, type_ingest_uncommitted).Observe(uncommitted_ingest);
```
JaySon-Huang, line 184: what about the uncommitted bytes after `applySnapshot`

dbms/src/Storages/KVStore/Region.h
CalvinNeo: `splitInto`, seems the friend class declared here never works.

dbms/src/Storages/KVStore/MultiRaft/IngestSST.cpp
CalvinNeo: We can find some clue in `tiflash_raft_ongoing_snapshot_total_bytes` here if one snapshot has too many uncommitted data, by `type_raft_snapshot - type_dt_total`.
Of course, this metric is not as good as `tiflash_raft_write_flow_bytes` here. However, `tiflash_raft_write_flow_bytes` is a histogram metric which results in many buckets, I am not really sure if it is worthy here?

JaySon-Huang: I'm afraid we can not use `type_raft_snapshot - type_dt_total` to get the remain bytes after applying snapshot. Because `type_dt_total` is the decoded "column" bytes in memory, while `type_raft_snapshot` is the bytes of key-values read from SSTReader.

I think this metrics is to find out the reason why "CurrentMetrics::MemoryTrackingKVStore" grows. There are three known ways and missing it for the apply snapshot is weird:
* normal write raft log
* uncommitted bytes after apply snapshot
* uncommitted bytes after ingest sst

dbms/src/Storages/KVStore/Region.h
JaySon-Huang: because `TEST_F(A, B)` will generate a subclass `class A_B` that inherits `class A` using macro and run the test body. So we need to workaround in another way. I've push a commit to remove the hack.

dbms/src/Storages/KVStore/tests/gtest_new_kvstore.cpp
```diff
@@ -12,18 +12,113 @@
...
+    {
+        root_of_kvstore_mem_trackers->reset();
+        RegionPtr region = tests::makeRegion(702, start, end, proxy_helper.get());
+        region->insert("default", TiKVKey::copyFrom(str_key), TiKVValue::copyFrom(str_val_default));
+        ASSERT_EQ(root_of_kvstore_mem_trackers->get(), str_key.dataSize() + str_val_default.size());
+        tryPersistRegion(kvs, 1);
+        reloadKVSFromDisk();
+        ASSERT_EQ(root_of_kvstore_mem_trackers->get(), str_key.dataSize() + str_val_default.size());
+    }
+    return;
```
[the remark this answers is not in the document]
CalvinNeo: sorry, will fix that

JaySon-Huang:
Also commit the grafana changes to the PR?

CalvinNeo:
> Also push the grafana changes to this PR?

This pr is held until the grafana is pushed.

If other things are OK, could you firstly lgtm this PR? Lots of work taken to build such a metric with about 4+ new panels.

dbms/src/Storages/KVStore/MultiRaft/IngestSST.cpp
CalvinNeo: Yes, you are right, I will add another metric

CalvinNeo:
image
image
image

JaySon-Huang (approved):
LGTM

PR #8514, opened 2023-12-13
WIP Introduce `not_null` pointer

### LINKED ISSUE #8500: Introduce `not_null` and a safe `std::optional` to avoid dereferencing nullptrs
## Enhancement

Currently, codes are prone to panic due to undefined behavior if:
1. We derefing an `std::shared_ptr` of nullptr.
2. We derefing an `std::optional` of nullopt.

If these happens, it will be segment fault, and it is very hard to find what causes the panic. https://github.com/pingcap/tiflash/issues/8249 is an example. It is also not worthy to check by `if` everytime we try to deref.

So I suggest we introduce [`not_null`](https://github.com/microsoft/GSL/blob/main/include/gsl/pointers) and [`safe_option`](https://gist.github.com/H2CO3/52613a9a200f742b1aea07fec655df96) in TiFlash.

CalvinNeo:

Issue Number: close #8500

Problem Summary:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/NotNullPtr.h
+++ b/dbms/src/Common/NotNullPtr.h
@@ -1,0 +1,320 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+// Copyright (c) 2015 Microsoft Corporation. All rights reserved.
+//
+// This code is licensed under the MIT License (MIT).
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


dbms/src/Common/NotNullPtr.h
```diff
@@ -0,0 +1,320 @@
+///////////////////////////////////////////////////////////////////////////////
```
JaySon-Huang, line 1: put it to contrib lik `contrib/cpptoml`

CalvinNeo:
Use https://github.com/pingcap/tiflash/pull/8802

PR #8430, opened 2023-11-28
Disagg: Fetch MemTableSet by streaming

JinheLin:
### What problem does this PR solve?

Issue Number: close #8429

Problem Summary:

Currently, th data of delta cache is responded in the Establish stage, instead of in the FetchPages stage.

We could improve it to respond in the FetchPages stage, so that the parallel FetchPages could start as soon as possible and reducing memory usage of Establish stage.

### What is changed and how it works?

- Make Serializer to support not serializing data of ColumnFileInMemory and can deserializing ColumnFileInMemory without data.
  - dbms/src/Storages/DeltaMerge/Remote/Serializer.h
  - dbms/src/Storages/DeltaMerge/Remote/Serializer.cpp
- Make WNEstablishDisaggTaskHandler not return data of ColumnFileInMemory by default.
  - dbms/src/Flash/Disaggregated/WNEstablishDisaggTaskHandler.cpp
- Make WNFetchPagesStreamWriter sending MemTableSet in streaming. Also refine some of its code for writing unit-tests.
  - dbms/src/Flash/Disaggregated/WNFetchPagesStreamWriter.h
  - dbms/src/Flash/Disaggregated/WNFetchPagesStreamWriter.cpp
- Make SegmentReadTask support receiving MemTableSet from WN. Also refine some of its code for writing unit-tests.
  - dbms/src/Storages/DeltaMerge/SegmentReadTask.h
  - dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8523, opened 2023-12-14
Storages: fix segmentfault when `fetchPages` failed.

### LINKED ISSUE #8515: all tiflash compute crash when kill one of tiflash wn
## Bug Report

1、run ch
2、 kill one of tiflash wn （tc-tiflash-1）

no crash

all tiflash compute crash when kill one of tiflash wn

img_v3_0264_1440726f-45dc-435b-9bfc-2bba81a4569g

git hash：6f4deccd645ecdf9cc341c6e1bf4fcc3a1bbd55b

JinheLin:

Issue Number: close #8515

Problem Summary:
When network partitioned, CN fetches pages and memtableset from WN will fail. However, we forgot to check the status of grpc after response stream returns false, and the function `FetchPages()` may think that the pages fetching was successful (but actually not).

Before [PR8430](https://github.com/pingcap/tiflash/pull/8430), [some](https://github.com/pingcap/tiflash/blob/master/dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp#L674-L678)[ condition checks](https://github.com/pingcap/tiflash/blob/master/dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp#L674-L678) will discover these anomalies, and throw exceptions. In the case that all pages are cached, this condition check takes no effect. 

In [PR8430](https://github.com/pingcap/tiflash/pull/8430), `FetchPages()` needs to fetch `MemTableSet` too. If grpc fails, the data of `MemTableSet` is empty (pointer of cache is nullptr), and causes segmentfault when reading data.

- We need to call `Finish()` and check the status of grpc after response stream returns false.
- For safety, check if the data of `ColumnFileInMemory` is ready after fetching pages.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
    - Run queries and restart WN, check server status of CNs.
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp
+++ b/dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp
@@ -502,6 +502,18 @@ void SegmentReadTask::checkMemTableSet(const ColumnFileSetSnapshotPtr & mem_tabl
     }
 }
 
+void SegmentReadTask::checkMemTableSetReady() const
+{
+    const auto & mem_table_snap = read_snapshot->delta->getMemTableSetSnapshot();
+    for (auto & cf : mem_table_snap->getColumnFiles())
+    {
+        if (auto * in_mem_cf = cf->tryToInMemoryFile(); in_mem_cf)
+        {
+            RUNTIME_CHECK(in_mem_cf->getCache() != nullptr);
+        }
+    }
+}
+
 bool SegmentReadTask::needFetchMemTableSet() const
 {
     // Check if any object of ColumnFileInMemory does not contain data.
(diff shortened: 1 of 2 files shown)


CalvinNeo (approved):
lgtm

dbms/src/Storages/DeltaMerge/SegmentReadTask.cpp
```diff
@@ -543,12 +555,32 @@ void SegmentReadTask::doFetchPages(const disaggregated::FetchDisaggPagesRequest
...
+            {
+                return true;
+            }
+            else
+            {
+                auto status = stream_resp->Finish();
+                stream_resp.reset(); // Reset to avoid calling `Finish()` repeatedly.
+                RUNTIME_CHECK_MSG(
+                    status.ok(),
+                    "status={} message={}",
```
JaySon-Huang, line 578: Add some messages like L709 `Failed to fetch all pages (from {})...`

JaySon-Huang (approved):

Issue #8506, opened 2023-12-12
Unstable tests `fullstack-test2/ddl/flashback/flashback_database.test`


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)
```
[2023-12-12T13:23:18.666Z] fullstack-test2/ddl/flashback/flashback_database.test: Running
[2023-12-12T13:23:33.477Z]   File: fullstack-test2/ddl/flashback/flashback_database.test
[2023-12-12T13:23:33.477Z]   Error line: 115
[2023-12-12T13:23:33.477Z]   Error: set session tidb_isolation_read_engines='tiflash'; select * from d1_new.t1 order by a;
[2023-12-12T13:23:33.477Z]   Result:
[2023-12-12T13:23:33.477Z]     <nothing>
[2023-12-12T13:23:33.477Z]   Expected:
[2023-12-12T13:23:33.477Z]     +------+------+
[2023-12-12T13:23:33.477Z]     | a    | b    |
[2023-12-12T13:23:33.477Z]     +------+------+
[2023-12-12T13:23:33.477Z]     |    1 | NULL |
[2023-12-12T13:23:33.477Z]     |    2 |    2 |
[2023-12-12T13:23:33.477Z]     +------+------+
```
https://ci.pingcap.net/job/tiflash-ghpr-integration-tests/15401/artifact/fullstack-test2-logs.tar.gz
https://ci.pingcap.net/job/tiflash-ghpr-integration-tests/15396/artifact/fullstack-test2-logs.tar.gz

### 4. What is your TiFlash version? (Required)
master

ref https://github.com/pingcap/tiflash/issues/8450

JaySon-Huang:
```

[2023-12-15T08:10:44.558Z]   File: fullstack-test2/ddl/flashback/flashback_database.test
[2023-12-15T08:10:44.559Z]   Error line: 162
[2023-12-15T08:10:44.559Z]   Error: set session tidb_isolation_read_engines='tiflash'; select * from d1_new.t2 order by id;
[2023-12-15T08:10:44.559Z]   Result:
[2023-12-15T08:10:44.559Z]     id   name    b
[2023-12-15T08:10:44.559Z]     1    abc NULL
[2023-12-15T08:10:44.559Z]     2    cde NULL
[2023-12-15T08:10:44.559Z]   Expected:
[2023-12-15T08:10:44.559Z]     +----+------+------+
[...]

PR #8536, opened 2023-12-15
ddl: Fix corner case of `FLASHBACK DATABASE`

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8506

Problem Summary:

If the Region snapshot arrived TiFlash after the database has been dropped, then we can not successfully create the Storage instance because we can not fetch it from TiKV. And also `DatabaseInfo` can not be fetch by "mvcc get".
https://github.com/pingcap/tiflash/blob/a46271dd7db654ec0ab59f77915706863bf91e0b/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L1454-L1465

To successfully create a Storage instance (`applyCreateStorageInstance`), we don't need the `DatabaseInfo`, a `DatabaseID` is sufficient. We are rewriting methods in `SchemaBuilder` to replace the methods with `DatabaseInfo` by `DatabaseID`. So that we can create those Storage instances and decode the Raft logs or snapshots after a database is dropped and the data can be "flashback" later.
What's more, avoiding fetching the `DatabaseInfo` from TiKV can speed up syncing schema.

For those methods that need the `DatabaseInfo` to get the database name in TiDB to rename tables, we fetch the `DatabaseInfo` from the local cache because the `applyDiff` that calls those methods should be applied before the "drop database" SchemaDiff.

Add a failpoint to pause pre-handling snapshot for testing.

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/tests/fullstack-test2/ddl/flashback/flashback_database.test
+++ b/tests/fullstack-test2/ddl/flashback/flashback_database.test
@@ -97,6 +97,7 @@ func> wait_table d1 t1
 mysql> alter table d1.t1 add column b int;
 
 >> DBGInvoke __enable_fail_point(pause_before_apply_raft_cmd)
+>> DBGInvoke __enable_fail_point(pause_before_prehandle_snapshot)
 
 # exactly write until fail point "pause_before_apply_raft_cmd" to be disable
 mysql> insert into d1.t1 values(2,2);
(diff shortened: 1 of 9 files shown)


tests/fullstack-test2/ddl/flashback/flashback_database.test
```diff
@@ -145,15 +147,18 @@ mysql> create table d1.t2(id INT NOT NULL,name VARCHAR(30)) PARTITION BY RANGE (
 mysql> insert into d1.t2 values(1, 'abc'),(2, 'cde'),(53, 'efg');
 
 >> DBGInvoke __enable_fail_point(pause_before_apply_raft_cmd)
+>> DBGInvoke __enable_fail_point(pause_before_prehandle_snapshot)
 mysql> alter table d1.t2 set tiflash replica 1;
 mysql> alter table d1.t2 add column b int;
 
 mysql> drop database d1;
 
+SLEEP 5
```
JaySon-Huang, line 156:
```suggestion
```

hongyunyan (approved):
LGTM

Lloyd-Pottiger (approved):

Issue #8351, opened 2023-11-10
Support encryption at rest under disaggregate mode


Lloyd-Pottiger:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Encryption at rest provides data protection for stored data (at rest). In disaggregate TiFlash, we still do not support it.

**Describe the feature you'd like:**

Each keyspace, each file uses its own encryption key. 

- Different files use different keys.
- For data belonging to different keyspaces and stored in the same file, should use different keys.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #8540, opened 2023-12-18
Storage: add keyspaceID to DMFile

Lloyd-Pottiger:

Issue Number: ref #8351

Problem Summary:

I will modify `struct EncryptionPath` into the following:

```cpp
/// 1. In OP, we use tikv proxy, all encryption info are stored in one file called file.dict.
/// Every update of file.dict will sync the whole file.
/// So when the file is too large, the update cost increases.
/// To keep the file size as small as possible, we reuse the encryption info among a group of related files.(e.g. the files of a DMFile)
/// For security reason, the same `iv` is not allowed to encrypt two different files,
/// so we combine the `iv` fetched from file.dict with the hash value of the file name to calculate the real `iv` for every file.
/// 2. In cloud, we use cse proxy, all encryption info are stored in page storage.
/// Each keyspace has only a encryption info.
/// We will use full_path and file_name to generate `iv`.
struct EncryptionPath
{
    EncryptionPath(const String & full_path_, const String & file_name_)
        : full_path{full_path_}
        , file_name{file_name_}
    {}

    EncryptionPath(const String & full_path_, const String & file_name_, UInt32 keyspace_id_)
        : full_path{full_path_}
        , file_name{file_name_}
        , keyspace_id{keyspace_id_}
    {}

    // If it is a file, full_path is the full path of the file.
    // If it is a directory, full_path is the full path of the directory.
    const String full_path;
    const String file_name;
    // The id of the keyspace which the file belongs to.
    const UInt32 keyspace_id = 0;
};
```

so we need to store keyspaceID in DMFile.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Lloyd-Pottiger:
> Should we store KeyspaceID in DMFile?

1. DMFile can only belong to a keyspace, so it is ok to store KeyspaceID.
2. We need KeyspaceID to get the encryption key.

CalvinNeo:
A DmFile can be mapped to a region by range, a region is currently tagged with keyspace id, so why don't we just use the keyspace in Region?

Lloyd-Pottiger:
> A DmFile can be mapped to a region by range, a region is currently tagged with keyspace id, so why don't we just use the keyspace in Region?

Store a uint32 keyspaceID is not a big deal, we just need a simple way to get keyspaceID in DMFile's member functions like

```
EncryptionPath DMFile::encryptionDataPath(const FileNameBase & file_name_base) const
{
    return EncryptionPath(encryptionBasePath(), file_name_base + details::DATA_FILE_SUFFIX);
}```

CalvinNeo (approved):
lgtm

PR #8553, opened 2023-12-19
Add mutex to protect concurrent prepareMPPTasks

yibin87:
### What problem does this PR solve?

Issue Number: ref #8285

Problem Summary:
Executor id is the field of global instance context. And it is passed as ref during prepareMPPTasks for TestUtils, and will be increased by 1 each time. Add mutex to protect it and pass tsan check.
Increase the sleep time, since in tsan mode, the execution time will be longer than before due to the mutex lock operation.
### Check List

Tests 

- [x] Unit test
- [] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

Issue #8285, opened 2023-11-01
TSan new failed compute test case ComputeServerRunner.aggWithColumnPrune


yibin87:
Use 2023-10-31 latest version, error message:
[2023-10-31T18:37:38.841Z] WARNING: ThreadSanitizer: data race (pid=20500)
[2023-10-31T18:37:38.841Z]   Read of size 8 at 0x7b6000030208 by main thread:
[2023-10-31T18:37:38.841Z]     #0 memcpy /llvm-project/compiler-rt/lib/tsan/../sanitizer_common/sanitizer_common_interceptors.inc:827:5 (gtests_dbms+0x2ba6ce0)
[2023-10-31T18:37:38.841Z]     #1 core::mem::maybe_uninit::MaybeUninit$LT$T$GT$::assume_init::h2f3b2e6915ec01bd /rustc/96ddd32c4bfb1d78f0cd03eb068b1710a8cebeef/library/core/src/mem/maybe_uninit.rs:627:38 (gtests_dbms+0xfe05c16)
[2023-10-31T18:37:38.841Z]     #2 core::ptr::read::h5d271dd8cc4a87b2 /rustc/96ddd32c4bfb1d78f0cd03eb068b1710a8cebeef/library/core/src/ptr/mod.rs:1140:9 (gtests_dbms+0xfe05c16)
[2023-10-31T18:37:38.841Z]     #3 alloc::vec::Vec$LT$T$C$A$GT$::remove::h77cb7968e7379f6d /rustc/96ddd32c4bfb1d78f0cd03eb068b1710a8cebeef/library/alloc/src/vec/mod.rs:1522:23 (gtests_dbms+0xfe05c16)
[2023-10-31T18:37:38.841Z]     #4 backtrace::symbolize::gimli::Cache::mapping_for_lib::h8e9c09979b3359f8 /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.64/src/symbolize/gimli.rs:312:29 (gtests_dbms+0xfe05c16)
[2023-10-31T18:37:38.841Z]     #5 backtrace::symbolize::gimli::resolve::_$u7b$$u7b$closure$u7d$$u7d$::hc070d62f1c0573e8 /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.64/src/symbolize/gimli.rs:354:24 (gtests_dbms+0xfe05c16)
[...]

PR #8565, opened 2023-12-22
Fix gtest data race detected by tsan check

yibin87:

Issue Number: ref #8285 #8553

Problem Summary:
In #8553, mutex is added to protect prepareMPPTasks. However, the DAGRequestBuilder parameter is constructed outside the protection, and the context's index will be updated during the construction. This is caused by the two-phase compilation: first for operators without exchange, second for exchange operators. Thus adjust the mutex protection scope in this pr.

Set server's mock storage at beginning to avoid harmless data races.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Flash/tests/gtest_compute_server.cpp
```diff
@@ -147,11 +147,16 @@ class ComputeServerRunner : public DB::tests::MPPTaskTestUtils
             BlockInputStreamPtr stream;
             try
             {
-                auto tasks = prepareMPPTasks(
```
SeaRise, line 150: how about `prepareMPPTasks([&] () { return context.xxx }, properties)`?
and then `DAGRequestBuilder` will under lock

SeaRise: Maybe `MockDAGRequestContext::index` should be atomic_size_t?

yibin87: Sounds great, I'll change it.

SeaRise: Maybe `thread_local size_t index`? but seems that mac support `thread_local` :(

SeaRise (approved):

windtalker (approved):
LGTM

PR #8567, opened 2023-12-22
Storages: limit memory usage of data sharing

### LINKED ISSUE #8564: Data sharing may use more memory than expected
img_v3_026b_758a591a-291e-431f-b853-d18c475950fg

img_v3_026b_73732574-1c95-4c8a-ad51-77e1a74395bg

JinheLin:

Issue Number: close #8564

Problem Summary:
In terms of design, we hope that by scheduling tasks, the shared data will be immediately read and deleted, so under normal circumstances, it will not occupy additional memory.
However, there is a scenario that would break this assumption: If two requests A and B are merged, request A reads faster and request B reads slowly. 

A will share the data with B, but request B may skip directly without reading and deleting these shared data, because it finds its queue is full (https://github.com/pingcap/tiflash/blob/11c056718a90d4fb283890ee0dc842ffa32555f8/dbms/src/Storages/DeltaMerge/ReadThread/MergedTask.cpp#L81-L84), which may result in consuming much more memory than expected.

This PR adds limitations on the size of shared data. It includes two aspects:
- Global shared data size limit.
- Limit on the count of shared data held by each DMFileReader.
  - Why not limit the size of shared data held by each DMFileReader?
  - First, it just holds a smart pointer. This pointer can be shared between multiple requests. How many bytes should it take? That needs some considerations.
  - Second, make changes smaller to pick this PR to other LST branches.
- Also, it supports disabling data sharing and doesn't affect read thead.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  - Run a fast tablescanning query and a slow tablescanning query concurrently.
  - In the image below, 
    - on the left is the code of this PR, the limitation of shared data size is 1GB.
    - on the right is the code of master.
  
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


Lloyd-Pottiger:
Can we adjust the execution order of merged tasks to let the slow task run first? Maybe we can simply order by read_columns.

dbms/src/Storages/DeltaMerge/tests/gtest_column_sharing_cache.cpp
```diff
@@ -50,7 +50,7 @@ TEST(ColumnSharingCacheTest, AddAndGet)
     ColumnSharingCache cache;
 
     auto col = createColumn(8);
-    cache.add(1, 8, col);
+    cache.add(1, 8, col, std::numeric_limits<UInt64>::max());
```
JaySon-Huang, line 53: Add some case about reaching the limit, then get/del

dbms/src/Common/TiFlashMetrics.h
```diff
@@ -582,6 +582,8 @@ static_assert(RAFT_REGION_BIG_WRITE_THRES * 4 < RAFT_REGION_BIG_WRITE_MAX, "Inva
       F(type_ru_exhausted, {"type", "ru_exhausted"}),                                                                               \
       F(type_add_cache_succ, {"type", "add_cache_succ"}),                                                                           \
       F(type_add_cache_stale, {"type", "add_cache_stale"}),                                                                         \
+      F(type_add_cache_count_limit, {"type", "add_cache_count_limit"}),                                                             \
```
Lloyd-Pottiger, line 585:
```suggestion
      F(type_add_cache_reach_count_limit, {"type", "add_cache_count_limit"}),                                                             \
```

maybe `type_add_cache_reach_count_limit` is better

dbms/src/Storages/DeltaMerge/File/DMFileBlockInputStream.cpp
```diff
@@ -63,6 +65,20 @@ DMFileBlockInputStreamPtr DMFileBlockInputStreamBuilder::build(
...
+    if (!enable_read_thread || max_sharing_column_bytes_for_all <= 0)
+    {
+        // Disable data sharing.
+        max_sharing_column_count = 0;
+    }
+    else if (
+        shared_column_data_mem_tracker != nullptr
+        && std::cmp_greater_equal(shared_column_data_mem_tracker->get(), max_sharing_column_bytes_for_all))
+    {
+        // Memory limitation.
```
JaySon-Huang, line 77:
```suggestion
        // The memory used reaches the limitation by running queries, disable the data sharing for this DMFile
```

Lloyd-Pottiger (approved):

JaySon-Huang (approved):

JaySon-Huang:
Changed to release note to be "Fix a bug that the memory usage rises significantly when there are slow queries"

Issue #8583, opened 2023-12-25
The lock in `RegionPersister::doPersist` slow down raft apply when meets jitter


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

https://github.com/pingcap/tiflash/blob/5980908e68629f40c2191ae2d100f41d554dce44/dbms/src/Storages/KVStore/MultiRaft/RegionPersister.cpp#L99-L125

`RegionPersister::doPersist` acquires a lock and calls `PageStorage::write` with the lock. This means when we're persisting Region meta, all other threads can not do the persistence for other Region meta. Which is not reasonable.

### 4. What is your TiFlash version? (Required)
4.0~master

PR #8584, opened 2023-12-25
KVStore: Reduce lock contention in `RegionPersister::doPersist`

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8583

Problem Summary:

https://github.com/pingcap/tiflash/blob/5980908e68629f40c2191ae2d100f41d554dce44/dbms/src/Storages/KVStore/MultiRaft/RegionPersister.cpp#L99-L125

`RegionPersister::doPersist` acquires a lock and calls `PageStorage::write` with the lock. This means when we're persisting Region meta, all other threads can not do the persistence for other Region meta. Which is not reasonable.

There is a `region_task_lock` to ensure that only one thread could call `RegionPersister::doPersist` at the same time. And `PageStorage` support multi-thread reading and writing for different page_id. There is no need to acquire another lock in `RegionPersister::doPersist` that could slowdown the raft apply process

And also remove calling `KVStore::persistRegion` with a null `region_task_lock`. So we can simplify codes and remove the member `RegionPersister:: RegionManager`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce the impact of disk performance jitter on write throughput
```

### DIFF
--- a/dbms/src/Storages/KVStore/KVStore.cpp
+++ b/dbms/src/Storages/KVStore/KVStore.cpp
@@ -55,9 +55,8 @@ extern const char force_not_clean_fap_on_destroy[];
 
 KVStore::KVStore(Context & context)
     : region_persister(
-        context.getSharedContextDisagg()->isDisaggregatedComputeMode()
-            ? nullptr
-            : std::make_unique<RegionPersister>(context, region_manager))
+        context.getSharedContextDisagg()->isDisaggregatedComputeMode() ? nullptr
+                                                                       : std::make_unique<RegionPersister>(context))
     , raft_cmd_res(std::make_unique<RaftCommandResult>())
     , log(Logger::get())
     , region_compact_log_min_rows(40 * 1024)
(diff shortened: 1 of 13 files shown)


dbms/src/Storages/KVStore/KVStore.cpp
```diff
@@ -375,33 +374,25 @@ void KVStore::setRegionCompactLogConfig(UInt64 rows, UInt64 bytes, UInt64 gap, U
...
+    auto reason_id = magic_enum::enum_underlying(reason);
+    std::string caller = fmt::format("{} {}", PersistRegionReasonMap[reason_id], extra_msg);
+    LOG_INFO(
+        log,
+        "Start to persist {}, cache size: {} bytes for `{}`",
+        region.getDebugString(),
+        region.dataSize(),
+        caller);
+    region_persister->persist(region, region_task_lock);
+    LOG_DEBUG(log, "Persist {} done, cache size: {} bytes", region.toString(false), region.dataSize());
```
CalvinNeo, line 395: Use `region size` for `region.dataSize()`. Though it's quite misleading that we have both `approx_mem_cache_bytes` and `region.dataSize()`. I think maybe we optimize it later.

CalvinNeo (approved):
lgtm

Lloyd-Pottiger (approved):

JaySon-Huang:
Change the release note from "write throughput" to "read latency"

Tests with some failpoint injects (99%-10ms and %0.7-100ms and %0.2-1000ms) latency on `RegionPersister::doPersist` show that the changes mainly affect the read latency caused by `wait index` but not the write throughput

image

PR #8593, opened 2023-12-26
KVStore: Reduce lock contention in `RegionPersister::doPersist` (release-6.5)

JaySon-Huang:
cherry-pick of https://github.com/pingcap/tiflash/pull/8584

Issue Number: close https://github.com/pingcap/tiflash/issues/8583

Problem Summary:

https://github.com/pingcap/tiflash/blob/5980908e68629f40c2191ae2d100f41d554dce44/dbms/src/Storages/KVStore/MultiRaft/RegionPersister.cpp#L99-L125

`RegionPersister::doPersist` acquires a lock and calls `PageStorage::write` with the lock. This means when we're persisting Region meta, all other threads can not do the persistence for other Region meta. Which is not reasonable.

There is a `region_task_lock` to ensure that only one thread could call `RegionPersister::doPersist` at the same time. And `PageStorage` support multi-thread reading and writing for different page_id. There is no need to acquire another lock in `RegionPersister::doPersist` that could slowdown the raft apply process

And also remove calling `KVStore::persistRegion` with a null `region_task_lock`. So we can simplify codes and remove the member `RegionPersister:: RegionManager`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce the impact of disk performance jitter on write throughput
```

### DIFF
--- a/dbms/src/Common/FailPoint.cpp
+++ b/dbms/src/Common/FailPoint.cpp
@@ -107,7 +107,8 @@ namespace DB
     M(pause_before_apply_raft_snapshot)        \
     M(pause_until_apply_raft_snapshot)         \
     M(pause_after_copr_streams_acquired_once)  \
-    M(pause_before_register_non_root_mpp_task)
+    M(pause_before_register_non_root_mpp_task) \
+    M(pause_when_persist_region)
 
 #define APPLY_FOR_PAUSEABLE_FAILPOINTS(M) \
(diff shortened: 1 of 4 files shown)


Lloyd-Pottiger (approved):

CalvinNeo (approved):
lgtm

Issue #8650, opened 2024-01-03
The lock in `BlobStore::removePosFromStats` introducing performance jitter


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

When all page data in a BlobFile gets removed in the GC thread, `removePosFromStats` will remove the BlobFile from disk. However, it holds a lock on `mtx_blob_files` when removing the file from disk. If `BlobStore::write` happens to write a page data to a new BlobFile, `BlobStore::getBlobFile` will be blocked by the GC thread.

According to local tests, removing a BlobFile takes tens of milliseconds. Thus introducing performance jitter.

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L803-L814

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L632-L636

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L1494-L1503

### 4. What is your TiFlash version? (Required)
master

PR #8651, opened 2024-01-03
PageStorage: Optimize the critical region when removing a BlobFile from disk

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8650

Problem Summary:

When all page data in a BlobFile gets removed in the GC thread, `removePosFromStats` will remove the BlobFile from disk. However, it holds a lock on `mtx_blob_files` when removing the file from disk. If `BlobStore::write` happens to write a page data to a new BlobFile, `BlobStore::getBlobFile` will be blocked by the GC thread.

According to local tests, removing a BlobFile takes tens of milliseconds. Thus introducing performance jitter.

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L803-L814

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L632-L636

https://github.com/pingcap/tiflash/blob/63eb295f0664ec4251907f00f335976225ddf337/dbms/src/Storages/Page/V3/BlobStore.cpp#L1494-L1503

Keep a pointer within the lock on `mtx_blob_files`, and remove the blob file from disk without the lock

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Reduce the effect of background GC tasks on foreground read/write tasks
```

### DIFF
--- a/dbms/src/Storages/Page/V3/BlobStore.cpp
+++ b/dbms/src/Storages/Page/V3/BlobStore.cpp
@@ -800,17 +800,25 @@ void BlobStore<Trait>::removePosFromStats(BlobFileId blob_id, BlobFileOffset off
         auto lock_stats = blob_stats.lock();
         blob_stats.eraseStat(std::move(stat), lock_stats);
     }
+
+    BlobFilePtr blob_file_to_removed;
     {
-        // Remove the blob file from disk and memory
+        // Remove the blob file from memory
+        // If the blob_id does not exist, the blob_file is never
+        // opened for read/write. It is safe to ignore it.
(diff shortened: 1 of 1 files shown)


Lloyd-Pottiger (approved):
lgtm～

Issue #8695, opened 2024-01-17
Mis-reuse StoragePool::max_data_page_id cause data corruption after changing tiflash replica number


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

1. Create a table, insert some data, add tiflash replica
2. Set tiflash replica to 0, and wait for its data on TiFlash to be GCed
3. Set tiflash replica to 1 again, and do some queries

### 3. What did you see instead (Required)

```
# table is tombstoned after setting tiflash replica to 0
[2024/01/11 14:05:08.373 +08:00] [INFO] [SchemaBuilder.cpp:1203] ["Tombstone table db_2855.testdemo begin, table_id=5578"] [source="keyspace=4294967295"] [thread_id=434]
[2024/01/11 14:05:08.382 +08:00] [INFO] [SchemaBuilder.cpp:1225] ["Tombstone table db_2855.testdemo end, table_id=5578"] [source="keyspace=4294967295"] [thread_id=434]
...
# table get physically removed
[2024/01/11 14:20:14.622 +08:00] [INFO] [SchemaSyncService.cpp:218] ["Schema GC begin, last_safepoint=446943112160346112 safepoint=446943269446746112"] [source="keyspace=4294967295"] [thread_id=58]
[2024/01/11 14:20:14.622 +08:00] [INFO] [SchemaSyncService.cpp:256] ["Detect stale table, database_name=db_2855 table_name=t_5578 database_tombstone=0 table_tombstone=446943227634515973 safepoint=446943269446746112"] [thread_id=58]
[2024/01/11 14:20:14.623 +08:00] [INFO] [SchemaSyncService.cpp:295] ["Physically dropping table, table_tombstone=446943227634515973 safepoint=446943269446746112 baocai.testdemo, database_id=2855 table_id=5578"] [source="keyspace=4294967295"] [thread_id=58]
[2024/01/11 14:20:14.646 +08:00] [INFO] [SchemaSyncService.cpp:306] ["Physically dropped table baocai.testdemo, database_id=2855 table_id=5578"] [source="keyspace=4294967295"] [thread_id=58]
[2024/01/11 14:20:14.646 +08:00] [INFO] [SchemaSyncService.cpp:383] ["Schema GC done, tables_removed=1 databases_removed=0 safepoint=446943269446746112"] [source="keyspace=4294967295"] [thread_id=58]
...
[...]

PR #8698, opened 2024-01-17
Storage: Fix max-id being mis-reused cause data corruption after changing tiflash replica number

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8695

Problem Summary: As the issue describe: https://github.com/pingcap/tiflash/issues/8695#issuecomment-1895280095

Introduce a `GlobalPageIdAllocator` to avoid a page_id from being reused in one physical_table_id scope. Specifically,

* Introduced a `GlobalPageIdAllocator` to allocate page_id for segment_id, delta-layer and stable-layer for all IStorage instances instead of each `StoragePool` allocate the page_id for its own IStorage instance independently
* Each IStorage instance restored from disk, it will first try raise the lower bound of the `GlobalPageIdAllocator`. After that, a table won't be able to reuse the allocated id
* If an IStorage instance being physically removed (cause by setting tiflash replica to 0) and re-create again, it will still allocated the page_id from `GlobalPageIdAllocator`, so no page_id will be reused too.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
Load tpc-ds 50
Set tiflash replica 1 for database tpcds50 and wait for all tables progress == 1
Set tiflash replica 0 and wait for gc
Set tiflash replica 1 and wait for all tables progress == 1
Run queries on tpcds 50 through tiflash
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the tiflash replica data may be corrupted after setting the tiflash replica to 0 and add it back later
```

### DIFF
--- a/dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
+++ b/dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
@@ -95,7 +95,8 @@ typename PageDirectoryFactory<Trait>::PageDirectoryPtr PageDirectoryFactory<Trai
     const String & storage_name,
     FileProviderPtr & file_provider,
     PSDiskDelegatorPtr & delegator,
-    PageEntriesEdit & edit)
+    PageEntriesEdit & edit,
+    UInt64 filter_seq)
 {
     auto [wal, reader] = WALStore::create(storage_name, file_provider, delegator, WALConfig());
     (void)reader;
@@ -108,7 +109,7 @@ typename PageDirectoryFactory<Trait>::PageDirectoryPtr PageDirectoryFactory<Trai
         r.version.sequence = ++mock_sequence;
     }
 
-    loadEdit(dir, edit, /*force_apply*/ true);
+    loadEdit(dir, edit, /*force_apply*/ false, filter_seq);
     // Reset the `sequence` to the maximum of persisted.
     dir->sequence = max_applied_ver.sequence;
     RUNTIME_CHECK(dir->sequence, mock_sequence);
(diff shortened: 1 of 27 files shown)


dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
```diff
@@ -362,6 +350,28 @@ void PageDirectoryFactory<Trait>::applyRecord(
...
+            || UniversalPageIdFormat::isType(r.page_id, StorageType::Meta))
+        {
+            dir->max_page_id = std::max(dir->max_page_id, Trait::PageIdTrait::getU64ID(r.page_id));
+        }
+    }
+    else
+    {
+        dir->max_page_id = std::max(dir->max_page_id, Trait::PageIdTrait::getU64ID(r.page_id));
+    }
+}
```
Lloyd-Pottiger, line 373: move to anonymous namespace?

Lloyd-Pottiger (approved):
LGTM~

dbms/src/Storages/Page/V3/PageDirectoryFactory.cpp
JaySon-Huang: This function depends on the `typename Trait` of PageDirectoryFactory. If we make it a static function outside the `PageDirectoryFactory` class, the function signature is complicated.

Lloyd-Pottiger: get it

PR #8702, opened 2024-01-18
Storages: Update the length of Block queue and the number of read threads

### LINKED ISSUE #8703: Update the length of Block queue and the number of read threads

JinheLin:

Issue Number: close #8703

- The length of Block queue:
  - For on-premise deployment, adjusting the length of the Block queue to 1.5 times of the current default value is expected to reduce the time required for simple queries by 3-6%.
  - For disaggregated deployment, adjusting the length of the Block queue to 1.5 times of the current default value is expected to reduce the time required for simple queries by 2-20%. The specific performance improvement depends on data caching. The less cached data, the more significant the improvement.
 
- The number of read threads:
  - For disaggregated deployment, if most of the data cannot hit the local disk cache, the network will become a bottleneck, and increasing the number of threads can significantly improve performance. If more than half of the data is expected to hit the local cache, adjusting the number of threads to 2 times of the current default value is enough.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):
LGTM

I think it should be picked back to release-7.5 at least. What about the release-6.5 and release-6.1?

Lloyd-Pottiger (approved):

Issue #8691, opened 2024-01-16
Refactor the IO related directory


Lloyd-Pottiger:
## Enhancement

split `dbms/src/IO` and `dbms/src/Encryption` into

- BaseFile: classes that manipulate files directly
- Encryption: encryption and checksum
- Compression: compression
- IO: all kinds of buffers, the objects that we manipulate directly when we want to do IO.

PR #8709, opened 2024-01-19
*: clean the Encryption directory

Lloyd-Pottiger:

Issue Number: ref #8691

Problem Summary:

1. [*: remove (Write|Read)BufferFromFileProvider](https://github.com/pingcap/tiflash/commit/50a98ac200e7530580d654f16b98ea0bfd347ee2)
2. [remove create(Read|Write)BufferFromFileBaseByFileProvider](https://github.com/pingcap/tiflash/commit/6391052b539d5dd867c142e65a8c53707f9f675a)
3. move checksum related code into `IO/Checksum` directory

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Encryption/ReadBufferFromRandomAccessFileBuilder.cpp
```diff
@@ -13,23 +13,24 @@
...
-#include <Encryption/ReadBufferFromFileProvider.h>
+#include <Encryption/ReadBufferFromRandomAccessFileBuilder.h>
+
 
 namespace DB
 {
-ReadBufferFromFileProvider::ReadBufferFromFileProvider(
-    const FileProviderPtr & file_provider_,
+
+ReadBufferFromRandomAccessFilePtr ReadBufferFromRandomAccessFileBuilder::build(
```
CalvinNeo, line 22: Why we return a Ptr rather an object here?

Lloyd-Pottiger: maybe it can reduce a copy

CalvinNeo: Could you give me an example?

CalvinNeo: Meanwhile, will it bring about an extra indirect access, for example, we need to use `*it` rather than `it`. Will it change the performance if we use unique_ptr?

Lloyd-Pottiger: I think the performance impact is small that we can just ignore.

Since `ReadBuffer` is a heavy class which usually own a large memory, copy & move are both expensive, I think use  pointer maybe a better idea.

Lloyd-Pottiger:
```cpp
ReadBufferFromRandomAccess ReadBufferFromRandomAccessFileBuilder::build() {
    return ReadBufferFromRandomAccess(xxx)
}

auto buf = ReadBufferFromRandomAccessFileBuilder::build(); // there is a copy, although it may be optimized.
```

CalvinNeo: No, there are exactly no copy in your given example, even copies that you may think been optimized out.

CalvinNeo: The C++ standard guarantees that no copy shall happen in this case, see https://en.cppreference.com/w/cpp/language/copy_elision

Lloyd-Pottiger: yeah, I know this is be optimized since c++11. But we can not guarantees there will be no copy at all? Copy a unique_ptr cost much less than copy a ReadBuffer.

And as a lib function, return unique_ptr bring more convince. cast unique_ptr<ReadBuffer> to ReadBuffer is easy, but we can not cast ReadBuffer to unique_ptr<ReadBuffer> directly.

CalvinNeo: > I know this is be optimized since c++11. But we can not guarantees there will be no copy at all?
Yes, we can. In fact the compiler MUST do all the RVO thing, since C++17.

> return unique_ptr bring more convince
Then return a not_null to reduce calling `if (xxx)` over and over again. Because it is hard to get the idea which unique_ptr is guaranteed to be not null.

CalvinNeo: Meanwhile, `ReadBuffer` will create a vtable, since it has some virtual functions. I wonder if it's better to avoid use any kind of pointer, if we know exactly what we are operate on?

CalvinNeo: Please check https://stackoverflow.com/questions/77925442/when-does-the-c-standard-permits-forces-the-compiler-to-devirtualization for the "use o->foo() rather than o.foo()" problem.

CalvinNeo (commented):
Conclusion:
1. Since C++, returning a **prvalue** of something like `WriteBufferFromWritableFile` is guaranteed to have no side effect. Feel free to do it.
2. Returning a pointer to ReadBuffer which has vtable is observed to have no side-effect. The virtual call is actually been wiped out due to devirtualization. The devirtualization is never specified by the standard, however, in the compilers we known(clang13/15/17/20 from O2 from C++17), it does happen.

The real problem here, is `unique_ptr<T>` does not really free us from potential copy of `ReadBuffer` and all its subclasses. So IMO, I'd like keep it as it is now, before we solve the problem.

CalvinNeo (approved):
lgtm

JaySon-Huang:
> move checksum related code into Encryption directory

I think we should keep the checksum related code in `IO` directory and move some class defined in `Encryption` to the `IO` directory?

Lloyd-Pottiger:
> > move checksum related code into Encryption directory
> 
> I think we should keep the checksum related code in `IO` directory and move some class defined in `Encryption` to the `IO` directory?

the work path is `compression --> checksum --> encryption`, so the directory relationship should be:

```
Compression/
  Encryption/
    IO/
      BaseFile/ 
```

So checksum should not be under `IO/`, a new directory `Checksum/` is an option.

And FileProvider should put under `Compression/`, but we do not have `Compression/` now, so put in `Encryption/`. Next PR I will add `Compression/`.

JaySon-Huang:
The directory hierarchy should not be organized based on how the program actually runs, but rather on how to layer the code conceptually. The `IO` provide some classes/functions as an abstraction for writing/reading file, including compression, checksum and an abstract layer of underlying filesystem. And `Encryption` provide some drop-in classes/functions that implement the details encrypted stuff.

From the implementation level, @lidezhu split those files into `Encryption` diectory but not placed them into `IO` because those files need to link with `tiflash-proxy`. And we don't want to link the `common_io` with `tiflash-proxy`. However, in that time, the include header of some files are not refined and we misplaced some files under `Encryption` but not `IO`.

I suggest the directory hierarchy as follows
```
IO/Compress/

CompressedReadBufferFromFileProvider.cpp
CompressedReadBufferFromFileProvider.h

IO/

FileProvider.cpp
FileProvider_fwd.h
FileProvider.h
createReadBufferFromFileBaseByFileProvider.cpp
createReadBufferFromFileBaseByFileProvider.h
createWriteBufferFromFileBaseByFileProvider.cpp
createWriteBufferFromFileBaseByFileProvider.h
ReadBufferFromFileProvider.cpp
ReadBufferFromFileProvider.h
WriteBufferFromFileProvider.cpp
WriteBufferFromFileProvider.h
KeyManager.h // an abstract class as a concept that we need to add in IO/FileProvider for Encryption. The implementation "DataKeyManager" is placed under "Encryption"
EncryptionPath.h // 

Encryption/

AESCTRCipher.cpp
AESCTRCipher.h
AESCTRCipherStream.cpp
AESCTRCipherStream.h
BlockAccessCipherStream.h
DataKeyManager.cpp
DataKeyManager.h
EncryptedRandomAccessFile.cpp
EncryptedRandomAccessFile.h
EncryptedWritableFile.cpp
EncryptedWritableFile.h
EncryptedWriteReadableFile.cpp
EncryptedWriteReadableFile.h
EncryptionKey.h
KeyspacesKeyManager.cpp
KeyspacesKeyManager.h
MasterKey.h
MockKeyManager.cpp
MockKeyManager.h
```

Lloyd-Pottiger:
How about this? @JaySon-Huang @JinheLin 

```
IO/Encryption/
├── AESCTRCipher.cpp
├── AESCTRCipher.h
├── AESCTRCipherStream.cpp
├── AESCTRCipherStream.h
├── BlockAccessCipherStream.h
├── DataKeyManager.cpp
├── DataKeyManager.h
├── EncryptedRandomAccessFile.cpp
├── EncryptedRandomAccessFile.h
├── EncryptedWritableFile.cpp
├── EncryptedWritableFile.h
├── EncryptedWriteReadableFile.cpp
├── EncryptedWriteReadableFile.h
├── EncryptionKey.h
├── KeyspacesKeyManager.cpp
├── KeyspacesKeyManager.h
├── MasterKey.h
├── MockKeyManager.cpp
├── MockKeyManager.h
├── tests

IO/Checksum/
├── ChecksumBuffer.cpp
├── ChecksumBuffer.h
├── ChecksumReadBufferBuilder.cpp
├── ChecksumReadBufferBuilder.h
├── ChecksumWriteBufferBuilder.cpp
├── ChecksumWriteBufferBuilder.h

IO/Compression
├── CompressedReadBufferFromFileProvider.cpp
├── CompressedReadBufferFromFileProvider.h
├── CompressedReadBufferBase.cpp
├── CompressedReadBufferBase.h
├── CompressedReadBuffer.cpp
├── CompressedReadBufferFromFile.cpp
├── CompressedReadBufferFromFile.h
├── CompressedReadBuffer.h
├── CompressedStream.h
├── CompressedWriteBuffer.cpp
├── CompressedWriteBuffer.h
├── CompressionSettings.cpp
├── CompressionSettings.h
├── .....

IO/BaseBuffer
├── ReadBufferAIO.cpp
├── ReadBufferAIO.h
├── ReadBuffer.cpp
├── ReadBufferFromFileBase.cpp
├── ReadBufferFromFileBase.h
├── ReadBufferFromFile.cpp
├── ReadBufferFromFileDescriptor.cpp
├── ReadBufferFromFileDescriptor.h
├── ReadBufferFromFile.h
├── ReadBufferFromIStream.h
├── ReadBufferFromMemory.h
├── ReadBufferFromPocoSocket.cpp
├── ReadBufferFromPocoSocket.h
├── ReadBufferFromRandomAccessFile.cpp
├── ReadBufferFromRandomAccessFile.h
├── ReadBufferFromString.h
├── ReadBuffer.h
├── WriteBufferAIO.cpp
├── WriteBufferAIO.h
├── WriteBufferFromArena.h
├── WriteBufferFromFileBase.cpp
├── WriteBufferFromFileBase.h
├── WriteBufferFromFile.cpp
├── WriteBufferFromFileDescriptor.cpp
├── WriteBufferFromFileDescriptor.h
├── WriteBufferFromFile.h
├── WriteBufferFromOStream.h
├── WriteBufferFromPocoSocket.cpp
├── WriteBufferFromPocoSocket.h
├── WriteBufferFromString.h
├── WriteBufferFromTemporaryFile.cpp
├── WriteBufferFromTemporaryFile.h
├── WriteBufferFromVector.h
├── WriteBufferFromWritableFile.cpp
├── WriteBufferFromWritableFile.h
├── WriteBuffer.h
├── ...

IO/
├── WriteBufferFromWritableFileBuilder.cpp
├── WriteBufferFromWritableFileBuilder.h
├── ReadBufferFromRandomAccessFileBuilder.cpp
├── ReadBufferFromRandomAccessFileBuilder.h
├── EncryptionPath.h
├── FileProvider.cpp
├── FileProvider_fwd.h
├── FileProvider.h
├── KeyManager.h

```

JaySon-Huang:
@Lloyd-Pottiger LGTM

Lloyd-Pottiger:
@JaySon-Huang I have moved checksum related files into `IO/Checksum/`, other changes leave it for next PR, otherwise this PR is too large.

JaySon-Huang (approved):

Issue #8710, opened 2024-01-20
Data may be dropped silently after replica is set to 0 but peer not removed


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Load some data into a table 
Add tiflash replica for the table
Set tiflash replica to zero, sometime PD didn't remove (all) related region peers from tiflash
Wait tiflash replica data be GCed
Add tiflash replica back

### 4. What is your TiFlash version? (Required)
nightly, v7.5.0

PR #8721, opened 2024-01-22
ddl: Fix the storage instance may be physically dropped when the region is not removed

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8710

Problem Summary:
There are some case that a region may not be removed from TiFlash when there are unhealth peers. However, after the tiflash replica is set to zero, the data still be physically dropped after gc safepoint exceed.

* Prevent the storage instance being physically dropped when the region is not removed
* Throw exception when the storage instance is not exist after schema synced

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that the tiflash replica data may be corrupted after setting the tiflash replica to 0 and add it back later
```


Lloyd-Pottiger (approved):

hongyunyan (approved):
LGTM

JaySon-Huang:
This PR also prevent data being unexpected dropped when the region is not removed like https://github.com/pingcap/tiflash/issues/8777

Issue #8736, opened 2024-01-29
 Support Pipelined DML


JinheLin:
- [ ] Support spill default cf to disk
- [ ] Support spill lock cf to disk
- [ ] Bypass the long-run resolve locks after pipelined DML to serve reading requests

PR #8739, opened 2024-01-31
KVStore: share some utils with Debug and refine definitions in KVStore

CalvinNeo:

Issue Number: ref #8736

Problem Summary:
1. Use `auto & ctx` rather than `auto ctx` in some tests.
1. Refine KVStore.h
1. Make some utils in region_helper.h to `Debug`
1. Move orphan key handling aside

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Debug/MockKVStore/MockUtils.h
```diff
@@ -0,0 +1,100 @@
...
+#pragma once
+
+#include <Storages/KVStore/MultiRaft/RegionMeta.h>
+#include <Storages/KVStore/MultiRaft/RegionPersister.h>
+#include <Storages/KVStore/TiKVHelpers/TiKVRecordFormat.h>
+
+namespace DB
+{
+namespace RegionBench
+{
```
Lloyd-Pottiger, line 24:
```suggestion
namespace DB:: RegionBench
{

```

dbms/src/Debug/MockKVStore/MockUtils.h
```diff
@@ -0,0 +1,100 @@
...
+            initialApplyState()),
+        proxy_helper);
+}
+
+inline RegionPtr makeRegion(RegionMeta && meta)
+{
+    return std::make_shared<Region>(std::move(meta), nullptr);
+}
+} // namespace RegionBench
+} // namespace DB
```
Lloyd-Pottiger, line 100:
```suggestion

} // namespace DB:: RegionBench

```

Lloyd-Pottiger (approved):
Others LGTM~

dbms/src/Storages/KVStore/tests/gtest_region_persister.cpp
```diff
@@ -199,7 +203,7 @@ TEST_F(RegionSeriTest, RegionOldFormatVersion)
 try
 {
     TableID table_id = 100;
-    auto region = std::make_shared<Region>(createRegionMeta(1001, table_id));
+    auto region = makeTmpRegion();
```
JaySon-Huang, line 206: The `makeTmpRegion()` create a region with `table_id == 1` but not the `table_id` here. It may cause test failure?

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -381,9 +393,9 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
         case WriteCmdType::Del:
         {
             auto tikv_key = TiKVKey(cmds.keys[i].data, cmds.keys[i].len);
-            if (unlikely(cf == ColumnFamilyType::Write))
+            if (unlikely(cf == ColumnFamilyType::Lock))
             {
-                del_key_count++;
+                lock_del_key_count++;
             }
```
JaySon-Huang, line 399: Why counting the `lock_del_key` is better than `write_del_key`?

CalvinNeo: `write_del_key` is almost always empty, because everytime a write key is written from a raft normal write command, it will immediately trigger `RegionTable::writeBlockByRegion` which will remove the write key in `RemoveRegionCommitCache`.
So briefly speaking, the "put write cf key" and "del write key" can't be observed in the situation above.
However, there are actually some cases where a "del write key" to its corresponding "put write cf key". One of this case is applying snapshot in which we will leaving some write cf keys if we get trouble in row2col.

dbms/src/Storages/KVStore/tests/gtest_region_persister.cpp
CalvinNeo: No, in this test suite, we only focus on KVStore, so we have no schema.

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
CalvinNeo: Either have `write_del_key` or not seems good to me. How do you think?

dbms/src/Storages/KVStore/tests/gtest_region_persister.cpp
JaySon-Huang: OK

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
JaySon-Huang: I get it. I'd like to keep this `write_del_key`

JaySon-Huang (approved):
lgtm

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
CalvinNeo: Ok, we will add it back

CalvinNeo:
Could not perform submodule update

PR #8753, opened 2024-02-02
Storages: Refine comparator of row key to operator<=>

JinheLin:

Issue Number: ref #6233

- Updating the comparsion code of RowKeyValueRef/RowKeyValue with operators that makes code more readable. 
- Rename comparator of `RowKeyValueRef` from `compare` to `operator<=>`, so that we can use operators >, <, >=, <= directly.
- Simply wrap `operator==` and `operator!=` with `operator<=>` because complier does not generate these functions when `operator<=>` if not default.
- Remove some unused overload operators.
- Refine comparator of `RowKeyValue`. Make it use the same comparator with `RowKeyValueRef`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/RowKeyRange.h
+++ b/dbms/src/Storages/DeltaMerge/RowKeyRange.h
@@ -170,11 +170,6 @@ struct RowKeyValue
         return std::make_shared<DecodedTiKVKey>(prefix + *value);
     }
 
-    bool operator==(const RowKeyValue & v) const
-    {
-        return is_common_handle == v.is_common_handle && (*value) == (*v.value) && int_value == v.int_value;
-    }
-
     /**
      * Returns the key so that the range [this, this.toPrefixNext()) contains
      * all keys with the prefix `this`.
@@ -240,14 +235,14 @@ struct RowKeyValue
         return RowKeyValue(is_common_handle, start_ptr);
     }
 
-    bool is_common_handle;
+    bool is_common_handle = false;
     /// In case of non common handle, the value field is redundant in most cases, except that int_value == Int64::max_value,
     /// because RowKeyValue is an end point of RowKeyRange, assuming that RowKeyRange = [start_value, end_value), since the
     /// end_value of RowKeyRange is always exclusive, if we want to construct a RowKeyRange that include Int64::max_value,
     /// just set end_value.int_value to Int64::max_value is not enough, we still need to set end_value.value a carefully
     /// designed value. You can refer to INT_HANDLE_MAX_KEY for more details
     HandleValuePtr value;
-    Int64 int_value;
+    Int64 int_value = 0;
 
     static const RowKeyValue INT_HANDLE_MIN_KEY;
     static const RowKeyValue INT_HANDLE_MAX_KEY;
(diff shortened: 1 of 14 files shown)


dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -307,76 +302,48 @@ inline int compare(const RowKeyValueRef & a, const RowKeyValueRef & b)
...
-inline int compare(Int64 a, const RowKeyValueRef & b)
+inline int operator<=>(Int64 a, const RowKeyValueRef & b)
 {
     RowKeyValueRef r_a{false, nullptr, 0, a};
-    return compare(r_a, b);
+    return r_a <=> b;
 }
 
-inline int compare(const RowKeyValueRef & a, Int64 b)
+inline int operator<=>(const RowKeyValueRef & a, Int64 b)
```
CalvinNeo, line 333: amazing

JaySon-Huang: 😳 Does clang 13 in our CI already support this operator? I though it was only supported since clang 15 or so.

CalvinNeo (approved):
lgtm

dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -260,7 +255,7 @@ using RowKeyValues = std::vector<RowKeyValue>;
 
 /// An optimized implementation that will try to compare IntHandle via comparing Int values directly.
 /// For common handles, per-byte comparison will be still used.
-inline int compare(const RowKeyValueRef & a, const RowKeyValueRef & b)
+inline int operator<=>(const RowKeyValueRef & a, const RowKeyValueRef & b)
```
JaySon-Huang, line 258: Should we make this return type as [`std::strong_ordering`](https://en.cppreference.com/w/cpp/utility/compare/strong_ordering)?

dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -170,11 +182,6 @@ struct RowKeyValue
         return std::make_shared<DecodedTiKVKey>(prefix + *value);
     }
 
-    bool operator==(const RowKeyValue & v) const
-    {
-        return is_common_handle == v.is_common_handle && (*value) == (*v.value) && int_value == v.int_value;
-    }
-
```
JaySon-Huang, line 177: Maybe recover these lines and remove `inline bool operator==(const RowKeyValue & a, const RowKeyValue & b)`

dbms/src/Storages/DeltaMerge/RowKeyRange.h
```diff
@@ -33,13 +33,25 @@ using HandleValuePtr = std::shared_ptr<String>;
 struct RowKeyRange;
 using RowKeyRanges = std::vector<RowKeyRange>;
 
-inline int compare(const char * a, size_t a_size, const char * b, size_t b_size)
+namespace
```
JaySon-Huang, line 36: Better to define an explicit name for this namespace because it is in a `.h` but not `.cpp` file. Maybe `CommonHandle`

JaySon-Huang (approved):
LGTM

PR #8756, opened 2024-02-05
Storages: Make DMFile ready for new column indexes/types

JaySon-Huang:

Issue Number: ref https://github.com/pingcap/tiflash/issues/6233, close https://github.com/pingcap/tiflash/issues/8768

Problem Summary:

In the near future, we may add new types (vector type) or new indexes. But currently the ColumnStat is not flexible enough for extending new fields for the index size, array size, etc.

* Add new type `ExtendColumnStat` meta block under DMFile meta v2, which accept protobuf base ColumnStat
* Add array_sizes_bytes/array_sizes_mark_bytes for `ColumnStat` to support vec type
* Try to merge array_size(`.size0.dat`) into X.merged file

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/File/DMFile.cpp
```diff
@@ -920,10 +938,16 @@ DMFile::MetaBlockHandle DMFile::writeMergedSubFilePosotionsToBuffer(WriteBuffer
...
-    std::array meta_block_handles = {
+    std::array meta_block_handles = { //
         writeSLPackStatToBuffer(tmp_buffer),
         writeSLPackPropertyToBuffer(tmp_buffer),
+#if 1
         writeColumnStatToBuffer(tmp_buffer),
+#else
+        // ExtendColumnStat is not enabled yet because it cause downgrade compatibility, wait
+        // to be released with other binary format changes.
+        writeExtendColumnStatToBuffer(tmp_buffer),
```
breezewish, line 949: How about we write the data at the end (and keep old format)? In this way it may be possible to keep downgrade compatibility as new format will be regarded as extra data and will be discarded.

JaySon-Huang: Good suggestion! The ColumnStat is small enough to ignore its impact on performance and data size. We can keep it until we actually need an incompatible storage format changed by other components

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
```diff
@@ -1207,15 +1251,21 @@ void DMFile::finalizeSmallFiles(
...
-            {
-                auto fname = colDataFileName(getFileNameBase(col_id, {IDataType::Substream::NullMap}));
-                auto fsize = stat.nullmap_data_bytes;
-                copy_file_to_cur(fname, fsize);
-                delete_file_name.push_back(fname);
-            }
+            auto fname = colDataFileName(getFileNameBase(col_id, {IDataType::Substream::NullMap}));
+            auto fsize = stat.nullmap_data_bytes;
+            copy_file_to_cur(fname, fsize);
+            delete_file_name.push_back(fname);
```
Lloyd-Pottiger, line 1259:
```suggestion
            delete_file_name.emplace_back(std::move(fname));
```

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
```diff
@@ -1207,15 +1251,21 @@ void DMFile::finalizeSmallFiles(
...
+        }
+
+        // check .size0.dat
+        if (stat.array_sizes_bytes > 0 && stat.array_sizes_bytes <= small_file_size_threshold)
+        {
+            auto fname = colDataFileName(getFileNameBase(col_id, {IDataType::Substream::ArraySizes}));
+            auto fsize = stat.array_sizes_bytes;
+            copy_file_to_cur(fname, fsize);
+            delete_file_name.push_back(fname);
```
Lloyd-Pottiger, line 1268:
```suggestion
            delete_file_name.emplace_back(std::move(fname));
```

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
```diff
@@ -540,31 +541,44 @@ void DMFileReader::readFromDisk(
...
-    }
+    auto iter = column_streams.find(stream_name);
+#ifndef NDEBUG
+    RUNTIME_CHECK_MSG(
+        iter != column_streams.end(),
+        "Can not find column_stream, column_id={} stream_name={}",
+        column_define.id,
+        stream_name);
+
+#endif
```
Lloyd-Pottiger, line 552: Why not just use `assert`?

dbms/src/Storages/DeltaMerge/File/DMFileWriter.cpp
```diff
@@ -128,6 +129,11 @@ void DMFileWriter::addStreams(ColId col_id, DataTypePtr type, bool do_index)
 
 void DMFileWriter::write(const Block & block, const BlockProperty & block_property)
 {
+#ifndef NDEBUG
+    // In the prod env, the #rows is checked in upper level
+    block.checkNumberOfRows();
+#endif
```
Lloyd-Pottiger, line 135: ditto

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
JaySon-Huang: After some tests, I found the downgrade compatibility can not be satisfied because there is a default branch 😂

https://github.com/pingcap/tiflash/blob/d9dc96bdf7f062ed79459f21143e09e731b4e1ef/dbms/src/Storages/DeltaMerge/File/DMFile.cpp#L1055-L1059

dbms/src/Storages/DeltaMerge/File/DMFileWriter.cpp
JaySon-Huang: assert can not be used in here because `block.checkNumberOfRows()` return void

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
JaySon-Huang: assert in here don't give any information about the column, using RUNTIME_CHECK helps us locate the bug more quickly under tests.

dbms/src/Storages/DeltaMerge/File/DMFile.h
```diff
@@ -367,7 +368,7 @@ class DMFile : private boost::noncopyable
         return Poco::File(colDataPath(file_name_base)).getSize();
     }
     size_t colIndexSize(ColId id);
-    size_t colDataSize(ColId id, bool is_null_map);
+    size_t colDataSize(ColId id, bool is_null_map, bool is_array_map);
```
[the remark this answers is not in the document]
JaySon-Huang: Addressed

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
```diff
@@ -247,15 +247,15 @@ size_t DMFile::colIndexSize(ColId id)
...
+size_t DMFile::colDataSize(ColId id, ColDataType type)
 {
     if (useMetaV2())
     {
         if (auto itr = column_stats.find(id); itr != column_stats.end())
         {
-            return is_null_map ? itr->second.nullmap_data_bytes
-                : is_array_map ? itr->second.array_sizes_bytes
-                               : itr->second.data_bytes;
+            return type == ColDataType::NullMap   ? itr->second.nullmap_data_bytes
```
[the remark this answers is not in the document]
JaySon-Huang: Addressed

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
```diff
@@ -1089,7 +1089,10 @@ void DMFile::parseExtendColumnStat(std::string_view buffer)
         ColumnStat stat;
         stat.mergeFromProto(msg);
         // replace the ColumnStat if exists
-        column_stats.emplace(stat.col_id, stat).first->second = stat;
+        if (auto [iter, inserted] = column_stats.emplace(stat.col_id, stat); unlikely(!inserted))
```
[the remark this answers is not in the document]
JaySon-Huang: This is a small optimization - if col_id is not exist, construct the `<col_id, stat>` pair in place. `columns_stats[stat.col_id] = stat` always create an `<col_id, empty_stat>` and then copy the stat to replace the empty_stat.

CalvinNeo (approved):
lgtm

Lloyd-Pottiger (approved):

dbms/src/Storages/DeltaMerge/File/DMFile.cpp
breezewish: Good idea. What kind of downgrade compatibility do we need to have? How about we first land the code that will not throw error (and will not change the format at all) for a few major versions, and then land the compatible format? In this way we will have downgrade compatibility for 1 major version.

JaySon-Huang: There is only downgrade compatibility requirement between patch versions under the same minor version. And no requirement across minor or major versions. But it is good to have.

I've removed the exception thrown branch in this PR because the correctness of bytes read is protected by checksum. I think ignoring unknown meta block types without exception or logging is an acceptable behavior.

@JinheLin @breezewish

Issue #8777, opened 2024-02-21
After upgrade from 5.4 to 7.5, my TiFlash data gone after the TiKV node disconnected


uzuki27:
## Bug Report

I’m facing an issue with TiFlash after upgrade my TiDB cluster from v5.4 to v7.5.
A days after the upgrade, I realized that the data became out of sync between TiKV and TiFlash. The query via TiFlash returned almost empty data. When show tiflash_replica, all of my table progress is back to 0 or something not 1 but avaiable still 1.

```
+----------+---------------+-----------+----------+
| TABLE_ID | REPLICA_COUNT | AVAILABLE | PROGRESS |
+----------+---------------+-----------+----------+
|   460166 |             2 |         1 |      0.5 |
|   396090 |             2 |         1 |        0 |
|   396886 |             2 |         1 |        0 |
|   397221 |             2 |         1 |        0 |
|   397295 |             2 |         1 |        0 |
+----------+---------------+-----------+----------+
```

After that, I decided to set all replica to 0, scale-in all 2 of my TiFlash node, waiting end of Pending Offline, prune its and scale-out them into cluster again. After set replica to 2 for my tables again, the replication is run normally. After replication of all tables completed, my query on TiFlash is on-sync again. But after about 1 days, the issue occurred again I did more 2 times scale-in scale-out, the issue still hasn’t been resolved.

### 1. Minimal reproduce step (Required)
After several attempts, I realized they all shared the same scenario. After all of tables had completed the replication about 1 days. 1 of TiKV node evicted all leader region and then rebalance again
![image](https://github.com/pingcap/tiflash/assets/54520021/ecb3ca22-59d7-47ae-bf3a-5467e3c5ab48)

The evicted TiKV node log raised some ERROR like this
```
[2024/02/21 08:37:27.278 +07:00] [ERROR] [peer.rs:5351] ["failed to send extra message"] [err_code=KV:Raftstore:Transport] [err=Transport(Full)] [target="id: 1020098876 store_id: 5"] [peer_id=1020785306] [region_id=1020098875] [type=MsgHibernateResponse] [thread_id=0x5]
[2024/02/21 08:37:27.280 +07:00] [ERROR] [peer.rs:5351] ["failed to send extra message"] [err_code=KV:Raftstore:Transport] [err=Transport(Full)] [target="id: 1065360228 store_id: 5"] [peer_id=977883503] [region_id=977883502] [type=MsgHibernateRequest] [thread_id=0x5]
```
[...]

PR #7437, opened 2023-05-08
Refactor DDL Module to support a large number of tables with frequent ddl actions

hongyunyan:
### What problem does this PR solve?

Issue Number: ref #7630 close #6532 #3777 #1664

Problem Summary:

Refactor DDL module to support a large number of tables with frequent ddl actions.

related PR:https://github.com/tikv/client-c/pull/146 

### What is changed and how it works?
1. new DDL architecture, basiclly including TiDB/Schema/xxx.
2. change to use new ddl architecture in the main logical, mainly includes StorageDeltaMerge.cpp, DeltaMergeStore.cpp
3. update and add a lot of tests for new DDL architecture 

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
Shorten TiFlash restart time, and be more efficient under multiple ddl
```

PR #8785, opened 2024-02-23
ddl: Fix database drop by accident when meet `regenerate_schema_map` flag

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8777

Problem Summary:

Normally, after users executed DDL statements, the TiDB server will generate some "SchemaDiff" to indicate what happens. And TiFlash will follow the "SchemaDiff" to figure out what DDL operations need to be executed.
But when there are special DDL `FLASHBACK CLUSTER TO` or `br restore point` (which generate `regenerate_schema_map == true`), TiFlash will try to list all existing databases and tables in the cluster and enter `syncAllSchema` to check all existing databases and tables one by one and execute proper DDL on them.

However, in the DDL framwork refactor since v7.2(https://github.com/pingcap/tiflash/pull/7437), there is a mistake in the implementation. In the following codes, only newly created database that does not exist in TiFlash local are put in `created_db_set`. In line 1383, if the database is not existed in `created_db_set`, then the database will be marked as tombstone unexpectedly. And after the gc_safepoint advance, the data on TiFlash replica get physically removed unexpectedly. :(

https://github.com/pingcap/tiflash/blob/fe6621befaefdfd03a6fa5c10318b7e6583fa3d2/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L1269-L1391

Though `syncAllSchema` will also run right after TiFlash restarts, but at that time, the `DatabaseInfoCache` is empty, so it do not trigger this bug.

Add the database name to set no matter the database is already present locally or not

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that TiFlash replica data dropped unexpectedly after PITR restore or `FLASHBACK CLUSTER TO` is executed
```

### DIFF
--- a/dbms/src/TiDB/Schema/SchemaBuilder.cpp
+++ b/dbms/src/TiDB/Schema/SchemaBuilder.cpp
@@ -1326,7 +1326,6 @@ public:
 /// syncAllSchema will be called when
 /// - a new keyspace is created
 /// - meet diff->regenerate_schema_map = true
-/// - exception thrown in applyDiff
 /// Thus, we should not assume all the map is empty during syncAllSchema.
 template <typename Getter, typename NameMapper>
 void SchemaBuilder<Getter, NameMapper>::syncAllSchema()
--- a/dbms/src/TiDB/Schema/SchemaBuilder.cpp
+++ b/dbms/src/TiDB/Schema/SchemaBuilder.cpp
@@ -52,6 +52,7 @@
 #include <mutex>
 #include <string_view>
 #include <tuple>
+#include <unordered_set>
 
 namespace DB
 {
@@ -1304,6 +1305,24 @@ void SchemaBuilder<Getter, NameMapper>::applyDropTable(
     applyDropPhysicalTable(name_mapper.mapDatabaseName(database_id, keyspace_id), table_info.id, action);
 }
 
+class SyncedDatabaseSet
+{
+private:
+    std::unordered_set<String> nameset;
+    std::mutex mtx;
+
+public:
+    // Thread-safe emplace
+    void emplace(const String & name)
+    {
+        std::unique_lock lock(mtx);
+        nameset.emplace(name);
+    }
+
+    // Non thread-safe check.
+    bool nonThreadSafeContains(const String & name) const { return nameset.contains(name); }
+};
+
 /// syncAllSchema will be called when
 /// - a new keyspace is created
 /// - meet diff->regenerate_schema_map = true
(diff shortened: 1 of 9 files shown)


JaySon-Huang:
Though `syncAllSchema` will also run after TiFlash restarts, but at that time, the `DatabaseInfoCache` is empty, so it do not trigger this bug.

Lloyd-Pottiger:
You mean

```sql
# t0
create table t (xxx);
insert into t xxx;
# t1
xxx
# t2
# table t is still not replica to TiFlash
flash back to t1
```
And then t1 is dropped accidentally?

Lloyd-Pottiger (approved):
lgtm

Does this mean when meets `diff->regenerate_schema_map = true`, all databases in Database info cache will be dropped in previous implement?

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -1320,23 +1341,21 @@ void SchemaBuilder<Getter, NameMapper>::syncAllSchema()
         = ThreadPool(default_num_threads, default_num_threads / 2, default_num_threads * 2);
     auto sync_all_schema_wait_group = sync_all_schema_thread_pool.waitGroup();
 
-    std::mutex created_db_set_mutex;
-    std::unordered_set<String> created_db_set;
+    SyncedDatabaseSet latest_db_nameset;
```
Lloyd-Pottiger, line 1344:
```suggestion
    SyncedDatabaseSet latest_db_nameset;
    databases.clear();
```

Maybe just clear cache here?

JaySon-Huang:
> Does this mean when meets diff->regenerate_schema_map = true, all databases in Database info cache will be dropped in previous implement?

Yes. The v7.1 and oldest versions implement it correctly. 
https://github.com/pingcap/tiflash/blob/b7b6d006cf59a792c7d99dfba154681dc0d7a9d4/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L1361-L1470
The implementation in current master/7.5 branch is a mistake after DDL framework refactor since v7.2.

dbms/src/TiDB/Schema/SchemaBuilder.cpp
JaySon-Huang: I think clear cache and let those existing database re-created by the loop is hard to understand.

Just let the `nameset` record all the database name of database info fetch from tikv, and dropped the local database that is not exists in the nameset is more easy to understand.

PR #8792, opened 2024-02-27
fix anti semi join 

### LINKED ISSUE #8791: anti semi join with other condition may get wrong results
## Bug Report

During the join probe stage, in order to make sure that Intermediate block's size is not bigger than `max_block_size`, one probe block maybe probed multiple times, each time only [probe_process_info.start_row, probe_process_info.end_row) rows is processed. If the join has other condition, after probe the hash table, it need to call [`handleOtherConditions`](https://github.com/pingcap/tiflash/blob/v7.5.0/dbms/src/Interpreters/Join.cpp#L826) to further evaluate the other condition. In handleOtherConditions, the input argument `anti_filter` and `offsets_to_replicate` is expected to be a vector of size `probe_process_info.end_row - probe_process_info.start_row`, however, for anti-semi join, the caller only cut `offsets_to_replicate` to the correct size(https://github.com/pingcap/tiflash/blob/v7.5.0/dbms/src/Interpreters/Join.cpp#L1297-L1307), for `filter`, it is not handled, which will make `handleOtherConditions` return wrong result in some cases. 

This bug only affect v6.6 ~v7.5

windtalker:

Issue Number: close #8791

Problem Summary:

Ref #8791 for details

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


yibin87 (approved):
LGTM

dbms/src/Interpreters/Join.cpp
```diff
@@ -1304,6 +1304,9 @@ Block Join::doJoinBlockHash(ProbeProcessInfo & probe_process_info) const
                 offsets_to_replicate->assign(
                     offsets_to_replicate->begin() + probe_process_info.start_row,
                     offsets_to_replicate->begin() + probe_process_info.end_row);
+                if (isAntiJoin(kind) && filter != nullptr)
+                    filter->assign(filter->begin() + probe_process_info.start_row,
```
gengliqi, line 1308: Better not to use `assign` because it uses `memcpy` which has an undefined behavior if two memory locations are overlapped.

windtalker: I aggree, but since it is close to the release deadline, I suggest we keep `assign` in this pr and fix it later since `assign` is used many places, it will bring extra risk if we change too many in this pr.

gengliqi (approved):

JaySon-Huang:
I think you should update the release note to describe what bug has been fixed

windtalker:
> I think you should update the release note to describe what bug has been fixed

Done

Issue #8837, opened 2024-03-12
Disaggregated TiFlash crash when the process is shutting down


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

stale read and chaos
```
[2024/03/11 20:29:49.078 +08:00] [ERROR] [BaseDaemon.cpp:358] ["(from thread 48) Terminate called without an active exception
"] [source=BaseDaemon] [thread_id=552]
[2024/03/11 20:29:49.078 +08:00] [ERROR] [BaseDaemon.cpp:367] [########################################] [source=BaseDaemon] [thread_id=552]
[2024/03/11 20:29:49.078 +08:00] [ERROR] [BaseDaemon.cpp:368] ["(from thread 48) Received signal Aborted(6)."] [source=BaseDaemon] [thread_id=552]
[2024/03/11 20:29:49.080 +08:00] [ERROR] [BaseDaemon.cpp:560] ["
       0x77c17c1    faultSignalHandler(int, siginfo_t*, void*) [tiflash+125573057]
                    libs/libdaemon/src/BaseDaemon.cpp:211
  0x7fe0caf9fdb0    <unknown symbol> [libc.so.6+347568]
  0x7fe0cafec54c    __pthread_kill_implementation [libc.so.6+660812]
  0x7fe0caf9fd06    __GI_raise [libc.so.6+347398]
  0x7fe0caf737f3    abort [libc.so.6+165875]
       0x77c0eb4    terminate_handler() [tiflash+125570740]
                    libs/libdaemon/src/BaseDaemon.cpp:0
  0x7fe0cfbe4a13    std::__terminate(void (*)()) [libc++abi.so.1+236051]
  0x7fe0cfbe49a2    std::terminate() [libc++abi.so.1+235938]
  0x7fe0cfc5f8ce    std::__1::thread::~thread() [libc++.so.1+461006]
       0x809999c    DB::BackgroundProcessingPool::~BackgroundProcessingPool() [tiflash+134846876]
                    dbms/src/Storages/BackgroundProcessingPool.cpp:151
       0x874b644    DB::UniversalPageStorageService::~UniversalPageStorageService() [tiflash+141866564]
                    dbms/src/Storages/Page/V3/Universal/UniversalPageStorageService.cpp:297
[...]

PR #8848, opened 2024-03-15
disagg: Fix a potential crash when shutting down

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8837

Problem Summary:

There is a logging about "std::exception. Code: 1001, type: std::__1::system_error, e.what() = thread::join failed: Resource deadlock avoided". Usually it is caused by a thread is trying to call `join` for the thread itself.

There is a chance that the task handle (shared_ptr) is being held by the background pool thread
https://github.com/pingcap/tiflash/blob/0d13f398b566c7b06303747b117c3d3af0e5c779/dbms/src/Storages/BackgroundProcessingPool.cpp#L226-L228

And the task handle could hold a shared_ptr to `UniversalPageStorageService`. So there could be a chance that:

1. The main thread receive kill signal, `UniversalPageStorageService::shutdown` is called
2. But there is a background pool thread holding a shared_ptr of `remote_checkpoint_handle`, which contains a shared_ptr of `UniversalPageStorageService`
3. The shared_ptr of `UniversalPageStorageService` in the `ContextShared` is released
4. The background pool thread try to release to shared_ptr of `remote_checkpoint_handle`, which cause the shared_ptr of `UniversalPageStorageService` being released. But the thread is created by `UniversalPageStorageService::checkpoint_pool`. So the thread run into calling `join` on itself and causing the system_error

Use a `weak_ptr` of `UniversalPageStorageService` instead of `shared_ptr`. So that even when the `remote_checkpoint_handle` is still holding by the background pool thread, it won't cause the background pool thread calling `join` on itself. Because the share counter of `UniversalPageStorageService` must go down to 0 outside of the background pool thread.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a potential crash when shutting down under disagg arch
```


CalvinNeo (approved):
lgtm

CalvinNeo:
Meanwhile, I don't think it's a good idea to make a ThreadPool lives actually shorter than its tasks.
The holder should release the ThreadPool only after all its workers are idle.

PR #8857, opened 2024-03-21
Storages: Refine memory tracker of data sharing

### LINKED ISSUE #8856: Memory metric of data-sharing is incorrect
The default value of `dt_max_sharing_column_bytes_for_all` is 2GB https://github.com/pingcap/tiflash/blob/master/dbms/src/Interpreters/Settings.h#L222

But the metric in Grafana is beyond 10GB, the left part of the picture below:

image

JinheLin:

Issue Number: close #8856   

If `has_concurrent_reader` is true, we will consider the blocks to be read will being shared and set `current_memory_tracker` to null in https://github.com/pingcap/tiflash/blob/fdf8bf28ab86f21d2378f522d4e65bf5d0a8d422/dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp#L583-L606

But in reality, it may not true because there are also a `dt_max_sharing_column_count` limitations inside `ColumnSharingCacheMap`. This will cause `shared_column_data_mem_tracker` to count data that will not actually be shared.

This PR deprecated `dt_max_sharing_column_count` and only use `dt_max_sharing_column_bytes_for_all` as the memory limitation of data-sharing.

Only need to limit the sharing memory usage in `DMFileReader::readColumn`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  - Run `select * from lineitem` concurrently. 
  - The result is as the right part of the picture below
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang (approved):

JaySon-Huang:
Change the release note to "Fix a bug that the memory usage may not be correct when read operations of concurrent queries are merged"

Lloyd-Pottiger (approved):

Issue #7000, opened 2023-03-09
Split Context into multiples partials, and only change partials


breezewish:
## Enhancement

Currently we have a huge `Context`, which is imported in many `.cpp` files or even `.h` files. When we want to introduce something global, a large number of files (and nearly all test cases) need to be recompiled. It is very slow.

We can split Context into different partials, As the first step, we may split as:

- SharedContextDisagg: Global modules that related with disaggregated mode.
- SharedContextStorage: Global modules that related with storages, for example, PS instances, read & write limiters, etc.
- Context: Everything remains

TMTContext is untouched.

Hope this change could allow us changing `Context.h` much less frequently.

PR #8866, opened 2024-03-24
*: more forward decl

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/7000

Problem Summary:

Reduce the usage of these imports in headers by using forward decl as more as possible:

- Context.h
- DMContext.h
- ScanContext.h

Some class implementations were developed in .h files. They are now moved to .cpp files so that only forward decl is needed in .h files.

Before this PR, when introducing a new field to ScanContext, 140 files need to be recompiled. After this PR, only 35 will be re-compiled.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/SegmentReadTask.h
+++ b/dbms/src/Storages/DeltaMerge/SegmentReadTask.h
@@ -14,7 +14,7 @@
 
 #pragma once
 
-#include <Storages/DeltaMerge/DMContext.h>
+#include <Storages/DeltaMerge/DMContext_fwd.h>
 #include <Storages/DeltaMerge/Remote/DisaggTaskId.h>
 #include <Storages/DeltaMerge/Remote/Proto/remote.pb.h>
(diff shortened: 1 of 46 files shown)


zanmato1984 (approved):
LGTM

dbms/src/Storages/DeltaMerge/SegmentReadTask.h
```diff
@@ -115,6 +106,8 @@ struct SegmentReadTask
         return input_stream;
     }
 
+    String string() const;
```
JaySon-Huang, line 109: nit: `toString` maybe a more suitable name
```suggestion
    String toString() const;
```

JaySon-Huang (approved):
LGTM with a nit comment

hongyunyan (approved):
LGTM

Issue #7193, opened 2023-03-31
Clang 17 & C++23 Compatibility


breezewish:
## Enhancement

Clang 16.0.0 was officially released 2 weeks ago (March, 2023).

https://github.com/llvm/llvm-project/releases/tag/llvmorg-16.0.0

JaySon-Huang:
Change it to clang-17 that 17.0.4 was released on oct 31st

PR #8892, opened 2024-04-02
Static dispatch Arm-optimized-routines instead of dynamic dispatch

Lloyd-Pottiger:

Issue Number: ref #7193

Problem Summary:

hook syscall may make `getauxval` failed which may cause core dump.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


libs/libarm-optimized-routines/aor.h
```diff
@@ -0,0 +1,226 @@
...
+#elif __ARM_NEON
+    return __memcpy_aarch64_simd(dst_, src_, size);
+#endif
+    return __memcpy_aarch64(dst_, src_, size);
+#elif __arm__
+    return __memcpy_arm(dst_, src_, size);
+#else
+    return memcpy(dst_, src_, size);
+#endif
+}
```
JaySon-Huang, line 40: Can this file get the correct macro for `__ARM_FEATURE_SVE` / `__ARM_NEON`, etc?

JaySon-Huang: I'm not sure where these macro is defined. Are they defined in the arm system header?

Lloyd-Pottiger: Yes, I have verified.

image


They are introduced by flag, like "+simd" will add macro `__ARM_NEON`.

https://developer.arm.com/documentation/dui0774/l/Other-Compiler-specific-Features/Predefined-macros

JaySon-Huang (approved):
LGTM

PR #8906, opened 2024-04-08
do not mark spill if there is no data to spill in auto spill mode

### LINKED ISSUE #8905: Aggregation spill may stuck in infinite loop if an empty hash table is marked to spill
## Bug Report

#### Step 1 
In https://github.com/pingcap/tiflash/blob/9970e492dfdf2b4bee487fef4b27fee66131531f/dbms/src/Interpreters/AggSpillContext.cpp#L94-L139
Even if `per_thread_revocable_memories[index]` is 0(which mean the hash table is empty), it will be marked as `NEED_AUTO_SPILL`
#### Step 2 
In https://github.com/pingcap/tiflash/blob/9970e492dfdf2b4bee487fef4b27fee66131531f/dbms/src/Operators/AggregateBuildSinkOp.cpp#L20-L29
If there is no local data to build, it will first check `isTaskMarkedForSpill`
#### Step 3
In https://github.com/pingcap/tiflash/blob/9970e492dfdf2b4bee487fef4b27fee66131531f/dbms/src/Operators/AggregateContext.cpp#L57-L66
`isTaskMarkedForSpill` will call `updatePerThreadRevocableMemory`
#### Step 4
In https://github.com/pingcap/tiflash/blob/9970e492dfdf2b4bee487fef4b27fee66131531f/dbms/src/Interpreters/AggSpillContext.cpp#L50-L73
It will use cas to change `per_thread_auto_spill_status[thread_num]` to `WAIT_SPILL_FINISH` and return `true`
#### Step 5
In https://github.com/pingcap/tiflash/blob/9970e492dfdf2b4bee487fef4b27fee66131531f/dbms/src/Interpreters/Aggregator.cpp#L71-L75
If current hash table is empty, it will return `false`, which means no need to spill
#### Step 6
After the above 5 steps, now aggregation is expected to be in spill status from the view of `AggregationSpillContext`, but it actually is not in spill state from the view of `Aggregator`, in `Aggregator::executeOnBlock`, it will throw `ResizeException` but can not trigger spill because `updatePerThreadRevocableMemory` never returns true, so there is no change to call `tryMarkNeedSpill`. But`Aggregator::executeOnBlock` never aggregate data because once it need to resize hash table, it will meet `ResizeException`

windtalker:

Issue Number: close #8905

Problem Summary:

Don't mark spill if there is not data to spill in auto spill mode
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Interpreters/AggSpillContext.cpp
+++ b/dbms/src/Interpreters/AggSpillContext.cpp
@@ -52,6 +52,9 @@ bool AggSpillContext::updatePerThreadRevocableMemory(Int64 new_value, size_t thr
     if (!in_spillable_stage || !isSpillEnabled())
         return false;
     per_thread_revocable_memories[thread_num] = new_value;
+    if (new_value == 0)
+        // new_value == 0 means no agg data to spill
+        return false;
     if (auto_spill_mode)
     {
         AutoSpillStatus old_value = AutoSpillStatus::NEED_AUTO_SPILL;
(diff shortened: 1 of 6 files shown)


SeaRise (approved):

yibin87 (approved):
LGTM

Issue #8904, opened 2024-04-07
TiFlash result is wrong


lilinghai:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

prepare some data
do random dml and check `select count(*)` using tiflash and tikv
upgrade from 6.17 to master, when one tiflash version is 6.1.7 and another is master, query from tiflash is wrong
```
mysql> begin;
Query OK, 0 rows affected (0.04 sec)

mysql> set tidb_isolation_read_engines='tiflash';
Query OK, 0 rows affected (0.03 sec)

mysql> select count(*) from t;
+----------+
| count(*) |
+----------+
| 15284754 |
+----------+
1 row in set (0.17 sec)

mysql> set tidb_isolation_read_engines='tikv';
Query OK, 0 rows affected (0.03 sec)

mysql> select count(*) from t;
+----------+
| count(*) |
+----------+
| 16117328 |
+----------+
1 row in set (0.15 sec)

mysql> commit;
Query OK, 0 rows affected (0.04 sec)
```

JaySon-Huang:
Confirm it is introduced by https://github.com/pingcap/tiflash/pull/8909. Has been revert in the master

PR #8854, opened 2024-03-20
Storage: Refactor DMFileReader

Lloyd-Pottiger:
### What problem does this PR solve?

Issue Number: ref #6233

Problem Summary:

### What is changed and how it works?

Refactor DMFileReader to make life easier~

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #8925, opened 2024-04-10
Add insertMany interface for IColumn

Lloyd-Pottiger:

Issue Number: ref #8904

Problem Summary:

I used `insertManyFrom` in https://github.com/pingcap/tiflash/pull/8854, where the source column is a const column while the target column is not and it causes #8904. To fix this, add `insertMany` interface to avoid convert const column to full column.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Columns/ColumnVector.h
+++ b/dbms/src/Columns/ColumnVector.h
@@ -229,6 +229,11 @@ public:
             data[i + old_size] = src_container[position_vec[i]];
     }
 
+    void insertMany(const Field & field, size_t length) override
+    {
+        data.resize_fill(data.size() + length, static_cast<T>(field.get<T>()));
+    }
+
     void insertData(const char * pos, size_t /*length*/) override { data.push_back(*reinterpret_cast<const T *>(pos)); }
 
     bool decodeTiDBRowV2Datum(
@@ -243,7 +248,7 @@ public:
             {
                 throw Exception("Invalid float value length " + std::to_string(length), ErrorCodes::LOGICAL_ERROR);
             }
-            constexpr UInt64 SIGN_MASK = static_cast<UInt64>(1) << 63;
+            constexpr UInt64 SIGN_MASK = static_cast<UInt64>(1) << 63; // NOLINT(readability-identifier-naming)
             auto num = readBigEndian<UInt64>(raw_value.c_str() + cursor);
             if (num & SIGN_MASK)
                 num ^= SIGN_MASK;
--- a/dbms/src/Columns/IColumn.h
+++ b/dbms/src/Columns/IColumn.h
@@ -142,6 +142,13 @@ public:
     /// Appends disjunctive elements from other column with the same type.
     virtual void insertDisjunctFrom(const IColumn & src, const std::vector<size_t> & position_vec) = 0;
 
+    /// Appends one field multiple times. Can be optimized in inherited classes.
+    virtual void insertMany(const Field & field, size_t length)
+    {
+        for (size_t i = 0; i < length; ++i)
+            insert(field);
+    }
+
     /// Appends data located in specified memory chunk if it is possible (throws an exception if it cannot be implemented).
     /// Is used to optimize some computations (in aggregation, for example).
     /// Parameter length could be ignored if column values have fixed size.
--- a/dbms/src/Columns/IColumn.h
+++ b/dbms/src/Columns/IColumn.h
@@ -130,16 +130,20 @@ public:
 
     /// Appends n-th element from other column with the same type.
     /// Is used in merge-sort and merges. It could be implemented in inherited classes more optimally than default implementation.
+    /// Note: the source column and the destination column must be of the same type, can not ColumnXXX->insertFrom(ConstColumnXXX, ...)
     virtual void insertFrom(const IColumn & src, size_t n) { insert(src[n]); }
 
-    /// Appends range of elements from other column.
+    /// Appends range of elements from other column with the same type.
     /// Could be used to concatenate columns.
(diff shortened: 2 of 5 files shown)


dbms/src/Columns/IColumn.h
```diff
@@ -130,18 +130,29 @@ class IColumn : public COWPtr<IColumn>
...
     /// Appends one element from other column with the same type multiple times.
+    /// Note: the source column and the destination column must be of the same type, can not ColumnXXX->insertManyFrom(ConstColumnXXX, ...)
     virtual void insertManyFrom(const IColumn & src, size_t position, size_t length) = 0;
 
     /// Appends disjunctive elements from other column with the same type.
+    /// Note: the source column and the destination column must be of the same type, can not ColumnXXX->insertDisjunctFrom(ConstColumnXXX, ...)
     virtual void insertDisjunctFrom(const IColumn & src, const std::vector<size_t> & position_vec) = 0;
 
+    /// Appends one field multiple times. Can be optimized in inherited classes.
+    virtual void insertMany(const Field & field, size_t length)
```
yibin87, line 150: May reduce virtual function call like this:
https://github.com/pingcap/tiflash/blob/adc57e22149d04d76d26f5193134a05a938af7c7/dbms/src/Columns/IColumn.h#L476-L477

yibin87 (approved):
Others LGTM

dbms/src/Columns/IColumn.h
Lloyd-Pottiger: It should be called as `IColumn->insertMany(v)`

yibin87: Yeah, the insert virutal function is called inside the loop

Lloyd-Pottiger: compiler can help optimize this, no vtable lookup here https://gcc.godbolt.org/z/Y48rMoqqc

dbms/src/Columns/ColumnVector.h
```diff
@@ -229,6 +229,11 @@ class ColumnVector final : public COWPtrHelper<ColumnVectorHelper, ColumnVector<
             data[i + old_size] = src_container[position_vec[i]];
     }
 
+    void insertMany(const Field & field, size_t length) override
+    {
+        data.resize_fill(data.size() + length, static_cast<T>(field.get<T>()));
+    }
+
```
JaySon-Huang, line 236: /cc @breezewish vector may need to pick it back to the cse release-7.5 branch

dbms/src/Columns/IColumn.h
yibin87: I don't think so, think like this: different derived instances can call insertMany function, thus optimizer can't infer which insert function to be called next. Thus optimizer should lookup the vtable of current instance to find the actual insert function address. That's what "call    qword ptr [rax + 8]" do, I guess.

yibin87: Anyway, I think it is ok to improve it later.

dbms/src/Columns/tests/gtest_column_insertFrom.cpp
```diff
@@ -57,54 +57,82 @@ class TestColumnInsertFrom : public ::testing::Test
...
-            ColumnWithTypeAndName result(std::move(cols[1]), col_with_type_and_name.type, "");
-            compareColumn(ref, result);
+            for (size_t i = 0; i < 2; ++i)
+                cols[i] = column_ptr->cloneEmpty();
+            for (size_t i = 0; i < 6; ++i)
+                cols[0]->insertFrom(*column_ptr, 1);
+            if unlikely (
+                typeid_cast<const ColumnNothing *>(column_ptr.get())
+                || typeid_cast<const ColumnSet *>(column_ptr.get()))
+            {
```
JaySon-Huang, line 125:
```suggestion
            if (unlikely (
                typeid_cast<const ColumnNothing *>(column_ptr.get())
                || typeid_cast<const ColumnSet *>(column_ptr.get())))
            {
```

Better use `if ( unlikely(...) )`. `( ... )` is needed

dbms/src/Columns/ColumnVector.h
JaySon-Huang: `insertMany` is a new interface. I misunderstand it. We don't need to backport this in the cse branch.

JaySon-Huang (approved):
LGTM

dbms/src/Columns/ColumnVector.h
breezewish: Just curious,

> I used insertManyFrom in https://github.com/pingcap/tiflash/pull/8854, where the source column is a const column while the target column is not and it causes https://github.com/pingcap/tiflash/issues/8904.

Why insertManyFrom causes issues?

JaySon-Huang: Actually, it is caused by calling `ColumnVector::insertManyForm(ColumnConst, ... )`. So `static_cast<const Self &>(column_const)` just returns a random memory address and introduces bugs.

https://github.com/pingcap/tiflash/blob/8cec573af26adf5d3d974506ac4e9cbaecd7d5f6/dbms/src/Columns/ColumnVector.h#L216-L220

yibin87:
Caused by my tidb PR, https://github.com/pingcap/tidb/pull/52143, I'm fixing it now. Maybe you can temporarily bypass this case to pass the integration tests.

PR #8927, opened 2024-04-10
Storage: Refactor DMFileReader

### LINKED ISSUE #8904: TiFlash result is wrong

Lloyd-Pottiger:

Issue Number: close #8904

Problem Summary:

In the previous implementation, I used `insertManyFrom` to insert data in ConstColumn into the target column.

https://github.com/pingcap/tiflash/blob/adc57e22149d04d76d26f5193134a05a938af7c7/dbms/src/Columns/ColumnVector.h#L216-L220

In `insertManyFrom`, ColumnVector::insertManyFrom only expected calling src with exactly the same type of Self rather than any subclass of IColumn. If we do so, static_cast<const Self &>(src).getData() just returns a random memory location. It causes the del tag column contains too much non-zero value, leading to return less data than expected.

- Add a new interface `insertMany`  in `IColumn` https://github.com/pingcap/tiflash/pull/8925
- Use `insertMany` instead of `insertManyFrom`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


CalvinNeo:
Could you elaborate the bug in this issue

Lloyd-Pottiger:
> Could you elaborate the bug in this issue

done

JaySon-Huang:
> In insertManyFrom, the data is get from the container, but ConstColumn does not have any container.

Actually, `ColumnVector::insertManyFrom` only expected calling `src` with exactly the same type of `Self` rather than any subclass of `IColumn`. If we do so, `static_cast<const Self &>(src).getData()` just returns a random memory location.

Lloyd-Pottiger:
> > In insertManyFrom, the data is get from the container, but ConstColumn does not have any container.
> 
> Actually, `ColumnVector::insertManyFrom` only expected calling `src` with exactly the same type of `Self` rather than any subclass of `IColumn`. If we do so, `static_cast<const Self &>(src).getData()` just returns a random memory location.

updated

JaySon-Huang:
/hold
need to verify with coverage report

Lloyd-Pottiger:
image

image

release-centos7-llvm/gen_coverage.sh
```diff
@@ -54,7 +54,7 @@ cd -
 ## TODO: Add more testing binaries
 
 # Collect the prof raw data and generate cov report
-llvm-profdata merge -sparse "${COVERAGE_DIR}/*.profraw" -o "${COVERAGE_DIR}/merged.profdata"
+llvm-profdata merge -sparse ${COVERAGE_DIR}/*.profraw -o "${COVERAGE_DIR}/merged.profdata"
```
JaySon-Huang, line 57: why the `""` is removed?

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
```diff
@@ -142,156 +117,158 @@ bool DMFileReader::getSkippedRows(size_t & skip_rows)
...
+            else
+            {
+                for (size_t i = 0; i < block.columns(); ++i)
+                {
+                    columns[i]->insertRangeFrom(
+                        *block.getByPosition(i).column,
+                        0,
+                        block.getByPosition(i).column->size());
+                }
+            }
```
JaySon-Huang, line 253: We assume `readWithFilter` will never be called on the three hidden columns, so we don't need to take care about whether the `block.getByPosition(i).column` is ColumnConst or not?

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
```diff
@@ -142,156 +117,158 @@ bool DMFileReader::getSkippedRows(size_t & skip_rows)
...
-            blocks.emplace_back(std::move(block));
+            else
+            {
+                for (size_t i = 0; i < block.columns(); ++i)
+                {
+                    columns[i]->insertRangeFrom(
+                        *block.getByPosition(i).column,
+                        0,
+                        block.getByPosition(i).column->size());
+                }
```
JaySon-Huang, line 252: Are there any cases cover this branch?

release-centos7-llvm/gen_coverage.sh
Lloyd-Pottiger: add `""` will not be viewed as regrex, cause "{COVERAGE_DIR}/*.profraw no such file or directory"

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
Lloyd-Pottiger: image

clean read is not enabled when there are hidden columns.

dbms/src/Storages/DeltaMerge/File/DMFileReader.cpp
Lloyd-Pottiger: It is kind of hard to construct.... https://github.com/pingcap/tiflash/pull/8927#discussion_r1566627981 never do clean read here, so the source and target column are must be the same.

dbms/src/Storages/DeltaMerge/File/ColumnCache.cpp
```diff
@@ -55,8 +54,53 @@ RangeWithStrategys ColumnCache::getReadStrategy(size_t pack_id, size_t pack_coun
...
+    auto strategy = Strategy::Unknown;
+    size_t range_start = 0;
+    for (size_t cursor = target_range.first; cursor < target_range.second; ++cursor)
+    {
+        if (std::find(clean_read_pack_idx.cbegin(), clean_read_pack_idx.cend(), cursor) != clean_read_pack_idx.cend())
+        {
+            if (strategy == Strategy::Memory)
+            {
+                continue;
+            }
```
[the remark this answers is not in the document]
Lloyd-Pottiger: image

done

JaySon-Huang (approved):
LGTM

PR #8903, opened 2024-04-07
Support Adaptive Lightweight Compression Algorithm

Lloyd-Pottiger:
### What problem does this PR solve?

Issue Number: close #8982

Problem Summary:

### What is changed and how it works?

Support Adaptive Lightweight Compression Algorithm

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

Issue #8789, opened 2024-02-26
Support multiple compression method


Lloyd-Pottiger:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now we only support LZ4 and ZSTD. We need to support more compression methods.

**Describe the feature you'd like:**

We can introduce more compression method or encoding method, like delta. And we need to support need multiple compression method on the same data, like delta + LZ4.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #8969, opened 2024-04-19
Common: support bitpacking

Lloyd-Pottiger:

Issue Number: ref #8789

Problem Summary:

Using https://github.com/lemire/FastPFor to support Bit-Packing encoding. But it only supports 32bits and 64 bits, so most of the codes come from https://github.com/duckdb/duckdb/tree/main/third_party/fastpforlib

Part of https://github.com/pingcap/tiflash/pull/8903

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


contrib/fastpforlib/bitpacking.cpp
JaySon-Huang, line 1: why copying the files rather than adding them as submodule? It may bring trouble when we want to update the version of fastpfor

Lloyd-Pottiger: > Using https://github.com/lemire/FastPFor to support Bit-Packing encoding. But it only supports 32bits and 64 bits.

Lloyd-Pottiger: https://github.com/powturbo/TurboPFor-Integer-Compression is the best, but it is GPL

JaySon-Huang: Can we use the bitpacking in FastPFor? https://github.com/lemire/FastPFor/blob/master/headers/blockpacking.h

Lloyd-Pottiger: It only supports 32bits.

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+    constexpr static size_t getRequiredSize(size_t count, UInt8 width)
+    {
+        count = roundUpToAlgorithmGroupSize(count);
+        return ((count * width) / 8);
+    }
+
+    template <typename T>
+    constexpr static T roundUpToAlgorithmGroupSize(T num_to_round)
+    {
+        int remainder = num_to_round % BITPACKING_ALGORITHM_GROUP_SIZE;
```
CalvinNeo, line 130: Why int here?

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+        if (remainder == 0)
+        {
+            return num_to_round;
+        }
+
+        return num_to_round + BITPACKING_ALGORITHM_GROUP_SIZE - remainder;
```
CalvinNeo, line 136: AFAIK, if BITPACKING_ALGORITHM_GROUP_SIZE is a power of 2, there could be a faster algorithm

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+private:
+    template <typename T, bool is_signed, bool round_to_next_byte = false>
+    constexpr static UInt8 findMinimumBitWidth(const T * values, size_t count)
```
CalvinNeo, line 141: Again, where do we check if count is zero or value is empty?

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+            {
+                value = std::max(static_cast<T>(-min_value), max_value);
+            }
+        }
+        else
+        {
+            value = max_value;
+        }
+
+        if (value == 0)
```
CalvinNeo, line 187: Could it be moved before the `is_signed` test? signed and unsigned zero seems make no difference here

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    // Prevent compression at widths that are ineffective
+    template <typename T>
+    constexpr static UInt8 getEffectiveWidth(UInt8 width)
+    {
+        UInt8 bits_of_type = sizeof(T) * 8;
+        UInt8 type_size = sizeof(T);
+        if (width + type_size > bits_of_type)
```
CalvinNeo, line 249: I don't really get the idea of this formula

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,340 @@
...
+        else if constexpr (std::is_same<T, Int64>::value || std::is_same<T, UInt64>::value)
+        {
+            fastpforlib::fastpack(
+                reinterpret_cast<const UInt64 *>(values),
+                reinterpret_cast<UInt32 *>(dst),
+                static_cast<UInt32>(width));
+        }
+        else
+        {
+            throw Exception("Unsupported type for bitpacking");
```
CalvinNeo, line 289: I think it can actually be made into a compiler error?

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger:
```cpp
        T value;
        if (value == 0)
        {
            return 0;
        }

        if constexpr (is_signed)
        {
            if (min_value == std::numeric_limits<T>::min())
            {
                // handle special case of the minimal value, as it cannot be negated like all other values.
                return sizeof(T) * 8;
            }
            else
            {
                value = std::max(static_cast<T>(-min_value), max_value);
            }
        }
        else
        {
            value = max_value;
        }
```

like this?

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger:
```cpp
    // round up to nearest multiple of BITPACKING_ALGORITHM_GROUP_SIZE
    template <typename T>
    constexpr static T roundUpToAlgorithmGroupSize(T num_to_round)
    {
        constexpr T mask = BITPACKING_ALGORITHM_GROUP_SIZE - 1;
        return (num_to_round + mask) & ~mask;
    }
```

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: do not check, unexpected input.

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: sorry, not get your point

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger:
```cpp
            static_assert(false, "Unsupported type for bitpacking");
```

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: not a serious formula, mainly avoid some ineffective widths.

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,336 @@
...
+    constexpr static T roundUpToAlgorithmGroupSize(T num_to_round)
+    {
+        constexpr T mask = BITPACKING_ALGORITHM_GROUP_SIZE - 1;
+        return (num_to_round + mask) & ~mask;
+    }
+
+private:
+    template <typename T, bool is_signed, bool round_to_next_byte = false>
+    constexpr static UInt8 findMinimumBitWidth(const T * values, size_t count)
+    {
```
CalvinNeo, line 138: Could it have another name? There are woo many overloads of `findMinimumBitWidth`.

dbms/src/Common/BitpackingPrimitives.h
CalvinNeo: Okay, it is an private function. I think it is ok that we don't check.

dbms/src/Common/BitpackingPrimitives.h
CalvinNeo: I wonder why don't we put it at the very beginning?

dbms/src/Common/BitpackingPrimitives.h
CalvinNeo: Also static assert that `BITPACKING_ALGORITHM_GROUP_SIZE` is a power of 2.

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: `value` is not initialize

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger:
```cpp
    // Calculates the minimum required number of bits per value that can store all values
    template <typename T, bool is_signed = std::numeric_limits<T>::is_signed>
    constexpr static UInt8 minimumBitWidth(T value)
    {
        return findMinimumBitWidth<T, is_signed, BYTE_ALIGNED>(value, value);
    }

    // Calculates the minimum required number of bits per value that can store all values
    template <typename T, bool is_signed = std::numeric_limits<T>::is_signed>
    constexpr static UInt8 minimumBitWidth(const T * values, size_t count)
    {
        return findMinimumBitWidth<T, is_signed, BYTE_ALIGNED>(values, count);
    }

    // Calculates the minimum required number of bits per value that can store all values,
    // given a predetermined minimum and maximum value of the buffer
    template <typename T, bool is_signed = std::numeric_limits<T>::is_signed>
    constexpr static UInt8 minimumBitWidth(T minimum, T maximum)
    {
        return findMinimumBitWidth<T, is_signed, BYTE_ALIGNED>(minimum, maximum);
    }
```

I think they are impossible to resolve by mistake.

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: If the compression ratio is less than 8/7, then it is ineffective, just copy is better.

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,339 @@
...
+        T min_value = values[0];
+        T max_value = values[0];
+
+        for (size_t i = 1; i < count; i++)
```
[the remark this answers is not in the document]
Lloyd-Pottiger: `values` is not a container.

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: revert, keep `throw Exception` since clang13 does not support `static_assert(false)`

dbms/src/Common/BitpackingPrimitives.h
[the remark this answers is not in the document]
Lloyd-Pottiger: Thank~

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -0,0 +1,343 @@
...
+            fastpforlib::fastpack(
+                reinterpret_cast<const UInt64 *>(values),
+                reinterpret_cast<UInt32 *>(dst),
+                static_cast<UInt32>(width));
+        }
+        else
+        {
+            // TODO: use static_assert instead until the toolchain upgrade to clang 17.0
+            // static_assert(false, "Unsupported type for bitpacking");
+            throw Exception("Unsupported type for bitpacking");
```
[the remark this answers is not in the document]
Lloyd-Pottiger: done

dbms/src/Common/BitpackingPrimitives.h
CalvinNeo: got

dbms/src/Common/BitpackingPrimitives.h
CalvinNeo: OK

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: `width >= bits_of_type` is impossible. 

For example, `T` = UInt32, `width` = 31, uncompressed size = 1MB, then the compressed size is 0.97MB, the compression ration is quite low, there is no need to do bitpicking, just copy is much more effective.

CalvinNeo (approved):
lgtm

dbms/src/Common/BitpackingPrimitives.h
Lloyd-Pottiger: Any reasonable number is ok. `bits_of_type` is always equal to `type_size * 8`, here use `1/8` as threshold.

Lloyd-Pottiger: Add some comments.

Issue #8982, opened 2024-04-24
Support Adaptively Decide Compression Algorithm


Lloyd-Pottiger:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, TiFlash only supports general compression LZ4, ZSTD, and [DeflateQpl](https://github.com/intel/qpl) (not available in the release version). It is decided by the configuration file, which means all data in the same TiFlash instance will use the same compression algorithm.

By detecting specific patterns, specialized compression algorithms can be significantly more lightweight, providing much faster compression and decompression. In addition, they can be effective on much smaller data sizes. This allows us to decompress a few rows at a time, rather than requiring large blocks of data to be decompressed at once. These specialized compression algorithms can also offer efficient support for random seeks, making data access through an index significantly faster.

**Describe the feature you'd like:**

Support more lightweight compression algorithms and adaptively decide compression algorithms for each data part in TiFlash.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #8983, opened 2024-04-24
Compression: support FOR & replace Delta with DeltaFOR

Lloyd-Pottiger:

Issue Number: ref #8982

Problem Summary:

FOR: https://dbms-arch.fandom.com/wiki/Frame_of_Reference_(Compression_Scheme)

Delta is not a compression algorithm, but Delta + FOR be.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/IO/Compression/CompressionCodecDeltaPFor.cpp
```diff
@@ -0,0 +1,380 @@
...
+template <class T>
+void ApplyFrameOfReference(T * dst, T frame_of_reference, UInt32 count)
+{
+    if (!frame_of_reference)
+        return;
+
+    UInt32 i = 0;
+    UInt32 misaligned_count = count;
+#if defined(__AVX2__)
+    misaligned_count = count % (sizeof(__m256i) / sizeof(T));
```
CalvinNeo, line 106: What if size of T is greater than size of __m256i?

Lloyd-Pottiger: impossible, the largest is UInt64.

CalvinNeo: static assert that

Lloyd-Pottiger: done

CalvinNeo (approved):
lgtm

dbms/src/IO/Compression/CompressionCodecDeltaFor.h
```diff
@@ -19,10 +19,10 @@
 namespace DB
 {
 
-class CompressionCodecDelta : public ICompressionCodec
+class CompressionCodecDeltaFor : public ICompressionCodec
```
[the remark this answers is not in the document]
Lloyd-Pottiger: https://dbms-arch.fandom.com/wiki/Frame_of_Reference_(Compression_Scheme)

It is a kind of encoding algorithm.

dbms/src/IO/Compression/CompressionCodecDeltaFor.cpp
```diff
@@ -0,0 +1,394 @@
...
+UInt8 CompressionCodecDeltaFor::getMethodByte() const
+{
+    return static_cast<UInt8>(CompressionMethodByte::DeltaFor);
+}
+
+UInt32 CompressionCodecDeltaFor::getMaxCompressedDataSize(UInt32 uncompressed_size) const
+{
+    // 1 byte for bytes_size, x bytes for frame of reference, 1 byte for width.
+    const size_t count = uncompressed_size / bytes_size;
+    return 1 + bytes_size + sizeof(UInt8) + BitpackingPrimitives::getRequiredSize(count, bytes_size * 8);
```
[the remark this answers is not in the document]
Lloyd-Pottiger: yes

dbms/src/IO/Compression/CompressionCodecFor.cpp
```diff
@@ -0,0 +1,227 @@
...
+        source += sizeof(T);
+    }
+    T frame_of_reference = *std::min_element(values.cbegin(), values.cend());
+    // store frame of reference
+    unalignedStore<T>(dest, frame_of_reference);
+    dest += sizeof(T);
+    if (frame_of_reference != 0)
+    {
+        for (auto & value : values)
+            value -= frame_of_reference;
```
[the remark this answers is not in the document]
Lloyd-Pottiger: > A computation involving unsigned operands can never overflow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type. (ISO/IEC 9899:1999 (E) §6.2.5/9)

dbms/src/IO/Compression/CompressionCodecFor.cpp
```diff
@@ -0,0 +1,227 @@
...
+    while (source < source_end)
+    {
+        values.push_back(unalignedLoad<T>(source));
+        source += sizeof(T);
+    }
+    T frame_of_reference = *std::min_element(values.cbegin(), values.cend());
+    // store frame of reference
+    unalignedStore<T>(dest, frame_of_reference);
+    dest += sizeof(T);
+    if (frame_of_reference != 0)
```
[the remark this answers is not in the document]
Lloyd-Pottiger: T can only be an unsigned type, `frame_of_reference` can not be negative.

dbms/src/IO/Compression/CompressionCodecFor.cpp
```diff
@@ -0,0 +1,227 @@
...
+UInt32 CompressionCodecFor::compressData(const char * source, UInt32 source_size, char * dest)
+{
+    static_assert(std::is_integral<T>::value, "Integral required.");
+    const auto count = source_size / sizeof(T);
+    std::vector<T> values;
+    values.reserve(count);
+    const char * const source_end = source + source_size;
+    while (source < source_end)
+    {
+        values.push_back(unalignedLoad<T>(source));
```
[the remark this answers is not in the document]
Lloyd-Pottiger: done

dbms/src/IO/Compression/CompressionCodecFor.h
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include <IO/Compression/ICompressionCodec.h>
+
+namespace DB
+{
+
+class CompressionCodecFor : public ICompressionCodec
```
JaySon-Huang, line 22:
```suggestion
// The Frame of Reference (FOR) compression scheme for numeric values: Instead
// of compressing the actual value, use a value close to all others in the same
// range (for integers - often the minimum value, or the minimum without
// outliers/exceptionals) and encode all values using their difference from
// this reference. The differences typically need less bits to represent.
// One could think of this as an approximation of the data by a constant + residuals.
// Reference: https://dbms-arch.fandom.com/wiki/Frame_of_Reference_(Compression_Scheme)
class CompressionCodecFOR : public ICompressionCodec
```

IMO, we should use `FOR` instead of `For`. The `for` is a meaningful word, using it as a compression method name is confusing. While `FOR` can remind us that it is an abbreviation.

dbms/src/IO/Compression/CompressionCodecDeltaFor.h
JaySon-Huang, line 22:
```suggestion
class CompressionCodecDeltaFOR : public ICompressionCodec
```

dbms/src/IO/Compression/CompressionCodecFor.h
Lloyd-Pottiger: done

dbms/src/IO/Compression/CompressionCodecFor.cpp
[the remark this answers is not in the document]
Lloyd-Pottiger: Now we support signed types, but here is just an optimization, when `frame_of_reference` = 0, no need to do minus.

And for signed types, there is no need to consider overflow, because

```cpp
int count = std::numeric_limits<int>::min();
count -= std::numeric_limits<int>::max();
count += std::numeric_limits<int>::max();
std::cout << count << "\n"; // equal to std::numeric_limits<int>::min()
```

dbms/src/IO/Compression/CompressionCodecDeltaFOR.cpp
```diff
@@ -32,58 +35,59 @@ extern const int CANNOT_COMPRESS;
...
-void ordinaryDecompressData(const char * source, UInt32 source_size, char * dest, UInt32 output_size)
+UInt32 compressData(const char * source, UInt32 source_size, char * dest)
 {
-    if (source_size % sizeof(T) != 0)
-        throw Exception(
-            ErrorCodes::CANNOT_DECOMPRESS,
-            "Cannot decompress Delta-encoded data, data size {} is not aligned to {}",
-            source_size,
-            sizeof(T));
+    static_assert(std::is_integral<T>::value, "Integral required.");
```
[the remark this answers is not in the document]
Lloyd-Pottiger: done

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,250 @@
...
+            value = _mm256_add_epi32(value, _mm256_set1_epi32(frame_of_reference));
+        }
+        else if constexpr (sizeof(T) == 8)
+        {
+            value = _mm256_add_epi64(value, _mm256_set1_epi64x(frame_of_reference));
+        }
+        // Store the result back to memory
+        _mm256_storeu_si256(reinterpret_cast<__m256i *>(dst + i), value);
+    }
+#endif
```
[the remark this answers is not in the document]
Lloyd-Pottiger: When `__AVX2__` is set, handle misaligned first.

```
    // count = 10, T = UInt32
    UInt32 misaligned_count = count; // 10
#if defined(__AVX2__)
    misaligned_count = count % (sizeof(__m256i) / sizeof(T)); // 2
#endif
    for (; i < misaligned_count; ++i) // [0,2)
    {
        dst[i] += frame_of_reference;
    }
#if defined(__AVX2__)
    for (; i < count; i += (sizeof(__m256i) / sizeof(T))) // [2, 10)
```

dbms/src/Common/BitpackingPrimitives.h
```diff
@@ -72,6 +72,11 @@ class BitpackingPrimitives
         UInt8 width,
         bool skip_sign_extension = false)
     {
+        if (width == 0)
+        {
+            memset(dst, 0, count * sizeof(T));
+            return;
+        }
```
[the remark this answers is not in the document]
Lloyd-Pottiger: `unPackGroup` can handle the input with `width == 0`, add this to avoid loop.

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,254 @@
...
+    T frame_of_reference = *std::min_element(values.cbegin(), values.cend());
+    // store frame of reference
+    unalignedStore<T>(dest, frame_of_reference);
+    dest += sizeof(T);
+    if (frame_of_reference != 0)
+    {
+        for (auto & value : values)
+            value -= frame_of_reference;
+    }
+    T max_value = *std::max_element(values.cbegin(), values.cend());
```
JaySon-Huang, line 69: https://godbolt.org/z/rh16v69Mz

We don't need to copy the numbers from `source` to a temporary vector `values`. `std::min_element` and `std::max_element` support directly run on raw array

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,254 @@
...
+    return 1 + bytes_size + sizeof(UInt8) + BitpackingPrimitives::getRequiredSize(count, bytes_size * 8);
+}
+
+template <std::integral T>
+UInt32 CompressionCodecFOR::compressData(const T * source, UInt32 count, char * dest)
+{
+    std::vector<T> values(count);
+    for (UInt32 i = 0; i < count; ++i)
+        values[i] = source[i];
+    T frame_of_reference = *std::min_element(values.cbegin(), values.cend());
```
JaySon-Huang, line 60: Do we need to handle `count == 0` in the previous code? Or it will make a crash here

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,254 @@
...
+        reinterpret_cast<const unsigned char *>(source),
+        count,
+        width);
+    CompressionCodecFOR::applyFrameOfReference(reinterpret_cast<T *>(dest), frame_of_reference, count);
+}
+
+template <std::integral T>
+void CompressionCodecFOR::applyFrameOfReference(T * dst, T frame_of_reference, UInt32 count)
+{
+    if (!frame_of_reference)
```
JaySon-Huang, line 118:
```suggestion
    if (frame_of_reference == 0)
```

dbms/src/IO/Compression/CompressionCodecFOR.cpp
Lloyd-Pottiger:
```cpp
    if (frame_of_reference != 0)
    {
        for (auto & value : values)
            value -= frame_of_reference;
    }
```

`const char *` can not be changed.

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    UInt32 i = 0;
+    UInt32 misaligned_count = count;
+#if defined(__AVX2__)
+    misaligned_count = count % (sizeof(__m256i) / sizeof(T));
+#endif
+    for (; i < misaligned_count; ++i)
+    {
+        dst[i] += frame_of_reference;
+    }
```
JaySon-Huang, line 129: I think doing it after the `for (; i < count; i += (sizeof(__m256i) / sizeof(T)))` can bring better performance. Because  we can assume the memory address of `dst + 0` should be aligned with the cache line, but it is not always aligned when it starts with `dst + misaligned_count`

dbms/src/IO/Compression/CompressionCodecFOR.cpp
Lloyd-Pottiger:
```
UInt32 CompressionCodecFOR::doCompressData(const char * source, UInt32 source_size, char * dest) const
{
    if unlikely (source_size % bytes_size != 0)
        throw Exception(ErrorCodes::CANNOT_DECOMPRESS, "source size {} is not aligned to {}", source_size, bytes_size);
```

if `count == 0`, will throw exception here.

dbms/src/IO/Compression/CompressionCodecFOR.cpp
Lloyd-Pottiger: Good catch, fixed.

dbms/src/IO/Compression/CompressionCodecFOR.cpp
Lloyd-Pottiger: anyway, add a `RUNTIME_CHECK`

dbms/src/IO/Compression/CompressionCodecFOR.cpp
Lloyd-Pottiger: And we can not write the data into `dest` since the size of `dest` may be smaller than `uncompressed_size`.

JaySon-Huang: Get it.
Then I think we can replace the for-loop in line 58 with `values.assign(source, source + count);`, which can copy the bytes by `memcpy`

dbms/src/IO/Compression/CompressionCodecFOR.cpp
```diff
@@ -0,0 +1,252 @@
...
+     *|1 bytes               |bytes_size        |sizeof(UInt8)|required size |
+     */
+    const size_t count = uncompressed_size / bytes_size;
+    return 1 + bytes_size + sizeof(UInt8) + BitpackingPrimitives::getRequiredSize(count, bytes_size * 8);
+}
+
+template <std::integral T>
+UInt32 CompressionCodecFOR::compressData(const T * source, UInt32 count, char * dest)
+{
+    RUNTIME_CHECK(count > 0, count);
```
JaySon-Huang, line 57:
```suggestion
    assert(count > 0); // `doCompressData` ensure it
```
When the upper-level function ensures the checking, we can just use `assert` to avoid the extra performance penalty in the release code.

JaySon-Huang (approved):
LGTM

PR #8985, opened 2024-04-25
Optimize unnecessary column copy for HashAgg

### LINKED ISSUE #8891: Avoid duplicated first_row agg func
## Enhancement
Check the following example:
```
drop table if exists t1;
create table t1(c1 int, c2 varchar(100), c3 int);
insert into t1 values(1, 'a', 1), (2, 'b', 2), ..., (10000, 'xxx', 10000);
select sum(c1), c2, c3 from t1 group by c2, c3;
```
For tiflash HashAgg current implementation, the internal computation procedure includes:
1. For each row, insert column serialized c2+c2 into HashMap and update agg state in HashMap
2. After all rows are handled, start to read HashMap. Will copy data from HashMap to `Column`. And the real copy include:
    1. Copy result of `sum(c1)` to `ColumnDecimal`.
    2. Copy result of `first_row(c2)` to `ColumnString`.
    3. Copy result of `first_row(c3)` to `ColumnInt`.
    4. Copy result of `any(c2)` to `ColumnString`. 
    5. Copy result of `c3` to `ColumnInt`. 

For `2.i`, `2.ii` and  `2.iii`, they are corresponsding to the select item of `sum(c1)`, `c2` and `c3`.

For `2.iv` and `2.v`, they are corresponding to the group by column `c2` and `c3`.

But actually the copy of `2.ii` is duplicated with `2.iv`, and the copy of `2.iii` is duplicated with `2.v`. So we can do the following optimizations:
1. Eliminate the `first_row(c3)` agg func(a.k.a. `2.iii`) and make a pointer to reference `c3` directly. So we can avoid the computation of `first_row` and the copy of `first_row(c3)` to `result ColumnInt`.
2. Eliminate the `any(c2)` agg func (a.k.a. `2.v`) and make a pointer to reference `first_row(c2)`. So we can aovoid the computation of `any(c2)` (a.k.a. `2.iv`) and the copy of `any(c2)`

If c2 and c3 are high NDV, this optimization is significant.

guo-shaoge:

Issue Number: close #8891

Problem Summary:
When there are group by key in select item(a.k.a. first_row), tiflash have extra agg func, which cause unnecessary copy from HashMap to result column.

Basic idea:
image

**Optimization-1 (with collation)**:
What: For group by keys that with collation, there will be `first_row/any agg` func for it to keep original data. So no need to copy these keys from HashMap, instead just a pointer to reference its corresponding `first_row/any` result is enough.

How:
1. Detect if there is a `first_row` agg func in the select item. 
2. If so, ignore `any` agg func.  If not, add the 'any' agg func.
4. Also, set `key_from_agg_func` to indicate that this key is equivalent to `first_row/any` agg func, which can avoid copying this key from the HashMap in subsequent operations. (check `DAGExpressionAnalyzer::buildAggGroupBy`)
5. If all keys are included in `first_row/any` (which is rare, but still can happens), will skip copy keys(template argument `skip_serialize_key` is true)

**Optimization-2(no collation)**
What: When SQL query has agg func like: `first_row(group_by_key_col)`, and that group_by_key_col has no collation. Then we can eliminate the `first_row` agg func, just use a pointer to reference the group by key is enough.

**Results**
1. 25% improvement
2. workload:
    1. 20M rows, very high NDV.
    2. 3 varchar columns, 3 decimal columns, 3 int columns (that means Aggregator will use HashMethodSerialized)

before:
image

after:
image

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1403,8 +1503,11 @@ struct AggregatorMethodInitKeyColumnHelper<AggregationMethodOneKeyStringNoCache<
 
     void initAggKeys(size_t rows, std::vector<IColumn *> & key_columns)
     {
-        Method::initAggKeys(rows, key_columns[0]);
         index = 0;
+        if (key_columns.size() == 1)
+            Method::initAggKeys(rows, key_columns[0]);
+        else
+            throw Exception("unexpected key_columns size for AggMethodOneKeyString: {}", key_columns.size());
```
SeaRise, line 1510:
```suggestion
        RUNTIME_CHECK_MSG(key_columns.size() == 1, "unexpected key_columns size for AggMethodOneKeyString: {}", key_columns.size());
        Method::initAggKeys(rows, key_columns[0]);
```

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -575,6 +588,7 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
             /// need double check this assumption when we support agg with collation
             aggregation_keys.push_back(name);
             agg_key_set.emplace(name);
+            collators.push_back(nullptr);
```
SeaRise, line 591: It seems L591 and L605 have changed the original logic. Was this intended?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -555,13 +555,26 @@ void DAGExpressionAnalyzer::buildCommonAggFunc(
...
+    const AggregateDescriptions & aggregate_descriptions,
+    const String & arg_name)
+{
+    for (const auto & desc : aggregate_descriptions)
+    {
+        if (desc.function->getName() == "first_row" && desc.argument_names[0] == arg_name)
+            return std::make_pair(desc.column_name, desc.function->getReturnType());
+    }
+    return std::make_pair("", nullptr);
+}
```
SeaRise, line 568: Please move to `namespace {}` that from L76 to L386

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -588,22 +602,33 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
...
-                appendAggDescription(
-                    {name},
-                    {type},
-                    arg_collators,
-                    "any",
-                    aggregate_descriptions,
-                    aggregated_columns,
-                    false,
-                    context);
+                auto [first_row_name, first_row_type] = findFirstRow(aggregate_descriptions, name);
```
SeaRise, line 608: Is `first_row_type` necessary here? Are the `type` at L600 and `first_row_type` always the same? If so, can we just use `type`?

dbms/src/Flash/tests/bench_aggregation_hash_map.cpp
```diff
@@ -0,0 +1,331 @@
...
+#include <benchmark/benchmark.h>
+#include <google/protobuf/util/json_util.h>
+#include <tipb/executor.pb.h>
+
+#include <random>
+#include <string>
+
+namespace DB
+{
+namespace tests
```
SeaRise, line 38: Please also add unit tests in https://github.com/pingcap/tiflash/blob/8e50de84e6d6ecdcc108990217b70b6bb3f50271/dbms/src/Flash/tests/gtest_aggregation_executor.cpp

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
guo-shaoge: done

dbms/src/Interpreters/Aggregator.cpp
guo-shaoge: done

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.cpp
```diff
@@ -76,34 +76,63 @@ bool isGroupByCollationSensitive(const Context & context)
...
 }
 
-Aggregator::Params buildParams(
+std::shared_ptr<Aggregator::Params> buildParams(
     const Context & context,
     const Block & before_agg_header,
     size_t before_agg_streams_size,
     size_t agg_streams_size,
     const Names & key_names,
+    const KeyRefAggFuncMap & key_ref_agg_func,
```
guo-shaoge, line 85: Add comments about the meaning of above two map.

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.cpp
```diff
@@ -76,34 +76,63 @@ bool isGroupByCollationSensitive(const Context & context)
...
-    for (const auto & name : key_names)
+    ColumnNumbers keys(key_names.size(), 0);
+    size_t normal_key_idx = 0;
+    size_t agg_func_as_key_idx = key_names.size() - key_ref_agg_func.size();
+    // Put group by key that reference aggregate func after original key.
+    // For example: select sum(c0), first_row(c1), first_row(c3) group by c1, c2, c3
+    // Before: keys: c1 | c2 | c3
+    // After:  keys: c2 | c1 | c3
+    // By doing this, when deserialize group by keys from HashMap to columns,
+    // we only need to handle c2(convert_key_size == 1) and ignore c1/c3.
```
guo-shaoge, line 100: Update comment to clarify that still need to let c1/c3 reference first_row agg func.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
SeaRise: So, there was a bug in the previous code?

guo-shaoge: The orignal implementation is no bug. I change this code to make sure the collators and `key_ref_agg_func` are corresponding one by one.

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1062,24 +1076,53 @@ Block Aggregator::convertOneBucketToBlock(
...
+    FILLER_DEFINE(convert_key, false);
+    FILLER_DEFINE(skip_convert_key, true);
+#undef FILLER_DEFINE
+
+    // Ignore key optimization if in non-final mode(a.k.a. during spilling process).
+    // Because all keys are needed when insert spilled block back into HashMap during restore process.
+    size_t convert_key_size = final ? params.keys_size - params.key_ref_agg_func.size() : params.keys_size;
+
+    Block block;
+    if (final && params.key_ref_agg_func.size() == params.keys_size)
```
windtalker, line 1108:
```suggestion
    if (final && convert_key_size == 0)
```

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1093,24 +1136,50 @@ BlocksList Aggregator::convertOneBucketToBlocks(
...
+    };
+
+    FILLER_DEFINE(convert_key, false);
+    FILLER_DEFINE(skip_convert_key, true);
+#undef FILLER_DEFINE
+
+    BlocksList blocks;
+    size_t convert_key_size = final ? params.keys_size - params.key_ref_agg_func.size() : params.keys_size;
+
+    if (final && params.key_ref_agg_func.size() == params.keys_size)
```
windtalker, line 1164:
```suggestion
    if (final && convert_key_size == 0)
```

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1403,8 +1508,12 @@ struct AggregatorMethodInitKeyColumnHelper<AggregationMethodOneKeyStringNoCache<
 
     void initAggKeys(size_t rows, std::vector<IColumn *> & key_columns)
     {
-        Method::initAggKeys(rows, key_columns[0]);
         index = 0;
+        RUNTIME_CHECK_MSG(
+            key_columns.size() == 1,
```
windtalker, line 1513: This check will always be true because if no key column need to be convert, `filler_skip_convert_key` will be called?

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1474,45 +1597,68 @@ std::vector<std::unique_ptr<AggregatorMethodInitKeyColumnHelper<Method>>> initAg
...
+template <typename Method, typename Table, bool skip_convert_key>
 void NO_INLINE Aggregator::convertToBlocksImplFinal(
     Method & method,
     Table & data,
+    const Sizes & key_sizes,
     std::vector<std::vector<IColumn *>> && key_columns_vec,
     std::vector<MutableColumns> & final_aggregate_columns_vec,
     Arena * arena) const
 {
     assert(!key_columns_vec.empty());
```
windtalker, line 1609: If `skip_convert_key` is true, `key_columns_vec` maybe empty?

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1550,20 +1708,28 @@ void NO_INLINE Aggregator::convertToBlocksImplNotFinal(
...
+    size_t convert_key_size) const
 {
-    MutableColumns key_columns(params.keys_size);
+    MutableColumns key_columns(convert_key_size);
     MutableColumns aggregate_columns(params.aggregates_size);
     MutableColumns final_aggregate_columns(params.aggregates_size);
     AggregateColumnsData aggregate_columns_data(params.aggregates_size);
+    // Store size of keys that need to convert.
+    Sizes new_key_sizes;
+    new_key_sizes.reserve(key_sizes.size());
```
windtalker, line 1724:
```suggestion
    new_key_sizes.reserve(convert_key_size);
```

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1637,6 +1804,9 @@ BlocksList Aggregator::prepareBlocksAndFill(
     std::vector<AggregateColumnsData> aggregate_columns_data_vec;
     std::vector<MutableColumns> aggregate_columns_vec;
     std::vector<MutableColumns> final_aggregate_columns_vec;
+    // Store size of keys that need to convert.
+    Sizes new_key_sizes;
+    new_key_sizes.reserve(convert_key_size);
```
windtalker, line 1809: how about fill `new_key_sizes` here so we can remove L1834-L1835?

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1768,7 +1941,7 @@ BlocksList Aggregator::prepareBlocksAndFillWithoutKey(AggregatedDataVariants & d
         }
     };
 
-    BlocksList blocks = prepareBlocksAndFill(data_variants, final, rows, filler);
+    BlocksList blocks = prepareBlocksAndFill(data_variants, final, rows, filler, params.keys_size);
```
windtalker, line 1944:
```suggestion
    BlocksList blocks = prepareBlocksAndFill(data_variants, final, rows, filler, 0);
```

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -1779,35 +1952,56 @@ BlocksList Aggregator::prepareBlocksAndFillWithoutKey(AggregatedDataVariants & d
...
 #undef M
-    };
+#undef M_skip_convert_key
+#undef M_convert_key
+#undef FILLER_DEFINE
+
+    size_t convert_key_size = final ? params.keys_size - params.key_ref_agg_func.size() : params.keys_size;
 
-    return prepareBlocksAndFill(data_variants, final, rows, filler);
+    if (final && params.key_ref_agg_func.size() == params.keys_size)
```
windtalker, line 2000:
```suggestion
    if (final && convert_key_size == 0)
```

dbms/src/Interpreters/Aggregator.cpp
guo-shaoge: Yes. Add this check only for guard

dbms/src/Interpreters/Aggregator.cpp
guo-shaoge: `key_columns_vec` should be emtpy if `skip_convert_key` is true. But currently it's not, because in `prepareBlockAndFill`, it will fill `MutableColumns` even when `convert_key_size` is zero. Check [here](https://github.com/pingcap/tiflash/blob/bab33aa025adcc19b57f783229e63215930a0faa/dbms/src/Interpreters/Aggregator.cpp#L1643).

Maybe I should change it.

windtalker: I think it's better to change it, and keep `key_sizes.size() == key_columns_vec.size()` always true?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -588,22 +602,33 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
...
+                {
+                    /// if the column is a string with collation info, the `sort_key` of the column is used during
+                    /// aggregation, but we can not reconstruct the origin column by `sort_key`, so add an extra
+                    /// extra aggregation function any(group_by_column) here as the output of the group by column
+                    TiDB::TiDBCollators arg_collators{collator};
+                    appendAggDescription(
+                        {name},
+                        {type},
+                        arg_collators,
+                        "any",
```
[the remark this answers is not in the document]
guo-shaoge: Why change to `first_row`? I think `any` and `first_row` has no difference?

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.h
```diff
@@ -33,12 +33,13 @@ bool isFinalAgg(const tipb::Aggregation & aggregation);
 
 bool isGroupByCollationSensitive(const Context & context);
 
-Aggregator::Params buildParams(
+std::shared_ptr<Aggregator::Params> buildParams(
```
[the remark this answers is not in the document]
guo-shaoge: Because `Aggregator::Params` has no default copy assign operator(because it has two const member), and the benchmark needs store `Params` as member [here](https://github.com/pingcap/tiflash/pull/8985/files/c312346cd19e3a8dd570b17da5d1660295bd1118#diff-fe1f01075b1a6eb635a2258522ac620ca04e669ff64edcc4fd856164b8887cc8R272)

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -261,7 +262,15 @@ Block Aggregator::Params::getHeader(
     else
     {
         for (const auto & key : keys)
-            res.insert(src_header.safeGetByPosition(key).cloneEmpty());
+        {
```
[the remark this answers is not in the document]
guo-shaoge: Disable `intermediate_block` code path, because it's only used by `InterpreterSelectQuery`, which is not used by TiFlash

dbms/src/Interpreters/Aggregator.cpp
guo-shaoge: Add check for `key_sizes.size() == key_columns_vec[i].size()` for all four convert functions.

dbms/src/Interpreters/Aggregator.cpp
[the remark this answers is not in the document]
guo-shaoge: Delete related ctor

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.h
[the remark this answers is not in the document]
zanmato1984: Can you try declaring deleted copy/assign operator and defaulted move/assign operator to mitigate this?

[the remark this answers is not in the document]
guo-shaoge: done

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -588,22 +602,34 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
...
+                if (!first_row_name.empty())
+                {
+                    // Got here when this group by key has its corresponding first_row agg func.
+                    aggregated_columns.emplace_back(name, first_row_type);
+                }
+                else
+                {
+                    /// if the column is a string with collation info, the `sort_key` of the column is used during
+                    /// aggregation, but we can not reconstruct the origin column by `sort_key`, so add an extra
+                    /// extra aggregation function any(group_by_column) here as the output of the group by column
```
windtalker, line 619:
```suggestion
                    /// extra aggregation function first_row(group_by_column) here as the output of the group by column
```

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -615,6 +641,57 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
             aggregated_columns.emplace_back(name, actions->getSampleBlock().getByName(name).type);
         }
     }
+#ifndef NDEBUG
+    for (const auto & p : key_ref_agg_func)
+    {
+        LOG_TRACE(Logger::get(), "key_ref_agg_func optimization: {} ref {}", p.first, p.second);
```
windtalker, line 647:
```suggestion
        LOG_DEBUG(Logger::get(), "key_ref_agg_func optimization: {} ref {}", p.first, p.second);
```

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -615,6 +641,57 @@ void DAGExpressionAnalyzer::buildAggGroupBy(
...
+        {
+            tmp_aggregate_descriptions.push_back(desc);
+        }
+    }
+    aggregate_descriptions = tmp_aggregate_descriptions;
+
+#ifndef NDEBUG
+    for (const auto & p : agg_func_ref_key)
+    {
+        LOG_TRACE(Logger::get(), "agg_func_ref_key optimization: {} ref {}", p.first, p.second);
```
windtalker, line 692:
```suggestion
        LOG_DEBUG(Logger::get(), "agg_func_ref_key optimization: {} ref {}", p.first, p.second);
```

dbms/src/Flash/Planner/Plans/PhysicalAggregation.cpp
```diff
@@ -54,9 +54,20 @@ PhysicalPlanNodePtr PhysicalAggregation::build(
     NamesAndTypes aggregated_columns;
     AggregateDescriptions aggregate_descriptions;
     Names aggregation_keys;
+    // key_ref_agg_func and agg_func_ref_key are two optimizations for aggregation:
+    // 1. key_ref_agg_func: for group by key with collation, there will always be a agg func(first_row or any)
```
windtalker, line 58:
```suggestion
    // 1. key_ref_agg_func: for group by key with collation, there will always be a agg func(first_row)
```

dbms/src/Flash/Planner/Plans/PhysicalAggregation.cpp
```diff
@@ -54,9 +54,20 @@ PhysicalPlanNodePtr PhysicalAggregation::build(
...
+    //    to help keep the original column. For these key columns, no need to copy it from HashMap to result column,
+    //    instead a pointer to reference to their corresponding first_row/any agg func is enough.
```
windtalker, line 60:
```suggestion
    //    instead a pointer to reference to their corresponding first_row agg func is enough.
```

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
guo-shaoge: only for debug unit test, so no need to change

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.h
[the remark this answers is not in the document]
zanmato1984: What is done?

zanmato1984 (commented):
Consider using `unique_ptr` instead of `shared_ptr` for a factory method.

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.h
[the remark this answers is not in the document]
guo-shaoge: Since `Params` needs to be store in `bencharmrk` class as member, and will be copy assign in `SetUp`. But copy assign/move assign operator is not allowed for `Params`. So we have to make this function return shared_pointer.

So I just declare the copy assign/move assign operator as deleted. Basically nothing changed.

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.h
```diff
@@ -33,12 +33,14 @@ bool isFinalAgg(const tipb::Aggregation & aggregation);
...
```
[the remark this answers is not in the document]
guo-shaoge: changed to unique_ptr

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.cpp
```diff
@@ -76,34 +76,66 @@ bool isGroupByCollationSensitive(const Context & context)
...
     auto total_two_level_threshold_bytes
         = allow_to_use_two_level_group_by ? settings.group_by_two_level_threshold_bytes : SettingUInt64(0);
 
-    bool has_collator = std::any_of(begin(collators), end(collators), [](const auto & p) { return p != nullptr; });
+    bool has_collator = std::any_of(begin(reordered_collators), end(reordered_collators), [](const auto & p) {
+        return p != nullptr;
+    });
 
-    return Aggregator::Params(
+    return std::make_shared<Aggregator::Params>(
```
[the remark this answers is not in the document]
guo-shaoge: changed to make_unique

dbms/src/Flash/Coprocessor/AggregationInterpreterHelper.cpp
```diff
@@ -76,34 +76,66 @@ bool isGroupByCollationSensitive(const Context & context)
     return context.getSettingsRef().group_by_collation_sensitive || context.getDAGContext()->isMPPTask();
 }
 
-Aggregator::Params buildParams(
+std::shared_ptr<Aggregator::Params> buildParams(
```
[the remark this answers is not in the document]
guo-shaoge: ditto

windtalker (approved):
LGTM

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -383,6 +383,18 @@ void buildActionsAfterWindow(
...
+    const AggregateDescriptions & aggregate_descriptions,
+    const String & arg_name)
+{
+    for (const auto & desc : aggregate_descriptions)
+    {
+        if (desc.function->getName() == "first_row" && desc.argument_names[0] == arg_name)
+            return std::make_pair(desc.column_name, desc.function->getReturnType());
+    }
+    return std::make_pair("", nullptr);
+}
```
SeaRise, line 397:
```suggestion
std::optional<String> findFirstRow(
    const AggregateDescriptions & aggregate_descriptions,
    const String & arg_name,
    const DataTypePtr & arg_type)
{
    for (const auto & desc : aggregate_descriptions)
    {
        if (desc.function->getName() == "first_row" && desc.argument_names[0] == arg_name)
        {
            RUNTIME_CHECK(desc.function->getReturnType().requals(*arg_type));
            return desc.column_name;
        }
    }
    return {};
}
```

SeaRise: from https://github.com/pingcap/tiflash/pull/8985/files#r1590598257

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
```diff
@@ -685,19 +766,18 @@ std::tuple<Names, TiDB::TiDBCollators, AggregateDescriptions, ExpressionActionsP
     for (const auto & aggregation_key : aggregation_keys)
         step.required_output.push_back(aggregation_key);
 
-    source_columns = std::move(aggregated_columns);
```
SeaRise, line 688: why delete this line?

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
guo-shaoge: I think they are same. And add a check to make sure.

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
guo-shaoge: `source_columns` will be setup in `appendCopyColumnAfterAgg`

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
guo-shaoge: changed

dbms/src/Flash/Coprocessor/DAGExpressionAnalyzer.cpp
SeaRise: ok

SeaRise (approved):
lgtm

SeaRise:
comment `/unhold` to merge pr~
@guo-shaoge

Issue #9043, opened 2024-05-15
TSAN failed to run


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

introduce by https://github.com/pingcap/tiflash/pull/9003

```
[2024-05-14T18:08:54.907Z] set thread name ThreadMemoryTracer is too long and will be truncated by system
[2024-05-14T18:08:54.907Z] ==================
[2024-05-14T18:08:54.907Z] WARNING: ThreadSanitizer: data race (pid=25225)
[2024-05-14T18:08:54.907Z]   Read of size 8 at 0x000014ca2680 by main thread:
[2024-05-14T18:08:54.907Z]     #0 std::__1::ios_base::width() const /usr/local/bin/../include/c++/v1/ios:523:12 (gtests_dbms+0x2c78509)
[2024-05-14T18:08:54.907Z]     #1 std::__1::ostreambuf_iterator<char, std::__1::char_traits<char> > std::__1::__pad_and_output<char, std::__1::char_traits<char> >(std::__1::ostreambuf_iterator<char, std::__1::char_traits<char> >, char const*, char const*, char const*, std::__1::ios_base&, char) /usr/local/bin/../include/c++/v1/locale:1396:29 (gtests_dbms+0x2c78509)
[2024-05-14T18:08:54.907Z]     #2 std::__1::basic_ostream<char, std::__1::char_traits<char> >& std::__1::__put_character_sequence<char, std::__1::char_traits<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char const*, unsigned long) /usr/local/bin/../include/c++/v1/ostream:718:17 (gtests_dbms+0x2c783e1)
[2024-05-14T18:08:54.907Z]     #3 std::__1::basic_ostream<char, std::__1::char_traits<char> >& std::__1::operator<<<char, std::__1::char_traits<char>, std::__1::allocator<char> >(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) /usr/local/bin/../include/c++/v1/ostream:1041:12 (gtests_dbms+0x111be873)
[...]

PR #9044, opened 2024-05-15
*: Fix too long thread name make TSAN failed to run

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9043

Problem Summary:

As the issue describe
 
Shorten the thread name to be within const char[15]

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/KVStore/FFI/JointThreadAllocInfo.cpp
+++ b/dbms/src/Storages/KVStore/FFI/JointThreadAllocInfo.cpp
@@ -27,7 +27,7 @@ namespace DB
 JointThreadInfoJeallocMap::JointThreadInfoJeallocMap()
 {
     monitoring_thread = new std::thread([&]() {
-        setThreadName("ThreadMemoryTracer");
+        setThreadName("ThdMemTrace");
         while (true)
         {
             using namespace std::chrono_literals;


Lloyd-Pottiger (approved):

dbms/src/Storages/KVStore/FFI/JointThreadAllocInfo.cpp
```diff
@@ -27,7 +27,7 @@ namespace DB
 JointThreadInfoJeallocMap::JointThreadInfoJeallocMap()
 {
     monitoring_thread = new std::thread([&]() {
-        setThreadName("ThreadMemoryTracer");
+        setThreadName("ThdMemTrace");
```
zanmato1984, line 30: How does the name length cause the issue in #9043? Is it just a warning or some real race caused by this long name? Or is it a false-positive?

CalvinNeo (approved):
lgtm

dbms/src/Storages/KVStore/FFI/JointThreadAllocInfo.cpp
JaySon-Huang: The `std::cerr` is not thread safe, and it cause data race between std::cerr output here and that in other loggings cause by `DB::Logger`.

This data race make `/tiflash/gtests_dbms: Command '['/tiflash/gtests_dbms', '--gtest_list_tests']' returned non-zero exit status 66` and TSAN fails to run following commands

zanmato1984: Two things I'm still not clear:
1. How does this thread name have anything to do with `std::cerr`?
2. How could simply shortening the name avoid the race?

JaySon-Huang: The `setThreadName` here is a cross-platform version. It will print message to `std::err` when the thread name is larger than 15 bytes 😂
https://github.com/pingcap/tiflash/blob/de9f362397cd0ea65c5cc747b762e44d2e03cc9c/dbms/src/Common/setThreadName.cpp#L38-L43

zanmato1984: Got it, thanks.

Maybe we should replace `std::cerr` with a reasonable logger in the future.

Issue #8869, opened 2024-03-26
Use `notify` instead of `polling` in pipeline model


SeaRise:
## Enhancement

- [x] UnorderedSourceOp https://github.com/pingcap/tiflash/pull/8872
- [x] SharedQueue https://github.com/pingcap/tiflash/pull/8875
- [x] ResultQueue https://github.com/pingcap/tiflash/pull/9065
- [x] ExchangeReceiver https://github.com/pingcap/tiflash/pull/9073
- [x] ExchangeSender https://github.com/pingcap/tiflash/pull/9072
- [x] RNSegmentSourceOp
    - Since TiFlash now uses the data sharing pool in its storage-compute separation mode, there's no need to refine RNSegmentSourceOp anymore.
- [x] HashJoinProbeTransformOp https://github.com/pingcap/tiflash/pull/9081
- [x] AggregateRestoreSourceOp https://github.com/pingcap/tiflash/pull/9082

windtalker:
close as all subtask is done.

PR #9072, opened 2024-05-22
Pipeline: use notify instead of polling for `ExchangeSender`

SeaRise:
### What problem does this PR solve?

Issue Number: ref #8869

Problem Summary:

### What is changed and how it works?

use `waitForWritable()` to replace `isWritable()`
```
    auto res = writer->waitForWritable();
    switch (res)
    {
    case WaitResult::Ready:
        return OperatorStatus::NEED_INPUT;
    case WaitResult::WaitForPolling:
        return OperatorStatus::WAITING;
    case WaitResult::WaitForNotify:
        return OperatorStatus::WAIT_FOR_NOTIFY;
    }
```

### Check List

Tests 

Tested tpch50, no performance regression

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #9081, opened 2024-05-23
Pipeline: use notify instead of polling for join probe

SeaRise:

Issue Number: ref #8869

Problem Summary:

```commit-message
use notify for wait join build/probe finished in pipeline model.
```

- add `OneTimeFuture` to implement `join probe`'s `wait for notify`.
- remove `HashJoinProbeTransformOp::awaitImpl`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Operators/HashJoinProbeTransformOp.cpp
```diff
@@ -251,7 +220,7 @@ OperatorStatus HashJoinProbeTransformOp::executeIOImpl()
         probe_transform->flushMarkedSpillData();
         probe_transform->finalizeProbe();
         switchStatus(ProbeStatus::WAIT_PROBE_FINISH);
-        return OperatorStatus::WAITING;
+        return OperatorStatus::HAS_OUTPUT;
```
windtalker, line 223: Why original code return `WAITING`?

SeaRise: Because later, we need to poll in the wait reactor for other task probes to finish before continuing with the restore build.
Now, return to the CPU thread pool to trigger wait-for-notify.

windtalker: you mean return to cpu thread pool, if `isProbeFinishedForPipeline` return false, it will go to `wait-for-notify` state?

SeaRise: yes

dbms/src/Flash/Pipeline/Schedule/Tasks/OneTimeNotifyFuture.h
```diff
@@ -0,0 +1,57 @@
...
+#pragma once
+
+#include <Flash/Pipeline/Schedule/Tasks/NotifyFuture.h>
+#include <Flash/Pipeline/Schedule/Tasks/PipeConditionVariable.h>
+
+#include <mutex>
+
+namespace DB
+{
+class OneTimeNotifyFuture : public NotifyFuture
```
windtalker, line 24: can you add some comments to explain why it is named as `OneTimeNotifyFuture`

SeaRise: ok, comment added.

windtalker (approved):
LGTM

windtalker:
hold since we need to double check the affects of #9072 and #9429

SeaRise:
> hold since we need to double check the affects of #9072 and #9429

typo: #9424

dbms/src/Operators/HashJoinProbeTransformOp.cpp
gengliqi: `onWaitProbeFinishDone` will be called instead of triggering wait-for-notify because `finalizeProbe` is called.

gengliqi (approved):

dbms/src/Operators/HashJoinProbeTransformOp.cpp
SeaRise: > `onWaitProbeFinishDone` will be called instead of triggering wait-for-notify because `finalizeProbe` is called.

yes, you are right.

PR #8872, opened 2024-03-26
Pipeline: use `notify` instead of `polling` for `UnorderedSourceOp`

SeaRise:
### What problem does this PR solve?

Issue Number: ref #8869

Problem Summary:

### What is changed and how it works?

add a new status `WAIT_FOR_NOTIFY` to register task to a notify future.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #9082, opened 2024-05-23
Pipeline: use notify instead of polling for `AggregateRestoreSourceOp`

SeaRise:

Issue Number: ref #8869

Problem Summary:

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Operators/SharedAggregateRestorer.cpp
```diff
@@ -48,6 +48,22 @@ bool SharedSpilledBucketDataLoader::switchStatus(SharedLoaderStatus from, Shared
     return status.compare_exchange_strong(from, to);
 }
 
+bool SharedSpilledBucketDataLoader::checkCancelled()
+{
+    if unlikely (exec_context.isCancelled())
+    {
+        {
+            std::lock_guard lock(mu);
+            if (is_cancelled)
```
windtalker, line 57: looks like use cas here is enough?

SeaRise: https://github.com/pingcap/tiflash/pull/9082/files#diff-b54d5b998f3a9de97d9452c7c40a8016e79ecbc2251784e96ac0fd6ee840fa16R159-R163
It's to ensure SharedSpilledBucketDataLoader::registerTask can immediately notify the task when cancelled, rather than moving the task to pipe_read_cv.

SeaRise: Otherwise, it may cause the task to be permanently in pipe_read_cv, unable to be notified when the query is cancelled.

windtalker: I mean if you use a atomicbool for `is_cancelled`, then you don't need to acquire lock here

dbms/src/Operators/SharedAggregateRestorer.cpp
```diff
@@ -66,7 +82,7 @@ void SharedSpilledBucketDataLoader::storeBucketData()
     RUNTIME_CHECK(status == SharedLoaderStatus::loading);
 
     // Although the status will always stay at `SharedLoaderStatus::loading`, but because the query has stopped, it doesn't matter.
-    if unlikely (exec_context.isCancelled())
+    if (checkCancelled())
```
windtalker, line 85: Why `SharedQuery` and `ResultQueue` need to be cancelled explicitly in `PipelineExecutorContext::cancel`, but this don't need to be cancelled in `PipelineExecutorContext::cancel`?

SeaRise: Because it will not be stuck forever, when the query is canceled, checkCancel will always be called, because wait_for_notify must be because the queue is empty. When the queue is empty, it will trigger an asynchronous load bucket event, and the load bucket event will call checkCancel.

windtalker: If ` trigger an asynchronous load bucket event` fails, will it still work?

SeaRise: yes, `Event::finishImpl` is always called, regardless of failure or cancellation
https://github.com/pingcap/tiflash/blob/a4f0ec6ea19659062e2596c772a9bfcfab115c07/dbms/src/Flash/Pipeline/Schedule/Events/Impls/LoadBucketEvent.cpp#L29-L33

SeaRise: https://github.com/pingcap/tiflash/pull/9082/files#diff-b54d5b998f3a9de97d9452c7c40a8016e79ecbc2251784e96ac0fd6ee840fa16R119-R120

windtalker: OK. I want to confirm another unrelated question here, why `WorkQueue` in this [pr](https://github.com/pingcap/tiflash/pull/8872/files) don't need to be cancelled explicitly in `PipelineExecutorContext::cancel`?

SeaRise: When the data in the storage layer is fully read, notify will definitely be called.
When the data in the storage layer is not fully read, there will always be a block, and notify will also be called.

windtalker: When query meet error, the storage layer is not informed?

SeaRise: Yes, the storage layer is not explicitly canceled.

SeaRise: However, the query will not be stuck, because the storage layer will continuously generate blocks, which will trigger checkCancel, thus ending the query.

windtalker: Storage layer generate block => pipeline task is notified and finishes since it found the query is cancelled => `~UnorderedSourceOp()` is called, and in side `~UnorderedSourceOp()`, it will call `task_pool->decreaseUnorderedInputStreamRefCount()`, which eventually stop the storage read?

SeaRise: > Storage layer generate block => pipeline task is notified and finishes since it found the query is cancelled => `~UnorderedSourceOp()` is called, and in side `~UnorderedSourceOp()`, it will call `task_pool->decreaseUnorderedInputStreamRefCount()`, which eventually stop the storage read?

yes

windtalker (approved):
LGTM

PR #9115, opened 2024-06-04
*: Introduce MutexProtected and SharedMutexProtected

Lloyd-Pottiger:

Issue Number: ref #6233

Problem Summary:

**Introduce MutexProtected and SharedMutexProtected**

In C++, there is a RAII class for automatic locking & unlocking:

```cpp
{
    MutexLocker locker(thing->mutex);
    use(thing->field);
}
```

The MutexLocker locks the mutex when constructed and unlocks it when destroyed. No need for a manual call to mutex_unlock() anymore.

That’s already pretty good! However, the MutexLocker approach still has some major shortcomings:

1. You can still forget to lock the mutex and access field anyway.
2. Developers who are unfamiliar with this code may not realize that mutex and field have this important relationship.

So we introduce MutexProtected and SharedMutexProtected, which makes it significantly easier to use mutexes correctly

```cpp
struct Thing {
    MutexProtected<Field> field;
};

thing->field.with([&](Field& field) {
    use(field);
});
```

Essentially, MutexProtected<T> is a bundled mutex and T. However, you can’t access the T directly! The only way we’ll let you access the T is by calling with() and passing it a callback that takes a T& parameter.

When called, with() locks the mutex, then invokes the callback, and finally unlocks the mutex again before returning.

As you can see, we’ve now also solved the issue of someone forgetting to lock the mutex before accessing the field. And not only that, but since the mutex and field have been combined into a single variable, you no longer have to be aware of the relationship between the two. It’s been encoded into the type system!

Reference: https://awesomekling.github.io/MutexProtected-A-C++-Pattern-for-Easier-Concurrency/

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Common/SharedMutexProtected.h
```diff
@@ -0,0 +1,116 @@
...
+
+        ALWAYS_INLINE inline U const & get() const { return &m_value; }
+        ALWAYS_INLINE inline U & get()
+            requires(!std::is_const_v<U>)
+        {
+            return &m_value;
+        }
+
+    private:
+        U & m_value;
```
CalvinNeo, line 65: Name should conform to the current codebase. Remove the `m_` prefix.

CalvinNeo:
There are `MutexLockWrap` and `SharedMutexLockWrap`. What's the difference in `MutexProtected` and `SharedMutexProtected`? Can we unified them?

dbms/src/Common/SharedMutexProtected.h
```diff
@@ -0,0 +1,116 @@
...
+template <typename T>
+class SharedMutexProtected
+{
+    DISALLOW_COPY_AND_MOVE(SharedMutexProtected);
+
+private:
+    template <typename U, typename Lock>
+    class Locked
+    {
+        DISALLOW_COPY_AND_MOVE(Locked);
```
CalvinNeo, line 35: Why can't it move?

dbms/src/Common/MutexProtected.h
```diff
@@ -0,0 +1,101 @@
...
+    DISALLOW_COPY_AND_MOVE(MutexProtected);
+
+private:
+    template <typename U>
+    class Locked
+    {
+        DISALLOW_COPY_AND_MOVE(Locked);
+
+    public:
+        Locked(U & value, std::mutex & mutex)
```
CalvinNeo, line 36: Why introduce `U`, and why use `U &`?

Lloyd-Pottiger: 1. To distinct `T` and `const T`.
2. MutexProtected::Lock::value refer to MutexProtected::value

dbms/src/Common/SharedMutexProtected.h
Lloyd-Pottiger: `U` maybe non-moveable

dbms/src/Common/MutexProtected.h
CalvinNeo: Then can we use universal reference?

dbms/src/Common/SharedMutexProtected.h
CalvinNeo: U is not movable, a lock to U could be movable though IMO?

dbms/src/Common/MutexProtected.h
Lloyd-Pottiger: Sorry, what does universal reference mean? Now it does refer to outer class value.

dbms/src/Common/SharedMutexProtected.h
Lloyd-Pottiger: But it is not only a lock, It all contains the reference.

dbms/src/Common/MutexProtected.h
CalvinNeo: You can check "perfect forwarding" and "universal reference". I think what you want to do is something like "reference collapsing".

dbms/src/Common/SharedMutexProtected.h
CalvinNeo: What's the difference then? References and pointers can be safely elsewhere.

dbms/src/Common/MutexProtected.h
Lloyd-Pottiger: Sorry, still can not get your point, it is a lvalue reference.

Lloyd-Pottiger: `const T &` or `T &`

CalvinNeo: OK, I realized then...

dbms/src/Common/SharedMutexProtected.h
Lloyd-Pottiger:
```cpp
{
    SharedMutexProtected<> mp;
    auto lock = mp.lockShared();
    do_something;
    auto task = [std::move(lock)] () { // seems reasonable, because the lock can not be null in task.
       lock->xxx
   }
   add_task_to_thread_pool(task);
}

// thread_pool execute task here.
```

CalvinNeo: Yes, this is one case. However, IMO, preventing moving doesn't ensure the lock always refer to a valid pointer

Lloyd-Pottiger: The `value` can not be accessed. We only need to make sure return `Lock` is valid.

Lloyd-Pottiger:
> There are `MutexLockWrap` and `SharedMutexLockWrap` already. IMO, the new class here is also to make acquiring the lock the only way to access a field. It provides the `with` function though. So, can we unified them? It is confusing if we have two versions of utils which usages are similar.

`MutexLockWrap` and `SharedMutexLockWrap` seems only contain a mutex and provide some function to acquire a lock. By inherit one of them, we can get the mutex and these functions.

`MutexProtected` and `SharedMutexLockWrap` are mainly to warp a field and its mutex, and we do not need to worry about whether there are some places that access the field but forget to acquire a lock.

dbms/src/Common/SharedMutexProtected.h
CalvinNeo: > The value can not be accessed
I don't get that. I find `operator->` can access `value` in `Locked`. So I think the Locked should not outlive `value` which it is protected.

CalvinNeo (approved):

CalvinNeo:
> > There are `MutexLockWrap` and `SharedMutexLockWrap` already. IMO, the new class here is also to make acquiring the lock the only way to access a field. It provides the `with` function though. So, can we unified them? It is confusing if we have two versions of utils which usages are similar.
> 
> `MutexLockWrap` and `SharedMutexLockWrap` seems only contain a mutex and provide some function to acquire a lock. By inherit one of them, we can get the mutex and these functions.
> 
> `MutexProtected` and `SharedMutexLockWrap` are mainly to warp a field and its mutex, and we do not need to worry about whether there are some places that access the field but forget to acquire a lock.

I think they both aimed to provide some way to ensure a lock-raii sematics, which means a resource could only be accessed by acquiring a lock, and the lock will be released automaticlly when the resource is no longer accessed.

The difference is the `XXXProtected` methods are more elaberate.

Lloyd-Pottiger:
> I don't get that. I find operator-> can access value in Locked. So I think the Locked should not outlive value which it is protected.

Lock can access `value`, so Lock can not be moved, otherwise the Locke maybe outlive `value`.

CalvinNeo:
> value

IMO, Lock can still outlive value, even if it is pinned

Lloyd-Pottiger:
> > value
> 
> IMO, Lock can still outlive value, even if it is pinned

Could you please give an example?

CalvinNeo:
> > > value
> > 
> > 
> > IMO, Lock can still outlive value, even if it is pinned
> 
> Could you please give an example?

Consider the value referenced is not a auto-scoped variable on stack, but allocated on heap and managed by some `std::map` or `std::vector`. Then I can destroy the value in another thread.

Lloyd-Pottiger:
> > > > value
> > > 
> > > 
> > > IMO, Lock can still outlive value, even if it is pinned
> > 
> > 
> > Could you please give an example?
> 
> Consider the value referenced is not a auto-scoped variable on stack, but allocated on heap and managed by some `std::map` or `std::vector`. Then I can destroy the value in another thread.

The provided constructor does not allow this.

CalvinNeo:
> > > > > value
> > > > 
> > > > 
> > > > IMO, Lock can still outlive value, even if it is pinned
> > > 
> > > 
> > > Could you please give an example?
> > 
> > 
> > Consider the value referenced is not a auto-scoped variable on stack, but allocated on heap and managed by some `std::map` or `std::vector`. Then I can destroy the value in another thread.
> 
> The provided constructor does not allow this.

I don't quite understand

Lloyd-Pottiger:
> > > > > IMO, Lock can still outlive value, even if it is pinned
> > > > 
> > > > 
> > > > Could you please give an example?
> > > 
> > > 
> > > Consider the value referenced is not a auto-scoped variable on stack, but allocated on heap and managed by some `std::map` or `std::vector`. Then I can destroy the value in another thread.
> > 
> > 
> > The provided constructor does not allow this.
> 
> I don't quite understand

IMO, your example can not be implemented basing on this PR. Or could you please show the code?

CalvinNeo:
> > > > > > IMO, Lock can still outlive value, even if it is pinned
> > > > > 
> > > > > 
> > > > > Could you please give an example?
> > > > 
> > > > 
> > > > Consider the value referenced is not a auto-scoped variable on stack, but allocated on heap and managed by some `std::map` or `std::vector`. Then I can destroy the value in another thread.
> > > 
> > > 
> > > The provided constructor does not allow this.
> > 
> > 
> > I don't quite understand
> 
> IMO, your example can not be implemented basing on this PR. Or could you please show the code?

OK, I found that the `lockConst` method is private, so the only way to access the protected variable is through `with`.

JaySon-Huang (approved):

Issue #9132, opened 2024-06-07
Rename partitioned table may leave some partition on the old database


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

minimal reproduce step
```
CREATE TABLE employees (
    id INT NOT NULL,
    store_id INT NOT NULL
)PARTITION BY RANGE (store_id) (
    PARTITION p0 VALUES LESS THAN (6),
    PARTITION p1 VALUES LESS THAN (11),
    PARTITION p2 VALUES LESS THAN (16),
    PARTITION p3 VALUES LESS THAN (21)
);
alter table employees set tiflash replica 2;

-- (1,1),(2,2),(3,3) => p0
-- p1 is empty
-- (11,11) => p2
-- (16,16) => p3
insert into employees(id, store_id) values(1,1),(2,2),(3,3),(11,11),(16,16);
rename test.employees to mytest.employees;
```

original post on https://asktug.com/t/topic/1026991/34

### 2. What did you expect to see? (Required)

All partitions are renamed to the new database and tiflash run normally

### 3. What did you see instead (Required)

Some partition is not renamed to the new database. If the schema of partition table changed, tiflash may fail to restart.

### 4. What is your TiFlash version? (Required)
v7.5.2

PR #9133, opened 2024-06-07
ddl: Fix rename partitioned table is not atomic

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9132

Problem Summary:

In the previous implementation, when running `RENAME` for a partitioned table, it will skip renaming the following partitions if one of the IStorage instance does not exist.

https://github.com/pingcap/tiflash/blob/73929ad45d21e87ec71e7ce0d19e71a4cd5c50bf/dbms/src/TiDB/Schema/SchemaBuilder.cpp#L667-L691

That work well on the versions before v7.2. But after DDL framework refactor since v7.2, there could be a chance that the IStorage instance is not created.
If there are "p1, p2, p3" for a partitioned table, but "p2" is empty. Then the IStorage instance of "p2" is not created. When we need to rename the partitioned table to another database, it will make that "p1" is renamed to the new database, but "p3" remain in the old database. This cause inconsistency.

```commit-message

```

1. `applyRenameLogicalTable` will continue the rename for remaining partitions even if it happen to a partition that does not have `IStorage` instance
2. There could be a chance that tiflash restart when applying `applyRenameLogicalTable`, making some partitions is in the old database while some is in the new database. After restart, `syncAllSchema` will call `tryFixPartitionsBelongingDatabase` and check whether there are any partitions that does not belong to the database on disk. If so, rename the partition table to the correct database.

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that tiflash may panic after renaming a partitioned table with empty partition to a new database
```


dbms/src/Debug/MockTiDB.cpp
```diff
@@ -683,6 +725,45 @@ void MockTiDB::renameTable(const String & database_name, const String & table_na
...
+    RUNTIME_CHECK_MSG(
+        new_database != databases.end(),
+        "new_database is not exist in MockTiDB, new_database_name={}",
+        new_database_name);
+    table->database_id = new_database->second; // set new_database_id
+
+    TableInfo new_table_info = table->table_info;
+    new_table_info.name = new_table_name;
+    auto new_table
+        = std::make_shared<Table>(database_name, table->database_id, new_table_name, std::move(new_table_info));
```
Lloyd-Pottiger, line 751:
```suggestion
    auto new_table
        = std::make_shared<Table>(new_database_name, table->database_id, new_table_name, std::move(new_table_info));
```

typo?

dbms/src/Debug/MockTiDB.cpp
```diff
@@ -683,6 +725,45 @@ void MockTiDB::renameTable(const String & database_name, const String & table_na
     version_diff[version] = diff;
 }
 
+void MockTiDB::renameTableTo(
+    const String & database_name,
+    const String & table_name,
+    const String & new_database_name,
+    const String & new_table_name)
+{
+    std::lock_guard lock(tables_mutex);
```
CalvinNeo, line 734: Use scoped_lock

dbms/src/Debug/MockTiDB.cpp
JaySon-Huang: yes, fixed.

dbms/src/Debug/MockTiDB.cpp
JaySon-Huang: fixed

Lloyd-Pottiger (approved):

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -1439,23 +1453,122 @@ void SchemaBuilder<Getter, NameMapper>::syncAllSchema()
...
+            continue;
+        }
+
+        for (auto tbl_iter = db_ptr->getIterator(context); tbl_iter->isValid(); tbl_iter->next())
+        {
+            const auto table_name = tbl_iter->table()->getTableName();
+            auto opt_tbl_id = SchemaNameMapper::tryGetTableID(table_name);
+            if (!opt_tbl_id)
+            {
+                LOG_WARNING(
```
CalvinNeo, line 1507: In what case could this happen?

dbms/src/TiDB/Schema/SchemaBuilder.cpp
```diff
@@ -1439,23 +1453,122 @@ void SchemaBuilder<Getter, NameMapper>::syncAllSchema()
...
+
+            LOG_WARNING(
+                log,
+                "FixPartitionsDatabase: try to fix the partition that is not belong to this database, "
+                "db_name={} database_id={} table_id={} new_database_id={}",
+                db_name,
+                database_id,
+                *opt_tbl_id,
+                new_database_id);
+            auto managed_storage = std::dynamic_pointer_cast<IManageableStorage>(tbl_iter->table());
```
CalvinNeo, line 1543: Also protect the cast?

dbms/src/TiDB/Schema/SchemaBuilder.cpp
JaySon-Huang: Cases under "tests/delta-merge-test" may create table with the given table_name rather that "t_${table_id}"

dbms/src/TiDB/Schema/SchemaBuilder.cpp
JaySon-Huang: Added

Issue #8977, opened 2024-04-22
log densensitation in tiflash


JaySon-Huang:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

In https://github.com/pingcap/tiflash/issues/1277, we introduce `security.redact_info_log` to replace the rowkey by `?` in the tiflash logging.

TiDB introduce a "marker" mode for redact https://github.com/pingcap/tidb/issues/51306 since v8.1

Mark sensitive data with `‹..›`. when `security.redact_info_log=marker`. Note that `‹` is not "less" `<` but [U+2039](https://unicode.scarfboy.com/?s=U%2b2039). If the raw data contains `‹`, `‹` will be escaped to `‹‹`, similar to `›`. This is a new feature:

1. TiFlash will produce logs with sensitive data marked with `‹..›`
2. Cloud or OP users could desensitive logs on their own by tidb-server sub commands or whatever.

It is basically because that users want the original log, and we (primarily tech supports) should only see desensitived logs.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

JaySon-Huang:
* [] client-c: https://github.com/tikv/client-c/pull/184
* [] tiflash: https://github.com/pingcap/tiflash/pull/9136
* [] docs update: https://github.com/pingcap/docs-cn/pull/17627

Issue #1277, opened 2020-12-10
Log redact on TiFlash


JaySon-Huang:
Log redacted:  Avoid printing user data (mostly user keys) to info log. Replace the value into `?`.
[Spec](https://docs.google.com/document/d/1fBwsY5TJMchD2gRyvrcjK3_Bn3cHMGa0TiXmgtrboWo/edit#heading=h.g2e0meapu1x2)

Logs to be redacted:

- [ ] TiFlash process
  - [x] `TiKVKeyValue`,`RowKeyRange`,`HandleRange` https://github.com/pingcap/tics/pull/1282
  - [x] Value in RSFilter https://github.com/pingcap/tics/pull/1282
  - [ ]  Protobuf struct in logging
  - [x] Client-C resolve lock https://github.com/tikv/client-c/pull/59 https://github.com/pingcap/tics/pull/1306
- [x] ClusterManager
- [x] TiFlashProxy

Related issue in TiDB: https://github.com/pingcap/tidb/issues/18566
Related PR in TiKV: https://github.com/tikv/tikv/pull/8670

- Documents https://github.com/pingcap/docs/pull/4412, https://github.com/pingcap/docs-cn/pull/5094

JaySon-Huang:
close as the `Protobuf struct in logging` is not print to logging because the default level is INFO

PR #9136, opened 2024-06-09
security: Add "marker" mode for log densensitation

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/8977

Problem Summary:

In the previous PR, https://github.com/pingcap/tiflash/issues/1277, tiflash support setting `security.redact_info_log` to be set as "true" / "false".
If the configuration item is set to true, all user data in the log is replaced by `?`.

```commit-message
Support setting config "security.redact_info_log=marker" for log densensitation
```

Mark sensitive data with `‹..›`. when `security.redact_info_log=marker`. Note that `‹` is not "less" `<` but [U+2039](https://unicode.scarfboy.com/?s=U%2b2039). If the raw data contains `‹`, `‹` will be escaped to `‹‹`, similar to `›`.

Introduce `Redact::toMarkerString` to handle marking sensitive data with `‹..›` and escaping. 

* * *

* [x] Pending for https://github.com/tikv/client-c/pull/184

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support "marker" for log densensitation
```


dbms/src/Common/FieldVisitors.cpp
```diff
@@ -221,57 +221,121 @@ String FieldVisitorToString::operator()(const Tuple & x_def) const
 
 String FieldVisitorToDebugString::operator()(const Null &) const
```
Lloyd-Pottiger, line 222: Can we use template to reduce similar code?

JaySon-Huang: The `UInt64/Int64/DecimalField<...>` share the similar code, but other types are not.
We can try to eliminate it using `std::enable_if`, but it is not related to this PR change, maybe we can do it in another PR.

Lloyd-Pottiger (approved):

Issue #9233, opened 2024-07-15
Concurrency issue between rename table and `applyTable`


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 3. What did you see instead (Required)

https://ci.pingcap.net/blue/organizations/jenkins/tiflash-ghpr-integration-tests/detail/tiflash-ghpr-integration-tests/16561/pipeline/

[fullstack-test2-logs.tar.gz](https://github.com/user-attachments/files/16231251/fullstack-test2-logs.tar.gz)

```
[2024-07-12T04:46:15.000Z] fullstack-test2/ddl/rename_table_across_databases.test: Running
[2024-07-12T04:46:33.020Z]   File: fullstack-test2/ddl/rename_table_across_databases.test
[2024-07-12T04:46:33.020Z]   Error line: 117
[2024-07-12T04:46:33.020Z]   Error: set session tidb_isolation_read_engines='tiflash'; select * from test_new.part4 order by id;
[2024-07-12T04:46:33.020Z]   Result:
[2024-07-12T04:46:33.020Z]     ERROR 1105 (HY000) at line 1: other error for mpp stream: Code: 107, e.displayText() = DB::Exception: Cannot open file /tmp/tiflash/data/db/metadata/db_708/t_713.sql, errno: 2, strerror: No such file or directory, e.what() = DB::Exception,
[2024-07-12T04:46:33.020Z]   Expected:
[2024-07-12T04:46:33.020Z]     +----+----------+------+
[2024-07-12T04:46:33.020Z]     | id | store_id | c1   |
[2024-07-12T04:46:33.020Z]     +----+----------+------+
[2024-07-12T04:46:33.020Z]     |  1 |        1 | NULL |
[2024-07-12T04:46:33.020Z]     |  2 |        2 | NULL |
[2024-07-12T04:46:33.020Z]     |  3 |        3 | NULL |
[2024-07-12T04:46:33.020Z]     | 11 |       11 | NULL |
[2024-07-12T04:46:33.020Z]     | 16 |       16 | NULL |
[2024-07-12T04:46:33.020Z]     +----+----------+------+
```

### 4. What is your TiFlash version? (Required)
release-7.5

JaySon-Huang:
Thread-A enter `TiDBSchemaSyncer::syncSchemaDiffs` and run into `RenameTable`. That overwrite the id_mapping for table_id=710 old_database_id=2 new_database_id=708.
```
[...]

PR #9246, opened 2024-07-22
ddl: Atomic rename `DeltaMergeStore::db_name` in memory

JaySon-Huang:

Issue Number: ref https://github.com/pingcap/tiflash/issues/9233

Problem Summary:
A part of fixing #9233. In the previous code, updating `DeltaMergeStore::db_name` is not atomic.

```commit-message
Make renaming `DeltaMergeStore::db_name` to be atomic
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.h
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.h
@@ -277,8 +277,8 @@ public:
 
     void setUpBackgroundTask(const DMContextPtr & dm_context);
 
-    const String & getDatabaseName() const { return db_name; }
-    const String & getTableName() const { return table_name; }
+    String getDatabaseName() const { return table_meta.lockShared()->db_name; }
+    String getTableName() const { return table_meta.lockShared()->table_name; }
 
     void rename(String new_path, String new_database_name, String new_table_name);
 
@@ -796,8 +796,14 @@ public:
     Settings settings;
     StoragePoolPtr storage_pool;
 
-    String db_name;
-    String table_name;
+    struct TableMeta
+    {
+        String db_name;
+        String table_name;
+    };
+    SharedMutexProtected<TableMeta> table_meta;
+    // String db_name;
+    // String table_name;
 
     const KeyspaceID keyspace_id;
     const TableID physical_table_id;
@@ -838,7 +844,7 @@ public:
     mutable std::shared_mutex read_write_mutex;
 
     LoggerPtr log;
-}; // namespace DM
+};
 
 using DeltaMergeStorePtr = std::shared_ptr<DeltaMergeStore>;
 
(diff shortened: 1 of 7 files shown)


dbms/src/Storages/DeltaMerge/DeltaMergeStore.h
```diff
@@ -277,8 +277,17 @@ class DeltaMergeStore : private boost::noncopyable
...
+    struct TableMeta
+    {
+        String db_name;
+        String table_name;
+    };
+    TableMeta getTableMeta() const
+    {
+        auto meta = table_meta.lockShared();
+        return TableMeta{meta->db_name, meta->table_name};
+    }
```
Lloyd-Pottiger, line 289: separate to `getDbName` and `getTableName`?

JaySon-Huang: No, I find that the code we need "database_name" of an IManageable instance usually also requires the table name.  So I would rather remove the `IManageableStorage::getDatabaseName` to reduce the cost of acquiring the lock instead of calling `getDatabaseName` and `getTableName` separately (in later PR)

dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp
```diff
@@ -700,11 +700,7 @@ SegmentPtr DeltaMergeStore::gcTrySegmentMerge(const DMContextPtr & dm_context, c
     auto segment_bytes = segment->getEstimatedBytes();
     if (segment_rows >= dm_context->segment_limit_rows || segment_bytes >= dm_context->segment_limit_bytes)
     {
-        LOG_TRACE(
-            log,
-            "GC - Merge skipped because current segment is not small, segment={} table={}",
-            segment->simpleInfo(),
-            table_name);
+        LOG_TRACE(log, "GC - Merge skipped because current segment is not small, segment={}", segment->simpleInfo());
```
JaySon-Huang, line 703: Note: The logger of `DeltaMergeStore` contains the keyspace and table_id as a suffix. So we don't need to log down the table_name in these methods.
https://github.com/pingcap/tiflash/blob/ce3d9e124e7949d053603c0e6f7be5b30d64a14d/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp#L231

dbms/src/Storages/DeltaMerge/DeltaMergeStore.h
Lloyd-Pottiger: ok

PR #9255, opened 2024-07-25
Fix the wrong result bug when casting string as datetime with time zone or illegal chars

### LINKED ISSUE #8754: cast string as datetime with time zone results in loss of precision
## Bug Report

related to https://github.com/pingcap/tidb/issues/49555

```
use test;
drop table if exists t;
create table t(a varchar(50));
alter table test.t set tiflash replica 1;
insert into t values ('2020-01-01 12:00:00.123456 +0600 PST');
insert into t values ('2020-01-01 12:00:00.123456 -0600 PST');
set tidb_allow_mpp=1;set tidb_enforce_mpp=1; set tidb_isolation_read_engines='tiflash';
select cast(a as datetime(3)) from t;
```

```
+-------------------------+
| cast(a as datetime(3))  |
+-------------------------+
| 2020-01-01 12:00:00.123 |
| 2020-01-01 12:00:00.123 |
+-------------------------+
```
```
+-------------------------+
| cast(a as datetime(3))  |
+-------------------------+
| 2020-01-01 12:00:00.000 |
| 2020-01-01 12:00:00.000 |
+-------------------------+
```
v8.0.0-alpha-126-gc0476a0

solotzg:

Issue Number: close #8754

Problem Summary:

```commit-message
Fix the wrong result bug when casting string as datetime with time zone or illegal chars
- func `getFracIndex` ignore time zone.
- func `parseFrac` ignore illegal suffix after microsecond format.
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the wrong result bug when casting string as datetime with time zone or illegal chars
```


guo-shaoge (approved):

dbms/src/Common/MyTime.cpp
```diff
@@ -745,7 +745,7 @@ inline bool numberToDateTime(Int64 number, MyDateTime & result, bool allowZeroDa
 }
 
 // returns frac, overflow, matched. eg., "999" fsp=2 will overflow.
-std::tuple<UInt32, bool, bool> parseFrac(const String & str, int8_t fsp)
+std::tuple<UInt32, bool, bool> parseFrac(const std::string_view str, int8_t fsp)
```
windtalker, line 748: The caller don't need to be updated after the input argument type is changed?

solotzg: In this case, `const String &` and `std::string_view` have the same effect. The caller won't change the original string buffer.

windtalker (approved):
LGTM

PR #9328, opened 2024-08-15
Storage: let segment meta using protobuf format

Lloyd-Pottiger:
### What problem does this PR solve?

Issue Number: ref #9036

Problem Summary: 

### What is changed and how it works?

In order to provide flexible for building vector index in the DeltaVS, store the segment meta using protobuf format.

```commit-message
Storage: let segment meta using protobuf format
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

Issue #9036, opened 2024-05-11
Support lightweight ingesting optional index


JaySon-Huang:
## Enhancement

In some scenarios, we need to create new indexes for existing data to speed up queries. Typically this can be done by triggering a major compaction (Segment DeltaMerge) that creates new index data during the major compaction. However, this involves rewriting all the columns, which makes indexing slow and has a large impact on system IO and CPU usage.

To mitigate the impact of index creation on the system, we can create new indexes **only for the stable layer data** in a more lightweight way. Because the stable layer contains 95% of the whole dataset, usually indexing the stable layer can bring sufficient performance boosting.

The overall procedure:

1. Prepare - create a background task that reads only the columns relevant to index generation, then store the generated the index file to persisted storage.
2. Ingest - ingest the index file into stable's DMFile. In a DMFile that uses meta v2, this only requires modifying the information in the ExtendColumnStat block in meta v2. DMFile::ingestIndex creates a new meta v2 file and atomically replaces the old meta v2 file.
3. Apply - update the epoch value of the StableValueSpace in the Segment. Under store-compute separation, the ComputeNode needs to decide whether it needs to clean up the old DMFile meta file cache and re-download the file from S3 based on this value.

This ingesting index mechanism can help us handle indexes that are needed for Vector Search (https://github.com/pingcap/tiflash/issues/9032) / Full Text search, or other optional indexes.

JaySon-Huang:
Close as we have implemented it for supporting [vector search](https://github.com/pingcap/tiflash/issues/9032).

Checkout the following logic for details

- `DeltaMergeStore::segmentEnsureStableIndex`
- `DMFileV3IncrementWriter`

PR #9343, opened 2024-08-22
Storage: support restore segments from checkpoint with format version V7

Lloyd-Pottiger:

Issue Number: ref #9036

Problem Summary:

In https://github.com/pingcap/tiflash/pull/9328, we add a new segment format version V7, and support restore from local and remote source but do not support restore from checkpoint, this PR fix it.

```commit-message
Storage: support restore segments from checkpoint with format version V7
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFilePersisted.cpp
```diff
@@ -182,17 +177,15 @@ ColumnFilePersisteds deserializeSavedColumnFiles(
...
-        column_files = deserializeSavedColumnFilesInV4Format(context, segment_range, meta);
-        break;
+        return deserializeSavedColumnFilesInV4Format(context, segment_range, meta);
     }
     default:
         throw Exception(
             ErrorCodes::LOGICAL_ERROR,
             "Unexpected delta value version: {}, latest version: {}",
             version,
-            DeltaFormat::V3);
```
JaySon-Huang, line 193: maybe remove this default branch, so that the compiler can throw error when we miss adding the new branch for DeltaFormat?

Lloyd-Pottiger: No.

1. DeltaFormat is not enum type.
2. It is possible that the version is larger than 4 (when data is broken) during deserializing, and we should throw an exception.

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileBig.cpp
```diff
@@ -140,6 +138,23 @@ ColumnFilePersistedPtr ColumnFileBig::createFromCheckpoint(
...
+    UniversalPageStoragePtr temp_ps,
+    WriteBatches & wbs)
+{
+    UInt64 file_page_id = cf_pb.id();
+    size_t valid_rows = cf_pb.valid_rows();
+    size_t valid_bytes = cf_pb.valid_bytes();
+
+    auto remote_data_store = dm_context.global_context.getSharedContextDisagg()->remote_data_store;
+    auto dmfile = restoreDMFileFromCheckpoint(dm_context, remote_data_store, temp_ps, wbs, file_page_id);
+    auto * dp_file = new ColumnFileBig(dmfile, valid_rows, valid_bytes, target_range);
```
CalvinNeo, line 154: Why don't we use make_shared here?

Lloyd-Pottiger: private constructor

JaySon-Huang (approved):
LGTM

dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileBig.cpp
CalvinNeo: Why not use a static function?
```
struct A {
private:
    A() {}
public:
    static std::shared_ptr<A> makeA() {
        return std::make_shared<A>();
    }  
};
```

Lloyd-Pottiger: It is just the same as public constructor?

```cpp
struct A {
public:
    A() {}
};
```

CalvinNeo: OK... May be I should firstly ask the question that why the constructor is private?

Lloyd-Pottiger: IMHO, it is used to make sure the stat (`valid_rows`, `valid_bytes`, and `segment_range`) is right.

Lloyd-Pottiger: the public constructor will calculate stat:

```cpp
ColumnFileBig::ColumnFileBig(const DMContext & context, const DMFilePtr & file_, const RowKeyRange & segment_range_)
    : file(file_)
    , segment_range(segment_range_)
{
    calculateStat(context);
}

void ColumnFileBig::calculateStat(const DMContext & dm_context)
{
    auto index_cache = dm_context.global_context.getMinMaxIndexCache();

    auto pack_filter = DMFilePackFilter::loadFrom(
        file,
        index_cache,
        /*set_cache_if_miss*/ false,
        {segment_range},
        EMPTY_RS_OPERATOR,
        {},
        dm_context.global_context.getFileProvider(),
        dm_context.getReadLimiter(),
        dm_context.scan_context,
        /*tracing_id*/ dm_context.tracing_id,
        ReadTag::Internal);

    std::tie(valid_rows, valid_bytes) = pack_filter.validRowsAndBytes();
}
```

CalvinNeo: Alright, then this issue is better not to be addressed on this PR.

CalvinNeo (approved):
lgtm

PR #9073, opened 2024-05-22
Pipeline: use notify instead of polling for `ExchangeReceiver`

SeaRise:
### What problem does this PR solve?

Issue Number: ref #8869

Problem Summary:

### What is changed and how it works?

The cpu usage of wait reactor is reduced to 0 in tpch50.
c72a55fd-0cd5-4fe2-87e2-beb5e5631d0e

### Check List

Tests 

Tested tpch50, no performance regression

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #9424, opened 2024-09-09
Fix TiFlash hang issue after #9072

### LINKED ISSUE #9413: tpch query 10 hangs after #9072
## Bug Report

introduced by https://github.com/pingcap/tiflash/pull/9072

windtalker:

Issue Number: close #9413

Problem Summary:

The root casue of the hang issue introduced by #9072 is
1. `ExchangeSenderSinkOp` eventually write data to `LooseBoundedMPMCQueue`, although `LooseBoundedMPMCQueue` is named with `LooseBounded`, it actually has a upper bound
2. For the same `MPPTunnel`(which holds a `LooseBoundedMPMCQueue`), all the `ExchangeSenderSinkOp` will try to write data the the same `LooseBoundedMPMCQueue` concurrently
3. If current MPPTunnel is not writable(its `LooseBoundedMPMCQueue` is full), `ExchangeSenderSinkOp` will register itself to the pipeline notify furture(the `tunnelSender`)
4. pipeline notify future notify one task at a time when the consumer of `LooseBoundedMPMCQueue` read a message from `LooseBoundedMPMCQueue`
5. In current implementation, when an `ExchangeSenderSinkOp` is notified in stage 4, it is not 100% sure that the `ExchangeSenderSinkOp` will write data to `LooseBoundedMPMCQueue` because
    * `ExchangeSenderSinkOp` only calls `writer->write(block);` to write the data, and inside `writer->write(block)` the data can be cached in `writer` instead of writting to `LooseBoundedMPMCQueue`
    * If current `block` is empty, it will call `writer->flush()` to flush the cached data, and if there is no cached data, it will not write to `LooseBoundedMPMCQueue`
6. Consider a case that there is `M` `ExchangeSenderSinkOp`, and the `LooseBoundedMPMCQueue` has a limited size of `N`, where `M > N`, and all the `M` `ExchangeSenderSinkOp` tries to write to a full `LooseBoundedMPMCQueue` at the same time. Then all the `M` `ExchangeSenderSinkOp` are registered to the pipeline notify future. As described in stage 4, the pipeline notify future only notify one task at a time when the consumer of `LooseBoundedMPMCQueue` read a message from `LooseBoundedMPMCQueue`, then at most `N` `ExchangeSenderSinkOp` will be notified. If all the notified `N` `ExchangeSenderSinkOp` do not write data to `LooseBoundedMPMCQueue`, then the `LooseBoundedMPMCQueue` become empty queue, and there is still `M - N` `ExchangeSenderSinkOp` waiting on the pipeline notify future. Since `LooseBoundedMPMCQueue` is empty, all the `M - N` `ExchangeSenderSinkOp` have on chance to be notified. Then the whole query hangs.

There is 2 possible fix
* For each read from `LooseBoundedMPMCQueue`, triger a notification if current queue is empty
* Make sure that each time an `ExchangeSenderSinkOp` is notified, the `ExchangeSenderSinkOp` should either write data to `LooseBoundedMPMCQueue`, or try to notify another `ExchangeSenderSinkOp`

The first fix should be earier, but consider that `ExchangeReceiver` will also using the notify way after #9073, the first fix may not work in the furture, so this pr use the second fix. 

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Mpp/MPPTunnelSetHelper.cpp
+++ b/dbms/src/Flash/Mpp/MPPTunnelSetHelper.cpp
@@ -67,9 +67,6 @@ TrackedMppDataPacketPtr ToPacket(
 
 TrackedMppDataPacketPtr ToPacketV0(Blocks & blocks, const std::vector<tipb::FieldType> & field_types)
 {
-    if (blocks.empty())
-        return nullptr;
-
     CHBlockChunkCodec codec;
     auto codec_stream = codec.newCodecStream(field_types);
     auto tracked_packet = std::make_shared<TrackedMppDataPacket>(MPPDataPacketV0);
--- a/dbms/src/Common/GRPCQueue.h
+++ b/dbms/src/Common/GRPCQueue.h
@@ -138,6 +138,7 @@ public:
     }
 
     bool isWritable() const { return send_queue.isWritable(); }
+    void triggerPipelineNotify() { send_queue.triggerPipelineNotify(); }
 
     void registerPipeReadTask(TaskPtr && task) { send_queue.registerPipeReadTask(std::move(task)); }
     void registerPipeWriteTask(TaskPtr && task) { send_queue.registerPipeWriteTask(std::move(task)); }
@@ -299,6 +300,7 @@ public:
     }
 
     bool isWritable() const { return recv_queue.isWritable(); }
+    void triggerPipelineNotify() { return recv_queue.triggerPipelineNotify(); }
 
     void registerPipeReadTask(TaskPtr && task) { recv_queue.registerPipeReadTask(std::move(task)); }
     void registerPipeWriteTask(TaskPtr && task) { recv_queue.registerPipeWriteTask(std::move(task)); }
--- a/dbms/src/Common/GRPCQueue.h
+++ b/dbms/src/Common/GRPCQueue.h
@@ -138,7 +138,7 @@ public:
     }
 
     bool isWritable() const { return send_queue.isWritable(); }
-    void triggerPipelineNotify() { send_queue.triggerPipelineNotify(); }
+    void triggerPipelineWriterNotify() { send_queue.triggerPipelineWriterNotify(); }
 
     void registerPipeReadTask(TaskPtr && task) { send_queue.registerPipeReadTask(std::move(task)); }
     void registerPipeWriteTask(TaskPtr && task) { send_queue.registerPipeWriteTask(std::move(task)); }
@@ -300,7 +300,7 @@ public:
     }
 
     bool isWritable() const { return recv_queue.isWritable(); }
-    void triggerPipelineNotify() { return recv_queue.triggerPipelineNotify(); }
+    void triggerPipelineWriterNotify() { return recv_queue.triggerPipelineWriterNotify(); }
 
     void registerPipeReadTask(TaskPtr && task) { recv_queue.registerPipeReadTask(std::move(task)); }
     void registerPipeWriteTask(TaskPtr && task) { recv_queue.registerPipeWriteTask(std::move(task)); }
--- a/dbms/src/Common/LooseBoundedMPMCQueue.h
+++ b/dbms/src/Common/LooseBoundedMPMCQueue.h
@@ -19,6 +19,7 @@
 
 #include <condition_variable>
 #include <deque>
+#include <mutex>
 
 namespace DB
 {
@@ -218,6 +219,17 @@ public:
         if unlikely (status != MPMCQueueStatus::NORMAL)
             return true;
         return !isFullWithoutLock();
+    }
+
+    void triggerPipelineNotify()
+    {
+        auto should_notify = false;
+        {
+            std::lock_guard lock(mu);
+            should_notify = status != MPMCQueueStatus::CANCELLED && !isFullWithoutLock();
+        }
+        if (should_notify)
+            pipe_writer_cv.notifyOne();
     }
 
     MPMCQueueStatus getStatus() const
--- a/dbms/src/Common/LooseBoundedMPMCQueue.h
+++ b/dbms/src/Common/LooseBoundedMPMCQueue.h
@@ -19,7 +19,6 @@
 
 #include <condition_variable>
 #include <deque>
-#include <mutex>
 
 namespace DB
 {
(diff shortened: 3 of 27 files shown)


SeaRise:
> /cc @SeaRise

That's amazing!

fuzhe1989:
> have on chance to be notified

have no chance to be notified

SeaRise:
How about modifying the waitForWritable method of ExchangeWriter so that it returns WaitResult::Ready even if the tunnel is not ready, as long as the cache has not reached the limit? Then, when sink->write detects that the tunnel is not ready, it returns WaitResult::WAIT_FOR_NOTIFY. This way, when the sink is notified, it will definitely write to the mpmcqueue, preventing any hang-ups?

windtalker:
> How about modifying the waitForWritable method of ExchangeWriter so that it returns WaitResult::Ready even if the tunnel is not ready, as long as the cache has not reached the limit? Then, when sink->write detects that the tunnel is not ready, it returns WaitResult::WAIT_FOR_NOTIFY. This way, when the sink is notified, it will definitely write to the mpmcqueue, preventing any hang-ups?

BY "even if the tunnel is not ready" you mean the tunnel is not connected or the tunnel is full?

SeaRise:
Considering this case:
If it's hash-partitioned, and the concurrency is 2, both operators need to flush, and the number of tunnels is 2:

1. Operator1 and Operator2 are both waiting for a notification on tunnel[1].
2. Operator1 is notified by tunnel[1], and since it has data to flush, it won't call `triggerPipelineWriterNotify`.
3. When Operator1 flushes, just because Partition 1 has no data written, and only Partition 0 has data, tunnel[1] will not trigger a write at this time.
4. Because tunnel[1] has no data written to it, there is no consumer to consume from tunnel[1] to trigger Operator2, resulting in a hang.

So query will hang, right?

windtalker:
> 1. iting for a notifi

Current pr makes sure that operator will send data to all tunnels, even if there is no actually data to send.

windtalker:
Another possible fix from @SeaRise 
Basic idea: currently, `ExchangeSenderSinkOp` will first write data to `writer`'s internal cache, and if cache if full, then flush the cache to tunnelset, so another fix should be "only register ExchangeSenderSinkOp to pipeline notify furture if the writer wants to flush data to tunnelset".
Changes need to made:
1. call tunnel's `waitForWritable` only when writer wants to flush data to tunnelset, currently, we call tunnelset's `waitForWritable` in `ExchangeSenderSinkOp->prepare()`, we need to move this code inside each `writer`'s write function, the code should be something like this
```
function write()
{
     // write data to cache
    if (cache is full)
    {
        if (tunnelset is writable)
            // flush data to tunnelset
        else
            // return wait_for_notify or wait_for_polling
    }
}
```
2. Inside `ExchangeSenderSinkOp->prepare()`, it should check if current writer has remaining data to flush to tunnel, the code should be something like this
```
function prepare() 
{
    if (writer->hasDataToFlush())
    {
        if (tunnel is writable)
            // flush data to tunnelset
        else
            // return wait_for_notify or wait_for_polling
    }
    else
    {
        // return need_input status
    }
}
```
3. Usually, a tunnelset contains more than 1 tunnel. Unfortunately, there is no atomic way to check if all the tunnels are writable, so in currently code, we use `force_push` once all tunnels claim they are ok to write independently. In this fix, it should be re-considered that if we still need to use `force_push` to do this or we can maintain `flush data to tunnel` independently for all the tunnels in the same tunnelset

gengliqi (approved):

dbms/src/Flash/Mpp/MPPTunnelSetHelper.cpp
```diff
@@ -67,18 +68,18 @@ TrackedMppDataPacketPtr ToPacket(
...
     CHBlockChunkCodec codec;
     auto codec_stream = codec.newCodecStream(field_types);
     auto tracked_packet = std::make_shared<TrackedMppDataPacket>(MPPDataPacketV0);
     for (auto & block : blocks)
     {
-        codec_stream->encode(block, 0, block.rows());
-        block.clear();
-        tracked_packet->addChunk(codec_stream->getString());
-        codec_stream->clear();
+        if (block)
```
SeaRise, line 76:
```suggestion
        if likely (block && block.rows() > 0)
```

SeaRise (commented):
others lgtm

windtalker:
> Another possible fix from @SeaRise Basic idea: currently, `ExchangeSenderSinkOp` will first write data to `writer`'s internal cache, and if cache if full, then flush the cache to tunnelset, so another fix should be "only register ExchangeSenderSinkOp to pipeline notify furture if the writer wants to flush data to tunnelset". Changes need to made:
> 
> 1. call tunnel's `waitForWritable` only when writer wants to flush data to tunnelset, currently, we call tunnelset's `waitForWritable` in `ExchangeSenderSinkOp->prepare()`, we need to move this code inside each `writer`'s write function, the code should be something like this
> 
> ```
> function write()
> {
>      // write data to cache
>     if (cache is full)
>     {
>         if (tunnelset is writable)
>             // flush data to tunnelset
>         else
>             // return wait_for_notify or wait_for_polling
>     }
> }
> ```
> 
> 2. Inside `ExchangeSenderSinkOp->prepare()`, it should check if current writer has remaining data to flush to tunnel, the code should be something like this
> 
> ```
> function prepare() 
> {
>     if (writer->hasDataToFlush())
>     {
>         if (tunnel is writable)
>             // flush data to tunnelset
>         else
>             // return wait_for_notify or wait_for_polling
>     }
>     else
>     {
>         // return need_input status
>     }
> }
> ```
> 
> 3. Usually, a tunnelset contains more than 1 tunnel. Unfortunately, there is no atomic way to check if all the tunnels are writable, so in currently code, we use `force_push` once all tunnels claim they are ok to write independently. In this fix, it should be re-considered that if we still need to use `force_push` to do this or we can maintain `flush data to tunnel` independently for all the tunnels in the same tunnelset

After some offline discussion, we decide to use this pr as a quick fix for the hang issue, and will refine the whole tunnel write process using the above idea.

PR #9429, opened 2024-09-13
Add metrics about the last time a region is sent a snapshot

CalvinNeo:

Issue Number: ref #9241

Problem Summary:

The idea is when a region got stuck, TiKV leader could send it a snapshot. We will record how many time has passed since the last snapshot. If the time is short, then the region may have some problems here.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/KVStore/MultiRaft/ApplySnapshot.cpp
+++ b/dbms/src/Storages/KVStore/MultiRaft/ApplySnapshot.cpp
@@ -75,7 +75,7 @@ void KVStore::checkAndApplyPreHandledSnapshot(const RegionPtrWrap & new_region, 
         }
 
         {
-            LOG_INFO(log, "{} set state to `Applying`", old_region->toString());
+            LOG_INFO(log, "{} set state to `Applying`, lastAppliedTime={}", old_region->toString(), old_region->lastSnapshotAppliedTime());
             // Set original region state to `Applying` and any read request toward this region should be rejected because
             // engine may delete data unsafely.
             auto region_lock = region_manager.genRegionTaskLock(old_region->id());
@@ -351,7 +351,7 @@ void KVStore::onSnapshot(
 
         tmt.getRegionTable().shrinkRegionRange(*new_region);
     }
-
+    new_region->updateSnapshotAppliedTime();
     prehandling_trace.deregisterTask(new_region->id());
 }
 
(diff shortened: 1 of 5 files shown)


dbms/src/Storages/KVStore/MultiRaft/ApplySnapshot.cpp
```diff
@@ -351,7 +351,7 @@ void KVStore::onSnapshot(
 
         tmt.getRegionTable().shrinkRegionRange(*new_region);
     }
-
+    new_region->updateSnapshotAppliedTime();
```
JaySon-Huang, line 354: We always create a `new_region` instance to apply snapshot instead of modifying the `old_region` in-place. So I think the `tiflash_raft_long_term_event_duration_seconds, type_apply_snapshot_gap` is not get reported at all?

CalvinNeo: Yes, I have tried to fix this by fetch the `lastSNapshotAppliedTime()` from the old_region(if any) to compute.

dbms/src/Storages/KVStore/MultiRaft/Disagg/CheckpointIngestInfo.h
```diff
@@ -42,6 +42,7 @@ struct CheckpointIngestInfo
     UInt64 regionId() const { return region_id; }
     UInt64 peerId() const { return peer_id; }
     UInt64 beginTime() const { return begin_time; }
+    UInt64 createdTime() const { return begin_time; }
```
JaySon-Huang, line 45: `return created_time`?

JaySon-Huang: And add comments about what is the difference between `beginTime` and `createdTime`

CalvinNeo: Changed this PR back to DNM, because some new functionalities is added into this PR.

CalvinNeo:
After https://github.com/pingcap/tiflash/pull/9434

dbms/src/Common/Stopwatch.h
```diff
@@ -69,6 +69,8 @@ class Stopwatch
         is_running = true;
     }
 
+    UInt64 getStartMillis() { return start_ns / 1000000UL; }
```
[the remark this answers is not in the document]
CalvinNeo: I would just use Milliseconds because other methods use this too...

Issue #9032, opened 2024-05-11
Vector Search


breezewish:
## Enhancement

Support Vector Search.

TiDB tracking issue: https://github.com/pingcap/tidb/issues/54245

- [x] #9600
- [x] #9637

PR #9476, opened 2024-09-26
Storage: Add error message when fail to build local index

Lloyd-Pottiger:

Issue Number: ref #9032

Problem Summary:

* Add 2 columns to `system.dt_local_indexes`
  * `column_name` is removed from the `system.dt_local_indexes`, let the tidb fill the index_name
  * `error_message` when it contains non-empty string, it means the index can not be built on tiflash
* An std::unordered_map<IndexID, String> fatal_error_when_building_local_index is added to each Segment:
  * The error message will be set when:
    * When the task can not be pushed to the LocalIndexScheduler, for example, need to large memory to be built
    * When the task throws any exception in `DeltaMergeStore::segmentEnsureStableIndex`
  * The error message will be clear when:
    * New local tasks is pushed to the LocalIndexScheduler with the same index_id
    * The index is dropped from tidb
  * `generateLocalIndexInfos` now will also return the index_ids that is dropped, so that we can cleanup the error message of the dropped indexes

```commit-message
Storage: Add error message when fail to build local index
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

```
create table x(e vector(3) not null);
set sql_mode='';
alter table x set tiflash replica 1;
insert into x values(); -- now this will insert a vector with dimension=0

# create vector index on tidb, fail to build index
ALTER TABLE x ADD VECTOR INDEX idx_e_l2 ((VEC_L2_DISTANCE(e))) USING HNSW;

# check the error message is returned to `information_schema.tiflash_indexes`
select * from information_schema.tiflash_indexes where tidb_table = 'x';
```
or:
```
# starts a tifalsh server with tcp_port enabled
> /path/to/tiflash server --config-file=... --tcp_port=9099

# connect to tidb and create the table and vector index
CREATE TABLE `t` (  `v` vector(5) DEFAULT NULL,  `v2` vector(3) DEFAULT NULL);
INSERT INTO t VALUES ('[8.7, 5.7, 7.7, 9.8, 1.5]','[1, 2, 3]')

# starts a tiflash client to enable the failpoint of building local index
>  /path/to/tiflash client --host=127.0.0.1 --port=9099
TiFlash :) DBGInvoke enable_fail_point(exception_build_local_index_for_file)

# create vector index on tidb
ALTER TABLE t ADD VECTOR INDEX idx_v2_l2 ((VEC_L2_DISTANCE(v2))) USING HNSW;
# check the error message is returned to `information_schema.tiflash_indexes`
select * from information_schema.tiflash_indexes where tidb_table = 't';
```

image

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Common/FailPoint.cpp
+++ b/dbms/src/Common/FailPoint.cpp
@@ -71,6 +71,8 @@ namespace DB
     M(force_remote_read_for_batch_cop_once)                       \
     M(exception_new_dynamic_thread)                               \
     M(force_wait_index_timeout)                                   \
+    M(force_local_index_task_memory_limit_exceeded)               \
+    M(exception_build_local_index_for_file)                       \
     M(force_not_support_vector_index)                             \
     M(sync_schema_request_failure)
 
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -39,7 +39,7 @@
 #include <Storages/DeltaMerge/File/DMFile.h>
 #include <Storages/DeltaMerge/Filter/PushDownFilter.h>
 #include <Storages/DeltaMerge/Filter/RSOperator.h>
-#include <Storages/DeltaMerge/Index/IndexInfo.h>
+#include <Storages/DeltaMerge/Index/LocalIndexInfo.h>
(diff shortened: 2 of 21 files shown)


JaySon-Huang (approved):
LGTM

PR #9484, opened 2024-09-27
storage: fix block rows not match when filter column is the first non-empty column in the block

### LINKED ISSUE #9472: query count(*) failed
A table stored in TiFlash, using this query sometimes works, sometimes not.
image
image

Lloyd-Pottiger:

Issue Number: ref #9472

Problem Summary:

Try to fix #9472, failed to reproduce the issue. But when the filter column is the first non-empty column in the block, block.rows() = filter_column->size(), but filter_column is not filtered, so the rows may not match.

**But filter_column is usually added to the back of the block, so it is kind of weird.**

```commit-message
storage: fix block rows not match when filter column is the first non-empty column in the block
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Core/Block.cpp
+++ b/dbms/src/Core/Block.cpp
@@ -531,7 +531,7 @@ static ReturnType checkBlockStructure(const Block & lhs, const Block & rhs, cons
         }
     }
 
-    return ReturnType(true);
+    return static_cast<ReturnType>(true);
 }
 
 /// join blocks by columns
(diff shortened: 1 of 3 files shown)


JaySon-Huang (approved):
lgtm

Lloyd-Pottiger:
fullstack-test2/ddl/rename_table_across_databases.test

PR #9504, opened 2024-10-09
Storage: Abort adding vector index if encryption enabled

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9505, ref https://github.com/pingcap/tiflash/issues/9032

Problem Summary:

Encryption-at-rest is not compatible with building vector-index. We need more tests.

```commit-message
Storage: Abort adding vector index if encryption enabled
```

Encryption-at-rest is not compatible with building vector-index. We need more tests. Abort adding vector-index when encryption-at-rest is enabled in tiflash for now.

One of the reason is `DMFileV3IncrementWriter::writeAndIncludeMetaFile` call `WriteBufferFromWritableFileBuilder::buildPtr(..., /* create_new_encryption_info */ true, ...)` and it renew the encryption_info of the DMFile. It causes the encryption_info mismatch after generating the new DMFile meta v1.

There is still something wrong after fixing it. Still under investigation.

Other:
Add info level logging message when vector search is performed in tiflash.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalSegment.cpp
@@ -543,6 +543,14 @@ bool DeltaMergeStore::segmentEnsureStableIndexAsync(const SegmentPtr & segment)
     if (!build_info.indexes_to_build || build_info.indexes_to_build->empty() || build_info.dm_files.empty())
         return false;
 
+    if (auto encryption_enabled = global_context.getFileProvider()->isEncryptionEnabled(); encryption_enabled)
+    {
+        segment->setIndexBuildError(
+            build_info.indexesIDs(),
+            "Encryption-at-rest on TiFlash is enabled, which does not support building vector index");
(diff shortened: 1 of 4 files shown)


breezewish:
Is it possible to still create the vector index file while encyption is enabled? Looks like setting create_new_encryption_info may resolve the issue I guess.

JaySon-Huang:
> Is it possible to still create the vector index file while encyption is enabled? Looks like setting create_new_encryption_info may resolve the issue I guess. I'm worried about aborting the index may affect dedicated tier or serverless.

@breezewish  After fixing the `WriteBufferFromWritableFileBuilder::buildPtr` in `DMFileV3IncrementWriter::writeAndIncludeMetaFile`, I still meet errors when trying to create vector index on a larger vector dataset. I'm still investigating the root cause of it.

Issue #9529, opened 2024-10-15
The `gcTrySegmentMergeDelta` on common handle table can not GC data correctly in timely manner


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

PR #9530, opened 2024-10-15
Storage: Fix `getSquashDeleteRange` does not return correctly squashed key-range when using common-handle

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9529

Problem Summary:

we get the "suqash delete range" to estimate whether there are sufficient delete rows that we need to compact data in background thread.
https://github.com/pingcap/tiflash/blob/ce0d411d87be44243d5ff9eec71948e57b2bbb8b/dbms/src/Storages/DeltaMerge/DeltaMergeStore_InternalBg.cpp#L526-L528

In `ColumnFileSetSnapshot::getSquashDeleteRange`, it use the member variable `is_common_handle`, `rowkey_column_size` to generate the rowkey range.
But actually those two variable are not set correctly. `is_common_handle` is always false in `ColumnFileSetSnapshot`
https://github.com/pingcap/tiflash/blob/ce0d411d87be44243d5ff9eec71948e57b2bbb8b/dbms/src/Storages/DeltaMerge/ColumnFile/ColumnFileSetSnapshot.cpp#L21-L30

So when the table use common-handle, `shouldCompactDeltaWithStable` always returns false, and the segment background gc is not GC in timely

```commit-message
Storage: Fix `getSquashDeleteRange` does not return correctly squashed key-range when using common-handle
```

* Remove unused variable `ColumnFileSetSnapshot::{is_common_handle, rowkey_column_size}`
* Pass the correct is_common_handle, rowkey_column_size in function `shouldCompactDeltaWithStable`
* Add check for rowkey-range is_common_handle in `DMFilePackFilter`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Improve the data background GC speed on clustered-index table
提升聚簇索引的表后台回收过期数据的速度
```


Lloyd-Pottiger (approved):
lgtm

PR #9523, opened 2024-10-14
Refine logical op in TiFlash

windtalker:
### What problem does this PR solve?

Issue Number: close #9146

Problem Summary:
According to #9146, the logical operator in TiFlash is kind of low efficient.
### What is changed and how it works?
This pr refine logical operator using the following ways
1. use a new algorithm to handle 3-value logical op, which can avoid data copy when the input column is Nullable(UInt8)
2. Add new struct `NullableAssociativeOperationImpl` to handle nullable column specifically, so the original `AssociativeOperationImpl` don't need to consider null value
3. If the input column has both not null and nullable columns, always handle the not null column first
4. add some quick path if the input contains constant value
5. Optimize `NullableAssociativeOperationImpl/AssociativeOperationImpl` for the case that input column is 2

benchmark results
Run on (72 X 3300 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x36)
  L1 Instruction 32 KiB (x36)
  L2 Unified 1024 KiB (x36)
  L3 Unified 25344 KiB (x2)
Load Average: 23.98, 20.10, 14.61
```
-----------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                                   Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------------------------------
LogicalOpBench/LegacyLogical_And_not_null_uint8_1_constant_true/iterations:1000         36373 ns        36248 ns         1000
LogicalOpBench/OptLogical_And_not_null_uint8_1_constant_true/iterations:1000             9398 ns         9361 ns         1000
LogicalOpBench/LegacyLogical_And_not_null_uint8_1_constant_false/iterations:1000        36602 ns        36473 ns         1000
LogicalOpBench/OptLogical_And_not_null_uint8_1_constant_false/iterations:1000             324 ns          323 ns         1000
LogicalOpBench/LegacyLogical_And_not_null_uint8_1_constant_null/iterations:1000         43667 ns        43514 ns         1000
LogicalOpBench/OptLogical_And_not_null_uint8_1_constant_null/iterations:1000            11052 ns        11014 ns         1000
LogicalOpBench/LegacyLogical_And_nullable_uint8_1_constant_true/iterations:1000         82794 ns        82510 ns         1000
LogicalOpBench/OptLogical_And_nullable_uint8_1_constant_true/iterations:1000             8390 ns         8362 ns         1000
[...]

Issue #9146, opened 2024-06-13
optimize logical operator function in TiFlash


windtalker:
## Enhancement
Running query 
```
explain analyze select count(ol_i_id), count(ol_supply_w_id), count(ol_delivery_d) from order_line where ol_i_id <= 100000 and ol_supply_w_id <= 1500 and (ol_delivery_d >= "2021-12-27 09:25:34" or ol_delivery_d is NULL)
```
Where `order_line` has 452038522 rows, the profile shows logical operator is very slow in TiFlash(takes up almost 70% cpu time), which is not expected.
<img width="1795" alt="Screenshot 2024-06-13 at 16 10 56" src="https://github.com/pingcap/tiflash/assets/1916264/6f789c53-e258-4925-b9e2-4d3998fc8745">

PR #9560, opened 2024-10-28
Add two value logical op

windtalker:

Issue Number: ref #9146

Problem Summary:

This is a follow up pr for #9523. #9523 optimize logical op in TiFlash, but if the logical op returns a nullable(UInt8) column for  filter executor, in `FilterDescription`, it still need to merge null_map and the data into a new filter vector:
https://github.com/pingcap/tiflash/blob/b36199ea6f2a8ec754ee17db86dd3727e0524b14/dbms/src/Columns/FilterDescription.cpp#L91-L93

This pr add a special logic function called `two_value_and`, and inside `two_value_and`, it will treat null as false, so it always return UInt8 column. This function should only be used for filter executor. In this pr, it only uses `two_value_and` for `PhysicalFilter::build` and `executePushedDownFilter`.

For the same query in #9146 
Orignal time ~0.42s
After #9523 ~0.26s
After this pr ~0.23s 
```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Functions/FunctionsLogical.h
```diff
@@ -587,13 +654,23 @@ class FunctionAnyArityLogical : public IFunction
...
-                vec_res[i] = static_cast<bool>(col_data[i]);
+                for (size_t i = 0; i < rows; ++i)
+                {
+                    vec_res[i] = !(*null_maps[0])[i] && static_cast<bool>(col_data[i]);
+                }
+            }
+            else
+            {
+                // according to https://github.com/pingcap/tiflash/issues/5849
+                // need to cast the UInt8 column to bool explicitly
```
gengliqi, line 667: These comments can be moved before `if constexpr (null_as_false)`

dbms/src/Functions/FunctionsLogical.h
```diff
@@ -772,24 +874,48 @@ class FunctionAnyArityLogical : public IFunction
...
+                    {
+                        vec_res[i] = static_cast<bool>(col_data[i]);
+                    }
+                    vec_res_is_null.assign(*null_maps[0]);
                 }
-                vec_res_is_null.assign(*null_maps[0]);
             }
             else
             {
+                if (!result_is_nullable)
```
gengliqi, line 895: How about `if constexpr (null_as_false)`?

windtalker: Done

dbms/src/Functions/FunctionsLogical.h
windtalker: Done

gengliqi (approved):
LGTM

yibin87 (approved):
LGTM

PR #9566, opened 2024-10-30
Add pipeline wait time of minTSO and pipeline breaker in execution summary

### LINKED ISSUE #9489: The execution time in `pipeline` mode is distortion sometimes
## Bug Report

Running the same query in pipeline mode and pull mode, the overall query peformance is almost the same, but the execution time for each executor differs a lot between pipeline mode and pull mode

In pipeline mode
Screenshot 2024-09-30 at 13 07 44

HashAgg: 2.23s
Projection: 2.71s
Selection: 1.75s

Cpu flame graph
Screenshot 2024-09-29 at 13 46 26

And in pull mode
Screenshot 2024-09-29 at 13 51 35
HashAgg: 4.4s 
Projection: 2.8s
Selection: 0.26s

Cpu flame graph
Screenshot 2024-09-29 at 13 48 31

And compared with cpu flame graph, the executor time in pull mode is much more reasonable.

yibin87:

Issue Number: close #9489 

Problem Summary:

This PR mainly does three things:

1. Add minTSO schedule time in execution summary for source ops.
2. Add pipeline breaker wait time info in execution summary.
3. Add the queue waiting time only on source ops instead of distribute the time averagely among all operators.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
Use tpch 100 data set, test result is shown in https://github.com/pingcap/tidb/pull/57058
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Common/Stopwatch.h
```diff
@@ -24,13 +24,17 @@
 #include <common/apple_rt.h>
 #endif
 
+constexpr UInt64 SECOND_TO_NANO = 1000000000ULL;
+constexpr UInt64 MILLISECOND_TO_NANO = 1000000UL;
+constexpr UInt64 MICROSECOND_TO_NANO = 1000UL;
```
SeaRise, line 29:
```suggestion
static constexpr UInt64 SECOND_TO_NANO = 1000000000ULL;
static constexpr UInt64 MILLISECOND_TO_NANO = 1000000UL;
static constexpr UInt64 MICROSECOND_TO_NANO = 1000UL;
```

dbms/src/Flash/Coprocessor/ExecutionSummary.cpp
```diff
@@ -36,7 +44,14 @@ void ExecutionSummary::merge(const ExecutionSummary & other)
 
 void ExecutionSummary::merge(const tipb::ExecutorExecutionSummary & other)
 {
-    time_processed_ns = std::max(time_processed_ns, other.time_processed_ns());
+    bool time_updated = time_processed_ns < other.time_processed_ns();
+    if (time_updated)
```
SeaRise, line 48:
```suggestion
    if (time_processed_ns < other.time_processed_ns())
```

dbms/src/Flash/Executor/QueryExecutor.h
```diff
@@ -58,7 +58,7 @@ class QueryExecutor
     virtual String getExtraJsonInfo() const { return "{}"; }
 
 protected:
-    virtual ExecutionResult execute(ResultHandler &&) = 0;
+    virtual ExecutionResult execute(ResultHandler &&, UInt64 minTSO_wait_time_in_ns) = 0;
```
SeaRise, line 61: why not fill minTso_wait_time in `mpp_task_statistics.genExecutionSummaryResponse()`, seems hack here.

dbms/src/Flash/Coprocessor/ExecutionSummary.cpp
yibin87: ok

dbms/src/Common/Stopwatch.h
yibin87: ok

dbms/src/Flash/Executor/QueryExecutor.h
yibin87: Yes, it's a little track here to pass minTSO time as argument. The reason I don't handle it in mpp_task_statistics.genExecutionSummaryResponse() is: I want to explicitly show the minTSO time in source operator only, it is easier to do this in PipelineExec.

SeaRise: Seems add a function like `ExecutorStatisticsCollector::addScheduleTimeForSource` is ok? And then the schedule time can also be displayed in the non-pipeline summary

yibin87: I see, I'll change it.

yibin87: Updated now.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
```diff
@@ -311,6 +311,7 @@ void PhysicalJoin::buildPipeline(PipelineBuilder & builder, Context & context, P
     join_build_builder.build();
 
     // Join probe pipeline.
+    builder.setHasPipelineBreakerWaitTime(true);
```
SeaRise, line 314: Also for aggregation?

dbms/src/Flash/Pipeline/Pipeline.h
```diff
@@ -90,6 +90,8 @@ class Pipeline : public std::enable_shared_from_this<Pipeline>
 
     String getFinalPlanExecId() const;
 
+    void setHasPipelineBreakerWaitTime(bool value) { has_pipeline_breaker_wait_time = value; }
```
SeaRise, line 93: It seems that the type of source physical plan can be used to determine if there is a pipeline breaker time, without needing external input.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
yibin87: For aggregation, I think the wait time is not "real wait", since the aggregation calculation is actually always going.

dbms/src/Flash/Pipeline/Pipeline.h
yibin87: Yes, you're right. I think it's natural to set the flag during pipeline building process also.

dbms/src/Flash/Planner/Plans/PhysicalJoin.cpp
SeaRise: If so, then the real pipeline breaker is actually just join, and it feels like it can be directly named join-build-wait-time.
And then `hasPipelineBreakerWaitTime` is useless because only `source->type == Join` will add the wait-time

yibin87: Confirmed with SeaRise offline, we can't use source->type == Join to check. Since it only works for JoinProbe pipeline, will finally show to users like "join wait build time" in execution summary string.

yibin87:
Please let me know if you have any other comments :) @SeaRise

SeaRise (approved):
lgtm

windtalker (approved):
lgtm

Issue #9433, opened 2024-09-14
task execution time should have an upper limit in pipeline model


windtalker:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
In pipeline model, when task is scheduled, it will be executed in fixed size thread pool, so for each task, the per-round execution time must have an upper limit, otherwise, it could block all the other queries.
Normally, the per-round exectuion time should be predictable and short enough because it is by design that in each round of exection, the task will only process one block of data(normally less than 60k rows), so we can just assume it already has a reasonable upper limit, but still in some special cases, some operators may take very very long time to process one block of the data, we need to handle this explicitly. Special attention should be spent on the following kind of operators:
* HashJoinProbeTransformOp

This issue aims to identify and improve operators that may seriously affect the normal operation of the system. For other operators, since their timing is predictable, they are not the focus of this issue.

SeaRise:
In velox, the operator gets task->execute_time to determine whether it has timed out.
Perhaps we can put the Task pointer in thread_local to get the execution time

windtalker:
> In velox, the operator gets task->execute_time to determine whether it has timed out. Perhaps we can put the Task pointer in thread_local to get the execution time

But we can not simply return error if execution time exceed the limit?

SeaRise:
> > In velox, the operator gets task->execute_time to determine whether it has timed out. Perhaps we can put the Task pointer in thread_local to get the execution time
> 
> But we can not simply return error if execution time exceed the limit?

Ah, yes, the need for a join probe that can return after probing a portion of the data (more thorough than before

PR #9668, opened 2024-11-25
`WindowTransformAction::tryCalculate()` calculate at most 1 block result at a time

windtalker:

Issue Number: ref #9433

Problem Summary:
Before this pr, `WindowTransformAction::tryCalculate()` will try to calculate as many as possible results at a time. This makes the calling time of `WindowTransformAction::tryCalculate()` unpredictable.
This pr let `WindowTransformAction::tryCalculate()` to calculate at most 1 block result at a time, so the calling time of `WindowTransformAction::tryCalculate()` won't be too long, and be more friendly for pipeline model.

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/DataStreams/WindowBlockInputStream.cpp
+++ b/dbms/src/DataStreams/WindowBlockInputStream.cpp
@@ -1325,6 +1325,7 @@ void WindowTransformAction::appendBlock(Block & current_block)
 
 void WindowTransformAction::tryCalculate()
 {
+    auto start_block_index = current_row.block;
     // Start the calculations. First, advance the partition end.
     for (;;)
     {
@@ -1396,6 +1397,11 @@ void WindowTransformAction::tryCalculate()
             first_not_ready_row = current_row;
             frame_ended = false;
             frame_started = false;
+            // each `tryCalculate()` will calculate at most 1 block's data
+            // this is to make sure that in pipeline mode, the execution time
+            // of each iterator won't be too long
+            if (current_row.block != start_block_index)
+                return;
         }
 
         if (input_is_finished)
--- a/dbms/src/DataStreams/WindowBlockInputStream.cpp
+++ b/dbms/src/DataStreams/WindowBlockInputStream.cpp
@@ -31,6 +31,7 @@
 #include <magic_enum.hpp>
 #include <tuple>
 #include <type_traits>
+#include "Core/Block.h"
 
 namespace DB
 {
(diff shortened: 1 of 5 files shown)


dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -316,27 +316,30 @@ bool WindowBlockInputStream::returnIfCancelledOrKilled()
...
 Block WindowBlockInputStream::readImpl()
 {
-    const auto & stream = children.back();
-    while (!action.input_is_finished)
-    {
-        if (returnIfCancelledOrKilled())
-            return {};
+    // first try to get result without reading from children
+    auto block = action.tryGetOutputBlock();
+    if (block || action.input_is_finished)
```
xzhangxian1008, line 321: I think it's unnecessary to check `action.input_is_finished` here as it will be check in the wile-loop.

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -1237,6 +1242,9 @@ void WindowTransformAction::writeOutCurrentRow()
 
 Block WindowTransformAction::tryGetOutputBlock()
 {
+    // first try calculate the result based on current data
+    tryCalculate();
```
xzhangxian1008, line 1246: In `WindowTransformOp::transformImpl`, the function always calls `tryCalculate()` before `tryGetOutputBlock`, maybe we can delete them.

windtalker: Before this pr, `tryCalculate()` will calculate as many blocks as possible, so `tryGetOutputBlock()` can be call multiple times after one call to `tryCalculate()`, after this pr, `tryCalculate()` will calculate at most 1 result block, which means if you don't call `tryCalculate()`, `tryGetOutputBlock()` will return 1 block at the first call, and all the subsequent call of `tryGetOutputBlock()`  will return `{}`, which is meaningless and even worse, in pull model, it may cause wrong results. In order to make things right, we should always call `tryCalculate()` before call `tryGetOutputBlock()`, which also means they are coupled together after this pr, so I change `tryCalculate()` to a private method, and call it inside `tryGetOutputBlock()`

xzhangxian1008: > Before this pr, `tryCalculate()` will calculate as many blocks as possible, so `tryGetOutputBlock()` can be call multiple times after one call to `tryCalculate()`, after this pr, `tryCalculate()` will calculate at most 1 result block, which means if you don't call `tryCalculate()`, `tryGetOutputBlock()` will return 1 block at the first call, and all the subsequent call of `tryGetOutputBlock()` will return `{}`, which is meaningless and even worse, in pull model, it may cause wrong results. In order to make things right, we should always call `tryCalculate()` before call `tryGetOutputBlock()`, which also means they are coupled together after this pr, so I change `tryCalculate()` to a private method, and call it inside `tryGetOutputBlock()`

Not delete `tryCalculate` in L1246, I mean that we may delete `tryCalculate` call in L47 and L54 in WindowTransformOp.cpp

windtalker: I don't get your point, I've already deleted all the all of `tryCalculate` in `WindowTransformOp.cpp`

xzhangxian1008 (approved):
LGTM

dbms/src/DataStreams/WindowBlockInputStream.cpp
```diff
@@ -316,27 +316,30 @@ bool WindowBlockInputStream::returnIfCancelledOrKilled()
 
 Block WindowBlockInputStream::readImpl()
 {
-    const auto & stream = children.back();
-    while (!action.input_is_finished)
-    {
-        if (returnIfCancelledOrKilled())
-            return {};
+    // first try to get result without reading from children
```
gengliqi, line 319: Is a `if (returnIfCancelledOrKilled())` check needed here?

windtalker: Why? Usually, `readImpl` does not need be aware of cancel signal.

PR #9736, opened 2024-12-20
ExchangeSenderOp check writer status only if there is data to flush to MPPTunnel

windtalker:

Issue Number: ref #6233

Problem Summary:
Currently, `ExchangeSenderSinkOp` check the writer status in `ExchangeSenderSinkOp::prepareImpl()`, so even if there is currently no data to write, the `ExchangeSenderSinkOp` need to check the writer status, and the pipeline task maybe blocked if the writer is not ready. In this pr, `ExchangeSenderSinkOp` only check the writer status if it has data to flush to MPPTunnel.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Operators/ExchangeSenderSinkOp.cpp
```diff
@@ -28,15 +28,34 @@ void ExchangeSenderSinkOp::operateSuffixImpl()
...
-    if (!block)
+    assert(!buffer);
+    buffer.emplace(std::move(block));
+    return tryFlush();
+}
+
+OperatorStatus ExchangeSenderSinkOp::tryFlush()
+{
+    assert(buffer);
+    auto res = waitForWriter();
```
SeaRise, line 39: Perhaps moving `waitForWriter` into `Writer::write` and `Writer::flush` would be better. This way, even if the tunnel is busy, write can still batch in memory.

windtalker: You mean if the block's size is small, it will be cached in the writer without writting to tunnel?

SeaRise: yes, if the total block rows is smaller than `batch_send_min_limit`, MPPTunnel::write would not be call.
https://github.com/pingcap/tiflash/blob/master/dbms/src/Flash/Mpp/BroadcastOrPassThroughWriter.cpp#L103-L109

windtalker: Good idea, I reimplement this pr by moving `waitForWriter` to `Writer::flush`. But `waitForWriter` is still needed in `ExchangeSenderSinkOp::await`

SeaRise: lgtm

dbms/src/Operators/ExchangeSenderSinkOp.cpp
```diff
@@ -55,11 +65,28 @@ OperatorStatus ExchangeSenderSinkOp::waitForWriter() const
 
 OperatorStatus ExchangeSenderSinkOp::prepareImpl()
 {
-    return waitForWriter();
+    // if there is cached data, flush it
+    if (need_flush)
+    {
+        // the writer must has data to flush
+        assert(writer->hasDataToFlush());
```
gengliqi, line 72: How about using `RUNTIME_CHECK`?

gengliqi (approved):
LGTM

SeaRise (approved):
lgtm
comment `/unhold` to merge pr.

dbms/src/Operators/ExchangeSenderSinkOp.h
```diff
@@ -49,6 +49,7 @@ class ExchangeSenderSinkOp : public SinkOp
     OperatorStatus waitForWriter() const;
 
 private:
+    bool need_flush = false;
```
yibin87, line 52: Better move this flag into writer.

windtalker: Done

PR #9780, opened 2025-01-10
KVStore: Record decoded memory of each Region

CalvinNeo:

Issue Number: ref #9722

Problem Summary:

This is part 1 of the change to record table-wise memory usage in KVStore.

Record payload and decoded memory in RegionData:
- payload memory is the data of the 3 cfs that flows from proxy into TiFlash.
- decoded memory is mainly the decoded lock value cached to reduce the cost of parsing lock when reading.

We now record the decoded memory of large txns because it could consume significant amount of memory if there are lots of locks in the lock cf.

We also simplify the codes in RegionData.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/KVStore/Decode/RegionTable.h
```diff
@@ -78,11 +78,15 @@ class RegionTable : private boost::noncopyable
             , range_in_table(range_in_table_)
         {}
 
+        void updateRegionCacheBytes(size_t);
+
         RegionID region_id;
         std::pair<DecodedTiKVKeyPtr, DecodedTiKVKeyPtr> range_in_table;
         bool pause_flush = false;
```
Lloyd-Pottiger, line 85: Let `pause_flush` be the last member to save memory.

CalvinNeo: Good idea!

dbms/src/Storages/KVStore/Decode/RegionTable.h
```diff
@@ -78,11 +78,15 @@ class RegionTable : private boost::noncopyable
             , range_in_table(range_in_table_)
         {}
 
+        void updateRegionCacheBytes(size_t);
+
         RegionID region_id;
         std::pair<DecodedTiKVKeyPtr, DecodedTiKVKeyPtr> range_in_table;
+        Timepoint last_flush_time = Clock::now();
```
JaySon-Huang, line 85: Can `last_flush_time` be removed? Seems we now only set it but never read it.

CalvinNeo: Quite so...
Meanwhile, the usage of `pause_flush` seems also ambiguous to me. However, removing it is not trival, so I will deliberate on it.

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
```diff
@@ -44,16 +73,18 @@ struct RegionCFDataBase
...
-    RegionDataRes insert(TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
+    RegionDataMemDiff insert(TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
 
     static size_t calcTiKVKeyValueSize(const Value & value);
-
     static size_t calcTiKVKeyValueSize(const TiKVKey & key, const TiKVValue & value);
+    static RegionDataMemDiff calcTotalKVSize(const Value & value);
+
+    size_t calcDecodedKeyValueSize(const Value & value);
+    size_t calcDecodedKeyValueSize(const TiKVKey & key, const TiKVValue & value);
```
JaySon-Huang, line 85:
```suggestion
```

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
```diff
@@ -44,16 +73,18 @@ struct RegionCFDataBase
...
     static const TiKVKey & getTiKVKey(const Value & val);
-
     static const TiKVValue & getTiKVValue(const Value & val);
 
-    RegionDataRes insert(TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
+    RegionDataMemDiff insert(TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
 
     static size_t calcTiKVKeyValueSize(const Value & value);
-
     static size_t calcTiKVKeyValueSize(const TiKVKey & key, const TiKVValue & value);
```
JaySon-Huang, line 81: Change the line 315 in `Region::tryCompactionFilter` from 
```
    data.cf_data_size -= RegionWriteCFData::calcTiKVKeyValueSize(write_map_it->second);
```
to
```
    data.cf_data_size -= RegionWriteCFData::calcTotalKVSize(write_map_it->second).payload;
```

... and then we can move these two methods to be "private"

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.cpp
```diff
@@ -124,7 +90,40 @@ RegionDataRes RegionCFDataBase<Trait>::insert(std::pair<Key, Value> && kv_pair,
...
+    UNUSED(mode);
+    RegionDataMemDiff delta;
+    Pair kv_pair = RegionLockCFDataTrait::genKVPair(std::move(key), std::move(value));
+    const auto & decoded = std::get<2>(kv_pair.second);
+    bool is_large_txn = decoded->isLargeTxn();
+    {
+        auto iter = data.find(kv_pair.first);
+        if (iter != data.end())
+        {
+            // Could be a perssimistic lock is overwritten, or a old generation large txn lock is overwritten.
```
JaySon-Huang, line 109:
```suggestion
            // Could be a pessimistic lock is overwritten, or a old generation large txn lock is overwritten.
```

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.cpp
```diff
@@ -322,22 +324,24 @@ size_t RegionCFDataBase<Trait>::serialize(WriteBuffer & buf) const
...
         auto key = TiKVKey::deserialize(buf);
         auto value = TiKVValue::deserialize(buf);
-        cf_data_size += new_region_data.insert(std::move(key), std::move(value));
+        auto r = new_region_data.insert(std::move(key), std::move(value));
+        cf_data_size += r.payload;
+        decoded_data_size += r.decoded;
     }
-    return cf_data_size;
+    return {cf_data_size, decoded_data_size};
 }
```
JaySon-Huang, line 345:
```
template <typename Trait>
RegionDataMemDiff RegionCFDataBase<Trait>::deserialize(ReadBuffer & buf, RegionCFDataBase & new_region_data)
{
    auto size = readBinary2<size_t>(buf);
    RegionDataMemDiff total_diff;
    for (size_t i = 0; i < size; ++i)
    {
        auto key = TiKVKey::deserialize(buf);
        auto value = TiKVValue::deserialize(buf);
        auto diff = new_region_data.insert(std::move(key), std::move(value));
        total_diff.add(diff);
    }
    return total_diff;
}
```

dbms/src/Storages/KVStore/MultiRaft/RegionData.h
```diff
@@ -130,8 +129,10 @@ class RegionData
     RegionLockCFData lock_cf;
     OrphanKeysInfo orphan_keys_info;
 
-    // Size of data cf & write cf, with lock cf.
+    // Size of 3 cfs, reflects size of real palyload flows to KVStore.
```
JaySon-Huang, line 132:
```suggestion
    // Size of 3 cfs, reflects size of real payload flows to KVStore.
```

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
```diff
@@ -25,7 +25,36 @@ namespace DB
...
+    RegionDataMemDiff(T1 payload_, T2 decoded_)
+        : payload(payload_)
+        , decoded(decoded_)
+    {}
+    RegionDataMemDiff()
+        : payload(0)
+        , decoded(0)
+    {}
+
+    RegionDataMemDiff operator-() const { return {-payload, -decoded}; }
```
JaySon-Huang, line 44: I'd like to keep a function naming `negative` but not overload the `operator-`. Because the other functions in `RegionDataMemDiff` use `add` `sub` but not overload the related operators
```suggestion
    RegionDataMemDiff negative() const { return {-payload, -decoded}; }
```

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
CalvinNeo: move `tryCompactionFilter ` into RegionData

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.cpp
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
CalvinNeo: fixed

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -50,43 +50,49 @@ void RegionData::reportDelta(size_t prev, size_t current)
...
+    decoded_data_size += delta.decoded;
+    if (delta.payload > 0)
+    {
+        root_of_kvstore_mem_trackers->alloc(delta.payload, false);
+    }
+    else if (delta.payload < 0)
+    {
+        root_of_kvstore_mem_trackers->free(-delta.payload);
+    }
+}
```
JaySon-Huang, line 65: `decoded_data_size` also reflect the occupied memory, why not add it to `root_of_kvstore_mem_trackers`? Or you want to do it in next PR?

JaySon-Huang:
... and some functions in `RegionCFDataBase` is not necessary. We can merge them and reduce the code complexity. Checkout these changes pls https://github.com/CalvinNeo/tiflash/pull/13/files

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
CalvinNeo: Yes, it is actually a dilemma here.

Since if I add it to `root_of_kvstore_mem_trackers`, it breaks the behavior. So when we want to check the metric on grafana board, we have to firstly make sure what this metric actually represents according to the TiFlash version. We could do this, but it takes some modifications and I want to leave them in some later PR.

However, I think we can savage out previous misunderstanding in a more elegant way, at least I think it so. The idea is that it is hard to get how the memory usage of payload, because it is a nested structure. However, it is quite easy to get the memory usage of the decoded lock, which has fixed size. So in https://github.com/pingcap/tiflash/blob/a8f72d428eec025b946c8c062369083b298e8031/dbms/src/Storages/KVStore/TiKVHelpers/DecodedLockCFValue.cpp#L167, I have recorded the count of `DecodedLockCFValue::Inner`, by which we could easily compute the cached size.

Thus, we can introduce another series in the "KVStore memory" panel. And it is quite easy to distinguish it from the payload series.

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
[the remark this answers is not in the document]
CalvinNeo: The intuition is that I want to hint what change would be taken to the KVStore.
However, you are right here, because using `RegionRataDiff` we no longer cares about the memory changes. It is all settled now.

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -131,22 +109,23 @@ RegionData::WriteCFIter RegionData::removeDataByWriteIt(const WriteCFIter & writ
...
         }
     }
 
-    auto delta = RegionWriteCFData::calcTiKVKeyValueSize(write_it->second);
-    cf_data_size -= delta;
-    reportDealloc(delta);
+    delta.sub(RegionWriteCFData::calcTotalKVSize(write_it->second));
+    cf_data_size += delta.payload;
+    decoded_data_size += delta.decoded;
+    recordMemChange(delta);
```
JaySon-Huang, line 128:
```suggestion
    recordMemChange(delta);
```
cf_data_size and decoded_data_size has been changed in `recordMemChange(delta)`

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -258,38 +237,54 @@ std::shared_ptr<const TiKVValue> RegionData::getLockByKey(const TiKVKey & key) c
...
+    return cf_data_size + decoded_data_size;
+}
+
+void RegionData::assignRegionData(RegionData && rhs)
 {
-    default_cf = std::move(new_region_data.default_cf);
-    write_cf = std::move(new_region_data.write_cf);
-    lock_cf = std::move(new_region_data.lock_cf);
-    orphan_keys_info = std::move(new_region_data.orphan_keys_info);
+    recordMemChange(cf_data_size.negative());
```
JaySon-Huang, line 275:
```suggestion
    recordMemChange(RegionDataMemDiff{-cf_data_size, -decoded_data_size});
```

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -305,19 +300,22 @@ size_t RegionData::serialize(WriteBuffer & buf) const
...
 
-    region_data.cf_data_size = total_size;
-    reportAlloc(total_size);
+    size_changed.add(RegionDefaultCFData::deserialize(buf, region_data.default_cf));
+    size_changed.add(RegionWriteCFData::deserialize(buf, region_data.write_cf));
+    size_changed.add(RegionLockCFData::deserialize(buf, region_data.lock_cf));
+
+    region_data.cf_data_size = size_changed.payload;
+    region_data.decoded_data_size = size_changed.decoded;
+    recordMemChange(size_changed.payload);
```
JaySon-Huang, line 311:
```suggestion
    region_data.recordMemChange(size_changed);
```
`region_data.cf_data_size` and `region_data.decoded_data_size` will be changed in `region_data.recordMemChange`

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -342,33 +342,57 @@ bool RegionData::isEqual(const RegionData & r2) const
...
+    , decoded_data_size(data.decoded_data_size.load())
 {}
 
-
 RegionData::~RegionData()
 {
-    reportDealloc(cf_data_size);
+    recordMemChange(cf_data_size.negative());
     cf_data_size = 0;
+    decoded_data_size = 0;
```
JaySon-Huang, line 357:
```suggestion
    recordMemChange(RegionDataMemDiff{-cf_data_size, -decoded_data_size});
```

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
```diff
@@ -342,33 +342,57 @@ bool RegionData::isEqual(const RegionData & r2) const
...
+                        write_map_it = write_map.erase(write_map_it);
+                        continue;
+                    }
+                }
+            }
+        }
+        ++write_map_it;
+    }
+    // No need to check default cf. Because tikv will gc default cf before write cf.
+    return del_write;
```
JaySon-Huang, line 395: Seems that the free cf_data_size in this function does not change `root_of_kvstore_mem_trackers`. We should handle it as following

```suggestion
RegionDataMemDiff delta;
    for (auto write_map_it = write_map.begin(); write_map_it != write_map.end();)
    {
        const auto & decoded_val = std::get<2>(write_map_it->second);
        const auto & [pk, ts] = write_map_it->first;

        if (decoded_val.write_type == RecordKVFormat::CFModifyFlag::PutFlag)
        {
            if (!decoded_val.short_value)
            {
                if (auto data_it = default_map.find({pk, decoded_val.prewrite_ts}); data_it == default_map.end())
                {
                    // if key-val in write cf can not find matched data in default cf and its commit-ts < gc-safe-point, we can clean it safely.
                    if (ts < safe_point)
                    {
                        del_write += 1;
                        delta.sub(RegionWriteCFData::calcTotalKVSize(write_map_it->second));
                        write_map_it = write_map.erase(write_map_it);
                        continue;
                    }
                }
            }
        }
        ++write_map_it;
    }
    // No need to check default cf. Because tikv will gc default cf before write cf.
    recordMemChange(delta);
    return del_write;
```

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
CalvinNeo: I have tried your `recordMemChange`, and now I think we'd better not combined the `root_of_kvstore_mem_trackers` with `cf_data_size += delta.payload;`. A change in memory in a certain region doesn't necessarily change the total memory usage. IMO it doesn't make the code clearer. We still have to handle the `cf_data_size` anywhere else.

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
CalvinNeo: ditto

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
CalvinNeo: ditto

dbms/src/Storages/KVStore/MultiRaft/RegionData.cpp
CalvinNeo: fixed

JaySon-Huang (approved):
LGTM

PR #9813, opened 2025-01-22
Re-enable jemalloc's background thread by default

### LINKED ISSUE #9812: The jemalloc's background thread is disabled
## Bug Report

The current implementation is
```
    auto * malloc_conf = getenv("MALLOC_CONF");

    if (!malloc_conf && !old_b)
    {
        // enable background thread
    }
```

In previous version, there is not MALLOC_CONF set, so the background thread is enabled by default.

However, after https://github.com/pingcap/tiup/commit/bac3c2e61555ab6dd2b69ad18bc01cf8dfb7e700, TiFlash is bootstrapped with a given `MALLOC_CONF` env string which does not specify `background_thread` at all. So the background thread will not be enabled.

If the background thread is not enabled, dirty pages could not be purged in the expected time, if using jemalloc 5.2.1.

There are mainly several ways to dirty pages in jemalloc:
- Manual purge called by `mallctl`
- The value of `dirty_decay_ms` is changed
- A counter named `decay_ticker` will decrease from 1000, on alloc/dealloc events. Once it goes to 0, it would check if the current epoch should be advanced according to deadline. If so, a purge will trigger.
- The background thread is enabled, and it would trigger purge in a given interval.

The problem is, if there is no frequent alloc/dealloc events after one dealloc, the `decay_ticker` will not be decreased. thus the epoch is not checked, and the memory is not released.

So, we'd better enable background_thread again.

CalvinNeo:

Issue Number: close #9812 ref #9722 

Problem Summary:

See issue 9812

```commit-message
Re-enable jemalloc's background thread by default
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix an issue that the memory may stay in high level after large amount of data ingested
```

### DIFF
--- a/dbms/src/Server/Setup.cpp
+++ b/dbms/src/Server/Setup.cpp
@@ -112,7 +112,7 @@ void setupAllocator([[maybe_unused]] const LoggerPtr & log)
 
     LOG_INFO(log, "Got jemalloc config: opt.background_thread {}, opt.max_background_threads {}", old_b, old_max_thd);
 
-    bool not_config_bg = !malloc_conf || strstr(malloc_conf, "background_thread") == NULL;
+    bool not_config_bg = !malloc_conf || strstr(malloc_conf, "background_thread") == nullptr;
     if (not_config_bg && !old_b)
     {
         // If the user doesn't explicitly set the background_thread opt, and it is actually false, then set it to true.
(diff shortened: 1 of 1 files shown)


Lloyd-Pottiger (approved):

JaySon-Huang (approved):
lgtm

PR #9835, opened 2025-01-27
KVStore: Record table-wise memory usage and warn when it goes beyond the threshold

CalvinNeo:

Issue Number: ref #9722

Problem Summary:

- Introduce a struct `RegionTableSize` which is shared by a `Table` and all its related `Region`s. Once registered, a region will update its memory change to this struct, so the table would learn.
- Introduce `maybeWarnMemoryLimitByTable` and `resetWarnMemoryLimitByTable`, to log when the memory of a table goes beyond threshold.

Both of the systems works under the path that:
- When a region is created, either for prehandling or inserting, it would register it to RegionTable, and get its shared `RegionTableSize`.
- Ever since, the insert/delete from the Region would also reflect on this shared `RegionTableSize`.
- After an insertion, if the memory consumed by a table associated with the given region is reported to go beyond the threshold propotion, and the total memory used is also beyond the given memory limit(which is a very corner case), then we will issue a log.
- Afterwards, if the memory usage somehow goes upper, we don't print the log again for the same table.
- If the memory goes down by committing txns, or applying snapshots, we will reset the warning. So if the table still consumes much memory, a new log will be printed.
- If the region is splitted or merged, the table's memory usage is not affected. If the region is destroyed, the table's memory would be decresed.

Also:
- Remove some useless structures in RegionTable.
- Make safe ts related codes into SafeTsMgr.
- Rename some methods in `Region` to reflect what they actually do.

```commit-message
Record table-wise memory usage and warn when it goes beyond threshold
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add logging when the uncommitted in-memory data of a table reach a threshold
```


dbms/src/Storages/KVStore/KVStore.cpp
```diff
@@ -495,8 +503,13 @@ RegionPtr KVStore::genRegionPtr(metapb::Region && region, UInt64 peer_id, UInt64
...
     });
-
-    return std::make_shared<Region>(std::move(meta), proxy_helper);
+    auto new_region = std::make_shared<Region>(std::move(meta), proxy_helper);
+    if (register_to_table)
+    {
+        auto & region_table = tmt.getRegionTable();
+        region_table.addPrehandlingRegion(*new_region);
+    }
+    return new_region;
```
JaySon-Huang, line 512: `TMTContext` is a complicated class. Passing the TMTContext to `KVStore::genRegionPtr` and register it to region_table will make this method hard to understand.

Maybe add a method like 
```
TMTContext::genRegionPtr(
    metapb::Region && region,
    UInt64 peer_id,
    UInt64 index,
    UInt64 term)
{
    // ...
    auto region = kvstore->genRegionPtr(...);
    region_table.addPrehandlingRegion(*region);
    return region;
}
```

CalvinNeo: Yes, this is a problem, and is confusing.

However, because `TMTContext` is really a complicated class, so IMO it is also kind of strange to me if we create a Region in `TMTContext`. It seems to me a better idea to make RegionTable into KVStore, but it cost some refactoring.

So in general, I think it is better to use `std::optional` over `RegionTable`. Once given, then we will handle the `region_table`.

dbms/src/Storages/KVStore/Decode/RegionTable.cpp
```diff
@@ -118,17 +115,17 @@ RegionTable::RegionTable(Context & context_)
 
 void RegionTable::clear()
 {
-    regions.clear();
+    region_infos.clear();
     tables.clear();
-    safe_ts_map.clear();
+    safe_ts_mgr.safe_ts_map.clear();
```
JaySon-Huang, line 120:
```suggestion
    safe_ts_mgr.clear();
```

dbms/src/Storages/KVStore/Decode/SafeTsMgr.h
```diff
@@ -0,0 +1,77 @@
...
+    UInt64 getSelfSafeTS(UInt64 region_id) const;
+
+    void remove(UInt64 region_id)
+    {
+        std::unique_lock write_lock(rw_lock);
+        safe_ts_map.erase(region_id);
+    }
+
+    SafeTsMap safe_ts_map;
+    mutable std::shared_mutex rw_lock;
```
JaySon-Huang, line 75:
```suggestion
    void clear()
    {
        std::unique_lock write_lock(rw_lock);
        safe_ts_map.clear();
    }

private:
    SafeTsMap safe_ts_map;
    mutable std::shared_mutex rw_lock;
```

dbms/src/Storages/KVStore/Decode/RegionTable.cpp
```diff
@@ -401,18 +427,37 @@ void RegionTable::shrinkRegionRange(const Region & region)
...
+        // `old_region` will no longer contribute to the memory of the table.
+        auto keyspace_id = region_range_keys->getKeyspaceID();
+        auto table_id = region_range_keys->getMappedTableID();
+        auto & table = getOrCreateTable(keyspace_id, table_id);
+        old_region->resetRegionTableCtx();
+        if unlikely (!new_region->getRegionTableCtx())
+        {
+            // For most of the cases, the region is prehandled, so the ctx is set at that moment.
+            new_region->setRegionTableCtx(table.ctx);
+        }
```
JaySon-Huang, line 448: `getOrCreateTable` should be called under lock protection
```suggestion
        std::lock_guard lock(mutex);
        auto & table = getOrCreateTable(keyspace_id, table_id);
        old_region->resetRegionTableCtx();
        if unlikely (!new_region->getRegionTableCtx())
        {
            // For most of the cases, the region is prehandled, so the ctx is set at that moment.
            new_region->setRegionTableCtx(table.ctx);
        }
```

dbms/src/Storages/KVStore/Decode/RegionTable.cpp
```diff
@@ -68,16 +63,18 @@ RegionTable::Table & RegionTable::getOrCreateTable(const KeyspaceID keyspace_id,
 RegionTable::InternalRegion & RegionTable::insertRegion(Table & table, const Region & region)
 {
     const auto range = region.getRange();
-    return insertRegion(table, *range, region.id());
+    return insertRegion(table, *range, region);
 }
```
JaySon-Huang, line 67: Make line 108 in `RegionTable::getOrInsertRegion` as `return insertRegion(table, *region.getRange(), region);` and remove this redundant function?

dbms/src/Storages/KVStore/MultiRaft/IngestSST.cpp
```diff
@@ -113,7 +113,7 @@ RegionPtr KVStore::handleIngestSSTByDTFile(
         auto meta_region = region->cloneMetaRegion();
         auto meta_snap = region->dumpRegionMetaSnapshot();
         auto peer_id = meta_snap.peer.id();
-        tmp_region = genRegionPtr(std::move(meta_region), peer_id, index, term);
+        tmp_region = genRegionPtr(std::move(meta_region), peer_id, index, term, std::nullopt);
```
JaySon-Huang, line 116:
```suggestion
        // the uncommitted data will be merged from `tmp_region` to the existing region. So here ignore the region_table
        tmp_region = genRegionPtr(std::move(meta_region), peer_id, index, term, std::nullopt);
```

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
```diff
@@ -52,6 +52,10 @@ struct RegionDataMemDiff
         payload -= other.payload;
         decoded -= other.decoded;
     }
+
+    Type total() const { return payload + decoded; }
+
+    size_t totalSize() const { return payload + decoded; }
```
JaySon-Huang, line 58: Why we need two method for the same result? Maybe just adding explicit cast when want to assign the result to differnet type?

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +401,48 @@ Timestamp Region::getLastObservedReadTso() const
...
+        {
+            if (!setRegionTableWarned(true))
+            {
+                // If it is the first time.
+#ifdef DBMS_PUBLIC_GTEST
+                tmt.getKVStore()->debug_memory_limit_warning_count++;
+#endif
+                LOG_INFO(
+                    log,
+                    "Memory limit exeeded, current={} limit={} table_id={} keyspace_id={} region_id={} from={}",
```
JaySon-Huang, line 431:
```suggestion
                    "Memory limit exceeded, current={} limit={} table_id={} keyspace_id={} region_id={} from={}",
```

dbms/src/Storages/KVStore/tests/gtest_region_persister.cpp
```diff
@@ -567,15 +567,19 @@ try
 
         auto region = makeRegion(createRegionMeta(region_100, table_id));
         TiKVKey key = RecordKVFormat::genKey(table_id, region_100, diff++);
-        region->insert(ColumnFamilyType::Default, TiKVKey::copyFrom(key), TiKVValue("value1"));
-        region->insert(ColumnFamilyType::Write, TiKVKey::copyFrom(key), RecordKVFormat::encodeWriteCfValue('P', 0));
-        region->insert(
+        region->insertFromSnap(tmt, ColumnFamilyType::Default, TiKVKey::copyFrom(key), TiKVValue("value1"));
```
JaySon-Huang, line 570: why use `insertFromSnapshot` rather than `insertDebug` for these test cases?

dbms/src/Storages/KVStore/Region.h
```diff
@@ -133,8 +134,19 @@ class Region : public std::enable_shared_from_this<Region>
     Region() = delete;
     ~Region();
 
-    void insert(const std::string & cf, TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
-    void insert(ColumnFamilyType type, TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
+    void insertDebug(const std::string & cf, TiKVKey && key, TiKVValue && value, DupCheck mode = DupCheck::Deny);
+    void insertFromSnap(
```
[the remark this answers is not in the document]
CalvinNeo: Because this insertion happens only when decoding a raft snapshot

dbms/src/Storages/KVStore/Decode/RegionTable.cpp
CalvinNeo: Yes, thx

dbms/src/Storages/KVStore/Region.h
```diff
@@ -228,6 +242,17 @@ class Region : public std::enable_shared_from_this<Region>
     RegionData::OrphanKeysInfo & orphanKeysInfo() { return data.orphan_keys_info; }
     const RegionData::OrphanKeysInfo & orphanKeysInfo() const { return data.orphan_keys_info; }
 
+    // Bind a region to a RegionTable. It could not be bound to another table any more.
+    // All memory changes to this region would relect to the binded table.
```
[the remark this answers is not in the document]
CalvinNeo: nice catch, would fix this typo

dbms/src/Storages/KVStore/tests/gtest_region_persister.cpp
CalvinNeo: Due to convenience. But yes, it is confusing, I will change it to debug

dbms/src/Storages/KVStore/Decode/RegionTable.cpp
CalvinNeo: OK

dbms/src/Storages/KVStore/MultiRaft/IngestSST.cpp
CalvinNeo: Will merge to the previous comment

dbms/src/Storages/KVStore/Decode/RegionTable_fwd.h
```diff
@@ -0,0 +1,34 @@
...
+namespace DB
+{
+struct RegionTableCtxInner
+{
+    // KVStore size of all regions of this table, including the prehandling one.
+    // So, this size may be larger than the table's real size.
+    std::atomic_int64_t table_size;
+    std::atomic_bool warned;
+};
+using RegionTableCtx = std::shared_ptr<RegionTableCtxInner>;
```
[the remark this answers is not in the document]
CalvinNeo: OK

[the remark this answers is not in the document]
CalvinNeo: Just keep those function name to make them shorter?

dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.h
CalvinNeo: `total()` needs no casting, but is not used very much,
`totalSize()` needs casting, but is widely used in test, so it is we just write it here to eliminate some `static_cast`ing

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +401,48 @@ Timestamp Region::getLastObservedReadTso() const
...
+void Region::setRegionTableCtx(RegionTableCtxPtr ctx) const
+{
+    data.setRegionTableCtx(ctx);
+}
+
+void Region::maybeWarnMemoryLimitByTable(TMTContext & tmt, const char * from)
+{
+    // If there are data flow in, we will check if the memory is exhaused.
+    auto limit = tmt.getKVStore()->getKVStoreMemoryLimit();
+    auto current = real_rss.load();
```
JaySon-Huang, line 413: * If KVStoreMemoryLimit (max_memory_usage_for_all_queries) is explicitly set to 0, now it  will not print any logging. IMO, for this case we should use the server memory capacity * 80% as the warning limit.
* Only print log when the region_table_size > 50% * current is too large. I think we may easily miss the logging for  track. I recommend to lower the table level memory threshold to min(20% * current, 10GiB) to increase the recall rate for logging down info.

CalvinNeo: - OK
- We'd better substract "start_time" memory.

CalvinNeo:
New method
```
[2025/02/14 18:59:53.199 +08:00] [INFO] [Server.cpp:975] ["Global memory status: computed_limit 4294967297 tranquil_time_rss 737959936, cur_virt_size 2888970240, capacity 403755778048"] [thread_id=1]

[2025/02/14 19:09:16.426 +08:00] [INFO] [Region.cpp:444] ["Memory limit exceeded, current=3889836032 limit=4294967297 limit_size=945562828.8 table_size=641429928 table_id=183 keyspace_id=4294967295 region_id=4226 from=put"] [thread_id=368]
```

image

dbms/src/Server/Server.cpp
```diff
@@ -955,7 +957,25 @@ int Server::main(const std::vector<std::string> & /*args*/)
...
+        LOG_INFO(
+            log,
+            "Global memory status: computed_limit {} tranquil_time_rss {}, cur_virt_size {}, capacity {}",
+            limit,
+            tranquil_time_rss,
+            cur_virt_size,
+            server_info.memory_info.capacity);
+        if (limit == 0)
+            limit = server_info.memory_info.capacity * 0.8;
+        proxy_machine.initKVStore(global_context->getTMTContext(), store_ident, limit);
```
JaySon-Huang, line 978:
```suggestion
        auto kvs_watermark = settings.max_memory_usage_for_all_queries.getActualBytes(server_info.memory_info.capacity);
        if (kvs_watermark == 0)
            kvs_watermark = server_info.memory_info.capacity * 0.8;
        LOG_INFO(
            log,
            "Global memory status: kvstore_high_watermark={} tranquil_time_rss={} cur_virt_size={} capacity={}",
            limit,
            tranquil_time_rss,
            cur_virt_size,
            server_info.memory_info.capacity);

        proxy_machine.initKVStore(global_context->getTMTContext(), store_ident, kvs_watermark);
```

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +402,56 @@ Timestamp Region::getLastObservedReadTso() const
...
+{
+    data.setRegionTableCtx(ctx);
+}
+
+void Region::maybeWarnMemoryLimitByTable(TMTContext & tmt, const char * from)
+{
+    // If there are data flow in, we will check if the memory is exhaused.
+    auto limit = tmt.getKVStore()->getKVStoreMemoryLimit();
+    size_t current = real_rss.load() > 0 ? real_rss.load() : 0;
+    if likely (limit == 0 || current == 0)
```
JaySon-Huang, line 415:
```suggestion
    if unlikely (limit == 0 || current == 0)
```

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +402,56 @@ Timestamp Region::getLastObservedReadTso() const
...
+            if (!setRegionTableWarned(true))
+            {
+                // If it is the first time.
+#ifdef DBMS_PUBLIC_GTEST
+                tmt.getKVStore()->debug_memory_limit_warning_count++;
+#endif
+                LOG_INFO(
+                    log,
+                    "Memory limit exceeded, current={} limit={} table_limit={} table_size={} table_id={} keyspace_id={} "
+                    "region_id={} from={}",
```
JaySon-Huang:
```suggestion
                    "Memory limit exceeded, current={} limit={} table_limit={} table_in_mem_size={} table_id={} keyspace_id={} "
                    "region_id={} from={}",
```

dbms/src/Storages/KVStore/MultiRaft/Persistence.cpp
```diff
@@ -222,14 +222,17 @@ bool KVStore::canFlushRegionDataImpl(
         LOG_INFO(
             log,
             "{} flush region due to tryFlushRegionData, index {} term {} truncated_index {} truncated_term {}"
-            " gap {}/{}",
+            " gap {}/{} table_size {} table_id={} keyspace_id={}",
```
JaySon-Huang, line 225:
```suggestion
            " gap {}/{} table_in_mem_size={} table_id={} keyspace={}",
```

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +402,57 @@ Timestamp Region::getLastObservedReadTso() const
...
+    /// Region management such as split/merge doesn't change the memory consumed by a table in KVStore.
+    /// The only cases memory is reduced in a table is removing regions, applying snaps and commiting txns.
+    /// The only cases memory is increased in a table is inserting kv pairs and applying snaps.
+    /// So, we only print once for a table, until one memory reduce event will happen.
+    if unlikely (current >= limit * 0.9)
+    {
+        auto table_size = getRegionTableSize();
+        auto grown_memory = current > tranquil_time_rss ? current - tranquil_time_rss : 0;
+        // 15% of the total non-tranquil-time memory, but not exceed 10GB.
+        auto table_memory_limit = std::min(grown_memory * 0.15, 10 * 1024 * 1024 * 1024);
```
JaySon-Huang, line 426:
```suggestion
        auto table_memory_limit = std::min(grown_memory * 0.15, 10 * 1024ULL * 1024 * 1024);
```

dbms/src/Storages/KVStore/Region.cpp
```diff
@@ -387,4 +402,57 @@ Timestamp Region::getLastObservedReadTso() const
...
+            {
+                // If it is the first time.
+#ifdef DBMS_PUBLIC_GTEST
+                tmt.getKVStore()->debug_memory_limit_warning_count++;
+#endif
+                LOG_INFO(
+                    log,
+                    "Memory limit exceeded, current={} limit={} table_limit={} table_size={} table_id={} "
+                    "keyspace_id={} "
+                    "region_id={} from={}",
```
JaySon-Huang, line 439:
```suggestion
                    "Memory limit exceeded, current={} limit={} table_limit={} table_in_mem_size={} table_id={} "
                    "keyspace={} "
                    "region_id={} from={}",
```

JaySon-Huang (approved):
lgtm

We need to investigate why "It shows a byte cost in KVStore will result in 2 bytes cost in memory usage." later

dbms/src/Server/Server.cpp
```diff
@@ -107,6 +108,7 @@
 #include <fiu.h>
 #endif
 
+std::atomic<UInt64> tranquil_time_rss;
```
CalvinNeo, line 111: TODO could be moved into KVStore to avoid this atomic. Not necessarily in this PR.

Issue #9750, opened 2024-12-30
TiFlash compute node crashes after executing `ALTER RANGE` in TiDB


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Reported from https://asktug.com/t/topic/1037727

```
ALTER PLACEMENT POLICY evict_sata_dw CONSTRAINTS="[-disk=sata-new, -disk=dw-ssd]" SURVIVAL_PREFERENCES="[host]";
ALTER RANGE global PLACEMENT POLICY evict_sata_dw;
```

### 3. What did you see instead (Required)

```
[2024/12/28 12:52:05.786 +08:00] [ERROR] [BaseDaemon.cpp:370] [########################################] [source=BaseDaemon] [thread_id=41]
[2024/12/28 12:52:05.788 +08:00] [ERROR] [BaseDaemon.cpp:371] ["(from thread 42) Received signal Segmentation fault(11)."] [source=BaseDaemon] [thread_id=41]
[2024/12/28 12:52:05.788 +08:00] [ERROR] [BaseDaemon.cpp:401] ["Address: 0x10"] [source=BaseDaemon] [thread_id=41]
[2024/12/28 12:52:05.788 +08:00] [ERROR] [BaseDaemon.cpp:407] ["Access: read."] [source=BaseDaemon] [thread_id=41]
[2024/12/28 12:52:05.788 +08:00] [ERROR] [BaseDaemon.cpp:416] ["Address not mapped to object."] [source=BaseDaemon] [thread_id=41]
[...]

PR #9849, opened 2025-02-07
disagg: Add exclusive label for tiflash compute node

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9750

Problem Summary:

When creating a placement policy through tidb, tidb will create a placement-rule with `"engine" notIn "tiflash"` https://github.com/pingcap/tidb/pull/22065
However, under tiflash disaggregated compute and storage arch, the compute node will register store with label `{"engine": "tiflash_compute"}`.

The logic in PD of choosing store for rule: [pkg/schedule/placement/label_constraint.go @ pd](https://github.com/tikv/pd/blob/7d33065019f78d9150a8c89ddb4593f81e6ff9b3/pkg/schedule/placement/label_constraint.go#L81-L95)

1. If the store's label.key is equal to "engine" (or "exclusive," or starts with '$'), and the rule's constraint does not contain a rule with the same label.key, then do not schedule the peer to this store.
2. Otherwise, further check if the rule's constraint matches the store's label. If it matches, schedule the peer to this store.

So the PD would pick tiflash compute node as target store to place the Region peer.

```commit-message
Add a label with {"exclusive": "no-data"}, so that PD won't select TiFlash compute node stores for placing Region
```

We avoid Region data being placed to TiFlash compute node in 2 ways:
* After https://github.com/pingcap/tidb/issues/58633, tidb will NOT set the placement rule with "engine" notIn "tiflash" explicitly.
* This PR in TiFlash compute node, it will add a label with {"exclusive": "no-data"}, so that PD won't select this store for placing Region
  * Remove `--engine-role-label` from proxy args, use `--labels` instead
  * Proxy args `--labels` won't overwrite the `server.labels` in tiflash_tikv.toml, instead, proxy append the extra labels to `server.labels`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a bug that after executing `ALTER TABLE ... PLACEMENT POLICY ...` Region peers may be unexpectedly added to the TiFlash Compute Node in the disaggregated storage and compute architecture
```

### DIFF
--- a/dbms/src/Storages/KVStore/ProxyStateMachine.h
+++ b/dbms/src/Storages/KVStore/ProxyStateMachine.h
@@ -148,9 +148,29 @@ private:
             else
                 args_map["advertise-engine-addr"] = args_map["engine-addr"];
             args_map["engine-label"] = getProxyLabelByDisaggregatedMode(disaggregated_mode);
-            // For tiflash write node, it should report a extra label with "key" == "engine-role-label"
+            String extra_label;
             if (disaggregated_mode == DisaggregatedMode::Storage)
-                args_map["engine-role-label"] = DISAGGREGATED_MODE_WRITE_ENGINE_ROLE;
+            {
+                // For tiflash write node, it should report a extra label with "key" == "engine-role-label"
+                // to distinguish with the node under non-disagg mode
+                extra_label = fmt::format("engine_role={}", DISAGGREGATED_MODE_WRITE_ENGINE_ROLE);
+            }
+            else if (disaggregated_mode == DisaggregatedMode::Compute)
+            {
+                // For compute node, explicitly add a label with `exclusive=no-data` to avoid Region
+                // being placed to the compute node.
+                // Related logic in pd-server:
+                // https://github.com/tikv/pd/blob/v8.5.0/pkg/schedule/placement/label_constraint.go#L69-L95
+                extra_label = "exclusive=no-data";
+            }
+            if (args_map.contains("labels"))
+                extra_label = fmt::format("{},{}", args_map["labels"], extra_label);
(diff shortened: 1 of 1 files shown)


JaySon-Huang:
Depends on https://github.com/pingcap/tidb-engine-ext/pull/420

JaySon-Huang:
/unhold

The proxy change has been merged

JaySon-Huang:
@CalvinNeo @Lloyd-Pottiger @JinheLin PTAL

CalvinNeo (approved):
lgtm

Lloyd-Pottiger (approved):

dbms/src/Server/Server.cpp
```diff
@@ -242,9 +242,29 @@ struct TiFlashProxyConfig
             else
                 args_map["advertise-engine-addr"] = args_map["engine-addr"];
             args_map["engine-label"] = getProxyLabelByDisaggregatedMode(disaggregated_mode);
-            // For tiflash write node, it should report a extra label with "key" == "engine-role-label"
+            String extra_label;
             if (disaggregated_mode == DisaggregatedMode::Storage)
-                args_map["engine-role-label"] = DISAGGREGATED_MODE_WRITE_ENGINE_ROLE;
+            {
```
CalvinNeo, line 247: Wrap it with `#if SERVERLESS_PROXY == 1`?

JaySon-Huang: @CalvinNeo PTAL again

CalvinNeo (approved):
lgtm

PR #9852, opened 2025-02-08
vector: Reduce logging under debug level

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9854

Problem Summary:

When we try to add a vector index on a relative large table, there are much useless logging under debug level

* After DDL owner try to add a vector index, the tidb ddl owner will get the add index progress by fetching `system.dt_local_indexes` from all tiflash instances
* `StorageSystemDTLocalIndexes::read` call `getLocalIndexesStatsFromStorage`. For non-inited `DeltaMergeStore`, the function calls `genLocalIndexStatsByTableInfo` -> `initLocalIndexInfos` -> `generateLocalIndexInfos` and generate useless logging like `"Local index info does not changed, ..."` for each `StorageDeltaMerge`. This will cause much debug logging when there are lots of tables in the cluster

```commit-message
* Move the logging from `generateLocalIndexInfos` to `DeltaMergeStore::applyLocalIndexChange`. Reduce logging that is not cause by DDL changed
* Return the keep/added/dropped index ids in `LocalIndexInfosChangeset` and generate the changeset logging string by `LocalIndexInfosChangeset::toString`
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
Deploy a cluster with tiflash and enable debug level logging
Load a dataset with 1m vector data
Add vector index and wait
```
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore.cpp
@@ -2047,7 +2047,7 @@ void DeltaMergeStore::applyLocalIndexChange(const TiDB::TableInfo & new_table_in
     // no index is created or dropped
     if (!changeset.new_local_index_infos)
     {
-        LOG_DEBUG(log, "Local index info does not changed, {}", changeset.toDebugString(/*simple*/ false));
+        LOG_DEBUG(log, "Local index info does not changed, {}", changeset.toDebugString());
         return;
     }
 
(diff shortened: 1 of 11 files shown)


CalvinNeo (approved):
lgtm

Lloyd-Pottiger (approved):

PR #9914, opened 2025-02-25
FFI: Fix NPE when GetLockByKey is called when region is not exist

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9915

Problem Summary:
There could be a chance that GetLockByKey is called when region is not exist
And returning `BaseBuffView` to rust is not memory safe

Pending on https://github.com/pingcap/tidb-engine-ext/pull/426

```commit-message
FFI: Fix NPE when GetLockByKey is called when region is not exist
FFI: Use CppStrWithView as the return type of fn_get_lock_by_key
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
+++ b/dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
@@ -1226,7 +1226,7 @@ UInt64 DeltaMergeStore::ingestSegmentsFromCheckpointInfo(
 
     auto restored_segments = checkpoint_info->getRestoredSegments();
     auto updated_segments = ingestSegmentsUsingSplit(dm_context, range, restored_segments);
-    auto estimated_bytes = 0;
+    size_t estimated_bytes = 0;
 
     for (const auto & segment : restored_segments)
     {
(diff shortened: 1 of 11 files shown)


dbms/src/Storages/DeltaMerge/DeltaMergeStore_Ingest.cpp
```diff
@@ -1226,7 +1226,7 @@ UInt64 DeltaMergeStore::ingestSegmentsFromCheckpointInfo(
 
     auto restored_segments = checkpoint_info->getRestoredSegments();
     auto updated_segments = ingestSegmentsUsingSplit(dm_context, range, restored_segments);
-    auto estimated_bytes = 0;
+    size_t estimated_bytes = 0;
```
JaySon-Huang, line 1229: `auto` in here unexpectedly lead to negative value because the compiler use `int32` for the auto. And collecting the bytes of restored_segment could larger than 2GiB.

PR #9920, opened 2025-02-26
fix virtual column not found when remote read happens

### LINKED ISSUE #9561: Not found column generated_column_xxx in block when remote read is triggered
## Bug Report

```
CREATE TABLE `t` (
`id` int(11) NOT NULL AUTO_INCREMENT COMMENT '?',
`name` varchar(100) NOT NULL DEFAULT '' COMMENT '?',
KEY `idx_basic_lower_name` ((lower(`name`)))
) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_bin AUTO_INCREMENT = 32409068
/*T! SHARD_ROW_ID_BITS=4 */
COMMENT = '?';

insert into t (name)values("201801");
insert into t (name)values("201801");
insert into t (name)values("xxxxx");

insert into t (select * from t);

...
```

select a generated column`explain analyze SELECT * FROM t a WHERE lower(a.name) like '%201801%';` and  triggere remote read.

no error

get error: Not found column generated_column_xxx in block

nightly: 02784a4c7e9

guo-shaoge:

Issue Number: close #9561

Problem Summary:
```
select col, gen_col from t;
```
gen_col is a virtual generated column. The current approach is that TiFlash generates an empty column and places it in the block. Then, it passes the block to TiDB, allowing TiDB to calculate and populate the column.

I’ve named this column generated_xxx.

The information about this column is marked in tipb_table_scan.columns with a flag. When TiFlash sees this flag, it automatically adds a generated_xxx column.

However, this situation has an issue when a remote read occurs: the block returned from the remote read, when parsed, assigns names to each column based on RemoteRequest.schema. For example, if TiFlash-1 receives a block from TiFlash-2, it will parse it as <col, gen_col>.

But TiFlash-1, in the table scan's projection, expects the block to be <col, generated_xxx>. So, when projecting the block from the remote read, an error occurs because it cannot find the generated_xxx column.

make the column name of local read and remote read be same

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
fix virtual column not found when remote read happens
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/TablesRegionsInfo.cpp
+++ b/dbms/src/Flash/Coprocessor/TablesRegionsInfo.cpp
@@ -71,10 +71,12 @@ static void insertRegionInfoToTablesRegionInfo(
     Int64 table_id,
     TablesRegionsInfo & tables_region_infos,
     std::unordered_set<RegionID> & local_region_id_set,
-    const TMTContext & tmt_context,
-    bool force_remote [[maybe_unused]])
+    const TMTContext & tmt_context)
 {
     auto & table_region_info = tables_region_infos.getOrCreateTableRegionInfoByTableID(table_id);
+#ifndef NDEBUG
+    size_t i = 0;
+#endif
     for (const auto & r : regions)
     {
         RegionInfo region_info(
@@ -100,14 +102,15 @@ static void insertRegionInfoToTablesRegionInfo(
         /// 4. The remote read will fetch the newest region info via key ranges. So it is possible to find the region
         ///    is served by the same node (but still read from remote).
         bool duplicated_region = local_region_id_set.contains(region_info.region_id);
-
 #ifndef NDEBUG
-        // Only for failpoint test.
-        if (force_remote)
-        {
-            table_region_info.remote_regions.push_back(region_info);
-            continue;
-        }
+        fiu_do_on(FailPoints::force_random_remote_read_for_partition_table, {
+            if ((i % 2) != 0)
+            {
+                table_region_info.remote_regions.push_back(region_info);
+                continue;
+            }
+        });
+        ++i;
 #endif
 
         if (duplicated_region || needRemoteRead(region_info, tmt_context))
@@ -134,26 +137,18 @@ TablesRegionsInfo TablesRegionsInfo::create(
             InvalidTableID,
             tables_regions_info,
             local_region_id_set,
-            tmt_context,
-            false);
+            tmt_context);
     else
     {
-        for (int i = 0; i < table_regions.size(); ++i) // NOLINT
+        for (const auto & table_region : table_regions)
         {
-            const auto & table_region = table_regions[i];
             assert(table_region.physical_table_id() != InvalidTableID);
-            bool force_remote = false;
-            fiu_do_on(FailPoints::force_random_remote_read_for_partition_table, {
-                if ((i % 2) != 0)
-                    force_remote = true;
-            });
             insertRegionInfoToTablesRegionInfo(
                 table_region.regions(),
                 table_region.physical_table_id(),
                 tables_regions_info,
                 local_region_id_set,
-                tmt_context,
-                force_remote);
+                tmt_context);
         }
         assert(static_cast<UInt64>(table_regions.size()) == tables_regions_info.tableCount());
     }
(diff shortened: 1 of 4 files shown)


tests/fullstack-test/mpp/remote_read_virtual_column.test
```diff
@@ -0,0 +1,41 @@
...
+mysql> alter table test.test_vir set tiflash replica 1;
+mysql> set @@cte_max_recursion_depth = 10000000; insert into test.test_vir (id, str) with recursive cte1 as (select 1 c1, 'a' c2 union select c1+1 c1, substr(concat(c2, 'B'), 10) c2 from cte1 limit 10000) select * from cte1;
+
+func> wait_table test test_vir
+mysql> split table test.test_vir by (100), (500), (1000);
+TOTAL_SPLIT_REGION	SCATTER_FINISH_RATIO
+3	1
+
+=> DBGInvoke __enable_fail_point(force_random_remote_read_for_partition_table)
+mysql> set @@tidb_isolation_read_engines='tiflash'; set @@tidb_enforce_mpp=1; select * from test.test_vir order by 1, 2, 3 limit 5;
```
solotzg, line 30: It seems that the table must have at least 2 regions, and part of them trigger remote read, while others do not.

dbms/src/Flash/Coprocessor/TablesRegionsInfo.cpp
```diff
@@ -73,6 +74,9 @@ static void insertRegionInfoToTablesRegionInfo(
     const TMTContext & tmt_context)
 {
     auto & table_region_info = tables_region_infos.getOrCreateTableRegionInfoByTableID(table_id);
+#ifndef NDEBUG
+    size_t i = 0;
+#endif
     for (const auto & r : regions)
```
solotzg, line 80: for (size_t i = 0; i< regions.size(); i++ ) {
auto & r = regions[i];
}

dbms/src/Flash/Coprocessor/TablesRegionsInfo.cpp
```diff
@@ -98,6 +102,16 @@ static void insertRegionInfoToTablesRegionInfo(
         /// 4. The remote read will fetch the newest region info via key ranges. So it is possible to find the region
         ///    is served by the same node (but still read from remote).
         bool duplicated_region = local_region_id_set.contains(region_info.region_id);
+#ifndef NDEBUG
+        fiu_do_on(FailPoints::force_random_remote_read, {
+            if ((i % 2) != 0)
+            {
+                table_region_info.remote_regions.push_back(region_info);
+                continue;
```
solotzg, line 110: `continue;` can not work here, please try to set `duplicated_region` to true.

dbms/src/Flash/Coprocessor/TablesRegionsInfo.cpp
```diff
@@ -98,6 +102,16 @@ static void insertRegionInfoToTablesRegionInfo(
...
+            }
+        });
+        ++i;
```
solotzg, line 113:
```suggestion
```

tests/fullstack-test/mpp/remote_read_virtual_column.test
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Preparation.
+=> DBGInvoke __init_fail_point()
+
+mysql> drop table if exists test.test_vir;
```
solotzg, line 18: Can you add some test cases about partition tables ?

guo-shaoge: I found a bug of partition table when virtual column is used, so we cannot recovery this bug on master for partition table. (but we can recovery on 7.5.5)

solotzg (approved):
lgtm

PR #9921, opened 2025-02-27
 KVStore: Defer deleting locks until after the committed data has been written to storage

### LINKED ISSUE #9922: Transaction locks can only be removed after committed data has been written to the storage.

JinheLin:

Issue Number: close #9922

Problem Summary:

If we directly remove the logic of the read thread writing to the storage engine during learner reads in the existing code, it would cause concurrency control problems that could result in missing some of the latest data. The concurrency logic is as follows:  
1. **Raft thread**: Receives a write log and removes the lock.  
2. **Read thread**: Detects no lock and begins reading (missing the data written in step 3).  
3. **Raft thread**: Writes the data.  

The root cause of this issue is that the Raft thread removes the lock **before** writing the data to the storage engine.

- Removes locks after committed data has been written to the storage.

- Rename `Region::remove` to `Region:removeDebug` because this function only used to tests.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
  - Consistency test: 
image

- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -348,6 +348,8 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
 
     auto is_v2 = this->getClusterRaftstoreVer() == RaftstoreVer::V2;
 
+    std::vector<TiKVKey> deleting_lock_keys;
+    deleting_lock_keys.reserve(cmds.len / 2); // write + lock
```
CalvinNeo, line 352: Could also be:
- lots of locks put, e.g. big txn
- some default cfs, e.g. large value

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -348,6 +348,10 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
 
     auto is_v2 = this->getClusterRaftstoreVer() == RaftstoreVer::V2;
 
+    std::vector<TiKVKey> deleting_lock_keys;
```
CalvinNeo, line 351: How about also move the description here as comments?

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -520,6 +527,13 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
             }
         }
 
+        if (!deleting_lock_keys.empty())
+        {
+            std::unique_lock<std::shared_mutex> lock(mutex);
+            for (const auto & tikv_key : deleting_lock_keys)
+                doRemove(ColumnFamilyType::Lock, tikv_key);
```
JaySon-Huang, line 534: wrap it with 

```
            try {
               // ...
            }
            catch (Exception & e)
            {
                LOG_ERROR(
                    log,
                    "{} catch exception: {}, while applying `CmdType::Delete` on [term {}, index {}], key in hex: {}, "
                    "CF {}",
                    toString(),
                    e.message(),
                    term,
                    index,
                    tikv_key.toDebugString(),
                    CFToName(cf));
                e.rethrow();
            }
}
```

JaySon-Huang (commented):
Should we apply the change of `remove write storage in doLearnerRead` in the same PR?

Rest LGTM

CalvinNeo:
> Should we apply the change of `remove write storage in doLearnerRead` in the same PR?
> 
> Rest LGTM

I think it is better that we firstly merge this PR, and let it run for a week for all relevant tests to make sure this change is less likely to have problems.

If we remove the `remove write storage in doLearnerRead`, then there will be no concurrent decoding, so some problem may not emerge

dbms/src/Storages/KVStore/MultiRaft/RaftCommands.cpp
```diff
@@ -520,6 +534,31 @@ std::pair<EngineStoreApplyRes, DM::WriteResult> Region::handleWriteRaftCmd(
...
+            {
+                std::unique_lock<std::shared_mutex> lock(mutex);
+                for (; i < deleting_lock_keys.size(); ++i)
+                    doRemove(ColumnFamilyType::Lock, deleting_lock_keys[i]);
+            }
+            catch (Exception & e)
+            {
+                LOG_ERROR(
+                    log,
+                    "{} catch exception: {}, while applying `CmdType::Delete` on [term {}, index {}], key in hex: {}, "
```
CalvinNeo, line 550:
```suggestion
                    "{} catch exception: {}, while applying `CmdType::Delete` on lock cf at [term {}, index {}], key in hex: {}, "
```

CalvinNeo (approved):
lgtm

JaySon-Huang (approved):
lgtm

PR #9987, opened 2025-03-11
Columns: optimize ColumnString filter when selectivity is high

Lloyd-Pottiger:

Issue Number: ref #6233

Problem Summary:

```commit-message
optimize the performance of ColumnString filter when the selectivity of filter is high:

For example, when filter is `0111111111111111011111111111111101111111111111110111111111111111`, 
the mask will be `11111111111111110111111111111111101111111111111111011111111111111110`, 
since it does not be `[0]*[1]+` or `[1]+[0]*`, we need to copy each selected row one by one.

Now, we can copy 15 rows at once.

The total elapsed time of TPC-H 50 reduce from 42.9s to 41.1s.
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Optimize the performance of ColumnString filter when the selectivity of filter is high. The total elapsed time of TPC-H 50 reduce 4%.
```

### DIFF
--- a/dbms/src/Columns/filterColumn.cpp
+++ b/dbms/src/Columns/filterColumn.cpp
@@ -134,31 +134,14 @@ void filterArraysImplGeneric(
     while (filt_pos < filt_end_aligned)
     {
         auto mask = ToBits64(filt_pos);
-        if likely (0 != mask)
+        while (mask)
         {
-            if (const auto prefix_to_copy = prefixToCopy(mask); 0xFF != prefix_to_copy)
-            {
-                copy_chunk(offsets_pos, prefix_to_copy);
-            }
-            else
-            {
-                if (const auto suffix_to_copy = suffixToCopy(mask); 0xFF != suffix_to_copy)
-                {
-                    copy_chunk(offsets_pos + FILTER_SIMD_BYTES - suffix_to_copy, suffix_to_copy);
-                }
-                else
-                {
-                    while (mask)
-                    {
-                        // 100011111000 -> index: 3, length: 5, mask: 100000000000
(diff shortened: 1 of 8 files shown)


dbms/src/Columns/filterColumn.cpp
```diff
@@ -251,28 +240,17 @@ inline void filterImplAligned(
...
-                    }
-                }
-            }
+            // 100011111000 -> index: 3, length: 5, mask: 100000000000
+            size_t index = std::countr_zero(mask);
+            size_t length = std::countr_one(mask >> index);
+            res_data.insert(data_pos + index, data_pos + index + length);
+            // bitshifts are undefined for 64, so we need to check if we are at the end
+            if (index + length == FILTER_SIMD_BYTES)
+                break;
```
gengliqi, line 251: How about adding a predefined mask table?
```cpp
constexpr uint64_t MASKS[65] = [] {
    uint64_t masks[65] = {};
    for (int i = 0; i < 64; ++i) {
        masks[i] = ~((1ULL << i) - 1);
    }
    masks[64] = 0;
    return masks;
}();
mast &= MASKS[index + length];
```

Lloyd-Pottiger: Good idea!

windtalker (commented):
Do we have some existing test for this?

gengliqi (approved):

windtalker (approved):
lgtm

PR #10004, opened 2025-03-14
Fix spill crash

### LINKED ISSUE #9999: Filter operator does not return constant column for filter column, which may crash TiFlash when sort spill is triggered.
## Bug Report

Filter operator should always return constant 1 column for filter column, but in current implementation, it may not be true if a block with all rows be kept:
https://github.com/pingcap/tiflash/blob/7262f960c72ae473943f2f59133f0889ad269867/dbms/src/DataStreams/FilterTransformAction.cpp#L140-L147

I believe it is just a typo because the comments shows "Replace the column with the filter by a constant."

Filter not return constant column for filter column may make sort spill crash because in sort operator, it will remove constant column:
https://github.com/pingcap/tiflash/blob/7262f960c72ae473943f2f59133f0889ad269867/dbms/src/Operators/MergeSortTransformOp.cpp#L151

If some block returned by filter operator contants constant column while others does not contain constant column, the block in [`sorted_blocks`](https://github.com/pingcap/tiflash/blob/7262f960c72ae473943f2f59133f0889ad269867/dbms/src/Operators/MergeSortTransformOp.h#L111) will have different schemas, which will cause a series of troubles, the worst case is make TiFlash crash.

windtalker:

Issue Number: close #9999

Problem Summary:

The root cause is described in #9999, this pr fix the error and add some check in sort operator to avoid potential undefined fail when the input block of sort is not well generated.
```commit-message
Fix sort spill crash in some corner cases.
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix sort spill crash in some corner cases
```

### DIFF
--- a/dbms/src/DataStreams/MergeSortingBlockInputStream.cpp
+++ b/dbms/src/DataStreams/MergeSortingBlockInputStream.cpp
@@ -89,7 +89,7 @@ Block MergeSortingBlockInputStream::readImpl()
 
             SortHelper::removeConstantsFromBlock(block);
             if (block.columns() != header_without_constants.columns())
-                throw Exception("Unexpected number of constant columns in block in MergeSortingBlockInputStream.");
+                throw Exception(ErrorCodes::LOGICAL_ERROR, "Unexpected number of constant columns in block in MergeSortingBlockInputStream, n_block={} n_header={}", block.columns(), header_without_constants.columns());
 
             blocks.push_back(block);
             sum_bytes_in_blocks += block.estimateBytesForSpill();
--- a/dbms/src/DataStreams/MergeSortingBlockInputStream.cpp
+++ b/dbms/src/DataStreams/MergeSortingBlockInputStream.cpp
@@ -88,6 +88,8 @@ Block MergeSortingBlockInputStream::readImpl()
                 return block;
 
             SortHelper::removeConstantsFromBlock(block);
+            if (block.columns() != header_without_constants.columns())
+                throw Exception("Unexpected number of constant columns in block in MergeSortingBlockInputStream.");
 
             blocks.push_back(block);
             sum_bytes_in_blocks += block.estimateBytesForSpill();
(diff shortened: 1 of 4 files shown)


xzhangxian1008 (approved):
LGTM

Lloyd-Pottiger (approved):

dbms/src/Operators/MergeSortTransformOp.cpp
```diff
@@ -149,6 +149,8 @@ OperatorStatus MergeSortTransformOp::transformImpl(Block & block)
 
         // store the sorted block in `sorted_blocks`.
         SortHelper::removeConstantsFromBlock(block);
+        if (block.columns() != header_without_constants.columns())
+            throw Exception(ErrorCodes::LOGICAL_ERROR, "Unexpected number of constant columns in block in MergeSortTransformOp, n_block={} n_header={}", block.columns(), header_without_constants.columns());
```
gengliqi, line 153: ditto

dbms/src/DataStreams/MergeSortingBlockInputStream.cpp
```diff
@@ -88,6 +88,8 @@ Block MergeSortingBlockInputStream::readImpl()
                 return block;
 
             SortHelper::removeConstantsFromBlock(block);
+            if (block.columns() != header_without_constants.columns())
+                throw Exception(ErrorCodes::LOGICAL_ERROR, "Unexpected number of constant columns in block in MergeSortingBlockInputStream, n_block={} n_header={}", block.columns(), header_without_constants.columns());
```
gengliqi, line 92: Better to use `RUNTIME_CHECK_MSG`

windtalker: Done

Issue #10052, opened 2025-03-28
TiFlash fails to restart when a row is larger than 16MiB


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
# tidb.toml
[performance]
txn-entry-size-limit = 125829120
# tiup edit
#    performance.txn-entry-size-limit: 125829120
```
```
# tikv.toml
[raftstore]
raft-entry-max-size = '120MiB'
# tiup edit
#     raftstore.raft-entry-max-size: '120MiB'
```

```
set global max_allowed_packet = 1073741824;
CREATE TABLE `t` (
  `id` bigint(21) unsigned NOT NULL,
  `text` longtext DEFAULT NULL
);
```

### 3. What did you see instead (Required)

Get the stack trace after fixing https://github.com/pingcap/tiflash/pull/9903

```
[2025/03/28 14:24:25.632 +08:00] [ERROR] [Exception.cpp:96] ["Code: 49, e.displayText() = DB::Exception: Too large string size, size=33624387 max_size=16777215: restoring region_id=2493078499 page_size=33624794, e.what() = DB::Exception, Stack trace:
       0x8ae80ef    DB::Exception::Exception<unsigned int&, unsigned long long>(int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int&, unsigned long long&&) [tiflash+145653999]
                    dbms/src/Common/Exception.h:53
       0x8ae7f46    std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > DB::readBinary2<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, false>(DB::ReadBuffer&) [tiflash+145653574]
                    dbms/src/Storages/KVStore/Utils/SerializationHelper.h:70
       0x8ae5bb3    DB::RegionCFDataBase<DB::RegionDefaultCFDataTrait>::deserialize(DB::ReadBuffer&, DB::RegionCFDataBase<DB::RegionDefaultCFDataTrait>&) [tiflash+145644467]
                    dbms/src/Storages/KVStore/MultiRaft/RegionCFDataBase.cpp:307
[...]

PR #10055, opened 2025-03-28
KVStore: Fix fail to restore tikv value larger than 16MiB

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/10052

Problem Summary:

```
# tidb.toml
[performance]
txn-entry-size-limit = 125829120
# tiup edit
#    performance.txn-entry-size-limit: 125829120
```
```
# tikv.toml
[raftstore]
raft-entry-max-size = '120MiB'
# tiup edit
#     raftstore.raft-entry-max-size: '120MiB'
```

```
set global max_allowed_packet = 1073741824;
CREATE TABLE `t` (
  `id` bigint(21) unsigned NOT NULL,
  `text` longtext DEFAULT NULL
);
```

TiDB could allow a raft-log with 120MiB at max, while TiFlash `Region::deserialize` use `readBinary2<std::string>` only allow 16MiB at max. So if there are some raftlog larger than 16MiB and flushed to `RegionData`, then TiFlash could fail to restart when restoring Region data from disk.
If the raftlog is committed, the rows are decoded into `Block` and write down to `IStorage`. Reading/writing rows larger than `16MiB` in the `IStorage` layer does not cause any error.

```commit-message
KVStore: Fix restoring large tikv value
Add function `readTiKVStringBinary` for `TiKVValue::deserialize` and `TiKVKey::deserialize` to allow restoring tikv key/values at most 120MiB
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix an issue that TiFlash may fail to restart when user write rows larger than 16MiB
```

### DIFF
--- a/dbms/src/Storages/KVStore/Utils/SerializationHelper.h
+++ b/dbms/src/Storages/KVStore/Utils/SerializationHelper.h
@@ -60,6 +60,14 @@ inline size_t writeBinary2(const std::string & s, WriteBuffer & buf)
     return 4 + s.size();
 }
 
+inline std::string readStringWithLength(ReadBuffer & buf, size_t length)
+{
+    std::string s;
+    s.resize(length);
+    buf.readStrict(&s[0], length);
+    return s;
+}
+
 template <>
 inline std::string readBinary2<std::string>(ReadBuffer & buf)
 {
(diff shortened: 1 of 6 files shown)


dbms/src/Storages/KVStore/Utils/SerializationHelper.h
```diff
@@ -72,18 +80,25 @@ inline std::string readBinary2<std::string>(ReadBuffer & buf)
...
 {
-    std::string s;
-    s.resize(length);
-    buf.readStrict(&s[0], length);
-    return s;
+    UInt32 size = 0;
+    readIntBinary(size, buf);
+
+    // According to `txn-entry-size-limit` in TiDB, the max size of a TiKV key/value is 120MB.
+    // https://docs.pingcap.com/tidb/stable/tidb-configuration-file/#txn-entry-size-limit-new-in-v4010-and-v500
```
CalvinNeo, line 93: Maybe move this to where `TIKV_MAX_VALUE_SIZE` is defined?

JaySon-Huang: fixed

CalvinNeo (approved):
lgtm

PR #10059, opened 2025-03-31
cherry pick #9736 to release-8.5

### LINKED ISSUE #10058: Query running in pipeline model may hang forever in some cases
## Bug Report

Original introduced by #9072, #9424 try to fix the hang by "whenever a task is notified from exchange sender, may sure it will write something to the exchange sender's send queue or notify next task in the sender's send queue", but #9424 assumes that if a task is notified by exchange sender's send queue, it will always call `sink->write()` at the end of current run, this assuming is wrong in sometimes, for example, for a left join with left side as the build side, the runnig task may be trapped into wait notify state when probe is done and it need to wait for all task's probe finish. If a task is trapped into wait notify state, it will never call `sink->write()`.
#9736 had another fix for #9072, and it does not rely on that assumaption, but #9736 does not cherry pick to release-8.5 branch, so this bug is fixed in master/9.0 but still exsits in release-8.5

windtalker:
ExchangeSenderOp check writer status only if there is data to flush to MPPTunnel (#9736)

Issue Number: close #10058

Problem Summary:

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


yibin87 (approved):
LGTM

gengliqi (approved):

Issue #10080, opened 2025-04-07
Keyword Search


breezewish:
## Feature Request

Support Keyword Search, which can be used with Vector Search to fulfill RAG scenarios.

- [ ] Basic Keyword Search with Ranking
- [ ] Optimization for Keyword Search filter (without ranking)

PR #10081, opened 2025-04-07
Storage: Add FullTextIndex readers, writers and streams

breezewish:

Issue Number: ref https://github.com/pingcap/tiflash/issues/10080

Problem Summary:

- Add libclara
- Add FullTextIndex readers and writers
- Add FullTextIndex streams

Major differences with VectorIndex:

- in FullTextIndex, when index is not built, it still returns a stream with correct score (calculated by brute force), because upper layers cannot calculate a BM25 score (which depend on information from other rows).
- FullTextIndex always returns an extra column (_INTERNAL_FTS_SCORE) for scoring. In future we may optimize the case when scoring is not needed.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support keyword search indexes
```


Lloyd-Pottiger (approved):

dbms/src/Storages/DeltaMerge/Index/FullTextIndex/Stream/Ctx.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <Storages/DeltaMerge/Index/FullTextIndex/Stream/Ctx.h>
+#include <TiDB/Decode/TypeMapping.h>
+#include <TiDB/Schema/TiDB.h>
+
+
+namespace DB::DM
+{
+
+const ColumnDefine FullTextIndexStreamCtx::VIRTUAL_SCORE_CD = ColumnDefine( //
+    -2050,
```
JaySon-Huang, line 29: We'd better define this magic number in `dbms/src/Storages/MutableSupport.h` along with other special column_id

JaySon-Huang (approved):
LGTM with a small comment that can be refine later

Issue #9060, opened 2024-05-17
Implement a new hash join to speed up join operator


gengliqi:
## Enhancement

We should implement a new hash join to speed up join operator.
The key part of new hash join is
1. row format of build table
2. chain hash table
3. using prefetch technology to hide cache miss latency

PR #10133, opened 2025-04-24
Support (left outer) (anti) semi join in hash join v2

gengliqi:

Issue Number: ref https://github.com/pingcap/tiflash/issues/9060

Problem Summary:

```commit-message
Support (left outer) (anti) semi join in hash join v2
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
```diff
@@ -0,0 +1,886 @@
...
+        if constexpr (kind == Semi || kind == Anti)
+            return genResultBlockForSemi(ctx);
+        else
+            return genResultBlockForLeftOuterSemi(ctx, has_other_eq_cond_from_in);
+    }
+
+    return join->output_block_after_finalize;
+}
+
+static constexpr UInt16 INITIAL_PACE = 4;
```
guo-shaoge, line 248: the original implementation is 2 for `INITIAL_PACE`, why change it to 4?

gengliqi: Because the join_v2 implementation needs more context data than the join_v1 one, especially for the prefetch version. Make it a little larger to reduce the saving context data overhead.

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
```diff
@@ -0,0 +1,886 @@
...
+            ptr = getNextRowPtr(ptr);
+            if (ptr == nullptr)
+                break;
+        }
+        if (prev_offset == current_offset)
+        {
+            setNotMatched<kind, false>(ctx, idx);
+            continue;
+        }
+        probe_list->append(idx);
```
guo-shaoge, line 382: If `left_table_row[idx]` has match all rows from right table (a.k.a. current_offset > prev_offset && ptr == nullptr) in the second loop in L323. This row will still be added to `probe_list`.

And in the first loop of L278, this row will be checked again and set to not matched in L306, because its next ptr is null, which makes `prev_offset == current_offset`. Do i understand correctly?

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
```diff
@@ -0,0 +1,886 @@
...
+            ++idx;
+            break;
+        }
+    }
+    ctx.current_row_idx = idx;
+}
+
+enum class ProbePrefetchStage : UInt8
+{
+    None,
```
guo-shaoge, line 401: `NextLeftRow` is better?

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
```diff
@@ -0,0 +1,886 @@
...
+    RUNTIME_CHECK(wd.result_block.columns() == left_columns + right_columns);
+    for (size_t i = 0; i < left_columns; ++i)
+    {
+        if (left_required_flag_for_other_condition[i])
+            exec_block.insert(wd.result_block.getByPosition(i));
+    }
+    for (size_t i = 0; i < right_columns; ++i)
+        exec_block.insert(wd.result_block.getByPosition(left_columns + i));
+
+    non_equal_conditions.other_cond_expr->execute(exec_block);
```
guo-shaoge, line 682: what happens if there is no other condition?

guo-shaoge:
other LGTM

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
gengliqi: Each time there is a matched row from right table, the `current_offset` will be increased by one.

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
gengliqi: I prefer `None` because it indicates the current stage of the prefetching slot, meaning there is nothing in this slot.

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
gengliqi: If there are no other conditions, the semi join will be executed in `JoinProbeHelper` instead of `SemiJoinProbeHelper`. Additionally, there is a `RUNTIME_CHECK(join->has_other_condition)` in the constructor of `SemiJoinProbeHelper`.

guo-shaoge (approved):

dbms/src/Interpreters/JoinV2/HashJoin.cpp
```diff
@@ -498,7 +498,10 @@ void HashJoin::workAfterBuildRowFinish()
     fiu_do_on(FailPoints::force_join_v2_probe_enable_lm, { late_materialization = true; });
     fiu_do_on(FailPoints::force_join_v2_probe_disable_lm, { late_materialization = false; });
 
-    join_probe_helper = std::make_unique<JoinProbeHelper>(this, late_materialization);
+    if (SemiJoinProbeHelper::isSupported(kind, has_other_condition))
+        semi_join_probe_helper = std::make_unique<SemiJoinProbeHelper>(this);
```
windtalker, line 502: so `semi_join_probe_helper` only handle semi join with other condition, for semi join without other condition, it will be handled in `join_probe_helper`?

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
```diff
@@ -0,0 +1,886 @@
...
+        probeFillColumns<KeyGetter, kind, has_null_map, has_other_eq_cond_from_in, tagged_pointer>(
+            ctx,
+            wd,
+            added_columns);
+    }
+    wd.probe_hash_table_time += watch.elapsedFromLastTime();
+
+    // Move the mutable column pointers back into the wd.result_block, dropping the extra reference (ref_count 2→1).
+    // Alternative: added_columns.clear(); but that is less explicit and may misleadingly imply the columns are discarded.
+    for (size_t i = 0; i < right_columns; ++i)
```
windtalker, line 218: maybe only move the mutable column back if `wd.selective_offsets.empty()` is not true will make it easier to understand?

dbms/src/Interpreters/JoinV2/HashJoin.cpp
gengliqi: Yes.

dbms/src/Interpreters/JoinV2/SemiJoinProbe.cpp
gengliqi: Regardless of whether `wd.selective_offsets` is empty, the reference count of each column pointer is already 2.

Issue #9891, opened 2025-02-19
TiFlash panics with `Data type Nullable(Nothing) cannot be used in tables` while syncing expression index schema `((null))`


solotzg:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

```
CREATE TABLE `t` (
  `id` int,
  KEY `idx_name` ((null))
);
alter table t set tiflash replica 1;
insert t values(0);
```

### 3. What did you see instead (Required)

tiflash panics;

error logs:
```
Code: 370, e.displayText() = DB::Exception: Data type Nullable(Nothing) cannot be used in tables
```

### 4. What is your TiFlash version? (Required)

`>= v6.1.0`

JaySon-Huang:
The logging when reproduce. Seems that the expression index `KEY  idx_name ((null))` will create a column with `"Tp": 6`

```
{
	"cols": [{
		"id": 1,
		"name": {
			"L": "id",
			"O": "id"
		},
		"offset": 0,
		"state": 5,
		"type": {
			"Charset": "binary",
			"Collate": "binary",
			"Decimal": 0,
			"Flag": 0,
			"Flen": 11,
			"Tp": 3
		}
	}, {
		"id": 2,
		"name": {
			"L": "_v$_idx_name_0",
			"O": "_v$_idx_name_0"
		},
		"offset": 1,
		"state": 5,
		"type": {
			"Charset": "binary",
			"Collate": "binary",
			"Decimal": 0,
			"Flag": 136,
			"Flen": 0,
			"Tp": 6
		}
	}],
	"id": 242807,
	"index_info": [],
	"is_common_handle": false,
	"keyspace_id": 4294967295,
	"name": {
		"L": "t",
		"O": "t"
	},
	"pk_is_handle": false,
	"schema_version": -1,
	"state": 5,
	"tiflash_replica": {
		"Available": false,
		"Count": 1
	},
	"update_timestamp": 456163970651521027
}
```

```
[...]

PR #10162, opened 2025-05-14
ddl: Fix compatibility with null expression index

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/9891

Problem Summary:

```
CREATE TABLE `t` (
  `id` int,
  KEY `idx_name` ((null))
);
insert t values(0);
```

If the user create an expression index as above, TiDB will generate a "column" for the expression with `"Tp": 6`. This type of column will generate `ColumnNothing` which is not expected to be created at the IStorage, thus causing exception thrown during the DDL syncing when TiFlash handling the RaftLog from the table.
Because TiFlash-proxy has persisted the RaftLog but get exception when decoding the RaftLog and apply it to the Storage layer, TiFlash panic.

https://github.com/pingcap/tiflash/blob/5c59a514c5fec73b26e6773b57b968f2f59afb7f/dbms/src/TiDB/Schema/TiDBTypes.h#L33
https://github.com/pingcap/tiflash/blob/5c59a514c5fec73b26e6773b57b968f2f59afb7f/dbms/src/Storages/StorageFactory.cpp#L39-L47

```commit-message
ddl: Fix TiFlash panic when meets expression index with format `((NULL))`
Using `ColumnNothing` in IStorage may bring unexpected behavior when we try to write or read the column. So TiFlash create `ColumnInt8` instead of `ColumnNothing` when `TiDB::ColumnInfo.Tp == Null`.

However, we don't expected the column is used when reading or writing to TiFlash.

* `((null))` is not a normal expression index, this could be created by the user in accident (e.g. executing wrong SQL statement to create the expression index), TiDB should not write or read from it
* Even for normal expression index, TiDB does not use expression when accessing to TiFlash
```

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that TiFlash panic when meets expression index with format `((NULL))`
```


CalvinNeo (approved):
lgtm

PR #10238, opened 2025-06-10
Fix the issue that window function returns incorrect result when range type frame end is preceding

### LINKED ISSUE #10236: Window function returns incorrect result when range type frame end is preceding
## Bug Report

1. `create table t (p int, o int, v int);`
2. Insert many same rows into table and ensure that we need at least two blocks to contain them when executing
3. `select p, o, v, count(v) over (partition by p order by o range between 0 preceding and 0 preceding) as c from t;`

all rows have same result

some rows get different result

nightly(1d5ca44e2776ed32)

xzhangxian1008:

Issue Number: close #10236

Problem Summary:

```commit-message
Remove a wrong judgement
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that window function returns incorrect result when range type frame end is preceding
```

### DIFF
--- a/dbms/src/DataStreams/WindowTransformAction.cpp
+++ b/dbms/src/DataStreams/WindowTransformAction.cpp
@@ -660,7 +660,7 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToStartForRangeFrame()
 
 std::tuple<RowNumber, bool> WindowTransformAction::stepToEndForRangeFrame()
 {
-    if ((!end_preceding || (end_preceding && frame.end_offset == 0)) && !partition_ended)
+    if (!partition_ended)
         // If we find the frame end and the partition_ended is false.
         // Some previous blocks may be dropped, this is an unexpected behaviour.
         // So, we shouldn't do anything before the partition_ended is true.
--- a/dbms/src/DataStreams/WindowTransformAction.cpp
+++ b/dbms/src/DataStreams/WindowTransformAction.cpp
@@ -646,7 +646,7 @@ void WindowTransformAction::stepToFrameEnd()
 
 std::tuple<RowNumber, bool> WindowTransformAction::stepToStartForRangeFrame()
 {
-    if (!window_description.frame.begin_preceding && !partition_ended)
+    if (!partition_ended)
         // If we find the frame end and the partition_ended is false.
         // The prev_frame_start may be equal to partition_end which
         // will cause the assert fail in advancePartitionEnd function.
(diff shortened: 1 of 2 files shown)


guo-shaoge (approved):

dbms/src/WindowFunctions/tests/gtest_agg.cpp
```diff
@@ -738,4 +738,32 @@ try
...
+        static_cast<Int64>(0));
+    frame.end = buildRangeFrameBound(
+        tipb::WindowBoundType::Preceding,
+        tipb::RangeCmpDataType::Int,
+        ORDER_COL_NAME,
+        false,
+        static_cast<Int64>(0));
+
+
+    executeFunctionAndAssert(
```
guo-shaoge: maybe change `profiles.default.max_block_size` to a smaller value? like 5?

dbms/src/DataStreams/WindowTransformAction.cpp
```diff
@@ -660,7 +660,7 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToStartForRangeFrame()
 
 std::tuple<RowNumber, bool> WindowTransformAction::stepToEndForRangeFrame()
 {
-    if (!window_description.frame.end_preceding && !partition_ended)
+    if (!partition_ended)
         // If we find the frame end and the partition_ended is false.
```
guo-shaoge: NOTE: for `expr` in `expr preceding`, `expr` can only be positive number, so no need to change `stepToStartForRangeFrame`

dbms/src/DataStreams/WindowTransformAction.cpp
```diff
@@ -660,7 +660,7 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToStartForRangeFrame()
 
 std::tuple<RowNumber, bool> WindowTransformAction::stepToEndForRangeFrame()
 {
-    if (!window_description.frame.end_preceding && !partition_ended)
+    if (!partition_ended)
```
guo-shaoge: Better add comments:
`
Still return false even when end_preceding is true, because there may still same rows as current_row in the next block.
So we cannot stop find the frame end until got the partition end(or got the first row that is greater than current_row).
`

xzhangxian1008: > Better add comments: `Still return false even when end_preceding is true, because there may still same rows as current_row in the next block. So we cannot stop find the frame end until got the partition end(or got the first row that is greater than current_row).`

added

dbms/src/DataStreams/WindowTransformAction.cpp
```diff
@@ -660,10 +660,15 @@ std::tuple<RowNumber, bool> WindowTransformAction::stepToStartForRangeFrame()
...
```
gengliqi: How about
```suggestion
    bool end_preceding = window_description.frame.end_preceding;
    if ((!end_preceding || (end_preceding && frame.end_offset == 0)) && !partition_ended)
```

contrib/tipb
gengliqi, line 1: Why update tipb?

Issue #10266, opened 2025-06-27
Graceful shutdown in TiFlash


gengliqi:
## Enhancement

When TiFlash is terminated, it will not wait for all MPP tasks to complete, so the SQL failure rate may be slightly higher.

gengliqi:
https://github.com/pingcap/tiflash/pull/10267 does not support a real graceful shutdown because the read index request will be canceled if setting `stopping` to store status, which is not expected and can cause MPP tasks to fail.

JaySon-Huang:
Note that when deployed with k8s/(tidb-operator), should set the `terminationGracePeriodSeconds`. See more details in  https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-flow

PR #10267, opened 2025-06-27
Support graceful shutdown in TiFlash

gengliqi:

Issue Number: close https://github.com/pingcap/tiflash/issues/10266

Problem Summary:

```commit-message
Support graceful shutdown in TiFlash
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/DAGDriver.cpp
+++ b/dbms/src/Flash/Coprocessor/DAGDriver.cpp
@@ -113,8 +113,6 @@ template <DAGRequestKind Kind>
 void DAGDriver<Kind>::execute()
 try
 {
-    if (context.getTMTContext().checkShuttingDown())
-        throw TiFlashException("TiFlash server is terminating", Errors::Coprocessor::Internal);
     auto start_time = Clock::now();
     DAGContext & dag_context = *context.getDAGContext();
     const auto & resource_group = dag_context.getResourceGroupName();


dbms/src/Storages/KVStore/ProxyStateMachine.h
```diff
@@ -367,33 +367,28 @@ struct ProxyStateMachine
...
     }
 
-    // Set KVStore to running, so that it could handle read index requests.
-    void runKVStore(TMTContext & tmt_context) const { tmt_context.setStatusRunning(); }
-
-    /// Stop all services in TMTContext and ReadIndexWorkers.
-    /// Then, inform proxy to stop by setting `tiflash_instance_wrap.status`.
-    void stopProxy(TMTContext & tmt_context)
+    /// Stopping services in TMTContext and ReadIndexWorkers.
+    void stoppingProxy(TMTContext & tmt_context)
```
CalvinNeo: How about naming it into something like `waitReadIndexTasksFinish`, so others can know we can't send any read request(namely read-index) to proxy after this call?

dbms/src/Server/Server.cpp
```diff
@@ -1167,8 +1167,6 @@ try
             }
         }
 
-        SCOPE_EXIT({ proxy_machine.stopProxy(tmt_context); });
```
CalvinNeo, line 1170: There is a FlashGrpcServerHolder which depends on a DiagnosticsService which will call proxy. So I think we keep putting it where it was originally.

gengliqi: I have changed to destroy `FlashGrpcServerHolder` first, and then calling `stopProxy`.

dbms/src/Storages/KVStore/ProxyStateMachine.h
gengliqi: Done.

windtalker (approved):
lgtm

dbms/src/Flash/Coprocessor/DAGDriver.cpp
```diff
@@ -112,6 +113,8 @@ template <DAGRequestKind Kind>
 void DAGDriver<Kind>::execute()
 try
 {
+    if (context.getTMTContext().checkShuttingDown())
+        throw TiFlashException("TiFlash server is terminating", Errors::Coprocessor::Internal);
```
JaySon-Huang: I wonder if tiflash return this error, can tidb internally retry the query by re-dispatching the query to another tiflash instance?

dbms/src/Server/FlashGrpcServerHolder.cpp
```diff
@@ -224,17 +224,19 @@ FlashGrpcServerHolder::~FlashGrpcServerHolder()
         *is_shutdown = true;
         // Wait all existed MPPTunnels done to prevent crash.
         // If all existed MPPTunnels are done, almost in all cases it means all existed MPPTasks and ExchangeReceivers are also done.
-        const int max_wait_cnt = 300;
+        constexpr int wait_step = 200;
+        constexpr int max_wait_cnt = 300 * 1000 / wait_step;
```
JaySon-Huang: Should we make `max_wait` configurable?

dbms/src/Server/Server.cpp
```diff
@@ -1250,9 +1247,51 @@ try
...
+        auto graceful_wait_before_shutdown
+            = global_context->getUsersConfig()->getUInt64(GRACEFUL_WIAT_BEFORE_SHUTDOWN, 60);
+        auto max_wait_time = start + std::chrono::seconds(graceful_wait_before_shutdown);
+        // Sleep before checking to reduce the chance of missing MPP tasks that are still in the process of being dispatched
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
+        while (true)
+        {
+            {
+                std::unique_lock lock(monitor->mu);
+                if (monitor->monitored_tasks.empty())
```
JaySon-Huang: Can we wrap acuquiring lock and checking monitor->monitored_tasks.empty() as a class function of monitor?

dbms/src/Server/Server.cpp
```diff
@@ -1250,9 +1247,51 @@ try
...
+        auto start = std::chrono::steady_clock::now();
+        // The maximum seconds TiFlash will wait for all current MPP tasks to finish before shutting down
+        static constexpr const char * GRACEFUL_WIAT_BEFORE_SHUTDOWN = "flash.graceful_wait_before_shutdown";
+        auto graceful_wait_before_shutdown
+            = global_context->getUsersConfig()->getUInt64(GRACEFUL_WIAT_BEFORE_SHUTDOWN, 60);
+        auto max_wait_time = start + std::chrono::seconds(graceful_wait_before_shutdown);
+        // Sleep before checking to reduce the chance of missing MPP tasks that are still in the process of being dispatched
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
+        while (true)
+        {
```
JaySon-Huang: maybe we can use 
```
do {
  std::this_thread::sleep_for(std::chrono::milliseconds(200));
  // check whether mpp task are all finished
} while (...)
```
instead of writing `std::this_thread::sleep_for(std::chrono::milliseconds(200));` twice

dbms/src/Server/Server.cpp
```diff
@@ -1250,9 +1247,51 @@ try
...
+
+            std::this_thread::sleep_for(std::chrono::milliseconds(200));
+        }
+
+        proxy_machine.waitAllReadIndexTasksFinish(tmt_context);
+
+        LOG_INFO(log, "Set store context status Terminated");
+        tmt_context.setStatusTerminated();
+
+        // Stop grpc server
```
JaySon-Huang:
```suggestion
        // Stop grpc server before proxy_machine because it depends on a DiagnosticsService which will call proxy
```

dbms/src/Server/Server.cpp
```diff
@@ -1250,9 +1247,51 @@ try
...
+        flash_grpc_server_holder.reset();
+        proxy_machine.stopProxy(tmt_context);
```
CalvinNeo, line 1267: The `LocalAdmissionController` and `TaskScheduler` is also reset before proxy is stopped in the original code. Do we need to move them? I'm not familiar with those two structs.

dbms/src/Server/FlashGrpcServerHolder.cpp
gengliqi: It's not necessary to be configurable because the wait time for MPPTunnels to finish should be minimal after all MPPTasks are done. Also, I changed it to 1 minute.

dbms/src/Server/Server.cpp
gengliqi: Good idea. Done.

dbms/src/Server/Server.cpp
gengliqi: Done.

dbms/src/Flash/Coprocessor/DAGDriver.cpp
gengliqi: Good catch. I copied these two lines of code from here so that the error can return quickly during shutdown. https://github.com/pingcap/tiflash/blob/43d58319468eeb564fa66efcd45138cf204168f9/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp#L894-L895
After checking the code from client-go and tidb, it seems tidb can not re-dispatch the query to another tiflash when tidb gets this error. I'll take a closer look later.

gengliqi: I checked the code and found that `BatchCop` supports retries when getting this error, but `Cop` does not. So I removed these two lines of code.

dbms/src/Server/Server.cpp
gengliqi: I checked the code. These two structures have nothing to do with the proxy.

dbms/src/Flash/Coprocessor/DAGDriver.cpp
JaySon-Huang: > BatchCop supports retries when getting this error, but Cop does not

What about `MPP`?

gengliqi: MPP will not report this error and cannot retry if the data is produced. That's why this PR will wait for all MPP tasks to finish before shutting down.

CalvinNeo (commented):
LGTM. @JaySon-Huang Would you please give a final check?

JaySon-Huang (approved):
Basically LGTM

Issue #10300, opened 2025-07-11
There could be large amount of remote request when querying on a large and hot-write table


JaySon-Huang:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Observed from a certain cluster

### 3. What did you see instead (Required)

```
-- first not match, rebuild the local read inputstreams
[2025/07/11 01:13:07.471 +00:00] [WARN] [DAGStorageInterpreter.cpp:748] ["RegionException after read from storage, regions [23874359,], message: EPOCH_NOT_MATCH, retry to read from local"] [source=MPP<query:459327769121916622,task:3>]
-- another not match happen, 21,700 regions need to be retry
[2025/07/11 01:13:07.649 +00:00] [WARN] [DAGStorageInterpreter.cpp:765] ["RegionException after read from storage, regions [24184535, .....,24109478,], message: EPOCH_NOT_MATCH"] [source=MPP<query:459327769121916622,task:3>] [thread_id=690]
[2025/07/11 01:13:07.699 +00:00] [INFO] [RemoteRequest.cpp:42] ["Start to build remote request for 21775 regions (23874359,24184535,...,24109478) for table 8132"] [source=MPP<query:459327769121916622,task:3>] [thread_id=690]
```

### 4. What is your TiFlash version? (Required)
v6.5.10

PR #10301, opened 2025-07-11
Raft: Increase max retry times to avoid too large remote requests

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/10300

Problem Summary:

In `DAGStorageInterpreter::buildLocalStreamsForPhysicalTable` we need to confirm the regions' keyranges not changed and regions are not removed before the storage snapshot is built. Otherwise the query result is incorrect.
https://github.com/pingcap/tiflash/blob/bec1390d8be06e97612404a61867f0b38b3627b2/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp#L1069-L1072

Building storage snapshot usually takes about a hundreds ms to few seconds. But the max retry time is only 1. If retry reach the max time, then all the Regions will be fetch though "RemoteRead". When querying a hot write large table, it could fail more than 1. And because the table is large, query contains large number of Regions, retrying all Regions in "RemoteRead" way could takes lots of network bandwidth and extra CPU, leading to TiFlash unstable.

```commit-message
Raft: Increase max retry times to avoid too large remote requests
  * Increase the max retry number between LearnerRead and acquiring snapshot from the storage layer by the number of query regions
```

Other changes:
* Avoid printing `ann_query_info`
* Enrich the error message when `ColumnFilePersistedSet::installCompactionResults` meet unexpected exception.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Increase the maximum retry count when acquiring storage snapshots to improve query stability for large tables
```

### DIFF
--- a/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
+++ b/dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
@@ -963,15 +963,12 @@ bool DAGStorageInterpreter::checkRetriableForBatchCopOrMPP(
     const TableID & table_id,
     const SelectQueryInfo & query_info,
     const RegionException & e,
-    int num_allow_retry)
+    const Int32 num_allow_retry)
 {
     const DAGContext & dag_context = *context.getDAGContext();
     assert((dag_context.isBatchCop() || dag_context.isMPPTask()));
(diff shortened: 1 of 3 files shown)


CalvinNeo (approved):
lgtm

PR #10299, opened 2025-07-10
Support graceful shutdown in TiFlash again

gengliqi:
### What problem does this PR solve?

Issue Number: close https://github.com/pingcap/tiflash/issues/10266

Problem Summary:

### What is changed and how it works?
Should merge https://github.com/pingcap/tiflash/pull/10298 first.

```commit-message
Support graceful shutdown in TiFlash again
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
Support graceful shutdown in TiFlash
```

PR #10313, opened 2025-07-18
Remove unnecessary wait for MPP tunnel destruction before shutdown

gengliqi:

Issue Number: ref https://github.com/pingcap/tiflash/issues/10266

Problem Summary:

After https://github.com/pingcap/tiflash/pull/10299, TiFlash will wait for all MPPTasks to finish before shutdown.

Whether all MPP tasks have finished is determined by checking if `monitored_tasks` is empty. In `MPPTask`, `MPPTaskMonitorHelper` is the last member to be destructed, and only when it is destructed will `removeMonitoredTask` be called to remove the task id from `monitored_tasks`. `MPPTunnel`s are stored in `MPPTunnelSet`, and although it is held by a shared_ptr, there are actually only two references: one directly in `MPPTask`, and the other inside the `std::unique_ptr<DAGContext>` also in MPPTask. Therefore, destroying the `MPPTask` will also destroy the `MPPTunnel`.

```commit-message
Remove unnecessary wait for MPP tunnel destruction during TiFlash shutdown
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Mpp/MPPTaskManager.h
+++ b/dbms/src/Flash/Mpp/MPPTaskManager.h
@@ -194,8 +194,6 @@ public:
         return monitored_tasks.find(task_unique_id) != monitored_tasks.end();
     }
 
-    void waitAllMPPTasksFinish(const std::unique_ptr<Context> & global_context);
-
     std::mutex mu;
     std::condition_variable cv;
     bool is_shutdown = false;
--- a/dbms/src/Flash/Mpp/MPPTaskManager.h
+++ b/dbms/src/Flash/Mpp/MPPTaskManager.h
@@ -194,6 +194,8 @@ public:
         return monitored_tasks.find(task_unique_id) != monitored_tasks.end();
     }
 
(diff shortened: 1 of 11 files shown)


dbms/src/Flash/Mpp/MPPTaskManager.h
```diff
@@ -194,7 +194,7 @@ struct MPPTaskMonitor
         return monitored_tasks.find(task_unique_id) != monitored_tasks.end();
     }
 
-    void waitAllMPPTasksFinish(const std::unique_ptr<Context> & global_context);
+    UInt64 waitAllMPPTasksFinish(const std::unique_ptr<Context> & global_context);
```
JaySon-Huang: It is better to add some comments about the meaning of return value.

gengliqi: Added

JaySon-Huang (approved):
LGTM

PR #10335, opened 2025-08-06
catch exception in EventTask:: doFinalizeImpl

### LINKED ISSUE #10334: catch exception in Operator::operatorSuffixImpl
## Enhancement
Operator::operatorSuffixImpl() may throw exception, which cause `EventTask::onTaskFinish()` will not be called in `EventTask::finalizeImpl()`.
And `PipelineExecutorContext::consume` will be stuck because not all task finishs.

guo-shaoge:

Issue Number: close #10334

Problem Summary: catch exception in EventTask:: doFinalizeImpl, otherwise PipelineExecutor maybe stuck because `PipelineExecutorContext::decActiveRefCount` will not be called.

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/Pipeline/Schedule/Tasks/Impls/EventTask.cpp
+++ b/dbms/src/Flash/Pipeline/Schedule/Tasks/Impls/EventTask.cpp
@@ -37,7 +37,14 @@ EventTask::EventTask(
 
 void EventTask::finalizeImpl()
 {
-    doFinalizeImpl();
+    try
+    {
+        doFinalizeImpl();
+    }
+    catch (...)
+    {
+        LOG_ERROR(log, getCurrentExceptionMessage(false, false));
+    }
     event->onTaskFinish(profile_info);
     event.reset();
 }
--- a/dbms/src/Flash/Executor/tests/gtest_pipeline_executor_context.cpp
+++ b/dbms/src/Flash/Executor/tests/gtest_pipeline_executor_context.cpp
@@ -16,14 +16,39 @@
 #include <Common/ThreadManager.h>
 #include <Flash/Executor/PipelineExecutorContext.h>
 #include <Flash/Executor/ResultQueue.h>
+#include <TestUtils/ExecutorTestUtils.h>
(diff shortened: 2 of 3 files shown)


dbms/src/Flash/Pipeline/Schedule/Tasks/Impls/EventTask.cpp
```diff
@@ -37,7 +37,14 @@ EventTask::EventTask(
 
 void EventTask::finalizeImpl()
 {
-    doFinalizeImpl();
+    try
+    {
+        doFinalizeImpl();
```
SeaRise: https://github.com/pingcap/tiflash/blob/5b5355cdb97da1471692208a3f12d990fc98e3c8/dbms/src/Flash/Pipeline/Schedule/Tasks/Task.cpp#L169-L176
企业微信截图_85e6196e-3b03-499f-be0b-f29626bfb0c5
But finalizeImpl has been try catch.

SeaRise: And if the suffix fails, the Task should also be marked as failed, not ignored.

guo-shaoge: 1. Even `finalizeImpl()` has been try catch, but I think should also catch inside `EventTask::finalizeImpl(), to make sure `onTaskFinish()` is always called.
2. Some Operators do prepare work in `Operator::operatorPrefixImpl()`, like `ExchangeSenderSinkOp`, it will call `DAGResponseWriter::prepare()`. I think we should also consider catch exception for `PipelineExec::executePrefix()`. Otherwise we may still got hang issue

windtalker (approved):
lgtm

gengliqi (approved):

PR #10364, opened 2025-08-25
Fix decimal floor/ceil (#10365)

### LINKED ISSUE #3496: Add unit test for Math function `floor`

### LINKED ISSUE #10365: Incorrect result of floor function
## Bug Report

```sql
create table t(id int, a decimal(15,0), b decimal(52,30));
insert into t values(1,78,144.00000000000000000000000000000000000000000000000000000000);
set tidb_isolation_read_engines='tidb,tiflash';
select floor(a*b), a*b from t;
```

```sql
mysql> set tidb_isolation_read_engines='tidb,tikv';
Query OK, 0 rows affected (0.02 sec)
mysql> select floor(a*b), a*b from t;
+------------+--------------------------------------+
| floor(a*b) | a*b                                  |
+------------+--------------------------------------+
|      11232 | 11232.000000000000000000000000000000 |
+------------+--------------------------------------+
1 row in set (0.02 sec)
```

```sql
mysql> set tidb_isolation_read_engines='tidb,tiflash';
Query OK, 0 rows affected (0.01 sec)
mysql> select floor(a*b), a*b from t;
+------------+--------------------------------------+
| floor(a*b) | a*b                                  |
+------------+--------------------------------------+
|      11231 | 11232.000000000000000000000000000000 |
+------------+--------------------------------------+
1 row in set (0.03 sec)
```

8.5.3

ChangRui-Ryan:

Issue Number: close #10365, close https://github.com/pingcap/tiflash/issues/3496, ref https://github.com/pingcap/tidb/issues/63086

Problem Summary:
 
```commit-message

```

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


JaySon-Huang:
https://do.pingcap.net/jenkins/blue/organizations/jenkins/pingcap%2Ftiflash%2Fpull_integration_test/detail/pull_integration_test/4670/pipeline
@ChangRui-Ryan Format the files using the script here pls
https://github.com/pingcap/tiflash?tab=readme-ov-file#contributing

dbms/src/Functions/FunctionsRound.h
```diff
@@ -327,12 +256,90 @@ struct IntegerRoundingComputation
...
+        using NativeType = T::NativeType;
+        // Currently, we only use DecimalRoundingComputation for floor/ceil.
+        // As for round/truncate, we always use tidbRoundWithFrac/tidbTruncateWithFrac.
+        // So, we only handle ScaleMode::Zero here.
+        if constexpr (scale_mode == ScaleMode::Zero)
+        {
+            using Op = IntegerRoundingComputation<NativeType, rounding_mode, ScaleMode::Negative>;
+            auto scale_factor = intExp10OfSize<NativeType>(decimal_scale);
+
+            if constexpr (std::is_same_v<T, OutputType>)
```
windtalker: In which case will the output be Int64, and in which case will the output type the same as T?

dbms/src/Functions/FunctionsRound.h
```diff
@@ -327,12 +256,90 @@ struct IntegerRoundingComputation
...
+            {
+                Op::compute(&in->value, scale_factor, &out->value);
```
windtalker: If the output type is decimal, will the return type of `ceil/floor` keep its input type's scale, or the result type's scale is always 0?

dbms/src/Functions/FunctionsRound.h
ChangRui-Ryan: For decimal(M, N), if M-N > 18, use T; otherwise use Int64. Because Int64 is sufficient to represent 18-digit  integers. TiDB will make decision.

dbms/src/Functions/FunctionsRound.h
ChangRui-Ryan: will keep its input type's scale.

windtalker: Are you sure of this? Because from TiDB's code, it seems if the return type is decimal, the scale is always 0:
https://github.com/pingcap/tidb/blob/ad52c8e05b49602888f08db4a81cfcffbb21e555/pkg/expression/builtin_math.go#L732-L735

ChangRui-Ryan: From getReturnTypeImpl of FunctionRounding class, we can see that return type is just arguments[0], and I didn't change that in this pr. Maybe some other places change the scale to 0 after. Let me try to find out.

windtalker: that is because after TiFlash's function, we will add extra cast to keep the return type the same as TiDB. But for this particular function, I think we can adopt TiDB's return type infer logic in TiFlash so we don't need to add extra cast after the function.

ChangRui-Ryan: OK, I will optimize that.

dbms/src/Functions/FunctionsRound.h
```diff
@@ -327,12 +256,75 @@ struct IntegerRoundingComputation
...
+        ScaleType decimal_scale)
+    {
+        static_assert(std::is_same_v<T, OutputType> || std::is_same_v<OutputType, Int64>);
+        using NativeType = T::NativeType;
+        // Currently, we only use DecimalRoundingComputation for floor/ceil.
+        // As for round/truncate, we always use tidbRoundWithFrac/tidbTruncateWithFrac.
+        // So, we only handle ScaleMode::Zero here.
+        if constexpr (scale_mode == ScaleMode::Zero)
+        {
+            auto scale_factor = intExp10OfSize<NativeType>(decimal_scale);
```
windtalker: why not use `scale_factor` as the last parameter of `compute`, so we don't need calcuate `scale_factor` for every input row

ChangRui-Ryan: Good suggestion, I will refactor this part.   
Originally, I didn’t want to make major change, like modifying function interface.

windtalker (approved):
lgtm

wshwsh12 (approved):

PR #10400, opened 2025-09-03
Improve like perf for utf8 ci collations

### LINKED ISSUE #10399: Improve like perf for utf8 CI collation
## Enhancement
# Performance Comparison: utf8*_ci vs. Bin Collations for LIKE Operations (TiFlash + TPCH Dataset)

## 1. Test Background & Objective
Verify the performance difference between **utf8*_ci collations (case-insensitive, UTF-8-aware)** and **bin collations (binary, byte-level)** when executing `LIKE` pattern matching queries in TiFlash (TiDB's columnar storage engine). The core focus is on comparing query latency under the two collation types, with specific attention to the overhead of UTF-8 processing in utf8*_ci collations.

## 2. Test Environment & Configuration
| Configuration Item          | Details                                                                 |
|------------------------------|-------------------------------------------------------------------------|
| Dataset                      | TPCH 10G Dataset (only the `orders` table is used, which contains the text field `o_comment` stored in UTF-8) |
| Execution Engine             | TiFlash (TiDB Columnar Storage Engine)                                 |
| Concurrency Control          | Restrict TiFlash to 1 concurrent thread via `set tidb_max_tiflash_threads=1;` to eliminate multi-threading interference |
| Test Field                   | `orders.o_comment` (UTF-8 text field used for `LIKE` pattern matching) |
| Collation Types              | Two collations are compared:<br>- utf8*_ci Collation: UTF-8-aware case-insensitive collations (e.g., `utf8mb4_general_ci`, `utf8mb4_unicode_ci`)<br>- Bin Collation: Binary collations (e.g., `utf8mb4_bin`) |

## 3. Test Steps
1. **Environment Initialization**: Load the TPCH 10G Dataset and ensure the `orders` table has complete, UTF-8-encoded data.
2. **Concurrency Configuration**: Execute the following command in the TiDB client to fix TiFlash's concurrency and isolate performance variables:
   ```sql
   set tidb_max_tiflash_threads=1;
   ```
3. **Query Execution (utf8*_ci Collation Scenario)**: Set the `o_comment` field to a utf8*_ci collation (e.g., `utf8mb4_general_ci`), then run the core `LIKE` query from TPCH Query 13 (Q13):
   ```sql
   SELECT COUNT(*) FROM orders WHERE o_comment LIKE "%special%requests%";
   ```
4. **Query Execution (Bin Collation Scenario)**: Change the `o_comment` field to a bin collation (e.g., `utf8mb4_bin`), then re-run the same query from Step 3 to ensure consistency.
5. **Latency Measurement**: Record the total execution time (latency) of the query in both scenarios (from request initiation to result return), averaging 3 runs to minimize variance.

## 4. Test Results
| Collation Type       | Collation Example       | Average Query Latency |
|----------------------|-------------------------|------------------------|
| utf8*_ci Collation   | `utf8mb4_general_ci`    | ~5 seconds             |
| Bin Collation        | `utf8mb4_bin`           | ~0.7 seconds           |

## 5. Conclusion
Under TiFlash's single-threaded concurrency, for the `LIKE "%special%requests%"` pattern matching query on the UTF-8-encoded `orders.o_comment` field:
- The latency of bin collations (~0.7 seconds) is only **14%** of that of utf8*_ci collations (~5 seconds), demonstrating a dramatic performance advantage.  
- Primary Root Cause: utf8*_ci collations incur far higher overhead because they require **UTF-8 decoding and character normalization** as a prerequisite for comparison. Raw UTF-8 byte sequences must first be decoded into logical Unicode characters; additional processing is then needed to handle character equivalence (e.g., case folding like `A` vs `a`, or accented character normalization like `é` vs `e´`). This two-step Unicode-aware processing adds substantial computational cost, especially for large text fields in bulk queries.  
- By contrast, bin collations bypass all Unicode-related overhead: they compare strings **directly as raw UTF-8 byte sequences**, with no decoding, normalization, or case-insensitive checks. This byte-level comparison logic is minimal and efficient, resulting in drastically lower latency for `LIKE` operations.

ChangRui-Ryan:

Issue Number: close #10399 

Problem Summary:
When performing a LIKE operation on strings in SQL, the underlying code logic varies depending on the collation used, leading to significant performance differences.
Currently, the LIKE operation performs well with BIN-type collations, but its performance with CI-type collations is poor—with a substantial gap compared to BIN-type collations.
##### 1. Prioritize ASCII_CI Comparison for LIKE Patterns  
Check if the LIKE pattern consists of only ASCII characters first. If so, prioritize case-insensitive (CI) comparison using the small (128-entry) ASCII character table (avoiding expensive UTF-8 decoding initially). Fall back to UTF-8 decoding only when non-ASCII characters are detected during matching.  

##### 2. Optimize Backtracking Logic for Patterns Starting with "%"  
When a pattern starts with `%`, restart backtracking directly from the characters *after* `%` (instead of restarting from the `%` itself). For scenarios where the first character after backtracking does not match, simplify the flow by immediately advancing the string pointer to skip redundant loops, significantly reducing total operations.
#### Benchmark 
To eliminate interference and facilitate comparison, a single-node TiFlash was used, and single concurrency was configured with the following setting:
`set tidb_max_tiflash_threads=1;`
For the query (under TPCH 10G data):
`SELECT COUNT(*) FROM orders WHERE o_comment LIKE ?;`
The comparison of execution time before and after optimization is as follows (unit: seconds):
| PATTERN | Collation | Before Optimization | After Optimization | Improvement Multiple |
| --- | --- | --- | --- | --- |
| `"%ZZZ%";` | utf8mb4_general_ci | 4.1 | 0.8 | 5.1 |
| `"%ZZZ%";` | utf8mb4_0900_ai_ci | 5.3 | 0.8 | 6.6 |
| `"%中国%";` | utf8mb4_general_ci | 4.0 | 1.1 | 3.6 |
| `"%中国%";` | utf8mb4_0900_ai_ci | 5.3 | 2.5 | 2.1 |
| `"%special%requests%" (From TPCH Q13)` | utf8mb4_general_ci | 5.0 | 1.7 | 2.9 |
| `"%special%requests%" (From TPCH Q13)` | utf8mb4_0900_ai_ci | 6.0 | 1.7 | 3.5 |

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/TiDB/Collation/Collator.cpp
+++ b/dbms/src/TiDB/Collation/Collator.cpp
@@ -26,6 +26,16 @@ extern const int LOGICAL_ERROR;
 
 namespace TiDB
 {
+
+const std::array<char, 128> weight_ascii_ci
+    = {0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F, 0x10, 0x11, 0x12,
+       0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x1A, 0x1B, 0x1C, 0x1D, 0x1E, 0x1F, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25,
+       0x26, 0x27, 0x28, 0x29, 0x2A, 0x2B, 0x2C, 0x2D, 0x2E, 0x2F, 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38,
+       0x39, 0x3A, 0x3B, 0x3C, 0x3D, 0x3E, 0x3F, 0x40, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68, 0x69, 0x6A, 0x6B,
+       0x6C, 0x6D, 0x6E, 0x6F, 0x70, 0x71, 0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79, 0x7A, 0x5B, 0x5C, 0x5D, 0x5E,
+       0x5F, 0x60, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x68, 0x69, 0x6A, 0x6B, 0x6C, 0x6D, 0x6E, 0x6F, 0x70, 0x71,
+       0x72, 0x73, 0x74, 0x75, 0x76, 0x77, 0x78, 0x79, 0x7A, 0x7B, 0x7C, 0x7D, 0x7E, 0x7F};
+
 TiDBCollators dummy_collators;
 std::vector<std::string> dummy_sort_key_contaners;
 std::string dummy_sort_key_contaner;
@@ -72,12 +82,13 @@ inline Rune decodeUtf8Char(const char * s, size_t & offset)
 template <typename Collator>
 void Pattern<Collator>::compile(const std::string & pattern, char escape)
 {
-    chars.clear();
+    pattern_weights.clear();
     match_types.clear();
 
-    chars.reserve(pattern.length() * sizeof(typename Collator::CharType));
+    pattern_weights.reserve(pattern.length() * sizeof(typename Collator::CharType));
     match_types.reserve(pattern.length() * sizeof(typename Pattern::MatchType));
 
+    auto last_tp = MatchType::Match;
     size_t offset = 0;
     while (offset < pattern.length())
     {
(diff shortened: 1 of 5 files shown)


dbms/src/TiDB/Collation/Collator.cpp
```diff
@@ -102,63 +113,270 @@ void Pattern<Collator>::compile(const std::string & pattern, char escape)
...
         }
-        chars.push_back(c);
+        pattern_weights.push_back(c);
         match_types.push_back(tp);
+        last_tp = tp;
+    }
+}
+
+template <typename Collator>
+void Pattern<Collator>::try_compile_ascii_ci(const std::string & pattern, char escape)
```
zanmato1984: We use camelCase for function names.

dbms/src/TiDB/Collation/Collator.cpp
```diff
@@ -102,63 +113,270 @@ void Pattern<Collator>::compile(const std::string & pattern, char escape)
...
+        auto ret = try_match_ascii_ci(s, length);
+        if (ret == 1)
+        {
+            return true;
+        }
+        else if (ret == 0)
+        {
+            return false;
+        }
+        // if ret == -1, means the string contains non-ASCII characters, continue to check
```
zanmato1984: Maybe we can simplify to something like:
```
if (auto ret = try_match_ascii_ci(s, length); ret >= 0) {
  return ret;
}
```

dbms/src/TiDB/Collation/Collator.h
```diff
@@ -77,6 +77,7 @@ class ITiDBCollator
         virtual ~IPattern() = default;
 
         virtual void compile(const std::string & pattern, char escape) = 0;
+        virtual void try_compile_ascii_ci(const std::string & pattern, char escape) = 0;
```
zanmato1984: camelCase

dbms/src/TiDB/Collation/Collator.cpp
```diff
@@ -102,63 +113,270 @@ void Pattern<Collator>::compile(const std::string & pattern, char escape)
...
+                }
+                else
+                {
+                    return 0;
+                }
+            }
+            else
+            {
+                if (s[str_idx] < 0)
+                {
```
windtalker, line 305: add a comment here?

dbms/src/TiDB/Collation/Collator.cpp
```diff
@@ -102,63 +113,270 @@ void Pattern<Collator>::compile(const std::string & pattern, char escape)
...
+                    return -1;
+                }
+                if ((match_types[p_idx] == Match
+                     && weight_ascii_ci[s[str_idx]] == weight_ascii_ci[ascii_ci_pattern[p_idx]])
```
windtalker: why not just save the converted pattern in `ascii_ci_pattern` during `try_compile_ascii`, then during `match`, there is no need to lookup the `weight_ascii_ci` for pattern chars?

dbms/src/TiDB/Collation/Collator.cpp
ChangRui-Ryan: OK

dbms/src/TiDB/Collation/Collator.cpp
ChangRui-Ryan: Good point

dbms/src/TiDB/Collation/Collator.cpp
ChangRui-Ryan: OK

dbms/src/TiDB/Collation/Collator.cpp
ChangRui-Ryan: Good point

dbms/src/TiDB/tests/gtest_tidb_collator.cpp
```diff
@@ -310,48 +361,55 @@ void testCollator()
...
-            {
-                pattern->compile(p, ESCAPE);
+
                 for (size_t idx = 0; idx < std::size(inner_cases); ++idx)
                 {
                     const auto & inner_c = inner_cases[idx];
-                    const std::string & s = inner_c.first;
-                    res[idx] = pattern->match(s.data(), s.length());
+                    bool ans = std::get<Collator::collation_case>(inner_c.second);
+                    std::cout << "Pattern case (" << p << ", " << inner_c.first << ", " << ans << ")" << std::endl;
```
windtalker, line 409: this line should be removed?

ChangRui-Ryan: This line of code is from the original line 352. To facilitate locating specific cases when errors occur, let's leave it unchanged this time?

windtalker: ok

windtalker (approved):
lgtm

Issue #10382, opened 2025-08-28
[next gen on cloud] tiflash wn oom during running simulated workload


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
6*tiflash-wn-16c/64g
1、run workload
sysbench --db-driver=mysql --mysql-user=root --mysql-host=xxx --mysql-port=4000 --mysql-password=12345678 --mysql-db=xxx --order_tables=38 --update_ratio=0.1 --cop_read=0.0002 --batch_insert=0.0002 --threads=500 --report-interval=10 --time=0 c66_dc run

### 2. What did you expect to see? (Required)
no oom

### 3. What did you see instead (Required)
tiflash wn oom
<img width="696" height="286" alt="Image" src="https://github.com/user-attachments/assets/2edaa09f-dcc2-4c3c-92df-bf2cffc46374" />

![Image](https://github.com/user-attachments/assets/1064e085-c376-4f7a-96ea-6663bef63d85)

### 4. What is your TiFlash version? (Required)
sh-5.1# ./tiflash/tiflash --version
TiFlash
Release Version: v9.0.0-beta.2.pre-32-gc077d8bde
Edition:         Enterprise
Git Commit Hash: c077d8bde9cd95885bb8b0b6217844238947a8a9
Git Branch:      HEAD
UTC Build Time:  2025-08-28 04:16:02
Enable Features: jemalloc sm4(GmSSL) mem-profiling asimd unwind thinlto next-gen hnsw.l2=sve hnsw.cosine=sve vec.l2=sve vec.cos=sve
Profile:         RELWITHDEBINFO
Compiler:        clang++ 17.0.6

Raft Proxy
Git Commit Hash:   d467c7143a4c38ef4244647bf380f27fbd8ed78c
Git Commit Branch: HEAD
UTC Build Time:    ""   
Rust Version:      rustc 1.77.0-nightly (89e2160c4 2023-12-27)
Storage Engine:    tiflash
Prometheus Prefix: tiflash_proxy_
Profile:           release
Enable Features:   "raftstore-proxy/external-jemalloc" pprof-fp portable test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure pprof-fp portable test-engine-kv-rocksdb test-engine-raft-raft-engine cloud-aws cloud-gcp cloud-azure pprof-fp

Issue #10278, opened 2025-07-03
[next gen on clould] one of tiflash wn continuous oom


Lily2025:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)
1、cluster node spec
2 * tiflash-write-8C32G
2 * tifllash-compute-128GB
keyspace1：2* tidb-8C16G
keyspace3：2* tidb-8C16G
tikv：6 * tikv-16C64G

2、run AP query and tp workload for keyspace1 and keyspace3

go-tpc ch run -D tpcc --host xxx --warehouses 2000 -T 0 --acThreads 1 --time 36000m --user root --password '' --interval '10s' --ignore-error all

go-tpc ch run -D tpcc --host xxx --warehouses 2000 -T 32 --acThreads 0 --time 36000m --user root --password '' --interval '10s'

### 2. What did you expect to see? (Required)
no oom 

### 3. What did you see instead (Required)
one of tiflash wn continuous oom  after a long time running workload

![Image](https://github.com/user-attachments/assets/8872ab0f-4c34-45a1-b573-23236d39a082)

![Image](https://github.com/user-attachments/assets/dea4711f-6cbb-491c-b44f-411aa17a4752)

### 4. What is your TiFlash version? (Required)
TiFlash
Release Version: v9.0.0-beta.2.pre-2-gb3a52b7ff
Edition: Enterprise
Git Commit Hash: https://github.com/pingcap/tiflash/commit/b3a52b7ff8989416907d3576fa8f10254cdef27a
Git Branch: HEAD
UTC Build Time: 2025-06-27 10:43:38
Enable Features: jemalloc sm4(GmSSL) mem-profiling asimd unwind thinlto next-gen hnsw.l2=sve hnsw.cosine=sve vec.l2=sve vec.cos=sve
Profile: RELWITHDEBINFO
Compiler: clang++ 17.0.6

Raft Proxy
Git Commit Hash: d467c7143a4c38ef4244647bf380f27fbd8ed78c
Git Commit Branch: HEAD
UTC Build Time: ""
Rust Version: rustc 1.77.0-nightly (89e2160c4 2023-12-27)
Storage Engine: tiflash
Prometheus Prefix: tiflash_proxy_
Profile: release
[...]

PR #10410, opened 2025-09-05
PageStorage: Fix tiflash wn oom issue by introducing DeltaTreeOnlySnapshot

JaySon-Huang:

Issue Number: close https://github.com/pingcap/tiflash/issues/10382, close https://github.com/pingcap/tiflash/issues/10278

Problem Summary:

Slow query requests prevent the release of PageStorage snapshots, causing persistent issues where multi-versioned data in PageStorage memory and multi-versioned columnar storage data on disk cannot be compacted. However, under the storage-compute separation architecture, TiFlash aims to support the capability of restoring new Region replicas from S3 , which motivates storing high-frequency write-and-delete data such as RaftLog into PageStorage. As a result, with Raft messages sustained at 12K QPS, a 10-minute-long query causes the number of Pages in PageStorage to increase by 7.2 million and the number of multi-versions to rise by 17.8 million. Due to the design of PageStorage, this leads to an additional 6 GB of memory consumption, and may also be accompanied by degraded read and write performance in PageStorage.

```commit-message
* UniversalPageStorage Provide two kinds of snapshot
  * GeneralSnapshot: protect all data, used for general-purpose reading on UniversalPageStorage
  * DeltaTreeOnlySnapshot: protect only data in DeltaTree engine, used for handling reading requests
* UniversalPageStorage GC
  * The page_ids from raft layer is not protected by `DeltaTreeOnlySnapshot`
  * If the page_id is from raft layer and there is no general snapshot, clean up entries more aggressively to reduce memory usage.
* WriteNode snapshot GC
  * The write node snapshot timeout is defined by compute node `EstablishDisaggTaskRequest`
  * The default snapshot timeout increase from 5 minutes to 30 minutes
```

#### Manual Testing result

```bash
# use br to restore a dataset with chbenchmark 1500
tiup br:v8.5.2 restore db --db=chbenchmark -s s3://benchmark/ch-1500 --check-requirements=false \
 --pd 10.2.12.81:6580 --keyspace-name ks1
# start tp workload with 32 tp threads, last for 90 minutes
tiup bench ch --host 10.2.12.81 -P 8081 --warehouses 1500 run -D chbenchmark -T 32 -t 0 --time 90m
# hack the code to mock that a reading request last for 15 minutes without releasing snapshot on write node, and `dt_snapshot_delta_tree_only = false`
https://github.com/pingcap/tiflash/pull/10410/commits/13e79da504d720506ea52ba52851b4128a022e14
# checkout the memory usage of tiflash write node
# update `dt_snapshot_delta_tree_only = false` and mock another reading request last for 15 minutes without releasing snapshot on write node
```

According to the test result

- 21:09~21:25, acquire a GeneralSnapshot as before this PR does.
  - The memory usage peak reach 17.4GB. Increase more than 12.7GB for the frequent insert and delete raft pages
  - The number of pages stored in UniPageStorage reach 19.6 million, and versioned entries reach 49.5 million
  - The disk usage of UniPageStorage increase from 6GB to 24GB
- 21:34~21:50, acquire a DeltaTreeOnlySnapshot
  - The memory usage peak remain less than 5.8GB. Increase only 1.7GB from the lower watermark
  - The number of pages stored in UniPageStorage is only 3.4 million, and versioned entries is 5.7 million
  - The disk usage of UniPageStorage increase from 6GB to 16GB

image

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
  described as above
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a OOM issue when there is slow query under compute and storage disagg arch
```

### DIFF
--- a/dbms/src/Storages/tests/gtest_path_pool.cpp
+++ b/dbms/src/Storages/tests/gtest_path_pool.cpp
@@ -427,12 +427,7 @@ TEST_F(PathCapacity, SingleDiskSinglePathTest)
 
     // Single disk with single path
     {
-        auto path_capacity = PathCapacityMetrics(
-            0,
-            {main_data_path},
-            {capacity},
-            {latest_data_path},
-            {capacity});
+        auto path_capacity = PathCapacityMetrics(0, {main_data_path}, {capacity}, {latest_data_path}, {capacity});
 
         path_capacity.addUsedSize(main_data_path, used);
         auto stats = path_capacity.getFsStats(DisaggregatedMode::None, false);


JinheLin (approved):

JaySon-Huang:
/hold
for adding metrics

CalvinNeo (approved):
lgtm

PR #10471, opened 2025-10-03
expression: fix float-to-bool conversion to match TiDB (and SQL standard)

### LINKED ISSUE #10470: TiFlash float-to-bool conversion in WHERE predicate is inconsistent with TiDB

On a hybrid TiDB + TiFlash deployment, run the following SQL:

```sql
CREATE TABLE t0 (c1 INT);
INSERT INTO t0 VALUES (65535);
ALTER TABLE t0 SET TIFLASH REPLICA 1;

-- TiDB
SELECT c1 FROM t0 WHERE (1/c1); -- returns 65535

-- TiFlash (MPP or engine isolation)
SET SESSION tidb_enforce_mpp = ON;
SELECT c1 FROM t0 WHERE (1/c1); -- returns empty set
```

TiDB and TiFlash should produce consistent results for WHERE predicates like (1/c1):  
As long as the expression evaluates to a non-zero value, the row should be returned.

TiDB returns the row (65535).  
TiFlash returns an empty set.  
This shows the float-to-bool conversion logic is inconsistent.

### Problem Analysis

- In TiDB (Go), float to bool conversion logic is: any non-zero value is treated as TRUE.
- In TiFlash (C++), the current behavior seems to have a special handling, or the conversion does not match the standard, so very small float results (like 0.0000152 from 1/65535) may be treated as FALSE.
- This leads to division predicates, aggregation, and operator results being inconsistent between MPP and TiDB, affecting data correctness.

v8.5.3

### Suggestion

Please refer to TiDB’s logic:  
Update float/double-to-bool conversion in TiFlash so that any non-zero value is treated as TRUE.  
This will ensure expression evaluation is consistent and eliminate result differences between distributed executions.

Reference: [TiFlash behaves differently from TiDB #63818](https://github.com/pingcap/tidb/issues/63818)

DengY11:

Issue Number: close #10470 

Problem Summary:

TiFlash’s float-to-bool conversion in WHERE predicates was inconsistent with TiDB. In TiDB, any non-zero floating point value is treated as TRUE in boolean context. In TiFlash, very small non-zero float values could be mistakenly considered FALSE due to differences in implementation. This could lead to result inconsistency between TiDB and TiFlash, especially for SQL expressions like `WHERE (1/c1)`.

This PR updates the boolean conversion logic for floating point types (float/double) in predicate evaluation. Now, TiFlash treats any non-zero float value as TRUE, matching TiDB and standard SQL behavior. The change uses an epsilon comparison, so values that are exactly zero are FALSE, and all others are TRUE.

#### Commit message:

```
expression: fix float-to-bool conversion to match TiDB (and SQL standard)

- For floating-point values, boolean conversion now treats all non-zero values as TRUE.
- Uses `std::abs(value) < std::numeric_limits<T>::epsilon()` for zero comparison, covering precision issues.
- Ensures WHERE predicates and filter conditions produce consistent results with TiDB.
```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Functions/FunctionsComparison.h
+++ b/dbms/src/Functions/FunctionsComparison.h
@@ -1476,6 +1476,10 @@ struct IsTrueTrait
     {
         if constexpr (IsDecimal<T>)
             return value.value == 0 ? 0 : 1;
+        else if constexpr (std::is_floating_point_v<T>)
+            // For TiDB compatibility, floating point values should be considered true if not zero
+            // This matches SQL standard behavior where non-zero values are true in boolean context
+            return std::abs(value) < std::numeric_limits<T>::epsilon() ? 0 : 1;
         else
             return value == 0 ? 0 : 1;
     }
@@ -1492,6 +1496,10 @@ struct IsFalseTrait
     {
         if constexpr (IsDecimal<T>)
             return value.value == 0 ? 1 : 0;
+        else if constexpr (std::is_floating_point_v<T>)
+            // For TiDB compatibility, floating point values should be considered false only if zero
+            // This matches SQL standard behavior where zero values are false in boolean context
+            return std::abs(value) < std::numeric_limits<T>::epsilon() ? 1 : 0;
         else
             return value == 0 ? 1 : 0;
     }


windtalker:
As I mentioned in https://github.com/pingcap/tiflash/issues/10470#issuecomment-3388878541, this pr actually does not fix the issue.

PR #10482, opened 2025-10-16
disable two-level hash table for fine-grained agg if spill is not needed

### LINKED ISSUE #10481: fine-grained agg should not enable two-level hash table if no spill is needed
## Bug Report

Currently, for a fine grained agg, if will convert the hash table to 2 level hash table if needed, even if spill is not enabled. This will hurt the potential performance since for fine grained agg, it only aggregate data using 1 thread for each stream independently, here is no need to use 2 level hash table.

windtalker:

Issue Number: close #10481

Problem Summary:
disable two-level hash table for fine-grained agg if spill is not needed.
In fact, for fine grained hash agg, we should also use a new spill algorithm for fine grained agg(we can actually spill data by threads instead of convert the hash table into two-level hash table), but this pr just keep the original behavior if spill is enabled.

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### DIFF
--- a/dbms/src/Flash/tests/gtest_split_tasks.cpp
+++ b/dbms/src/Flash/tests/gtest_split_tasks.cpp
@@ -75,7 +75,7 @@ try
            "    Expression: <before TopN>\n"
            "     Filter\n"
            "      Expression: <expr after aggregation>\n"
-           "       Aggregating\n"
+           "       Aggregating, enable_two_level_hashtable: false\n"
            "        MockExchangeReceiver"};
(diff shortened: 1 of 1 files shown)


gengliqi (approved):

xzhangxian1008 (approved):
LGTM

PR #10542, opened 2025-11-10
tici: support pushdown cmp function

wshwsh12:

Issue Number: close #xxx

Problem Summary:

```commit-message

```

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```


dbms/src/Storages/Tantivy/TantivyInputStream.h
```diff
@@ -267,8 +268,59 @@ class TantivyInputStream : public IProfilingBlockInputStream
...
+            case tipb::ScalarFuncSig::EQDecimal:
+            case tipb::ScalarFuncSig::NEDecimal:
+            case tipb::ScalarFuncSig::LTDecimal:
+            case tipb::ScalarFuncSig::LEDecimal:
+            case tipb::ScalarFuncSig::GTDecimal:
+            case tipb::ScalarFuncSig::GEDecimal:
+            {
+                ret.sig = expr.sig();
+                size_t col_idx = 0;
+                if (ret.children[0].tp != tipb::ExprType::ColumnRef)
```
windtalker: what about both children is not columnRef?

dbms/src/Storages/Tantivy/TantivyInputStream.h
```diff
@@ -267,8 +268,59 @@ class TantivyInputStream : public IProfilingBlockInputStream
...
+                auto & val_expr = ret.children[1 - col_idx];
+                for (auto & name_and_type : return_columns)
+                {
+                    if (name_and_type.name == column_ref.val.c_str())
+                    {
+                        // TiCI need scale and precision to encode decimal compared binary representation.
+                        auto type = removeNullable(name_and_type.type);
+                        auto scale = getDecimalScale(*type, 0);
+                        auto prec = getDecimalPrecision(*type, 0);
+                        val_expr.val = fmt::format("{:0>2}{:0>2}{}", scale, prec, val_expr.val.c_str());
```
windtalker: I think there is no dependence of column's type and the constant's type, so I don't understand why need to use column's scale and prec here

dbms/src/Storages/Tantivy/TantivyInputStream.h
```diff
@@ -284,10 +336,57 @@ class TantivyInputStream : public IProfilingBlockInputStream
...
         {
-            ret.tp = tipb::ExprType::String;
+            ret.tp = expr.sig();
             ret.val = expr.val();
             return {ret, {}};
         }
+        case tipb::ExprType::Int64:
+        {
+            ret.tp = expr.sig();
+            ret.val = fmt::format("{}", decodeDAGInt64(expr.val()));
```
windtalker: it looks to me convert a non string to string and convert it back is low efficient, do we have a better way to avoid this convert?

dbms/src/Storages/Tantivy/TantivyInputStream.h
wshwsh12: TiCI converts decimals into a binary representation and inserts them into Tantivy. When encoding this binary representation, you must provide precision and scale to ensure the generated key is fixed-length and comparable.

dbms/src/Storages/Tantivy/TantivyInputStream.h
wshwsh12: TiCi should check it, maybe TiFlash should simply pass the tipb through to tici unchanged. I remove this code.

dbms/src/Storages/Tantivy/TantivyInputStream.h
wshwsh12: Time, Decimal, and String types all require literal string-type inputs, so optimization is likely only possible for int and float types. For convenience, optimization has not been implemented yet.

dbms/src/Storages/Tantivy/TantivyInputStream.h
wshwsh12: I should obtain the precision and scale from the meta information in TiCI.  So I remove this code now. PTAL again

windtalker (approved):
lgtm

JinheLin (approved):

PR #10726, opened 2026-02-14
support nulleq in tiflash

windtalker:
### What problem does this PR solve?

Issue Number: close #xxx

Problem Summary:

### What is changed and how it works?

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added a null-aware equality operator covering more types (including vector float32) with proper string collation forwarding.

* **Behavior**
  * Query/filter parsing now recognizes null-safe equality and emits is-null predicates when comparing to NULL.

* **Tests**
  * Added unit tests validating mapping, runtime null propagation, collation behavior, and filter-parser null-safe cases.

PR #10769, opened 2026-03-24
DeltaMerge: fix rough-set semantics for NullEQ

windtalker:

Issue Number: None

Problem Summary:

Follow-up to #10726. DeltaMerge rough-set filtering currently reuses the normal `Equal` operator for `col <=> non_null_literal`, which is not null-safe under `NOT`. For nullable packs this can prune packs that contain `NULL` rows even though `NOT (col <=> k)` should match them.

```commit-message
DeltaMerge: fix rough-set semantics for NullEQ

- add a dedicated `NullEqual` rough-set operator instead of lowering to `Equal`
- teach minmax index to evaluate nullable `NullEQ` with null-safe semantics
- keep the old minmax-index compatibility path conservative when it cannot distinguish pure-null from mixed packs
- add regression coverage for nullable packs and old minmax-index compatibility
- update the new NullEQ test files introduced in #10726 to use copyright year 2026
```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

### Tests

- `./cmake-build-debug/dbms/gtests_dbms --gtest_filter='MinMaxIndexTest.ParseNullEQ:MinMaxIndexTest.CompatibleOldMinmaxIndex:MinMaxIndexTest.NullEQOrNotNullEQWithNullPack'`

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  * Added null-safe equality filtering so comparisons involving NULL are handled correctly and efficiently.
  * Extended query parsing and index rough-checks to recognize and optimize null-aware equality expressions.

* **Tests**
  * Added and updated tests to validate null-safe equality behavior, including NOT-wrapped and string/date cases.

* **Chores**
  * Minor copyright header updates.

### DIFF
--- a/dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
+++ b/dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
@@ -290,10 +290,10 @@ RSResults MinMaxIndex::checkNullableIn(
             size_t pos = i * 2;
             size_t prev_offset = pos == 0 ? 0 : offsets[pos - 1];
             // todo use StringRef instead of String
-            auto min = String(chars[prev_offset], offsets[pos] - prev_offset - 1);
+            auto min = String(reinterpret_cast<const char *>(&chars[prev_offset]), offsets[pos] - prev_offset - 1);
             pos = i * 2 + 1;
             prev_offset = offsets[pos - 1];
-            auto max = String(chars[prev_offset], offsets[pos] - prev_offset - 1);
+            auto max = String(reinterpret_cast<const char *>(&chars[prev_offset]), offsets[pos] - prev_offset - 1);
             auto value_result = RoughCheck::CheckIn::check<String>(values, type, min, max);
             results[i - start_pack] = addNullIfHasNull(value_result, i);
         }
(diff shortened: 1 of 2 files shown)


windtalker:
@pantheon-ai review

windtalker:
@pantheon-ai review

Mini256:
@pantheon-ai review

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
JaySon-Huang, line 1: The branch fixes the original rough-set bug for `NOT (col <=> k)`, but this new nullable-string path still has a correctness issue. In `checkNullableNullEqual()`, `String(chars[prev_offset], offsets[pos] - prev_offset - 1)` does not build a string from the underlying byte range; it uses the `(count, ch)` constructor shape instead, so the min/max boundary strings are decoded incorrectly. That means `Nullable(String)` packs can still get wrong rough-check results for `col <=> 'x'` and `NOT (col <=> 'x')`. Please switch this to the same `reinterpret_cast<const char *>(&chars[prev_offset])` pattern already used in the existing string minmax paths, and add a regression test for `Nullable(String)` with both a literal match and a NULL-containing pack.
Source: rule/JAYSONHUANG-RC-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

windtalker:
@pantheon-ai review

Mini256:
@pantheon-ai review

Mini256:
@pantheon-ai review

PR #10771, opened 2026-03-25
disagg: limit S3 read amplification on compute nodes

### LINKED ISSUE #10752: Avoid MPP queries exhaust tiflash-compute network under disagg arch
## Enhancement

A Dumpling export job failed with `Error 1105 (HY000): rpc error: code = Unavailable desc = keepalive ping failed to receive ACK within timeout` when the query was executed through TiFlash/Mpp. The issue was triggered by very high TiFlash network bandwidth at that time, which delayed the keepalive ACK and caused the MPP stream to fail.

### Reproduction Steps
This issue is not easy to reproduce in a stable way, because it depends on specific runtime conditions rather than on a simple deterministic query pattern alone.
A similar failure is more likely to happen when all of the following conditions are met:
- the query is executed on the TiFlash MPP path;
- the workload involves disaggregated reads with a relatively high number of S3 requests;
- TiFlash is under very high network load at that time; and
- the keepalive ACK from TiFlash is not returned to TiDB within the timeout window.
In this incident, the Dumpling export query hit exactly this kind of runtime condition and failed with keepalive ping failed to receive ACK within timeout.

### Analysis
The Dumpling log showed that the export failed with rpc error: code = Unavailable desc = keepalive ping failed to receive ACK within timeout. The same symptom was also observed on the TiDB side during MPP stream receiving, which indicates that the failure happened on the TiDB–TiFlash RPC path rather than in Dumpling logic itself. 

Image

Further investigation showed that the affected SQL was executed on TiFlash. At the incident time, the TiFlash node handling the query was under very high network load, with observed bandwidth reaching about 2.28 GiB/s. The instance type’s documented network bandwidth is about 12 / up to 25 Gbit/s, which is approximately 1.4 / up to 2.91 GiB/s after converting to GiB/s, so the observed traffic was already close to the upper end of the available bandwidth.

Image

The TiFlash metrics further showed that the high network traffic was mainly driven by a large number of S3 read requests at that time. This was related to an intentionally increased concurrency strategy on the disaggregated read path, which helps improve performance in some cache-miss scenarios(https://github.com/pingcap/tiflash/pull/10522), but does not sufficiently constrain network bandwidth under workloads such as wide-table reads, cold reads, and multi-segment access. In addition, the current synchronous I/O model and the lack of effective S3 bandwidth governance can further amplify the pressure on network bandwidth.
Taken together, the evidence indicates that the incident was caused by the TiFlash execution path under heavy S3-related network pressure, which eventually led to the keepalive timeout between TiDB and TiFlash.

JaySon-Huang:

Issue Number: close #10752

Problem Summary:

In disaggregated read workloads, compute nodes can generate uncontrolled S3 traffic when cold reads and repeated cache misses amplify remote reads on the same node. We need a production-safe way to bound S3 traffic, reduce duplicate downloads for the same object, and make the resulting behavior observable and tunable.

```commit-message
add a node-level S3 byte limiter and wire it into both direct reads and FileCache downloads
avoid duplicate S3 downloads for the same key by adding bounded wait on in-flight FileCache entries
add phase-2 remote cache observability for bounded wait and background download stages
refine FileCache and S3 read paths for maintainability, comments, failure handling, and reload consistency
add an English design document for the latest disagg S3 backpressure plan
```

More specifically:

- add `S3ReadLimiter` to cap node-level S3 read bytes for both direct reads and FileCache downloads
- add `storage.io_rate_limit.s3_max_read_bytes_per_sec`
- remove the earlier `s3_max_get_object_streams` stream-cap design after validation showed that a reader-lifetime stream token can stall disaggregated query progress
- publish the shared limiter through `ClientFactory` so all S3 clients on the node observe the same node-level byte budget
- keep the limiter object stable across config reloads; reloading to `0` now updates the existing limiter instance instead of replacing it, so in-flight readers observe the disable and the related metric is updated consistently
- make `S3RandomAccessFile` use limiter-aware chunked read/seek paths while sharing common read/seek tail logic
- make `FileCache` downloads reuse the same limiter, add `profiles.default.dt_filecache_wait_on_downloading_ms`, and let follower requests perform bounded wait on an in-flight same-key download instead of immediately creating more remote-read pressure
- handle bounded-wait download failures by removing failed placeholders eagerly so later retries can proceed
- add remote-cache observability for:
  - bounded wait result / bytes / wait-seconds by `result x file_type`
  - background download stage seconds by `stage x file_type`
  - reject counts for `too_many_download` by `file_type`
  - background download queue depth
- add an English design doc under `docs/design/2026-03-24-disagg-s3-node-level-backpressure-and-filecache-dedup.md`

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

#### Manual test
Tested with following py script
* [run_cold_disagg_read.py](https://github.com/user-attachments/files/26488370/run_cold_disagg_read.py)
```
# Round 1: cold-cache baseline without node-level S3 byte limiting and without FileCache bounded wait.
# Round 2: cold-cache baseline with two concurrent queries and no parameter changes.
# Round 3: cold-cache run with node-level S3 byte limiting enabled at 150 MiB/s.
# Round 4: keep the same S3 byte limit, enable FileCache bounded wait at 5000ms.
# Round 5: raise the S3 byte limit from 150 MiB/s to 200 MiB/s and rerun the cold-cache query.
# Round 6: keep the same config as round 5 and run the target query twice concurrently.
```

image
image
image
image

```
The six-round run_cold_disagg_read.py experiment clearly reproduces the target scenario described in the design doc: under cold-cache conditions, two concurrent queries without any limiter push the host bond1 receive peak to about 2.48 GB/s, while direct reads and FileCache downloads both surge at the same time. This is exactly the node-level S3 read amplification problem the proposal is meant to control. After enabling s3_max_read_bytes_per_sec, TiFlash shows sustained s3_read_byte pending activity, and the host-level network peak is consistently reduced to the few-hundred-MB/s range. That confirms the node-level byte limiter is effective not only in internal accounting but also at the real machine network level, turning the system from “unbounded amplification that can blow up the node network” into a bounded, observable, and tunable degraded state.

In this workload, 150 MiB/s is too conservative and significantly stretches TiFlash execution time. Raising the limit to 200 MiB/s noticeably improves both tiflash_task execution time and overall slowlog wall time in the single-query case, while still keeping network usage under control. For the current sample, s3_max_read_bytes_per_sec = 209715200 together with dt_filecache_wait_on_downloading_ms = 5000 is the most balanced configuration observed so far. Bounded wait helps mainly on the merged path in single-query runs, but under two-query concurrency it also starts producing substantial coldata hits/timeouts, which indicates that its benefit expands once overlap becomes strong enough. At this point, the dominant remaining bottleneck is no longer uncontrolled network traffic, but FileCache background-download supply: bg_downloading_count, bg_download_queue_count, and too_many_download stay high across multiple rounds. Therefore, the next optimization priority should be downloader concurrency / queue / admission behavior rather than further increasing the bounded-wait budget.
```

Side effects

Documentation

- [x] Affects user behaviors
- [ ] Contains syntax changes
  New configs https://github.com/pingcap/docs-cn/pull/21455
```
profiles.default.dt_filecache_wait_on_downloading_ms
storage.io_rate_limit.s3_max_read_bytes_per_sec
```

### Release note

```release-note
limit S3 read amplification on compute nodes
```

## Summary by CodeRabbit

* **New Features**
  * Node-level S3 read-rate limiter; bounded follower wait for in-flight FileCache downloads; two per-node HTTP cache-evict endpoints.

* **Improvements**
  * Limiter-aware chunked reads/seeks and downloads, background-download queue/stage timing, failed-placeholder cleanup, expanded remote-cache & I/O-limiter metrics, runtime wait tuning setting.

* **Tests**
  * New unit/integration coverage for limiter, FileCache wait semantics, background-failure and scheduling cleanup paths, and limiter publication.

* **Documentation**
  * Added design doc and HTTP API docs.

### DIFF
--- a/dbms/src/Storages/S3/FileCache.cpp
+++ b/dbms/src/Storages/S3/FileCache.cpp
@@ -244,10 +244,9 @@ void observeRemoteCacheRejectMetrics(FileType file_type)
         .Increment();
 }
 
-void updateBgDownloadStatusMetrics(Int64 bg_downloading_count)
+void updateBgDownloadStatusMetrics(Int64 bg_downloading_count, Int64 running_limit)
 {
     GET_METRIC(tiflash_storage_remote_cache_status, type_bg_downloading_count).Set(bg_downloading_count);
-    const auto running_limit = static_cast<Int64>(S3FileCachePool::get().getMaxThreads());
     GET_METRIC(tiflash_storage_remote_cache_status, type_bg_download_queue_count)
         .Set(std::max<Int64>(0, bg_downloading_count - running_limit));
 }
@@ -469,7 +468,7 @@ FileCache::FileCache(
     , log(Logger::get("FileCache"))
 {
     CurrentMetrics::set(CurrentMetrics::DTFileCacheCapacity, cache_capacity);
-    updateBgDownloadStatusMetrics(0);
+    updateBgDownloadStatusMetrics(0, /*running_limit*/ 0);
     prepareDir(cache_dir);
     restore();
 }
@@ -1380,7 +1379,8 @@ void FileCache::bgDownloadExecutor(
     const String & s3_key,
     FileSegmentPtr & file_seg,
     const WriteLimiterPtr & write_limiter,
-    std::chrono::steady_clock::time_point enqueue_time)
+    std::chrono::steady_clock::time_point enqueue_time,
+    Int64 running_limit)
 {
     observeBgDownloadStageMetrics(
         file_seg->getFileType(),
@@ -1411,13 +1411,17 @@ void FileCache::bgDownloadExecutor(
     {
         bg_download_succ_count.fetch_add(1, std::memory_order_relaxed);
     }
-    finishBgDownload(s3_key);
+    finishBgDownload(s3_key, running_limit);
 }
 
 void FileCache::bgDownload(const String & s3_key, FileSegmentPtr & file_seg)
 {
     bg_downloading_count.fetch_add(1, std::memory_order_relaxed);
-    updateBgDownloadStatusMetrics(bg_downloading_count.load(std::memory_order_relaxed));
+    // Capture the pool concurrency limit before scheduling. Background workers still update
+    // queue gauges while finishing, but tests may shut the global S3FileCachePool down at the
+    // same time. Re-reading the singleton from the worker tail would race with shutdown.
+    const auto running_limit = static_cast<Int64>(S3FileCachePool::get().getMaxThreads());
+    updateBgDownloadStatusMetrics(bg_downloading_count.load(std::memory_order_relaxed), running_limit);
     LOG_DEBUG(
         log,
         "downloading count {} => s3_key {} start",
@@ -1429,9 +1433,12 @@ void FileCache::bgDownload(const String & s3_key, FileSegmentPtr & file_seg)
     {
         FAIL_POINT_TRIGGER_EXCEPTION(FailPoints::file_cache_bg_download_schedule_fail);
         S3FileCachePool::get().scheduleOrThrowOnError(
-            [this, s3_key = s3_key, file_seg = file_seg, limiter = std::move(write_limiter), enqueue_time]() mutable {
-                bgDownloadExecutor(s3_key, file_seg, limiter, enqueue_time);
-            });
+            [this,
+             s3_key = s3_key,
+             file_seg = file_seg,
+             limiter = std::move(write_limiter),
+             enqueue_time,
+             running_limit]() mutable { bgDownloadExecutor(s3_key, file_seg, limiter, enqueue_time, running_limit); });
     }
     catch (...)
     {
@@ -1439,14 +1446,14 @@ void FileCache::bgDownload(const String & s3_key, FileSegmentPtr & file_seg)
         GET_METRIC(tiflash_storage_remote_cache, type_dtfile_download_failed).Increment();
         bg_download_fail_count.fetch_add(1, std::memory_order_relaxed);
         cleanupFailedDownload(s3_key, file_seg);
-        finishBgDownload(s3_key);
+        finishBgDownload(s3_key, running_limit);
     }
 }
 
-void FileCache::finishBgDownload(const String & s3_key)
+void FileCache::finishBgDownload(const String & s3_key, Int64 running_limit)
 {
     const auto count_after_finish = bg_downloading_count.fetch_sub(1, std::memory_order_relaxed) - 1;
-    updateBgDownloadStatusMetrics(count_after_finish);
+    updateBgDownloadStatusMetrics(count_after_finish, running_limit);
     LOG_DEBUG(log, "downloading count {} => s3_key {} finished", count_after_finish, s3_key);
 }
 
--- a/dbms/src/Storages/S3/FileCache.h
+++ b/dbms/src/Storages/S3/FileCache.h
@@ -370,8 +370,9 @@ public:
         const String & s3_key,
         FileSegmentPtr & file_seg,
         const WriteLimiterPtr & write_limiter,
-        std::chrono::steady_clock::time_point enqueue_time);
-    void finishBgDownload(const String & s3_key);
+        std::chrono::steady_clock::time_point enqueue_time,
+        Int64 running_limit);
+    void finishBgDownload(const String & s3_key, Int64 running_limit);
     void cleanupFailedDownload(const String & s3_key, FileSegmentPtr & file_seg);
     void downloadImpl(const String & s3_key, FileSegmentPtr & file_seg, const WriteLimiterPtr & write_limiter);
 


dbms/src/Storages/KVStore/FFI/ProxyFFIStatusService.cpp
```diff
@@ -579,6 +579,97 @@ RemoteCacheEvictRequest parseEvictRequest(
...
+    auto & global_ctx = server->tmt->getContext();
+    auto log = Logger::get("HandleHttpRequestLocalCacheEvict");
+
+    String err_msg;
+    auto cache_type = parseCacheEvictType(path, api_name, err_msg);
+    if (!cache_type.has_value())
+    {
+        auto body = fmt::format(R"json({{"status":"error","message":"{}"}})json", err_msg);
+        LOG_WARNING(log, "invalid local cache evict request, path={} api_name={}", path, api_name);
+        return buildRespWithCode(HttpRequestStatus::BadRequest, api_name, std::move(body));
```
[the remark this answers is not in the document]
JaySon-Huang: Currently the message is just plain text, resolve it later.

dbms/src/Storages/S3/FileCache.cpp
```diff
@@ -1000,17 +1265,35 @@ void downloadToLocal(
...
+        copyData(rbuf, wbuf, content_length);
+        if (s3_read_metrics_recorder != nullptr)
+            s3_read_metrics_recorder->recordBytes(rbuf.count(), S3::S3ReadSource::FileCacheDownload);
+        wbuf.sync();
+        return;
+    }
+
+    // The limiter-aware buffer preserves the old copyData/write-buffer path while charging the shared
+    // S3 budget before each refill from the remote body stream.
+    ReadBufferFromIStreamWithLimiter rbuf(istr, buffer_size, s3_read_limiter, S3::S3ReadSource::FileCacheDownload);
```
JaySon-Huang, line 1295: The limiter contract here assumes upper layers keep each request within one refill-period burst by using `getSuggestedChunkSize()`. This FileCache path still builds a limiter-aware buffer with a fixed refill size up to 128 KiB and then charges `requestBytes(internal_buffer.size())` on every refill. When `s3_max_read_bytes_per_sec` is small, each refill becomes an oversized request and falls back to the borrowing path, so FileCache downloads can substantially exceed the configured node-level budget and leave later direct reads paying back the debt. Please clamp the refill/buffer size with `getSuggestedChunkSize()` here, and add a non-zero low-limit regression test for FileCache downloads.

Source: rule/JAYSONHUANG-RC-005 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Server/Server.cpp
JaySon-Huang, line 1: The new `dt_filecache_wait_on_downloading_ms` setting is only applied in `FileCache::updateConfig(settings)`, but the startup path only calls `FileCache::initialize(...)` and never follows it with an initial `updateConfig(global_context->getSettingsRef())`. As a result, a non-default value from the config file stays at the in-memory default (`0`) until the first config reload. Please push the current settings into `FileCache` once right after initialization, and add a startup-path test so this setting does not depend on a later reload to take effect.

Source: rule/WJHUANG2016-RE-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Storages/S3/FileCache.cpp
JaySon-Huang: Re-checked on the latest PR head `9412be4ae7fd9cc22ca559f2ab337c1b7637d290`: this looks fixed. `downloadToLocal()` now clamps the limiter-aware refill size with `s3_read_limiter->getSuggestedChunkSize(...)` before constructing `ReadBufferFromIStreamWithLimiter`, so the FileCache download path no longer keeps issuing oversized 128 KiB borrowing requests under a low byte limit. There is also a targeted regression test, `BgDownloadUsesLimiterSuggestedChunkSize`, covering the low-limit FileCache download case.

Source: process/review-branch | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Server/Server.cpp
JaySon-Huang: Re-checked on the latest PR head `9412be4ae7fd9cc22ca559f2ab337c1b7637d290`: this also looks fixed. The startup path now calls `FileCache::instance()->updateConfig(global_context->getSettingsRef())` immediately after `FileCache::initialize(...)`, so startup-time values such as `dt_filecache_wait_on_downloading_ms` no longer wait for a later config reload to take effect.

Source: process/review-branch | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

metrics/grafana/tiflash_summary.json
```diff
@@ -20542,10 +20632,680 @@
...
+              "exemplar": true,
+              "expr": "sum(rate(tiflash_storage_remote_cache_reject{k8s_cluster=\"$k8s_cluster\", tidb_cluster=\"$tidb_cluster\", instance=~\"$instance\", instance=~\"$tiflash_role\"}[1m])) by (reason, file_type, $additional_groupby)",
+              "format": "time_series",
+              "instant": false,
+              "interval": "",
+              "intervalFactor": 2,
+              "legendFormat": "{{reason}}-{{file_type}} {{additional_groupby}}",
+              "refId": "A"
+            }
+          ],
```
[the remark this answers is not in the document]
JaySon-Huang: Fixed in https://github.com/pingcap/tiflash/pull/10771/changes/3e9bd94eb92c40e4bb26f60c30fdbbb2854cd1d9

JaySon-Huang:
@coderabbitai full review

dbms/src/Storages/S3/S3Common.h
```diff
@@ -146,6 +182,22 @@ class ClientFactory
 
     std::shared_ptr<TiFlashS3Client> sharedTiFlashClient();
 
+    void setS3ReadLimiter(const std::shared_ptr<S3ReadLimiter> & limiter)
+    {
+        std::unique_lock lock_init(mtx_init);
+        shared_s3_read_limiter = limiter;
+        if (shared_tiflash_client != nullptr)
+            shared_tiflash_client->setS3ReadLimiter(shared_s3_read_limiter);
+    }
```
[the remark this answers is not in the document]
JaySon-Huang: The nonzero -> 0 reload path is already handled by keeping the existing S3ReadLimiter object alive and applying updateConfig(0) in place, so readers that already snapped the shared limiter can observe disable without pointer recreation.
What still remains intentionally limited today is the first 0 -> nonzero transition when startup published no limiter object: previously constructed readers that snapped nullptr will not start throttling retroactively, and only newly created readers will see the newly published limiter. I added a code comment and TODO in RateLimiter.cpp to document this behavior and the possible follow-up of publishing a no-op limiter even at rate 0 if we want stronger reload semantics later.

JinheLin (approved):

JaySon-Huang:
> Please open a document PR on this

Related document PR: https://github.com/pingcap/docs-cn/pull/21455

PR #10871, opened 2026-05-26
columnar: support bucket parallel read in region

### LINKED ISSUE #10844: Support new columnar storage as data source
## Enhancement

Add new columnar storage data source for TiDB X.

The columnar storage data source is generated by TiKV next-gen with tikv-worker. So tiflash don't need to sync and decode the raftlog. So the disaggregated write path (tiflash write node) can be discard and can greatly reduce the burden of code synchronization between TIFlash and TIKV.
For deployed with columnar, only the read path from tiflash compute needs to be maintained. This is because tikv next-gen columnar only handles the generation of the columnar storage format. It still relies on tiflash compute's MPP computing framework to handle complex computational tasks.

yongman:

Issue Number: close #10844 

Problem Summary:
The read concurrency is limited in region level.

```commit-message

```
1. Support buckets level concurrency read in one region.
2. Optimize the streams dispatching for pipelines. Change read task pre-alloc to on-demand to avoid thread imbalance. 
3. Lazy create columnar reader & prefetch create reader in async thread.
4. Reuse SnapAccess for buckets concurrent read in one request.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **New Features**
  - PD-backed caching for region bucket keys and shared-snapshot de-duplication to reduce duplicate leader snapshot requests and improve read latency.
  - New engine interface hooks to fetch region bucket keys and clear shared-snapshot state.

* **Refactor**
  - Disaggregated columnar read pipeline rebuilt to a shared reader-task/work model with lazy, on-demand reader materialization for better parallelism and resource efficiency.
  - Server skips legacy read-thread initialization when columnar mode is enabled.
  - Source operator logging adds rows_per_sec and bytes_per_sec.

* **Tests**
  - Unit test added for cache eviction behavior during in-flight loads.

### DIFF
--- a/contrib/tiflash-columnar-hub/hub-runtime/src/cloud_helper.rs
+++ b/contrib/tiflash-columnar-hub/hub-runtime/src/cloud_helper.rs
@@ -69,7 +69,6 @@ const BACKOFF_RETRY_COUNT: usize = 5;
 
 const SNAPSHOT_CACHE_SIZE: usize = 10240; // 10k shards
 const SNAPSHOT_CACHE_CAP: u64 = 1024 * 1024 * 1024; // 1GB snapshot size with memtable
-const SHARED_SNAP_ACCESS_CACHE_SIZE: usize = 10240; // 10k request-local snapshot handles
 const SNAPSHOT_CACHE_CAPABILITY_HEADER: &str = "x-cse-snapshot-cache-version";
 
 #[derive(Debug, Error)]
@@ -234,7 +233,6 @@ pub struct CloudHelper {
     snapshot_cache: SnapCache,
     snapshot_cache_capable_stores: Arc<DashMap<u64, ()>>,
     shared_snap_access_cache: SharedSnapAccessCache,
-    shared_snap_access_loaders: Arc<DashMap<SharedSnapAccessKey, Arc<tokio::sync::Mutex<()>>>>,
     meta_file_cache: Arc<Cache<u64, Arc<dyn File>, MetaFileCacheWeighter>>,
     schema_files: Arc<DashMap<u64, SchemaFile>>,
     runtime: Arc<tokio::runtime::Runtime>,
@@ -282,8 +280,7 @@ impl CloudHelper {
         );
         let snapshot_cache = SnapCache::new(SNAPSHOT_CACHE_SIZE, SNAPSHOT_CACHE_CAP);
         let snapshot_cache_capable_stores = Arc::new(DashMap::new());
-        let shared_snap_access_cache = SharedSnapAccessCache::new(SHARED_SNAP_ACCESS_CACHE_SIZE);
-        let shared_snap_access_loaders = Arc::new(DashMap::new());
+        let shared_snap_access_cache = SharedSnapAccessCache::new();
 
         // Create a long-lived HTTP client for connection reuse
         let http_client = {
@@ -304,7 +301,6 @@ impl CloudHelper {
             snapshot_cache,
             snapshot_cache_capable_stores,
             shared_snap_access_cache,
-            shared_snap_access_loaders,
             meta_file_cache,
             schema_files: Arc::new(DashMap::new()),
             runtime,
@@ -421,7 +417,6 @@ impl CloudHelper {
         let snap_cache = self.snapshot_cache.clone();
         let snap_cache_capable_stores = self.snapshot_cache_capable_stores.clone();
         let shared_snap_access_cache = self.shared_snap_access_cache.clone();
-        let shared_snap_access_loaders = self.shared_snap_access_loaders.clone();
         let meta_file_cache = self.meta_file_cache.clone();
         let columnar_file_cache = self.columnar_file_cache.clone();
         let fts_cache = self.fts_cache.clone();
@@ -432,7 +427,6 @@ impl CloudHelper {
         self.runtime.spawn(async move {
             let snap = get_or_request_shared_snapshot(
                 shared_snap_access_cache,
-                shared_snap_access_loaders,
                 pd_client,
                 http_client,
                 dfs,
@@ -499,6 +493,20 @@ impl CloudHelper {
 
     pub fn get_region_bucket_keys(&self, region_id: u64, region_ver: u64) -> Vec<Vec<u8>> {
         self.pd_client.get_region_bucket_keys(region_id, region_ver)
+    }
+
+    pub fn clear_shared_snap_access_by_start_ts(&self, start_ts: u64) {
+        if start_ts == 0 {
+            return;
+        }
+
+        let (removed_cache_entries, removed_loader_entries) =
+            self.shared_snap_access_cache.remove_by_start_ts(start_ts);
+
+        info!(
+            "clear shared snapaccess by start_ts, start_ts: {}, removed_cache_entries: {}, removed_loader_entries: {}",
+            start_ts, removed_cache_entries, removed_loader_entries
+        );
     }
 }
 
@@ -936,48 +944,104 @@ impl Deref for SharedSnapAccessCache {
 }
 
 impl SharedSnapAccessCache {
-    pub fn new(size: usize) -> Self {
+    pub fn new() -> Self {
         Self {
-            core: Arc::new(SharedSnapAccessCacheCore::new(size)),
+            core: Arc::new(SharedSnapAccessCacheCore::new()),
         }
     }
 }
 
 #[derive(Clone)]
 pub struct SharedSnapAccessCacheCore {
-    cache: Arc<
-        Cache<
-            SharedSnapAccessKey,
-            Weak<kvengine::SnapAccessCore>,
-            SharedSnapAccessWeighter,
-            DefaultHashBuilder,
-        >,
-    >,
+    groups: Arc<DashMap<u64, Arc<SharedSnapAccessGroup>>>,
 }
 
 impl SharedSnapAccessCacheCore {
-    pub fn new(size: usize) -> Self {
-        let opts = quick_cache::OptionsBuilder::new()
-            .weight_capacity(size as u64)
-            .estimated_items_capacity(size)
-            .build()
-            .unwrap();
-
-        let cache = Arc::new(Cache::with_options(
-            opts,
-            SharedSnapAccessWeighter,
-            DefaultHashBuilder::default(),
-            DefaultLifecycle::default(),
-        ));
-        Self { cache }
+    pub fn new() -> Self {
+        Self {
+            groups: Arc::new(DashMap::new()),
+        }
     }
 
     pub fn get(&self, key: &SharedSnapAccessKey) -> Option<Weak<kvengine::SnapAccessCore>> {
-        self.cache.get(key)
+        self.groups
+            .get(&key.start_ts)?
+            .entries
+            .get(key)
+            .map(|entry| entry.clone())
     }
 
     pub fn insert(&self, key: SharedSnapAccessKey, entry: Weak<kvengine::SnapAccessCore>) {
-        self.cache.insert(key, entry);
+        let group = self
+            .groups
+            .entry(key.start_ts)
+            .or_insert_with(|| Arc::new(SharedSnapAccessGroup::new()))
+            .clone();
+        group.entries.insert(key, entry);
+    }
+
+    pub fn get_loader(&self, key: &SharedSnapAccessKey) -> Arc<tokio::sync::Mutex<()>> {
+        let group = self
+            .groups
+            .entry(key.start_ts)
+            .or_insert_with(|| Arc::new(SharedSnapAccessGroup::new()))
+            .clone();
+        let loader = group
+            .loaders
+            .entry(key.clone())
+            .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
+            .clone();
+        loader
+    }
+
+    pub fn remove_loader(&self, key: &SharedSnapAccessKey) -> bool {
+        let Some(group) = self.groups.get(&key.start_ts).map(|entry| entry.clone()) else {
+            return false;
+        };
+        let removed = group.loaders.remove(key).is_some();
+        self.try_remove_empty_group(key.start_ts, &group);
+        removed
+    }
+
+    pub fn remove_entry(&self, key: &SharedSnapAccessKey) -> bool {
+        let Some(group) = self.groups.get(&key.start_ts).map(|entry| entry.clone()) else {
+            return false;
+        };
+        let removed = group.entries.remove(key).is_some();
+        self.try_remove_empty_group(key.start_ts, &group);
+        removed
+    }
+
+    pub fn remove_by_start_ts(&self, start_ts: u64) -> (usize, usize) {
+        let Some((_, group)) = self.groups.remove(&start_ts) else {
+            return (0, 0);
+        };
+        (group.entries.len(), group.loaders.len())
+    }
+
+    fn try_remove_empty_group(&self, start_ts: u64, group: &Arc<SharedSnapAccessGroup>) {
+        if group.entries.is_empty() && group.loaders.is_empty() {
+            if let Some(entry) = self.groups.get(&start_ts) {
+                if Arc::ptr_eq(entry.value(), group) {
+                    drop(entry);
+                    let _ = self.groups.remove(&start_ts);
+                }
+            }
+        }
+    }
+}
+
+pub struct SharedSnapAccessGroup {
+    entries: DashMap<SharedSnapAccessKey, Weak<kvengine::SnapAccessCore>>,
+    loaders: DashMap<SharedSnapAccessKey, Arc<tokio::sync::Mutex<()>>>,
+}
+
+impl SharedSnapAccessGroup {
+    pub fn new() -> Self {
+        Self {
+            entries: DashMap::new(),
+            loaders: DashMap::new(),
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


JaySon-Huang:
# Bug: EPOCH_NOT_MATCH Infinite Retry on Region 125 in Disaggregated Columnar Read

## Summary

When running `fullstack-test2/clustered_index/ddl.test` against a next-gen columnar cluster, the test fails at line 79 with `Region error EPOCH_NOT_MATCH(125:68:1)`. The TiFlash compute node repeatedly retries with a stale region epoch (`shard_ver=68`) for ~60 seconds, never updating to the current epoch (`version=69`), causing the query to time out.

## Test Case

- **Test file:** `tests/fullstack-test2/clustered_index/ddl.test`
- **Failing line:** 79 — `set session tidb_isolation_read_engines='tiflash'; select * from test.t_3;`
- **Cluster:** next-gen columnar (tiflash-cn0 only, no tiflash-wn0)
- **Branch:** `bucket-read`

## Test Flow (simplified)

```sql
-- t_1 (int handle clustered): PASSED
create table test.t_1(a int primary key clustered, col int);
insert into test.t_1 values(1,2),(2,3);
alter table test.t_1 set tiflash replica 1;
-- wait_table → select → alter column → select → drop

-- t_2 (common handle clustered): PASSED
create table test.t_2(a varchar(10), b int, c int, primary key(a, b) clustered);
insert into test.t_2 values('1',2,3),('2',3,4);
alter table test.t_2 set tiflash replica 1;
-- wait_table → select → alter column → select → drop

-- t_3 (composite PRIMARY KEY clustered): FAILED
create table test.t_3 (A int, B varchar(20), C int, D int, PRIMARY KEY(A,C) CLUSTERED);
insert into test.t_3 values (1,'1',1,1),(2,'2',2,2);
alter table test.t_3 set tiflash replica 1;
-- wait_table → select → ERROR!
```

## Error

```
ERROR 1105 (HY000) at line 1: other error for mpp stream:
Poco::Exception. Code: 1000, e.code() = 0,
e.displayText() = Exception: Region error EPOCH_NOT_MATCH(125:68:1),
e.what() = Exception
```

## Log Analysis

All log files are under `tests/fullstack-test-next-gen-columnar/log/tiflash-cn0/`.

### Timeline

| Timestamp | Event | Source |
|-----------|-------|--------|
| `12:36:27.749` | **t_2 query**: `buildProxyReadTask, physical_table_id=14, region_ver_id={125,1,68}` | tiflash.log:362 |
| `12:36:27.750` | Proxy fetches snapshot: `shard_ver=68, start_key=...0E, end_key=...0F` — **SUCCESS** | tiflash_tikv.log:42 |
| `12:36:27.860` | **t_2 query** (second): `cache_hit: true` — **SUCCESS** | tiflash_tikv.log:58 |
| `12:36:27.932` | `drop table test.t_1` — triggers region split | test output |
| `12:36:27.998` | `drop table test.t_2` — region 125 epoch bumps 68→69 | test output |
| `12:36:30.168` | **t_3 query**: `buildProxyReadTask, physical_table_id=23, region_ver_id={125,1,68}` ← STALE! | tiflash.log:560 |
| `12:36:30.169` | Proxy request: `shard_ver=68` — **EPOCH_NOT_MATCH (current version=69)** | tiflash_tikv.log:66 |
| `12:36:30.169` | `try drop region, region_ver_id={125,1,68}` | tiflash.log:581 |
| `12:36:30.169` | `drop region because of send failure` | tiflash.log:582 |
| `12:36:30.169` ~ `12:37:30` | **Infinite retry loop**: same `shard_ver=68`, same EPOCH_NOT_MATCH, repeated ~40+ times | tiflash_error.log |
| `12:37:30` | Query times out after ~60s | test output |

### Region 125 — Before vs After Split

**Before (epoch=68, t_2 exists):**
```
start_key: 78FFFFFE74800000000000000E   → table_id=14 (t_2)
end_key:   78FFFFFE74800000000000000F
columnar_table_ids: 14
```

**After (epoch=69, t_2 dropped, t_3 created):**
```
start_key: 78FFFFFE74800000FF0000000017000000FC  → table_id=23 (t_3)
end_key:   78FFFFFE748000FFFFFFFFFFFFC1000000FC
region_epoch { conf_ver: 1 version: 69 }
```

Key changes:
1. **Epoch bumped from 68 to 69** — region split after dropping t_1/t_2
2. **Key encoding changed** — `78FFFFFE7480...0E` vs `78FFFFFE74800000FF...17` — the `FF` indicates a different key encoding for the new table's clustered index
3. **Key range completely different** — from table 14's range to table 23's range

---

## Root Cause

The bug is a **missing epoch refresh in the EPOCH_NOT_MATCH handling path**, spanning two layers:

### Architecture

```
C++ (StorageDisaggregatedColumnar.cpp)     Rust (cloud_helper.rs in hub-runtime)
─────────────────────────────────────     ─────────────────────────────────────
                                          ┌─────────────────────────────────┐
buildProxyReadTask()                      │ request_snapshot_from_leader()  │
  ↓ resolve key ranges from region_cache  │   while backoff.next_delay():   │
  ↓ captures region_ver=68                │     HTTP GET tikv0/.../125?     │
  ↓                                      │       shard_ver=68&start_key=.. │
reader_plans = [{125, 68, 1}, ...]       │                                  │
  ↓                                      │     if not_leader:               │
createColumnarReaderWithBackoff()        │       evict + retry ✅           │
  while(true):                           │                                  │
    ↓                                    │     if epoch_not_match:          │
    createProxyColumnarReader(plan)      │       return Err(...) ❌ ← BUG   │
      ↓ FFI with region_ver=68           │       (never retries!)           │
      ↓                                  │                                  │
      ← EPOCH_NOT_MATCH                  └─────────────────────────────────┘
      ← dropRegion() from cache
      ← throw RegionException
      ↓
    catch → backoff → RETRY with SAME plan (still ver=68!)
```

### Two Contributing Issues

#### Issue 1 (C++ layer): `reader_plan` is immutable in the retry loop

`StorageDisaggregatedColumnar.cpp:763-792` — `createColumnarReaderWithBackoff`:

```cpp
ColumnarReaderPtr RNProxyReadTask::createColumnarReaderWithBackoff(size_t reader_index) const
{
    const auto & reader_plan = reader_plans[reader_index];  // const ref — NEVER updated!
    pingcap::kv::Backoffer bo(pingcap::kv::copNextMaxBackoff);
    while (true)
    {
        try
        {
            return createProxyColumnarReader(*shared_reader_context, reader_plan);
            //                                  ^^^^^^^^^^^^^^ always the same {125, 68, 1}
        }
        catch (RegionException & e)
        {
            // Only backoff — reader_plan is NOT updated with the new epoch
            bo.backoff(pingcap::kv::boRegionMiss, pingcap::Exception(e.message(), e.code()));
        }
    }
}
```

The `reader_plan` is a const reference captured once before the loop. Every retry uses the same stale epoch.

#### Issue 2 (Rust layer — PRIMARY): EPOCH_NOT_MATCH is not retried

`hub-runtime/src/cloud_helper.rs:715-722` — `request_snapshot_from_leader`:

```rust
// NOT_LEADER: properly retried ✅
if delegate_resp.get_region_error().has_not_leader() {
    error!("{} request_snapshot_from_leader failed, not leader", tag);
    pd_client.evict_region_cache(shard_id);
    leader_changed = true;
    last_err = Some(Error::RegionError(delegate_resp.take_region_error()));
    tokio::time::sleep(next_delay).await;
    continue;  // ← retries!
}

// EPOCH_NOT_MATCH: immediately returned ❌
if delegate_resp.get_region_error().has_epoch_not_match() {
    // Return epoch not match error to TiDB to retry.  ← comment is wrong; caller is TiFlash CN
    error!("{} request_snapshot_from_leader failed, epoch not match, {:?}", ...);
    return Err(Error::RegionError(delegate_resp.take_region_error()));  // ← never retries!
}
```

Both `not_leader` and `epoch_not_match` are transient region errors that could be resolved by evicting the stale cache and retrying. Yet `not_leader` retries inside the same backoff loop, while `epoch_not_match` immediately gives up and returns the error to C++.

### Why Previous Tables Passed

Tables `t_1` and `t_2` worked because:
- The region cache was fresh (no region split had occurred yet)
- `shard_ver=68` was still current when their queries ran
- Their queries completed before the `drop table` triggered the region split

Table `t_3` failed because:
- The region split (epoch 68→69) happened between t_2's last query and t_3's first query
- `buildProxyReadTask` resolved t_3's key ranges and got the stale region info (ver=68) from cache
- The Rust layer refused to retry on EPOCH_NOT_MATCH

### The Infinite Loop

```
C++ retry loop           Rust request_snapshot_from_leader      TiKV
──────────────           ─────────────────────────────────      ────
build plan {125,68,1}
    ↓
create reader(68)  ──→  GET /snapshot/125?shard_ver=68  ──→  EPOCH_NOT_MATCH(69)
                         return Err(RegionError)          ←──  epoch {version:69}
    ← RegionException
dropRegion from cache
backoff & retry
    ↓
create reader(68)  ──→  GET /snapshot/125?shard_ver=68  ──→  EPOCH_NOT_MATCH(69)
    ↓                        ↓                                  ↓
  (repeats ~40+ times for ~60 seconds until timeout)
```

---

## Recommended Fix

**Fix in Rust layer (`hub-runtime/src/cloud_helper.rs:715-722`)** — update `shard_ver` from the error response and retry, matching the `not_leader` pattern:

```rust
if delegate_resp.get_region_error().has_epoch_not_match() {
    error!(
        "{} request_snapshot_from_leader failed, epoch not match, {:?}",
        tag,
        delegate_resp.get_region_error()
    );
    // Extract new shard_ver from the error and retry (same pattern as not_leader)
    let epoch_not_match = delegate_resp.get_region_error().get_epoch_not_match();
    if !epoch_not_match.get_current_regions().is_empty() {
        shard_ver = epoch_not_match.get_current_regions()[0]
            .get_region_epoch()
            .get_version();
    }
    pd_client.evict_region_cache(shard_id);
    last_err = Some(Error::RegionError(delegate_resp.take_region_error()));
    tokio::time::sleep(next_delay).await;
    continue;  // retry with updated shard_ver
}
```

### Why fix in Rust rather than C++:

1. **Precedent already exists** — `not_leader` is handled this way in the same function
2. **All needed info is available** — the new `shard_ver` is in the error response
3. **Follows TiKV client conventions** — client libraries handle epoch refresh internally
4. **Minimal change** — only ~5 lines, no C++ API changes needed
5. **Transparent to callers** — C++ retry loop works correctly without changes

## Affected Files

| File | Role |
|------|------|
| `dbms/src/Storages/StorageDisaggregatedColumnar.cpp:763-792` | C++ retry loop — never updates reader_plan |
| `dbms/src/Storages/StorageDisaggregatedColumnar.cpp:605-634` | C++ error handling — drops cache but doesn't update plan |
| `dbms/src/Storages/Columnar/RNProxyReaderPlan.h:30-36` | Plan struct — region_ver is a value type |
| `contrib/tiflash-columnar-hub/hub-runtime/src/cloud_helper.rs:715-722` | **★ PRIMARY BUG** — EPOCH_NOT_MATCH not retried |
| `contrib/tiflash-columnar-hub/hub-runtime/src/cloud_helper.rs:707-713` | Reference — `not_leader` is retried correctly |

## Log Files

- `tests/fullstack-test-next-gen-columnar/log/tiflash-cn0/tiflash.log` — C++ side (plan building, retry loop)
- `tests/fullstack-test-next-gen-columnar/log/tiflash-cn0/tiflash_error.log` — C++ warnings (repeated EPOCH_NOT_MATCH)
- `tests/fullstack-test-next-gen-columnar/log/tiflash-cn0/tiflash_tikv.log` — Rust proxy side (snapshot requests, epoch details)

---

## How Bucket-Read Branch Introduced This Bug

### Key Commits

| Commit | Author | Date | Description |
|--------|--------|------|-------------|
| `61d37c1750` | Ray Yan | May 22 | `*: refactor proxy to hub lib for columnar` — introduced `cloud_helper.rs` and `EPOCH_NOT_MATCH` non-retry (on **both** master and bucket-read) |
| `d20fc65bb4` | yongman | Jun 2 | `snapaccess cache` — wrapped `request_snapshot_from_leader` with `get_or_request_shared_snapshot` caching layer |
| `d0564d766b` | yongman | Jun 9 | `optimize snapaccess cache for buckets read` — refactored cache to `DashMap` groups by `start_ts`; added `clear_shared_snap_access_by_start_ts` FFI |
| `82f11a7bc9` | yongman | Jun 9 | `clear snap access in last owner drop` — changed cache clearing from eager (every context destruction) to lazy (only last owner) |
| `de6a182f67` | yongman | Jun 9 | `avoid inflight request leak snapaccess` — added terminal state to prevent inflight loader leaks after cache clear |

### Latent Bug vs. Exposure

The `EPOCH_NOT_MATCH` non-retry in `request_snapshot_from_leader` has existed since `61d37c1750` (which is on **both** master and bucket-read):

```rust
// hub-runtime/src/cloud_helper.rs:715-722
// Same on master AND bucket-read:
if delegate_resp.get_region_error().has_epoch_not_match() {
    // Return epoch not match error to TiDB to retry.  ← comment is wrong
    return Err(Error::RegionError(delegate_resp.take_region_error()));  // never retries
}
```

Compare with `not_leader` handling in the same backoff loop:
```rust
// NotLeader is properly retried:
if delegate_resp.get_region_error().has_not_leader() {
    pd_client.evict_region_cache(shard_id);
    leader_changed = true;
    tokio::time::sleep(next_delay).await;
    continue;  // retries ✅
}
```

**Why master doesn't reproduce:**

1. **No caching wrapper** — master calls `request_snapshot_from_leader` directly. Bucket-read interposes `get_or_request_shared_snapshot` which adds loader mutex serialization and SnapAccess caching by `(shard_id, shard_ver, start_ts, start_table_id, end_table_id)`.

2. **Different timing characteristics** — the caching layer changes request serialization patterns. Combined with lazy cache clearing (`82f11a7bc9`), stale region cache entries survive longer, making the race window between region split and query execution wider.

3. **Test path not exercised** — the `fullstack-test-next-gen-columnar` test infrastructure may not have been actively run against upstream/master.

### What Master and Bucket-Read Share

| Component | Master | Bucket-Read | Same? |
|-----------|--------|-------------|-------|
| `request_snapshot_from_leader` EPOCH_NOT_MATCH handling | `return Err` | `return Err` | ✅ Identical |
| C++ `createColumnarReaderWithBackoff` retry loop | const `reader_plan`, same stall epoch | const `reader_plan`, same stale epoch | ✅ Identical |
| C++ `createProxyColumnarReader` epoch handling | drops region cache, throws | drops region cache, throws | ✅ Identical |

### What Bucket-Read Added

| Component | Master | Bucket-Read |
|-----------|--------|-------------|
| `make_columnar_reader` → snapshot request | Direct call to `request_snapshot_from_leader` | Via `get_or_request_shared_snapshot` caching wrapper |
| SnapAccess caching | None | `SharedSnapAccessCache` by `(shard_id, shard_ver, start_ts, start_table_id, end_table_id)` |
| Cache clearing | N/A | Lazy (only on last owner drop via `StartTsClearRegistry`) |
| Loader serialization | None | Per-key `Mutex` via `get_loader()` |

### Conclusion

The **root cause** (EPOCH_NOT_MATCH non-retry in `request_snapshot_from_leader`) was introduced in `61d37c1750`, which is on both branches. The bucket-read **exposure** was caused by `d20fc65bb4` and `d0564d766b` which added the `get_or_request_shared_snapshot` caching layer, changing the timing characteristics enough to turn a latent race into a reproducible failure.

---

## Reproduction

```bash
cd tests/fullstack-test-next-gen-columnar
./compose.sh up -d
# wait for cluster ready
./compose.sh exec -T tiflash-cn0 bash -c \
  'cd /tests && ENABLE_NEXT_GEN=true verbose=true ./run-test.sh fullstack-test2/clustered_index/ddl.test'
```

dbms/src/Storages/StorageDisaggregatedColumnar.cpp
```diff
@@ -65,8 +67,264 @@ namespace ErrorCodes
...
+    for (const auto & [table_id, ranges] : physical_table_ranges)
+    {
+        for (const auto & range : ranges)
+        {
+            String current_start = range.start_key;
+            bool current_range_split = false;
+            for (const auto & bucket_key : bucket_keys)
+            {
+                const auto decoded_bucket_key
+                    = RecordKVFormat::decodeTiKVKey(TiKVKey(bucket_key.data(), bucket_key.size()));
```
JaySon-Huang: This optional bucket split path decodes every bucket key before checking whether the boundary is empty or in range. The raftstore bucket metadata includes the region start/end keys, and the first/last region can use an empty boundary; if that region also has internal bucket keys, this will throw from `decodeTiKVKey` and fail the query instead of falling back to region-level reading. Please skip empty encoded boundaries before decoding, and treat invalid bucket metadata as a safe no-bucket-split fallback.

Source: rule/JAYSONHUANG-RC-003 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

contrib/tiflash-columnar-hub/hub-runtime/src/cloud_helper.rs
```diff
@@ -171,11 +191,35 @@ impl PdClientWithCache {
...
+    pub fn get_region_bucket_keys(&self, region_id: u64, region_ver: u64) -> Vec<Vec<u8>> {
+        if let Some(bucket_entry) = self.region_bucket_cache.get(&region_id) {
+            match bucket_entry.region_ver.cmp(&region_ver) {
+                std::cmp::Ordering::Equal => return bucket_entry.keys.clone(),
+                std::cmp::Ordering::Greater => return Vec::new(),
+                std::cmp::Ordering::Less => {}
+            }
+        }
+
+        let Some(bucket_stat) = self.pd_client.get_buckets(region_id) else {
```
JaySon-Huang: This bucket metadata lookup is on an optimization path, but it calls the synchronous `PdClient::get_buckets`; the `RpcClient` implementation unwraps `get_buckets_async`, so a transient PD/header error can panic through the new `extern "C"` bucket-key FFI instead of just disabling bucket parallelism. Please make this path fallible and return empty keys on error after logging/evicting the cache.

Source: rule/XUHUAIYU-PR-003 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang: `pd_client.get_buckets` call unwraps on get_buckets_async does not related to the code change in this PR. But we'd better handle pd connection error better in cloud-storage-engine pd-client to avoid panic when connection to pd is unstable.

dbms/src/Storages/StorageDisaggregatedColumnar.cpp
```diff
@@ -642,157 +1215,137 @@ std::vector<RNProxyReadTaskPtr> RNProxyReadTask::buildProxyReadTask(
...
-                {
-                    std::lock_guard lock(output_lock);
-                    all_readers.push_back(reader_ptr);
-                }
-            });
-    }
+        table_scan.keepOrder(),
+        enable_bucket_parallel,
+        planned_reader_num,
+        total_max_reader_num);
```
JaySon-Huang: Merge it into one logging

```suggestion
LOG_INFO(
log,
        "read task built, region_num={}, table_num={}, num_streams={}, keep_order={}, enable_bucket_parallel={}, total_split_bucket_num={}, planned_reader_num={}, "
        "max_reader_num={}",
region_num,
physical_table_num,
num_streams,
        table_scan.keepOrder(),
        enable_bucket_parallel,
        total_split_bucket_num,
        planned_reader_num,
        total_max_reader_num);
```

Lloyd-Pottiger (approved):

JaySon-Huang (approved):

PR #10218, opened 2025-06-04
LAC: add keyspace_id in LAC

guo-shaoge:
### What problem does this PR solve?

Issue Number: close #10220

Problem Summary: add keyspace_id argument for all resource control related function.

### What is changed and how it works?

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory
- [ ] Breaking backward compatibility

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes
- [ ] Contains variable changes
- [ ] Contains experimental features
- [ ] Changes MySQL compatibility

### Release note

```release-note
None
```

PR #10929, opened 2026-06-25
LAC: fallback to legacy group when pd not support keyspace

### LINKED ISSUE #10939: columnar cannot find resource group:  default(keyspace=4294967295)
## Bug Report

There is compatible issue when run tiflash with pd & tidb with no keyspace LAC support.

yongman:

Issue Number: close #10939 

Problem Summary:
There is compatible issue when the tiflash runs with pd without keyspace resource group.
Ref: https://github.com/pingcap/tiflash/pull/10218

- This change teaches `LocalAdmissionController` to keep working when TiFlash runs with keyspace support enabled but PD still serves only legacy resource groups. It does that by:
  - falling back from a keyspace-scoped `GetResourceGroup` lookup to the legacy nullspace lookup,
  - synthesizing a local `"default"` group when PD does not persist one,
  - watching both the keyspace and legacy etcd paths so cached legacy groups still receive updates.
 
```commit-message
```

- `ResourceGroup` now has an `enable_gac` flag. Reserved local defaults are created with `enable_gac=false`, so normal token acquisition/reporting skips them. `buildRequestInfoIfNecessary()` returns no request when that flag is disabled. (`dbms/src/Flash/ResourceControl/LocalAdmissionController.h:80-90`, `dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:77-117`)
- `LocalAdmissionController` now resolves lookups through `findResourceGroupWithCompatWithoutLock()`: it first checks the exact `(keyspace_id, name)` cache entry, then falls back to `(NullspaceID, name)` for keyspace requests. All normal runtime callers (`getPriority`, `consumeResource`, `estWaitDuraMS`) now use that compatibility lookup. (`dbms/src/Flash/ResourceControl/LocalAdmissionController.h:566-599`)
- `warmupResourceGroupInfoCache()` changed from a single keyspace-scoped PD lookup into a compatibility sequence:
  - try `GetResourceGroup` with `keyspace_id`,
  - on `resp.has_error()`, retry without `keyspace_id`,
  - if that legacy lookup succeeds, cache the result under the resolved keyspace from the returned protobuf,
  - if both fail and the name is `"default"`, synthesize a reserved local default group.
  Returned protobufs are normalized so missing `keyspace_id` becomes `NullspaceID`. (`dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:334-420`)
- GAC request building now omits `keyspace_id` entirely for nullspace groups, which keeps fallback traffic compatible with legacy PD semantics. Response handling and metrics also use the cached group’s actual keyspace instead of the original request keyspace. (`dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:503-587`, `dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:694-840`)
- When `with_keyspace=true`, the constructor now starts two watch loops, one for `resource_group/keyspace/settings` and one for `resource_group/settings`. To support two concurrent watchers, the code replaced the single stored `grpc::ClientContext` with an `active_watch_gac_grpc_contexts` set so `stop()` can cancel every active watch stream. (`dbms/src/Flash/ResourceControl/LocalAdmissionController.h:367-388`, `dbms/src/Flash/ResourceControl/LocalAdmissionController.h:724-728`, `dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:843-936`, `dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp:1059-1099`)
- The new gtests cover:
  - reserved default fallback,
  - exact keyspace hit,
  - fallback to a legacy cached group,
  - normalization of legacy watch updates to nullspace.
  (`dbms/src/Flash/ResourceControl/tests/gtest_local_admission_controller.cpp:59-238`)

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

### Manual Test
1. Deploy cluster with legacy pd/tidb from pd-cse/tidb-cse.
2. Run tpcc workload with tiup bench.
3. There is no exception happens after this compatible fix.

## Summary

* **New Features**
  * Improved resource-group cache warmup with compatibility-aware keyspace normalization and backend-mode detection.
  * Enhanced GAC request/response handling to preserve legacy `keyspace_id` behavior and normalize token-bucket bookkeeping.
  * Updated watch-path behavior and lifecycle management for more reliable background processing.

* **Bug Fixes**
  * PUT updates for groups not yet warmed into the local cache are now ignored; warmed groups now recompute `max_ru_per_sec` and trigger refill when available.
  * Low-token reporting now uses the resolved cached group key rather than the originally requested one.
  * GAC watch shutdown now cancels the active gRPC context.

* **Tests**
  * Added/extended gtests covering legacy fallback/promotion, keyspace-scoped caching, and error propagation without compatibility fallback.

### DIFF
--- a/dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
+++ b/dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
@@ -15,7 +15,10 @@
 #include <Flash/ResourceControl/LocalAdmissionController.h>
 #include <etcd/rpc.pb.h>
 
+#include <algorithm>
+#include <cctype>
 #include <magic_enum.hpp>
+#include <string_view>
 
 namespace DB
 {
@@ -33,6 +36,20 @@ resource_manager::ResourceGroup normalizeResourceGroupKeyspace(
 KeyspaceID resolveResourceGroupKeyspace(const resource_manager::ResourceGroup & group_pb)
 {
     return group_pb.has_keyspace_id() ? group_pb.keyspace_id().value() : NullspaceID;
+}
+
+bool containsIgnoreCase(std::string_view haystack, std::string_view needle)
+{
+    return std::search(
+               haystack.begin(),
+               haystack.end(),
+               needle.begin(),
+               needle.end(),
+               [](char lhs, char rhs) {
+                   return std::tolower(static_cast<unsigned char>(lhs))
+                       == std::tolower(static_cast<unsigned char>(rhs));
+               })
+        != haystack.end();
 }
 } // namespace
 
@@ -326,10 +343,19 @@ resource_manager::ResourceGroup LocalAdmissionController::buildReservedDefaultRe
     return group_pb;
 }
 
-void LocalAdmissionController::addReservedDefaultResourceGroup(const KeyspaceID & keyspace_id)
+void LocalAdmissionController::addReservedDefaultResourceGroup()
 {
-    static_cast<void>(keyspace_id);
     addResourceGroup(NullspaceID, buildReservedDefaultResourceGroup(NullspaceID), false);
+}
+
+bool LocalAdmissionController::shouldFallbackToLegacyLookup(const resource_manager::GetResourceGroupResponse & resp)
+{
+    if (!resp.has_error())
+        return false;
+
+    const auto & err_msg = resp.error().message();
+    return containsIgnoreCase(err_msg, "resource group")
+        && (containsIgnoreCase(err_msg, "not found") || containsIgnoreCase(err_msg, "does not exist"));
 }
 
 void LocalAdmissionController::warmupResourceGroupInfoCache(const KeyspaceID & keyspace_id, const std::string & name)
@@ -342,7 +368,7 @@ void LocalAdmissionController::warmupResourceGroupInfoCache(const KeyspaceID & k
 
     {
         std::lock_guard lock(mu);
-        if (findResourceGroupWithCompatWithoutLock(keyspace_id, name).group != nullptr)
+        if (hasExactResourceGroupWithoutLock(keyspace_id, name))
             return;
     }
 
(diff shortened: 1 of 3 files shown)


dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -292,6 +329,35 @@ LACRUConsumptionDeltaInfo ResourceGroup::updateRUConsumptionDeltaInfoWithoutLock
...
+    addResourceGroup(NullspaceID, buildReservedDefaultResourceGroup(NullspaceID), false);
+}
+
+bool LocalAdmissionController::shouldFallbackToLegacyLookup(const resource_manager::GetResourceGroupResponse & resp)
+{
+    if (!resp.has_error())
+        return false;
+
+    const auto & err_msg = resp.error().message();
+    return containsIgnoreCase(err_msg, "resource group")
```
windtalker: This is very unstable, can we refine the `GetResourceGroupResponse` so we can check the error code instead of the error message here?

dbms/src/Flash/ResourceControl/LocalAdmissionController.h
```diff
@@ -559,10 +619,22 @@ class LocalAdmissionController final : private boost::noncopyable
 
         auto iter = keyspace_resource_groups.find({keyspace_id, new_group_pb.name()});
         if (iter != keyspace_resource_groups.end())
-            return;
+        {
```
windtalker: can you add some comments here to describ the logic inside `if (iter != keyspace_resource_groups.end())`

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -323,16 +391,71 @@ void LocalAdmissionController::warmupResourceGroupInfoCache(const KeyspaceID & k
             getCurrentExceptionMessage(false));
     }
 
-    RUNTIME_CHECK_MSG(
-        !resp.has_error(),
-        "warmupResourceGroupInfoCache: {}(keyspace={}) failed: {}",
-        name,
-        keyspace_id,
-        resp.error().message());
+    if (resp.has_error())
```
windtalker, line 345: I was wondering, isn't keyspace support a cluster-level configuration for PD? If a cluster's PD doesn't support keyspace, it likely won't support it for a long time. If that's the case, do we really need to handle the fallback at the RPC level?

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
yongman: There is no error code in proto GetResourceGroupResponse.
```
message GetResourceGroupResponse {
  Error error = 1;
  ResourceGroup group = 2;
}

message Error {
  string message = 1;
}
```

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
yongman: Here it means the pd dose not support keyspace scope resource group (not the keyspace itself).

JaySon-Huang: In the "starter" tier, currently the pd-cse is running with keyspace support but without keyspace scope resource group. And we don't actually know when pd-cse will be upgraded to pd. But we plan to upgrade the tiflash columnar binary on the "starter" tier to unify the codebase in late July, so tiflash need to adapt to the pd-cse.

dbms/src/Flash/ResourceControl/LocalAdmissionController.cpp
```diff
@@ -1027,14 +1081,67 @@ std::unique_ptr<LocalAdmissionController> LocalAdmissionController::global_insta
...
+
+    if (resource_group_backend_mode != ResourceGroupBackendMode::Unknown)
+    {
+        LOG_WARNING(
+            log,
+            "keep resource-group backend mode {} and ignore newly detected mode {}",
+            magic_enum::enum_name(resource_group_backend_mode),
+            magic_enum::enum_name(detected_mode));
+        return;
+    }
```
JaySon-Huang, line 1122: So if we update tiflash-compute columnar version and make it adapt to the existing LegacyGlobal mode. After pd-cse is upgraded to pd in the future, do we need to restart the tiflash-compute node to migrate its mode access to pd to resource group requests with keyspace?

yongman: Yes, it is by design. The tiflash reource group will keeps in legacy mode after pd upgrade to nextgen version. After pd upgrading done we restart tiflash to make keyspace scoped resource group take effects.

JaySon-Huang: OK

JaySon-Huang (approved):

Lloyd-Pottiger (approved):

PR #10946, opened 2026-07-01
Support TiDB NullEQ pushdown

### LINKED ISSUE #5102: Call For Participation: implement `NullEq` function push down
## Feature Request

We'd like to support `NullEq` function push down.

At least one integration test and several unit tests.

windtalker:

Issue Number: close #5102

Problem Summary:

TiDB can push down NullEQ expressions, but TiFlash does not fully recognize and execute the pushed-down NullEQ semantics yet, including rough-set filtering in DeltaMerge.

```commit-message
support nulleq in tiflash
DeltaMerge: fix rough-set semantics for NullEQ
```

This PR adds TiDB NullEQ function support in TiFlash, wires NullEQ pushdown handling, adds execution tests, and fixes DeltaMerge rough-set/min-max index semantics for NullEQ.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support TiDB NullEQ pushdown in TiFlash.
```

## Summary by CodeRabbit

* **New Features**
  * Added TiDB null-safe equality (`<=>`) support end-to-end: `tidbNullEQ` function and DeltaMerge `null_equal` rough filtering, including correct query parsing for `NULL` cases.

* **Bug Fixes**
  * Improved `NULL`-safe equality evaluation in MinMaxIndex for nullable columns, including mixed `NULL`/non-`NULL` packs.
  * Fixed nullable `String` min/max boundary extraction used by these checks.

* **Tests**
  * Added unit, parser, MinMaxIndex, and fullstack coverage for `tidbNullEQ` / `null_equal`, including negation and collator handling.


yy782:
Hi @windtalker and maintainers,
I noticed that PR #10946 also addresses the same issue (#5102). I was assigned to this issue and have been working on it independently. I've just submitted my implementation in this PR#10954.
Since we both have PRs for the same issue, I'd like to understand how we should resolve this overlap. I'm open to collaboration, but I also want to ensure that my work on this assigned issue is recognized.
Could you please advise on the best way forward? Should we compare implementations and consolidate, or are there other preferences from the maintainers?
Thanks!

JaySon-Huang:
Hi @yy782 , thanks a lot for your interest in TiFlash and for working on this issue.

Sorry that we did not notice earlier that this issue had been assigned to you, while another implementation was already ongoing on our side. This caused two overlapping PRs for the same issue, and I apologize for the confusion.

After reviewing both PRs, I would prefer to use #10946 as the base implementation. The main reason is that #10946 introduces a dedicated `tidbNullEQ` function and handles the NULL-safe equality semantics directly in the function implementation, including nullable inputs, NULL-only columns, nullable constants, and collator forwarding. It also adds a dedicated DeltaMerge `MinMaxIndex::checkNullEqual` path, which gives more precise rough-set behavior for nullable packs and is easier to maintain than reusing normal `Equal` semantics for `NullEQ`.

That said, your PR #10954 has very valuable test coverage, especially the richer type-combination tests and the fullstack end-to-end test. I think those tests would be a good supplement to #10946.

So my suggestion is to keep #10946 as the core implementation, and incorporate the additional test coverage from #10954. We will make sure your contribution is properly recognized when doing that. Does this sound reasonable to you?

yy782:
Thanks @JaySon-Huang! I'm happy to collaborate. Please let me know how you'd like me to help integrate the tests.

JaySon-Huang:
Hi @yy782,

Thanks for being willing to collaborate.

I think a good next step would be to use #10946 as the base implementation. You can first run:

```bash
gh pr checkout 10946
```

This will check out `windtalker:support_nulleq` locally. Then, based on the current commit of that branch, please port the useful test coverage from your PR, mainly from these files:

- `dbms/src/Functions/tests/gtest_nulleq.cpp`
- `dbms/src/Storages/DeltaMerge/Index/tests/gtest_dm_minmax_index.cpp`
- `tests/fullstack-test/expr/nulleq.test`

A few notes:

- In #10946, the registered function name is `tidbNullEQ`, not `nullEq`, so the function tests should call `tidbNullEQ`.
- Please try to keep your broader type-combination coverage, since that is a valuable supplement to #10946.
- Please make sure the related unit and fullstack tests can pass locally after the changes.

After that, you can either force-push the updated changes to your current `feature/null-eq-pushdown` branch, or create a new branch and open a new PR based on #10946.

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,214 @@
...
+            }
+        }
+
+        block.getByPosition(result).column = std::move(eq_mutable);
+    }
+
+private:
+    TiDB::TiDBCollatorPtr collator = nullptr;
+    std::shared_ptr<FunctionEquals> equals_function = std::make_shared<FunctionEquals>();
+};
```
JaySon-Huang, line 226: Seems that we can remove the `collator` from FunctionTiDBNullEQ
```suggestion
private:
    std::shared_ptr<FunctionEquals> equals_function = std::make_shared<FunctionEquals>();
};
```

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,214 @@
...
+    size_t getNumberOfArguments() const override { return 2; }
+
+    bool useDefaultImplementationForNulls() const override { return false; }
+    bool useDefaultImplementationForConstants() const override { return true; }
+
+    void setCollator(const TiDB::TiDBCollatorPtr & collator_) override
+    {
+        collator = collator_;
+        equals_function->setCollator(collator_);
+    }
```
JaySon-Huang:
```suggestion
    void setCollator(const TiDB::TiDBCollatorPtr & collator_) override
    {
        equals_function->setCollator(collator_);
    }
```

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,214 @@
...
 namespace DB
 {
+namespace ErrorCodes
+{
+extern const int NUMBER_OF_ARGUMENTS_DOESNT_MATCH;
+extern const int ILLEGAL_COLUMN;
+extern const int LOGICAL_ERROR;
+} // namespace ErrorCodes
+
+class FunctionTiDBNullEQ : public IFunction
```
JaySon-Huang, line 40:
```suggestion
/// TiDB null-safe equal (`<=>`, mapped as `tidbNullEQ`).
///
/// Difference from `equals`:
/// - `equals` only compares nested values and does not interpret NULL markers. For nullable
///   inputs, rows with NULL may still get 0/1 from placeholder nested data instead of SQL `=`
///   semantics (where `NULL = x` is unknown/NULL).
/// - `tidbNullEQ` applies TiDB `<=>` rules: `NULL <=> NULL` => 1, `NULL <=> x` => 0, and defers
///   to `equals` only when neither side is NULL. The result is always non-nullable UInt8.
class FunctionTiDBNullEQ : public IFunction
```

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -490,6 +490,134 @@ template RSResults MinMaxIndex::checkCmp<RoughCheck::CheckGreaterEqual>(
     const Field & value,
     const DataTypePtr & type);
 
+template <typename T>
+RSResults MinMaxIndex::checkNullableNullEqualImpl(
```
JaySon-Huang, line 499:
```suggestion
/// Rough-check `col <=> value` for a nullable column using per-pack min/max.
/// `value` must be non-null; `col <=> NULL` is handled by `checkIsNull` before reaching here.
///
/// Unlike `checkNullableCmp` (SQL `=`), null-safe equal treats `NULL <=> value` as definitely false
/// rather than unknown, so we never tag results with `has_null` (no `*Null` RSResult variants).
template <typename T>
RSResults MinMaxIndex::checkNullableNullEqualImpl(
```

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -490,6 +490,134 @@ template RSResults MinMaxIndex::checkCmp<RoughCheck::CheckGreaterEqual>(
...
+
+        auto min = minmaxes_data[i * 2];
+        auto max = minmaxes_data[i * 2 + 1];
+        auto value_result = RoughCheck::CheckEqual::check<T>(value, type, min, max);
+        if (has_null_marks[i] && value_result == RSResult::All)
+            results[i - start_pack] = RSResult::Some;
+        else
+            results[i - start_pack] = value_result;
+    }
+    return results;
```
JaySon-Huang, line 531:
```suggestion
    /// Default to Some: conservative when min/max cannot prove a tighter bound.
    RSResults results(pack_count, RSResult::Some);
    const auto & minmaxes_data = toColumnVectorData<T>(column_nullable.getNestedColumnPtr());
    for (size_t i = start_pack; i < start_pack + pack_count; ++i)
    {
        /// Min slot is NULL on indexes built before v6.4.0. For a pure-NULL pack,
        /// `NULL <=> value` is false for every row, so the pack can be skipped.
        if (details::minIsNull(null_map, i))
        {
            if (has_null_marks[i] && !has_value_marks[i])
                results[i - start_pack] = RSResult::None;
            continue;
        }

        auto min = minmaxes_data[i * 2];
        auto max = minmaxes_data[i * 2 + 1];
        auto value_result = RoughCheck::CheckEqual::check<T>(value, type, min, max);
        /// Non-null values may all equal `value`, but NULL rows still make `col <=> value` false.
        /// Downgrade All => Some so the pack is read and filtered row by row.
        if (has_null_marks[i] && value_result == RSResult::All)
            results[i - start_pack] = RSResult::Some;
        else
            results[i - start_pack] = value_result;
    }
    return results;
```

JaySon-Huang (approved):
LGTM

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,217 @@
...
+            }
+
+            const ColumnPtr & other_col = left_col->onlyNull() ? right_col : left_col;
+            if (other_col->isColumnNullable())
+            {
+                const auto & other_nullmap = assert_cast<const ColumnNullable &>(*other_col).getNullMapData();
+                auto res_col = ColumnUInt8::create();
+                auto & res_data = res_col->getData();
+                res_data.assign(other_nullmap.begin(), other_nullmap.end());
+                block.getByPosition(result).column = std::move(res_col);
```
gengliqi, line 109:
```suggestion
               block.getByPosition(result).column = assert_cast<const ColumnNullable &>(*other_col).getNullMapColumnPtr();
```

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,217 @@
...
+            }
+            else
+            {
+                block.getByPosition(result).column = ColumnUInt8::create(rows, 0);
```
gengliqi:
```suggestion
                block.getByPosition(result).column = DataTypeUInt8().createColumnConst(rows, static_cast<UInt64>(0));
```

dbms/src/Functions/FunctionsComparison.cpp
```diff
@@ -14,12 +14,217 @@
...
+                rows,
+                right_col->size());
+
+        /// Fast path for always-NULL columns (Nullable(Nothing)).
+        /// `NULL <=> x` equals to `isNull(x)`; `NULL <=> NULL` is always 1.
+        if (left_col->onlyNull() || right_col->onlyNull())
+        {
+            if (left_col->onlyNull() && right_col->onlyNull())
+            {
+                block.getByPosition(result).column = ColumnUInt8::create(rows, 1);
```
gengliqi:
```suggestion
                block.getByPosition(result).column = DataTypeUInt8().createColumnConst(rows, static_cast<UInt64>(1));
```

dbms/src/Functions/FunctionsComparison.cpp
windtalker: This is a little risk since nullmap maybe modified without check its ref_count

dbms/src/Storages/DeltaMerge/Filter/NullEqual.h
```diff
@@ -0,0 +1,58 @@
...
+    String name() override { return "null_equal"; }
+
+    RSResults roughCheck(size_t start_pack, size_t pack_count, const RSCheckParam & param) override
+    {
+        auto rs_index = getRSIndex(param, attr);
+        return rs_index ? rs_index->minmax->checkNullEqual(start_pack, pack_count, value, rs_index->type)
+                        : RSResults(pack_count, RSResult::Some);
+    }
+
+    ColumnRangePtr buildSets(const google::protobuf::RepeatedPtrField<tipb::ColumnarIndexInfo> & index_infos) override
```
gengliqi: The inverted index will ignore null values. So maybe it's better to return `UnsupportedColumnRange::create()`.

gengliqi (approved):

PR #10978, opened 2026-07-14
executor: Improve HashAgg string collation key caching

### LINKED ISSUE #10979: Optimize HashAgg by caching collation sort keys for repeated string keys
## Enhancement
For string `GROUP BY` keys using a non-trivial collation such as
`utf8mb4_general_ci`, HashAgg currently generates a collation sort key for
every input row.

When the input has many repeated raw string values, the same sort key is
generated repeatedly. For example:

```sql
SELECT l_shipmode, COUNT(*)
FROM lineitem
GROUP BY l_shipmode;
```
If l_shipmode uses utf8mb4_general_ci and has a low NDV, sort-key
generation can consume a significant portion of the aggregation time.

ChangRui-Ryan:

Issue Number: close #10979 

## Problem Summary

For string `GROUP BY` keys using a case-insensitive collation such as `utf8mb4_general_ci`, HashAgg needs to generate a collation sort key before probing or inserting into the hash table.

Previously, the sort key was generated for every input row. When the grouped column has a low number of distinct raw values, the same raw string is converted to the same sort key repeatedly, causing unnecessary collation processing and memory writes.

For example:

```sql
SELECT l_shipmode, COUNT(*)
FROM lineitem
GROUP BY l_shipmode;
```

`l_shipmode` usually has a very low NDV. If it uses `utf8mb4_general_ci`, repeatedly generating sort keys for millions of rows can consume a significant portion of the HashAgg execution time.

At the same time, unconditionally caching sort keys is not suitable for high-NDV inputs because maintaining the cache may introduce additional hash lookups and memory usage without enough cache hits. Therefore, the optimization needs to provide substantial benefits for low-NDV workloads while keeping the high-NDV overhead small and bounded.

This PR adds an always-on collation sort-key cache to the batch string-key processing path of HashAgg.

For each input Block, the cache maps a raw string value to its generated collation sort key:

```text
raw string -> collation sort key
```

When processing a row:

1. Look up the raw string in the cache.
2. If it is found, reuse the cached sort key.
3. If it is not found, generate the sort key once and insert it into the cache.
4. Use the resulting sort key for the existing hash-table lookup or insertion.

The cache is enabled only when all of the following conditions are satisfied:

- HashAgg uses the batch string-key processing path.
- The string key uses a supported case-insensitive collation.
- The current processing range contains at least 1,024 rows.

To bound the cost for high-NDV inputs, the maximum number of cached distinct values is calculated from the number of rows in the current Block:

```text
cache distinct limit = rows in the Block / 32
```

For a typical maximum runtime Block containing 65,536 rows, the cache can hold at most 2,048 distinct raw values.

If a cache miss occurs after the distinct-value limit has been reached, the current Block falls back to direct sort-key generation. The fallback state is also recorded in the worker's `AggregatedDataVariants`, so subsequent Blocks processed by the same aggregation worker do not attempt to initialize the cache again. This worker-level once-fallback behavior prevents high-NDV workloads from repeatedly paying the cache construction and lookup overhead for every Block.

The cache itself remains local to one Block. It only stores references to raw strings owned by that Block, so no input-column memory needs to be retained across Blocks. Only the fallback decision is preserved across Blocks.

The optimization does not change collation or grouping semantics. Hash-table keys are still the sort keys generated by the existing collator; the change only avoids regenerating an identical sort key for repeated raw string values.

The benchmark uses 128 Blocks with 65,536 rows per Block:

Workload | Baseline | With cache | Change
-- | -- | -- | --
NDV 64 | 27.16 M rows/s | 73.41 M rows/s | +170.3%
NDV 2,048 | 26.05 M rows/s | 55.71 M rows/s | +113.9%
NDV 2,049, fallback | 26.95 M rows/s | 26.85 M rows/s | -0.4%
NDV 65,536, fallback | 24.17 M rows/s | 23.99 M rows/s | -0.7%
First 64 Blocks low NDV, remaining Blocks high NDV | 25.40 M rows/s | 35.75 M rows/s | +40.8%

The results show substantial improvements for low-NDV and mixed workloads, while the worker-level fallback keeps the high-NDV overhead below 1% in the benchmark.

```commit-message

```

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Performance**
  * Improved aggregation performance for case-insensitive string data by reusing collation sort keys during batch processing.
  * Automatically switches to uncached processing when distinct values exceed the cache threshold, helping maintain predictable performance.

* **Bug Fixes**
  * Improved consistency when processing subsequent batches after cache fallback.

* **Tests**
  * Added coverage for collation-sensitive grouping, varying distinct-value counts, and fallback behavior.
  * Added benchmarks for cached and uncached aggregation scenarios.

### DIFF
--- a/dbms/src/Common/ColumnsHashing.h
+++ b/dbms/src/Common/ColumnsHashing.h
@@ -19,6 +19,7 @@
 #include <Columns/ColumnString.h>
 #include <Common/Arena.h>
 #include <Common/ColumnsHashingImpl.h>
+#include <Common/HashTable/HashMap.h>
 #include <Common/HashTable/HashTable.h>
 #include <Common/HashTable/HashTableKeyHolder.h>
 #include <Common/assert_cast.h>
@@ -38,6 +39,9 @@ extern const int LOGICAL_ERROR;
 
 namespace ColumnsHashing
 {
+static constexpr size_t min_rows_to_use_collation_sort_key_cache = 1024;
+static constexpr size_t max_collation_sort_key_cache_distinct_ratio = 32;
+
 /// For the case when there is one numeric key.
 /// UInt8/16/32/64 for any type with corresponding bit width.
 template <typename Value, typename Mapped, typename FieldType, bool use_cache = true>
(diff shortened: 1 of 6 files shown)


dbms/src/Common/ColumnsHashing.h
```diff
@@ -95,6 +99,69 @@ class KeyStringBatchHandlerBase
...
+    HashMap<StringRef, size_t, StringRefHash> raw_to_sort_key_index;
+    std::vector<String> cached_sort_keys{};
+
+    template <typename DerivedCollator>
+    StringRef getSortKeyWithCache(
+        const DerivedCollator & collator,
+        const StringRef & raw_key,
+        std::string & sort_key_container)
+    {
+        if (!collation_sort_key_cache_active)
```
windtalker: add `unlikely` hint for this `if`?

dbms/src/Interpreters/Aggregator.cpp
```diff
@@ -641,6 +641,8 @@ void Aggregator::executeImplInner(
 {
     auto * aggregates_pool = result.aggregates_pool;
     typename Method::State state(agg_process_info.key_columns, key_sizes, collators);
+    if (!result.string_collation_key_cache_fallback)
+        state.initCollationSortKeyCache(agg_process_info.end_row - agg_process_info.start_row);
```
windtalker, line 645: `agg_process_info.end_row - agg_process_info.start_row` is not always represent a whole block, which means the cache is not block level cache, is it expected?

windtalker (commented):
Looks like this cache only works for cases that `group by a single string column`, it does not work for something like `group by int_col, string_col`?

ChangRui-Ryan:
> Looks like this cache only works for cases that `group by a single string column`, it does not work for something like `group by int_col, string_col`?

Yes, this cache currently only applies to `HashMethodString`, which is used for grouping by a single string column. Composite keys such as `(int_col, string_col)` use `HashMethodSerialized` and are not covered by this optimization. We prefer to keep this change scoped.

dbms/src/Interpreters/Aggregator.cpp
ChangRui-Ryan: Good point. In the normal aggregation path, `start_row` is 0 and `end_row` is `block.rows()`, so the cache is initialized with the whole Block size and is reused across all mini-batches processed by the same `executeImplInner()` invocation.

The range can be smaller than a whole Block when aggregation is interrupted by a partial-processing path, such as spilling after a `ResizeException`. In that case, a subsequent `executeOnBlock()` invocation processes the remaining rows and creates a new local cache for that range.

This is expected. The cache is intentionally scoped to the current processing range rather than guaranteed to span the entire input Block. The cache only affects sort-key computation and does not affect correctness. The worker-level fallback state is kept separately in `AggregatedDataVariants::string_collation_key_cache_fallback`, so once fallback happens, later Blocks still skip cache attempts.

windtalker (approved):
lgtm

xzhangxian1008 (approved):

PR #10984, opened 2026-07-16
Columnar Hub: fix engine-addr CLI arg unconditionally overwriting config file value

### LINKED ISSUE #10983: Columnar Hub: engine-addr CLI arg unconditionally overwrites config file value, causing 0.0.0.0 registered as store address to PD

Deploy TiFlash in disaggregated compute mode (`tiflash_compute`) with the following configurations:

**config.toml** (TiFlash main config):
```toml
[flash]
  disaggregated_mode = "tiflash_compute"
  service_addr = "0.0.0.0:3930"

[flash.proxy]
  addr = "0.0.0.0:20170"
  advertise-addr = "tiflash-compute-node.example.com:20170"
  config = "/data0/proxy.toml"
```

Note: `flash.proxy.engine-addr` is NOT explicitly set — it falls back to `flash.service_addr` (`"0.0.0.0:3930"`) when building CLI args for the proxy/Hub.

**proxy.toml** (the proxy/Hub config file):
```toml
[server]
  engine-addr = "tiflash-compute-node.example.com:3930"
  status-addr = "0.0.0.0:20292"
  advertise-status-addr = "tiflash-compute-node.example.com:20292"
```

The store address registered to PD should be the routable FQDN `"tiflash-compute-node.example.com:3930"` because:
1. The Hub's config file (`proxy.toml`) explicitly sets `server.engine-addr` to the FQDN.
2. The CLI argument `--engine-addr=0.0.0.0:3930` from TiFlash (which is a wildcard bind address) should be treated as a fallback, not an override.

PD receives `address:"0.0.0.0:3930"` as the store address:

```
[2026/07/16 15:26:07.144 +08:00] [INFO] [grpc_service.go:858] ["put store ok"]
[store="id:1 address:\"0.0.0.0:3930\" labels:<key:\"engine\"
value:\"tiflash_compute\" > ..."]
```

`0.0.0.0` is the wildcard listen address and is not routable, so PD (and other nodes) cannot communicate with this store correctly.

v9.0.0-beta.2

---

## Root Cause Analysis

There are **two independent implementations** of `overwrite_config_with_cmd_args` that handle the `--engine-addr` CLI argument differently:

**Proxy Server path** (`contrib/tiflash-proxy/proxy_components/proxy_server/src/setup.rs:85-89`) — has `is_empty()` guard:
```rust
if proxy_config.server.engine_addr.is_empty() {
    if let Some(engine_addr) = matches.value_of("engine-addr") {
        proxy_config.server.engine_addr = engine_addr.to_owned();
    }
}
```

**Columnar Hub path** (`contrib/tiflash-columnar-hub/hub-runtime/src/run.rs:376-378`) — NO `is_empty()` guard:
```rust
if let Some(engine_addr) = matches.value_of("engine-addr") {
    config.server.engine_addr = engine_addr.to_owned();
}
```

The full data flow of the bug:

1. TiFlash reads `config.toml` — `flash.proxy.engine-addr` is not set, so `ProxyStateMachine.h` falls back to `flash.service_addr = "0.0.0.0:3930"`
2. TiFlash passes `--engine-addr=0.0.0.0:3930` as a CLI arg to the Columnar Hub
3. Columnar Hub loads `proxy.toml` → `config.server.engine_addr = "FQDN:3930"` (correct)
4. Columnar Hub calls `overwrite_config_with_cmd_args` → `config.server.engine_addr = "0.0.0.0:3930"` (unconditionally overwritten!)
5. `build_store()` → `resolve_engine_addr()` returns `"0.0.0.0:3930"`
6. Hub registers store to PD with the wrong address

## Suggested Fix

Add the `is_empty()` guard in the Hub's `overwrite_config_with_cmd_args` to match the Proxy Server behavior — only accept the CLI `--engine-addr` when the config file has not already provided a value:

In `contrib/tiflash-columnar-hub/hub-runtime/src/run.rs`:
```rust
if config.server.engine_addr.is_empty() {
    if let Some(engine_addr) = matches.value_of("engine-addr") {
        config.server.engine_addr = engine_addr.to_owned();
    }
}
```

**Note:** `advertise-engine-addr` should remain an unconditional override (as it currently is in both implementations), since its semantics are explicitly "override whatever was configured."

JaySon-Huang:

Issue Number: close #10983

Problem Summary:

When TiFlash is deployed in disaggregated compute mode (`tiflash_compute`) with the Columnar Hub, the store address registered to PD is `"0.0.0.0:3930"` (the wildcard bind address from `flash.service_addr`) instead of the routable FQDN configured in `proxy.toml`'s `server.engine-addr`.

Root cause: The Columnar Hub's `overwrite_config_with_cmd_args` unconditionally applies the `--engine-addr` CLI argument (which TiFlash derives from `flash.service_addr` as a fallback when `flash.proxy.engine-addr` is not explicitly set). This overwrites the correct FQDN that was already loaded from the config file. The Proxy Server path (`setup.rs`) correctly guards against this with an `is_empty()` check, but the Hub path was missing the same guard.

```commit-message
Columnar Hub: fix engine-addr CLI arg unconditionally overwriting config file value

In `overwrite_config_with_cmd_args`, add an `is_empty()` guard before
applying the `--engine-addr` CLI argument. This matches the existing
behavior in the Proxy Server path (setup.rs). The config file value now
takes precedence, while `--advertise-engine-addr` continues to
unconditionally override.
```

The fix follows the exact same pattern already used in `contrib/tiflash-proxy/proxy_components/proxy_server/src/setup.rs:85-89`:

```rust
// Before (Hub path):
if let Some(engine_addr) = matches.value_of("engine-addr") {
    config.server.engine_addr = engine_addr.to_owned();
}

// After (matches Proxy Server path):
if config.server.engine_addr.is_empty() {
    if let Some(engine_addr) = matches.value_of("engine-addr") {
        config.server.engine_addr = engine_addr.to_owned();
    }
}
```

### Check List

Tests

- [x] Unit test (existing tests continue to pass; `ConfigFile::default()` has empty `engine_addr`, so CLI arg still applies in tests)
- [ ] Integration test
- [x] Manual test (verified via `investigate.md` analysis that the config file value now survives CLI override)
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix the issue that Columnar Hub registers 0.0.0.0 as the store address to PD when
`flash.proxy.engine-addr` is not explicitly set, ignoring the FQDN in proxy.toml.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the engine address configured in the configuration file instead of always replacing it with the command-line value.
  * Continued to support command-line overrides when no engine address is configured.

### DIFF
--- a/contrib/tiflash-columnar-hub/hub-runtime/src/run.rs
+++ b/contrib/tiflash-columnar-hub/hub-runtime/src/run.rs
@@ -373,9 +373,18 @@ fn overwrite_config_with_cmd_args(config: &mut ConfigFile, matches: &clap::ArgMa
     if let Some(engine_git_hash) = matches.value_of("engine-git-hash") {
         config.server.engine_git_hash = engine_git_hash.to_owned();
     }
-    if let Some(engine_addr) = matches.value_of("engine-addr") {
-        config.server.engine_addr = engine_addr.to_owned();
+    // The special case is engine-addr:
+    // 1. If we have set our own engine-addr (from config file), we just ignore
+    //    what TiFlash gives us.
+    // 2. However, if we have not set our own value, we use what TiFlash gives us.
+    if config.server.engine_addr.is_empty() {
+        if let Some(engine_addr) = matches.value_of("engine-addr") {
+            config.server.engine_addr = engine_addr.to_owned();
+        }
     }
+
+    // advertise-engine-addr always overrides, regardless of what is in the
+    // config file. This is the same behavior as the proxy server path.
     if let Some(engine_addr) = matches.value_of("advertise-engine-addr") {
         config.server.engine_addr = engine_addr.to_owned();
     }


JaySon-Huang:
This is for compatibility with the TiFlash configuration generated by tidb-operator v1. tidb-operator v1 is still being used in internal testing; tidb-operator v2 generates `flash.service_addr` in the "FQDN:3930" form, so this issue does not exist there.

Lloyd-Pottiger (approved):

yongman (approved):

Issue #10985, opened 2026-07-16
TiFlash may repeatedly reuse stale read-index cache after resolving locks


windtalker:
### 1. Minimal reproduce step (Required)

1. Run an MPP / remote cop read against TiFlash where the target region has a secondary lock visible in TiFlash local LockCF.
2. TiFlash gets a read index for the region and waits to that index successfully.
3. During `RegionTable::resolveLocksAndWriteRegion`, TiFlash still finds the lock in local LockCF and returns a lock error to the caller.
4. The lock resolver resolves the lock on the TiKV leader, but the resolve-lock raft log has not been applied by the TiFlash learner yet.
5. The next read retry reuses the previous read-index cache for the same region / read ts, so TiFlash does not issue a fresh read-index request and does not wait to the newer resolve-lock raft log.
6. TiFlash may see the same local lock again and repeat the retry loop.

### 2. What did you expect to see? (Required)

After TiFlash finds a local LockCF lock after a successful wait-index, the cached read index for that region should be invalidated. The next retry should fetch a fresh read index and wait until TiFlash has applied the raft log that resolves the lock, unless the lock can be bypassed by the caller.

### 3. What did you see instead (Required)

TiFlash can reuse the stale cached read index after reporting the lock. This may cause repeated remote cop retries / lock errors for the same region and read ts, because the retry keeps reading before the resolve-lock raft log is applied on the TiFlash learner.

### 4. What is your TiFlash version? (Required)

Development branch. The issue was found while investigating the read-index cache behavior on TiFlash master / release-8.5 materialized-view development branches.

PR #10986, opened 2026-07-16
Invalidate read-index cache when local lock is found

windtalker:

Issue Number: ref #10985

Problem Summary:

TiFlash can repeatedly reuse a stale read-index cache after a local LockCF lock is found and the caller resolves the lock on TiKV. The next retry may skip fetching a fresh read index, so it may not wait for the resolve-lock raft log to be applied on the TiFlash learner and can hit the same lock again.

```commit-message
Invalidate read index cache when lock is found
```

When `RegionTable::resolveLocksAndWriteRegion` still finds a local lock after wait-index succeeds, TiFlash now invalidates the read-index worker history cache for that region and removes the current query-level read-index cache entry. The next retry will request a fresh read index instead of reusing the stale one.

A read-index worker unit test is added to verify that invalidating the cache forces a new read-index request for a smaller read ts that would otherwise use history.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a potential repeated TiFlash remote cop retry caused by reusing a stale read-index cache after resolving locks.
```

## Summary by CodeRabbit

* **New Features**
  * Added a way to explicitly invalidate per-region read-index caches.
* **Bug Fixes**
  * Improved cache consistency by clearing both read-index caches and related read-index result caches when lock resolution indicates cached data may be stale.
* **Tests**
  * Extended read-index tests to confirm cache invalidation returns the latest commit index while preserving expected history-success task behavior.
  * Disabled one fast-add peer test case.
* **Chores**
  * Enhanced lock exception debug messaging with more contextual details.
  * Updated the bundled C client submodule revision.

### DIFF
--- a/dbms/src/Storages/KVStore/KVStore.h
+++ b/dbms/src/Storages/KVStore/KVStore.h
@@ -235,6 +235,7 @@ public: // Raft Read
     void addReadIndexEvent(Int64 f) { read_index_event_flag += f; }
     Int64 getReadIndexEvent() const { return read_index_event_flag; }
     BatchReadIndexRes batchReadIndex(const std::vector<kvrpcpb::ReadIndexRequest> & req, uint64_t timeout_ms) const;
+    void invalidateReadIndexCache(RegionID region_id) const;
     /// Initialize read-index worker context. It only can be invoked once.
     /// `worker_coefficient` means `worker_coefficient * runner_cnt` workers will be created.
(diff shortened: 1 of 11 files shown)


solotzg (approved):

gengliqi (approved):
LGTM

PR #10999, opened 2026-07-26
Add multi-stage late materialization

windtalker:

Issue Number: ref #11003

Problem Summary:

For wide DeltaMerge table scans, the existing late materialization path can still read many output columns for rows that are later removed by residual Selection filters. This is inefficient when the residual filter only depends on a small subset of columns and has good selectivity.

```commit-message
Add multi-stage late materialization for DeltaMerge TableScan
```

This PR adds multi-stage late materialization for DeltaMerge TableScan when both pushed-down TableScan filters and residual Selection filters exist. The read path is split into:

- Stage 0: read pushed-filter columns and apply the pushed-down filter.
- Stage 1: read only residual-filter columns for rows that pass Stage 0, then evaluate the residual filter.
- Final stage: read the remaining output columns only for rows that pass previous stages.

The implementation is guarded by `dt_enable_multi_stage_late_materialization` and disables the optimization for unsupported or unsafe cases such as generated columns, keep-order scan, fast scan, `force_push_down_all_filters_to_scan`, and stage1 filters referencing virtual `_tidb_tid`. It also keeps TableScan / Selection `actRows` correct for local, remote-only, and mixed local/remote execution by reporting rows through MSLM runtime stats and delayed remote profile aggregation.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Unit tests run:

- `git diff --check`
- `ninja -C /Users/feixu/dev/pingcap/tiflash/cmake-build-debug dbms/gtests_dbms`
- `/Users/feixu/dev/pingcap/tiflash/cmake-build-debug/dbms/gtests_dbms --gtest_filter='ExecutorsWithDMTestRunner.MultiStageLateMaterialization*'`
- `/Users/feixu/dev/pingcap/tiflash/cmake-build-debug/dbms/gtests_dbms --gtest_filter='SkippableBlockInputStreamTest.MultiStageLateMaterialization*'`

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Add multi-stage late materialization for DeltaMerge TableScan to reduce unnecessary column reads when residual filters are selective.
```


dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -1724,23 +1763,114 @@ std::pair<Names, std::vector<UInt8>> DAGStorageInterpreter::getColumnsForTableSc
...
+    {
+        return disable(fmt::format(
+            "stage0 pushed filters cover all table scan columns, stage0_filter_col_cnt={}, table_scan_col_cnt={}",
+            stage0_filter_col_id_set.size(),
+            table_scan.getColumnSize()));
+    }
+
+    std::unordered_set<ColumnID> stage1_filter_col_id_set;
+    for (const auto & expr : filter_conditions.conditions)
+        getColumnIDsFromExpr(expr, table_scan.getColumns(), stage1_filter_col_id_set);
```
JaySon-Huang, line 1843: `stage1_filter_col_id_set` may include `ExtraTableIDColumnID` (`_tidb_tid`), but that virtual column is removed from `columns_to_read` and only injected by `AddExtraTableIDColumnInputStream` after the Segment stream. If a qualifying wide partition-table Selection references `_tidb_tid`, this gate enables multi-stage late materialization and `PushDownFilter::build` later fails its `columns_to_read_map.contains(cid)` check. Please disable multi-stage late materialization when Stage 1 references this virtual column (in both production and mock gates), and exclude it from the width heuristic.

Source: rule/WINDTALKER-PUSHDOWN-005 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Coprocessor/DAGStorageInterpreter.cpp
```diff
@@ -497,15 +512,20 @@ void DAGStorageInterpreter::executeImpl(
                 table_scan.getTableScanExecutorID(),
                 remote_builder.getCurIOProfileInfos(),
                 /*is_append=*/true);
+            auto remote_profile_infos = remote_builder.getCurProfileInfos();
+            dag_context.addProfileRowsToExecutorRowsOverride(table_scan.getTableScanExecutorID(), remote_profile_infos);
```
JaySon-Huang: `addProfileRowsToExecutorRowsOverride` snapshots `remote_profile_infos[*]->rows` here while the remote pipeline has only been built, not executed, so the value is still zero. Later operator updates do not propagate to the atomic override, and `ExecutorStatistics` replaces the normally aggregated rows with that override. Remote-only queries therefore report `actRows=0`, while mixed local/remote queries omit remote rows. Please retain and aggregate the remote profiles at statistics collection time, or update the override from the remote completion path; the regression test should use real remote-only and mixed execution paths instead of calling the helper after setting synthetic rows.

Source: rule/WJHUANG2016-RE-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (approved):

gengliqi (approved):

PR #11001, opened 2026-07-27
executor: skip probe for empty hash joins

### LINKED ISSUE #11000: Avoid unnecessary probe-side MPP work for empty-build hash joins
## Enhancement

For a hash join with an empty build side, the result can be known without reading the probe side, but TiFlash still scans it.

For example:

```sql
SELECT *
FROM orders
INNER JOIN lineitem
    ON l_comment = o_comment
WHERE o_clerk = 'Clerk#000000000';
```

The filter produces no build rows, but the profile shows:

```
Build actRows:                 0
Probe actRows:                 59,986,052
lineitem data_scanned_rows:    59,987,779
Query time:                    about 715 ms
```

It would be useful to skip the probe-side scan when the build side is empty. The same optimization opportunity also applies to eligible Semi Join cases.

ChangRui-Ryan:

Issue Number: close #11000 

## Problem Summary

When the build side of a hash join is empty after filtering, the join result may already be known without reading the probe side. However, the current execution pipeline can still scan the entire probe side and perform unnecessary hash lookups against an empty hash table.

For example:

```sql
SELECT *
FROM orders
INNER JOIN lineitem
    ON l_comment = o_comment
WHERE o_clerk = 'Clerk#000000000';
```

The filter produces an empty build side, but the profile shows:

```text
Build actRows:              0
Probe actRows:              59,986,052
lineitem data_scanned_rows: 59,987,779
Query time:                 about 715 ms
```

This causes unnecessary probe-side I/O, block processing, pipeline scheduling, and hash-table lookup overhead. The same problem also applies to eligible Semi Join cases.

This change allows supported hash joins to terminate the probe side early when the finalized build hash table is empty.

The implementation has two parts:

1. Track whether the finalized build side is empty.

   - V1 publishes the build-empty state after build finalization.
   - V1 only enables this state when no actual spill has happened.
   - V2 publishes the state after all build workers finish.
   - The check is based on effective hash-table entries, so a build input containing only NULL join keys is also treated as empty.

2. Skip probe processing at both the join and pipeline levels.

   - At the join level, incoming probe blocks are marked as processed without probing the hash table.
   - At the pipeline level, the probe transform can indicate that its source is no longer needed.
   - `PipelineExec` then avoids reading the probe source and finishes the transform normally, without relying on source EOF or repeatedly producing empty blocks.
   - This prevents unnecessary local probe scans and avoids repeated lookups against an empty hash table.

The optimization is enabled for:

```text
V1: Inner Join, Semi Join, RightSemi Join
V2: Inner Join, Semi Join
```

RightSemi Join is not currently supported by Hash Join V2 and therefore falls back to the V1 implementation.

Join types whose output still depends on every probe row, such as AntiSemi Join and LeftOuterSemi Join, are not included in this optimization. V1 also keeps the existing behavior when an actual external join spill has occurred.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Performance Improvements
* Improved hash join execution when the build side is empty by avoiding unnecessary probing and source reads.
* Reduced pipeline work when joins can complete without reading probe data.
* Preserved probe processing for join types that require it.

## Bug Fixes
* Improved completion and output handling when probe processing is skipped.
* Corrected behavior for spilled joins with empty hash tables.

## Tests
* Added coverage for empty and null-only build inputs across multiple join types.
* Added tests confirming skipped sources are not executed.

### DIFF
--- a/dbms/src/DataStreams/HashJoinProbeExec.cpp
+++ b/dbms/src/DataStreams/HashJoinProbeExec.cpp
@@ -128,6 +128,9 @@ PartitionBlock HashJoinProbeExec::getProbeBlock()
 
 Block HashJoinProbeExec::probe()
 {
+    if (probe_process_info.all_rows_joined_finish && join->shouldSkipProbe())
+        return {};
+
     if (probe_process_info.all_rows_joined_finish)
     {
         auto partition_block = getProbeBlock();
--- a/dbms/src/DataStreams/HashJoinProbeExec.cpp
+++ b/dbms/src/DataStreams/HashJoinProbeExec.cpp
@@ -128,7 +128,7 @@ PartitionBlock HashJoinProbeExec::getProbeBlock()
 
 Block HashJoinProbeExec::probe()
 {
-    if (probe_process_info.all_rows_joined_finish && join->shouldSkipProbe())
+    if (join->shouldSkipProbe())
         return {};
 
     if (probe_process_info.all_rows_joined_finish)
--- a/dbms/src/Flash/tests/gtest_spill_join.cpp
+++ b/dbms/src/Flash/tests/gtest_spill_join.cpp
@@ -267,7 +267,15 @@ try
     context.context->setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(10000)));
 
     WRAP_FOR_SPILL_TEST_BEGIN
-    ASSERT_COLUMNS_EQ_UR(empty_result, executeStreams(request, 10));
+    DAGContext dag_context(*request, "empty_hash_after_spill", 10);
+    ASSERT_COLUMNS_EQ_UR(empty_result, executeStreams(&dag_context));
+
+    const auto & join_execute_info = dag_context.getJoinExecuteInfoMap().at("Join_2");
+    ASSERT_TRUE(join_execute_info.join_profile_info->is_spilled);
+
+    // A spilled hash join must not skip its probe side even when no build row enters the hash table.
+    Expect expect{{"table_scan_0", {3, 10}}, {"table_scan_1", {8192, 10}}, {"Join_2", {0, 10}}};
+    testForExecutionSummary(request, expect);
     WRAP_FOR_SPILL_TEST_END
 }
 CATCH
--- a/dbms/src/Flash/tests/gtest_spill_join.cpp
+++ b/dbms/src/Flash/tests/gtest_spill_join.cpp
@@ -238,6 +238,40 @@ try
 }
 CATCH
 
+TEST_F(SpillJoinTestRunner, EmptyHashTableAfterSpillStillReadsProbe)
+try
+{
+    constexpr size_t build_rows = 8192;
+    /// All build keys are NULL, so the hash table has no entries. The payload makes the build side spill.
+    std::vector<std::optional<Int64>> null_keys(build_rows);
+    std::vector<std::optional<String>> payload(build_rows, String(128, 'x'));
+
+    context.addMockTable(
+        "empty_hash_after_spill",
+        "probe_table",
+        {{"a", TiDB::TP::TypeLong}},
+        {toNullableVec<Int32>("a", {1, 2, 3})});
+    context.addMockTable(
+        "empty_hash_after_spill",
+        "build_table",
+        {{"a", TiDB::TP::TypeLong}, {"payload", TiDB::TP::TypeString}},
+        {toNullableVec<Int32>("a", null_keys), toNullableVec<String>("payload", payload)});
+
+    auto request
+        = context.scan("empty_hash_after_spill", "probe_table")
+              .join(context.scan("empty_hash_after_spill", "build_table"), tipb::JoinType::TypeInnerJoin, {col("a")})
+              .build(context);
+    ColumnsWithTypeAndName empty_result;
+
+    context.context->getSettingsRef().enable_hash_join_v2 = false;
+    context.context->setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(10000)));
+
+    WRAP_FOR_SPILL_TEST_BEGIN
+    ASSERT_COLUMNS_EQ_UR(empty_result, executeStreams(request, 10));
+    WRAP_FOR_SPILL_TEST_END
+}
+CATCH
+
 TEST_F(SpillJoinTestRunner, ScanHashMapAfterProbeDataWithSpillEnabledAndSpillTriggered)
 try
 {
--- a/dbms/src/Interpreters/Join.cpp
+++ b/dbms/src/Interpreters/Join.cpp
@@ -1802,6 +1802,8 @@ void Join::workAfterBuildFinish(size_t stream_index)
 
         has_build_data_in_memory = !original_blocks.empty();
     }
+
+    build_side_empty.store(!isSpilled() && getTotalRowCount() == 0, std::memory_order_release);
 }
 
 void Join::finalizeNullAwareSemiFamilyBuild()
@@ -1984,6 +1986,13 @@ Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
         LOG_WARNING(log, "JoinBlock without non zero active_build_threads, return empty block");
         return {};
     }
+
+    if unlikely (shouldSkipProbe())
+    {
+        probe_process_info.updateEndRow<false>(probe_process_info.block.rows());
+        return output_block_after_finalize;
+    }
+
     std::shared_lock lock(rwlock);
 
     Block block{};
--- a/dbms/src/Interpreters/Join.cpp
+++ b/dbms/src/Interpreters/Join.cpp
@@ -1987,12 +1987,6 @@ Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
         return {};
     }
 
-    if unlikely (shouldSkipProbe())
-    {
-        probe_process_info.updateEndRow<false>(probe_process_info.block.rows());
-        return output_block_after_finalize;
-    }
-
     std::shared_lock lock(rwlock);
 
     Block block{};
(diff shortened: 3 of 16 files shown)


dbms/src/Flash/tests/gtest_spill_join.cpp
```diff
@@ -238,6 +238,40 @@ try
...
+              .join(context.scan("empty_hash_after_spill", "build_table"), tipb::JoinType::TypeInnerJoin, {col("a")})
+              .build(context);
+    ColumnsWithTypeAndName empty_result;
+
+    context.context->getSettingsRef().enable_hash_join_v2 = false;
+    context.context->setSetting("max_bytes_before_external_join", Field(static_cast<UInt64>(10000)));
+
+    WRAP_FOR_SPILL_TEST_BEGIN
+    ASSERT_COLUMNS_EQ_UR(empty_result, executeStreams(request, 10));
+    WRAP_FOR_SPILL_TEST_END
```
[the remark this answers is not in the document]
ChangRui-Ryan: Thanks. Addressed in d1f4dd7: the test now asserts `is_spilled` and verifies that probe `table_scan_0` produces 3 rows in the execution summary, while retaining the empty-result check.

dbms/src/DataStreams/HashJoinProbeExec.cpp
```diff
@@ -128,6 +128,9 @@ PartitionBlock HashJoinProbeExec::getProbeBlock()
 
 Block HashJoinProbeExec::probe()
 {
+    if (probe_process_info.all_rows_joined_finish && join->shouldSkipProbe())
```
windtalker: why need to check `probe_process_info.all_rows_joined_finish` here?

dbms/src/Operators/HashJoinV2ProbeTransformOp.cpp
```diff
@@ -80,6 +80,13 @@ OperatorStatus HashJoinV2ProbeTransformOp::tryOutputImpl(Block & block)
         block = {};
         return OperatorStatus::HAS_OUTPUT;
     }
+    if unlikely (probe_context.isAllFinished() && join_ptr->shouldSkipProbe())
```
windtalker: it looks to me that there is no need to check `probe_context.isAllFinished()`?

dbms/src/Operators/HashJoinProbeTransformOp.cpp
```diff
@@ -163,9 +163,12 @@ OperatorStatus HashJoinProbeTransformOp::tryOutputImpl(Block & block)
 {
     if (status == ProbeStatus::PROBE && probe_process_info.all_rows_joined_finish)
     {
-        if (auto ret = probe_transform->tryFillProcessInfoInProbeStage(probe_process_info);
-            ret != OperatorStatus::HAS_OUTPUT)
-            return ret;
+        if (!probe_transform->shouldSkipProbe())
```
windtalker, line 168: how about check `probe_transform->shouldSkipProbe()` at the begining of this method, and just set `block` to `{}` and return `OperatorStatus::HAS_OUTPUT` if `shouldSkipProbe` is true.

dbms/src/Operators/HashJoinV2ProbeTransformOp.cpp
```diff
@@ -80,6 +80,13 @@ OperatorStatus HashJoinV2ProbeTransformOp::tryOutputImpl(Block & block)
...
+    {
+        join_ptr->finishOneProbe(op_index);
+        probe_context.input_is_finished = true;
+        block = join_ptr->probeLastResultBlock(op_index);
```
windtalker: why not just set block to `{}`

dbms/src/Interpreters/JoinV2/HashJoin.cpp
```diff
@@ -622,6 +623,13 @@ Block HashJoin::probeBlock(JoinProbeContext & ctx, size_t stream_index)
     Stopwatch all_watch;
     SCOPE_EXIT({ probe_workers_data[stream_index].probe_time += all_watch.elapsedFromLastTime(); });
 
+    if unlikely (shouldSkipProbe())
```
windtalker: why `probeBlock` is called if `shouldSkipProbe` is true?

dbms/src/Interpreters/Join.cpp
```diff
@@ -1984,6 +1986,13 @@ Block Join::joinBlock(ProbeProcessInfo & probe_process_info) const
         LOG_WARNING(log, "JoinBlock without non zero active_build_threads, return empty block");
         return {};
     }
+
+    if unlikely (shouldSkipProbe())
```
windtalker: ditto

dbms/src/DataStreams/HashJoinProbeExec.cpp
ChangRui-Ryan: After tracing the execution order, this check is redundant in the current flow. HashJoinProbeExec enters the probe stage only after waitUntilAllBuildFinished(). If shouldSkipProbe() is true, the build side was already finalized as empty before the first probe call, so no probe block can have been loaded into probe_process_info; all_rows_joined_finish is therefore necessarily true. If the build side is non-empty, shouldSkipProbe() remains false for the whole probe stage.
I originally kept the condition to make the state-machine precondition explicit and defensive, but it is not needed for the current correctness. I will simplify it as suggested.

dbms/src/Operators/HashJoinV2ProbeTransformOp.cpp
ChangRui-Ryan: Ditto. Similar to the legacy HashJoinProbeExec path, this check is redundant under the current HashJoin V2 execution ordering. The probe pipeline is scheduled only after the build pointer-table event has completed and published `build_side_empty`. If `shouldSkipProbe()` is true when this probe transform starts, no probe block can have been accepted into `probe_context`, so `probe_context.isAllFinished()` is necessarily true. If the build side is non-empty, `shouldSkipProbe()` remains false throughout the probe stage.

I originally kept the condition to make the state-machine precondition explicit and defensive, but it is not needed for the current correctness. I will simplify it as suggested.

dbms/src/Operators/HashJoinProbeTransformOp.cpp
ChangRui-Ryan: This is different from the redundant state checks above. Returning `{}` directly here would bypass the completion state machine of the legacy `Join` implementation used by `HashJoinProbeTransformOp`.

In particular, it would skip `finishOneProbe()` for every probe worker and `finalizeProbe()` for the last one, so `active_probe_threads` would not reach zero and `wait_probe_finished_future` would never be finished. It would also bypass the existing post-probe transitions, such as scan-after-probe and restore handling.

For example, with an empty-build inner join and two probe workers, both probe transforms could emit EOF immediately while neither decrements the shared probe-worker count. The join would never observe probe completion.

The current code avoids reading any new probe input, but still enters `onOutput()` to reuse the normal completion and finalization path.

dbms/src/Operators/HashJoinV2ProbeTransformOp.cpp
ChangRui-Ryan: `probeLastResultBlock()` is needed by the normal source-EOF path to drain a result block buffered by earlier probe processing. In this empty-build fast path, however, the probe pipeline is scheduled only after the build pointer-table event, and the source is skipped before any probe block can be accepted into this transform. For the skip-eligible V2 join kinds (`Inner` and `Semi`), there can be no independently buffered final result in this case.

Therefore, `probeLastResultBlock()` is guaranteed to be empty here, and setting `block` to `{}` directly makes the logical EOF explicit. I will simplify it as suggested.

dbms/src/Interpreters/JoinV2/HashJoin.cpp
ChangRui-Ryan: Under the current HashJoin V2 execution ordering, this branch is unreachable when `shouldSkipProbe()` is true. The probe pipeline is scheduled only after the build pointer-table event, and `HashJoinV2ProbeTransformOp::tryOutputImpl()` completes the probe worker before any probe block can be accepted into `probe_context` or passed to `HashJoin::probeBlock()`.

This branch was added as a defensive fast path for a hypothetical caller that bypasses the probe transform and invokes `probeBlock()` directly. Such a caller would still produce the correct result by probing the empty hash table; it would only lose this extra optimization.

To keep the implementation focused on the actual pipeline path, I will remove this redundant branch as suggested.

dbms/src/Interpreters/Join.cpp
ChangRui-Ryan: Ditto for the legacy `Join::joinBlock()` path. `HashJoinProbeExec::probe()` and `HashJoinProbeTransformOp::tryOutputImpl()` both handle `shouldSkipProbe()` before a probe block is passed to `Join::joinBlock()`, so this branch is unreachable under the current execution flow.

It was only a defensive fast path for a hypothetical direct caller of `joinBlock()`. Removing it does not affect correctness; such a caller would still produce an empty result by probing the empty hash table, at the cost of unnecessary work.

I will remove it as well to keep the optimization focused on the actual probe transform and pipeline paths.

windtalker (approved):
lgtm

dbms/src/Operators/HashJoinProbeTransformOp.h
```diff
@@ -33,6 +33,8 @@ class HashJoinProbeTransformOp : public TransformOp
 
     String getName() const override { return "HashJoinProbeTransformOp"; }
 
+    bool shouldSkipSource() const override { return origin_join->shouldSkipProbe(); }
```
xzhangxian1008: I think it's better to call `probe_transform->shouldSkipProbe()`? Since HashJoinProbeTransformOp.cpp:L168 calls it.

xzhangxian1008 (approved):
LGTM

PR #11020, opened 2026-08-07
Support decimal rough set minmax filter

windtalker:

Issue Number: ref https://github.com/pingcap/tiflash/issues/11031

Problem Summary:

Decimal columns were not supported by rough set min/max filtering. The read-side comparison needs scale-aware decimal comparison, and the write-side must persist ordinary min/max indexes for decimal columns.

```commit-message
Support decimal rough set minmax filter
```

- Add decimal dispatch in min/max rough set check paths.
- Use existing execution-layer decimal comparison utility for Decimal-vs-Decimal comparisons across physical decimal types and scales.
- Keep Decimal-vs-non-Decimal comparisons conservative by returning `CanNotCompare`.
- Generate ordinary min/max index payloads for decimal columns in `DMFileWriter`.
- Add a dev note and regression coverage for decimal min/max filtering and DMFile write-side index generation.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Manual test:

```bash
ninja -C cmake-build-debug dbms/gtests_dbms
./cmake-build-debug/dbms/gtests_dbms --gtest_filter=MinMaxIndexTest.*:MinMaxIndexWriteTest.* > /tmp/minmax_decimal.log 2>&1
git diff --check
```

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Support rough set min/max filter for decimal columns in TiFlash.
```

## Summary by CodeRabbit

- **New Features**
  - Added min/max indexing and rough-set filtering for Decimal columns.
  - Enabled scale-aware Decimal comparisons across supported predicates and precisions.
  - Added Decimal index generation when writing data files.

- **Bug Fixes**
  - Improved nullable and non-nullable filtering for Decimal values.
  - Preserved conservative handling for unsupported Decimal comparisons.

- **Tests**
  - Expanded coverage for persisted indexes, comparisons, nullability, and varying Decimal widths.

- **Documentation**
  - Added guidance on Decimal filtering behavior and supported comparisons.

### DIFF
--- a/dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
+++ b/dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
@@ -55,6 +55,9 @@ inline bool isRoughSetFilterSupportType(const Int32 field_type)
     case TiDB::TypeDatetime:
     case TiDB::TypeTimestamp: // For timestamp, should take time_zone into consideration while parsing `literal`
         return true;
+    case TiDB::TypeDecimal:
+    case TiDB::TypeNewDecimal:
+        return true;
     // For these types, should take collation into consideration. Disable them.
     case TiDB::TypeVarchar:
     case TiDB::TypeJSON:
@@ -67,8 +70,6 @@ inline bool isRoughSetFilterSupportType(const Int32 field_type)
         return false;
     // Unknown.
     case TiDB::TypeTiDBVectorFloat32:
-    case TiDB::TypeDecimal:
-    case TiDB::TypeNewDecimal:
     case TiDB::TypeFloat:
     case TiDB::TypeDouble:
     case TiDB::TypeNull:
--- a/dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
+++ b/dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
@@ -12,12 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <Columns/ColumnDecimal.h>
 #include <Columns/ColumnString.h>
 #include <Columns/ColumnVector.h>
 #include <Columns/IColumn.h>
 #include <Common/Exception.h>
+#include <Common/assert_cast.h>
 #include <DataTypes/DataTypeDate.h>
 #include <DataTypes/DataTypeDateTime.h>
+#include <DataTypes/DataTypeDecimal.h>
 #include <DataTypes/DataTypeEnum.h>
 #include <DataTypes/DataTypeFixedString.h>
 #include <DataTypes/DataTypeMyDate.h>
@@ -38,6 +41,15 @@ static constexpr size_t NONE_EXIST = std::numeric_limits<size_t>::max();
 
 namespace details
 {
+template <typename T>
+ALWAYS_INLINE const auto & minMaxColumnData(const IColumn & column)
+{
+    if constexpr (IsDecimal<T>)
+        return assert_cast<const ColumnDecimal<T> &>(column).getData();
+    else
+        return toColumnVectorData<T>(column);
+}
+
 inline std::pair<size_t, size_t> minmax(
     const IColumn & column,
     const ColumnVector<UInt8> * del_mark,
@@ -285,7 +297,7 @@ RSResults MinMaxIndex::checkNullableInImpl(
     const DataTypePtr & type)
 {
     RSResults results(pack_count, RSResult::SomeNull);
-    const auto & minmaxes_data = toColumnVectorData<T>(column_nullable.getNestedColumnPtr());
+    const auto & minmaxes_data = details::minMaxColumnData<T>(*column_nullable.getNestedColumnPtr());
     for (size_t i = start_pack; i < start_pack + pack_count; ++i)
     {
         if (details::minIsNull(null_map, i))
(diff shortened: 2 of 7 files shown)


dbms/src/Storages/DeltaMerge/tests/gtest_dm_minmax_index.cpp
```diff
@@ -2489,6 +2270,122 @@ MinMaxIndexPtr createMinMaxIndex(const IDataType & col_type, const T & cases)
...
+            .column_data = {std::nullopt, std::nullopt},
+            .del_mark = {0, 0},
+        },
+    };
+
+    auto col_type = makeNullable(std::make_shared<DataTypeDecimal64>(20, 5));
+    auto minmax_index = std::make_shared<MinMaxIndex>(*col_type);
+    for (const auto & c : cases)
+    {
+        RUNTIME_CHECK(c.column_data.size(), c.del_mark.size());
```
JaySon-Huang: `RUNTIME_CHECK` currently uses `c.column_data.size()` as the condition, so every non-empty test case passes even when the delete-mark length differs. Please compare the sizes explicitly:

```suggestion
RUNTIME_CHECK(c.column_data.size() == c.del_mark.size());
```

Source: rule/JAYSONHUANG-RC-006 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Storages/DeltaMerge/Index/MinMaxIndex.cpp
```diff
@@ -481,6 +503,11 @@ RSResults MinMaxIndex::checkCmp(size_t start_pack, size_t pack_count, const Fiel
         return checkCmpImpl<Op, TYPE>(start_pack, pack_count, value, type);
     FOR_NUMERIC_TYPES(DISPATCH)
 #undef DISPATCH
+#define DISPATCH_DECIMAL(FIELD_TYPE, DATA_TYPE)   \
+    if (typeid_cast<const DATA_TYPE *>(raw_type)) \
+        return checkCmpImpl<Op, FIELD_TYPE>(start_pack, pack_count, value, type);
+    FOR_DECIMAL_TYPES(DISPATCH_DECIMAL)
```
JaySon-Huang, line 509: The new decimal dispatch is currently unreachable from production rough-set parsing. `cop::isRoughSetFilterSupportType` in `FilterParser.cpp` still returns `false` for `TypeNewDecimal`, so `parseTiCompareExpr` converts decimal predicates to `Unsupported` before `decodeLiteral` or `checkCmp` can run. Please enable `TypeNewDecimal` there and add a parser-level test that builds a decimal `tipb::Expr` and verifies that the resulting operator uses the min/max index.

Source: rule/JAYSONHUANG-RC-008 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (approved):
LGTM

xzhangxian1008 (approved):

dbms/src/Storages/DeltaMerge/FilterParser/FilterParser.cpp
```diff
@@ -55,6 +55,9 @@ inline bool isRoughSetFilterSupportType(const Int32 field_type)
     case TiDB::TypeDatetime:
     case TiDB::TypeTimestamp: // For timestamp, should take time_zone into consideration while parsing `literal`
         return true;
+    case TiDB::TypeDecimal:
+    case TiDB::TypeNewDecimal:
```
JaySon-Huang, line 59: **Non-blocking:** reject statically unsupported mixed-decimal predicates during parsing.

After decimal columns are admitted here, a predicate whose literal is encoded as `Int64`, `Uint64`, or `Float64` can still be parsed into a normal `Equal`/`Greater`/`In` operator. During `roughCheck`, `ValueComparision` then returns `CanNotCompare`, so every pack becomes conservatively `Some`. This makes the filter look enabled while providing no pruning, and it may load an index that cannot be used.

Could we detect this in `parseTiCompareExpr` and create `Unsupported` when exactly one side is decimal? Checking `child.tp() == tipb::ExprType::MysqlDecimal` is preferable to relying on `literal.field_type()`, because the encoded decimal value is the authoritative source for its scale. `NULL` should remain allowed, including inside `IN`. For example:

```diff
-    bool is_timestamp_column = false;
+    bool is_timestamp_column = false;
+    bool is_decimal_column = false;
     for (const auto & child : expr.children())
     {
         if (isColumnExpr(child))
         {
-            is_timestamp_column = (child.field_type().tp() == TiDB::TypeTimestamp);
+            const auto column_type = child.field_type().tp();
+            is_timestamp_column = (column_type == TiDB::TypeTimestamp);
+            is_decimal_column
+                = column_type == TiDB::TypeDecimal || column_type == TiDB::TypeNewDecimal;
             break;
         }
     }
@@
         else if (isLiteralExpr(child))
         {
             Field value = decodeLiteral(child);
+            const bool is_decimal_literal = child.tp() == tipb::ExprType::MysqlDecimal;
+            if (!value.isNull() && is_decimal_column != is_decimal_literal)
+                return createUnsupported(fmt::format(
+                    "Mixed decimal comparison is not supported, sig={} column_decimal={} literal_type={}",
+                    tipb::ScalarFuncSig_Name(expr.sig()),
+                    is_decimal_column,
+                    tipb::ExprType_Name(child.tp())));
```

`CanNotCompare -> Some` should still remain as the runtime safety net for mismatches that are not knowable during parsing.

Source: rule/WINDTALKER-TYPE-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

PR #11038, opened 2026-08-12
executor: Fix LIKE with non-ASCII patterns on unsigned-char platforms

### LINKED ISSUE #11037: Title: LIKE with CI collations silently returns wrong results for non-ASCII patterns on aarch64 builds (introduced by #10400)
Bug Report

Since [#10400](https://github.com/pingcap/tiflash/pull/10400) (tryCompileAsciiCi/tryMatchAsciiCi, merged 2025-09-19), LIKE on any case-insensitive collation (utf8mb4_general_ci, utf8mb4_unicode_ci, utf8mb4_0900_ai_ci, …) silently returns wrong results (typically 0 rows) for patterns containing non-ASCII characters (e.g. LIKE '%报告%') — on platforms where char is unsigned, i.e. all linux-aarch64 builds (AAPCS64 defaults char to unsigned; the build system does not pass -fsigned-char).

Root cause, in dbms/src/TiDB/Collation/Collator.cpp:

Pattern<Collator>::tryCompileAsciiCi: non-ASCII detection uses auto c = pattern[i]; if (c < 0) return; which is never true when char is unsigned. A CJK pattern is therefore mis-classified as an ASCII pattern, and weight_ascii_ci[c] (a 128-entry std::array<char,128>) is indexed with bytes up to 255 — an out-of-bounds read (confirmed with ASan: global-buffer-overflow).
Pattern<Collator>::tryMatchAsciiCi: the if (s[str_idx] < 0) return -1; fallback is likewise dead code on unsigned-char platforms, so it never falls back to the correct per-character matcher. Additionally match_types is compiled per character while ascii_ci_pattern is compiled per byte, so the shared index goes out of bounds on the match_types vector and the match deterministically returns "no match" for every row.
x86_64 builds are unaffected (char is signed there), which is presumably why CI didn't catch it.

Repro (standalone, extracting the exact Pattern code + CollationLUT.cpp from master): compile with -funsigned-char → '报告' LIKE '%报告%' returns 0; with -fsigned-char → 1. On a real cluster: any aarch64 TiFlash ≥ [#10400](https://github.com/pingcap/tiflash/pull/10400), utf8mb4_general_ci column, SELECT COUNT(*) ... WHERE col LIKE '%报告%' returns 0 via TiFlash vs correct count via TiKV.

Suggested fix: treat pattern/string bytes as unsigned char explicitly, e.g. if (static_cast<unsigned char>(c) >= 0x80) return;, and add CJK-pattern unit tests to an aarch64 CI job. Note there is also an index-misalignment edge case for patterns containing an escaped % immediately followed by % (e.g. a\%%b) because the consecutive-% dedup check pattern[i-1]=='%' ignores escaping.

ChangRui-Ryan:

Issue Number: close #11037 

### Problem Summary

`LIKE` pattern matching can produce incorrect results for non-ASCII patterns on platforms where `char` is unsigned, such as Linux AArch64.

The implementation used `char` values as sentinel states while scanning UTF-8 characters. Bytes greater than `0x7F` were interpreted differently when `char` is unsigned, causing the matcher to incorrectly handle escaped non-ASCII characters and patterns containing UTF-8 literals.

The initial correctness fix avoided the signedness-dependent behavior, but introduced an extra per-character decoding cost in the hot matching path. This could cause a measurable performance regression for common `LIKE` workloads on AArch64.

### What is changed and how it works

- Replace signedness-sensitive `char` sentinel handling in the `LIKE` matcher with an explicit representation that works consistently regardless of whether the compiler treats `char` as signed or unsigned.
- Add regression test coverage for non-ASCII UTF-8 patterns, including escaped wildcard cases, to ensure the matcher produces the same result on x86 and AArch64.
- Preserve the existing optimized ASCII fast path in the matcher. The revised implementation avoids unnecessary UTF-8 decoding for ordinary ASCII pattern matching while applying the correct handling only when non-ASCII bytes are encountered.

As a result, `LIKE` matching is correct for non-ASCII patterns on unsigned-`char` platforms, without giving up the substantial performance benefit of the prior optimized implementation.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved case-insensitive pattern matching for ASCII text.
  * Correctly handles escaped wildcard characters and consecutive wildcard symbols.
  * Prevents invalid or non-ASCII bytes from causing incorrect matching results.
  * Preserves expected matching behavior when processing Chinese text and mixed wildcard patterns.

* **Tests**
  * Added coverage for Chinese text, escaped percent signs, and repeated wildcard patterns.

### DIFF
--- a/dbms/src/TiDB/Collation/Collator.cpp
+++ b/dbms/src/TiDB/Collation/Collator.cpp
@@ -317,12 +317,12 @@ int Pattern<Collator>::tryMatchAsciiCi(const char * s, size_t length) const
             else
             {
                 // Can't handle non-ASCII characters
-                const auto c = static_cast<unsigned char>(s[str_idx]);
-                if (c >= weight_ascii_ci.size())
+                if (static_cast<unsigned char>(s[str_idx]) >= weight_ascii_ci.size())
                 {
                     return -1;
                 }
-                if ((match_types[p_idx] == Match && weight_ascii_ci[c] == ascii_ci_pattern[p_idx])
+                if ((match_types[p_idx] == Match
+                     && weight_ascii_ci[static_cast<unsigned char>(s[str_idx])] == ascii_ci_pattern[p_idx])
                     || match_types[p_idx] == One)
                 {
                     p_idx++;
@@ -358,13 +358,12 @@ int Pattern<Collator>::tryMatchAsciiCi(const char * s, size_t length) const
                 while (str_idx < length)
                 {
                     // Can't handle non-ASCII characters
-                    const auto c = static_cast<unsigned char>(s[str_idx]);
-                    if (c >= weight_ascii_ci.size())
+                    if (static_cast<unsigned char>(s[str_idx]) >= weight_ascii_ci.size())
                     {
                         return -1;
                     }
 
-                    if (weight_ascii_ci[c] != ascii_ci_pattern[p_idx_after_any])
+                    if (weight_ascii_ci[static_cast<unsigned char>(s[str_idx])] != ascii_ci_pattern[p_idx_after_any])
                     {
                         str_idx = ++backtrack_idx;
                     }
--- a/dbms/src/TiDB/Collation/Collator.cpp
+++ b/dbms/src/TiDB/Collation/Collator.cpp
@@ -134,44 +134,60 @@ template <typename Collator>
 void Pattern<Collator>::tryCompileAsciiCi(const std::string & pattern, char escape)
 {
     is_ascii_ci_pattern = false;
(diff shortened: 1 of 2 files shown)


windtalker (approved):
lgtm

xzhangxian1008 (approved):
lgtm

jy007:
效率太高了，点赞。

PR #11053, opened 2026-08-19
columnar: support late materialization

### LINKED ISSUE #11058: Support late materialization for columnar
## Enhancement

Add late materialization for columnar.

yongman:

Issue Number: close #11058 

Deps: https://github.com/tidbcloud/cloud-storage-engine/pull/5985

Problem Summary:
```
  The cloud Columnar read path fully materializes every requested column before evaluating the exact predicate in TiFlash.
  For wide tables with selective filters, this causes unnecessary L2 column-pack I/O, decompression, serialization, FFI
  transfer, and deserialization for rows that are filtered out immediately.

  Applying the predicate before global merge/MVCC is not correct, because rows from different sources may have the same
  handle and different versions. The optimization must therefore preserve the existing merge and MVCC semantics while
  postponing only late-column materialization.
```

```commit-message

```
```
  This PR adds an optional, L2-only late materialization path for cloud Columnar reads.

  The read flow is:

  1. Build an early projection containing handle/version columns, primary-key columns, and all columns required by the exact
     filter.

  2. Read early columns from all sources and perform the existing global merge and MVCC processing.
  3. Evaluate the exact predicate in TiFlash using the existing DAGExpressionAnalyzer and FilterTransformAction.
  4. Send the resulting row selection bitmap back to the Rust reader.
  5. Materialize late columns only for selected rows:
      - Memtable, L0, and L1 rows retain their late values in eager sidecars.
      - L2 rows retain stable physical file/row references.
      - Only L2 packs containing selected rows are loaded.
      - Late-column reads are performed concurrently with bounded parallelism.

  6. Assemble the original output block and continue through the existing downstream pipeline.

  A versioned optional FFI extension was added for the two-phase protocol, including early-block reads, selection submission,
  late-column reads, and batch finalization/discard. Older or incompatible readers automatically use the legacy full-
  materialization path.

  ## Default Behavior

  The feature is disabled by default:

  enable_columnar_l2_late_materialization = false

  When disabled, the existing full-materialization Columnar reader is used without behavioral changes.

  Even when the setting is enabled, the legacy path remains the fallback whenever the query or reader is not eligible, the
  reader does not support the protocol, or the initial selectivity probe shows insufficient benefit.

  ## Enablement Conditions

  Late materialization is enabled per reader only when all of the following conditions are satisfied:

  - enable_columnar_l2_late_materialization is set to true.
  - The scan has pushed-down exact filter conditions.
  - The filter does not reference unsupported columns or types, including generated columns, _tidb_tid, unsupported time
    types, or non-UTC timestamp conversions.

  - The reader plan does not contain multiple tables or unsupported concurrent aggregation.
  - The requested projection contains at least one late column.
  - The late-to-early column ratio is greater than columnar_l2_late_materialization_min_late_to_early_ratio (default: 10.0).
  - The Rust reader exposes a compatible version-1 late-materialization FFI interface.
  - The reader reports that late materialization is supported.
  - The schema contains valid handle and version columns.
  - At least one L2 file exists and the L2 files for the table are strictly non-overlapping.
  - The query is supported by the reader, excluding ANN/FTS cases.

  The first late-materialization batch is used as a selectivity probe. If the skipped-row ratio is lower than
  columnar_l2_late_materialization_min_selection_skip_ratio (default: 0.5), late materialization is disabled for the
  remainder of that reader and the legacy path resumes.
```

### Check List

Tests 

- [x] Unit test
- [x] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## New Features
- Added optional late materialization for cloud columnar reads, initially supporting L2 data.
- Filters are evaluated using early columns, with late columns loaded only for selected rows.
- Added settings to enable the feature and tune selection and column-ratio thresholds.
- Added capability detection and automatic fallback to legacy reads when unsupported or inefficient.
- The feature is disabled by default.

## Documentation
- Added design documentation covering behavior, compatibility, rollout, and operational considerations.


contrib/tiflash-columnar-hub/hub-runtime/src/columnar_impls.rs
```diff
@@ -249,6 +250,186 @@ pub unsafe extern "C" fn ffi_read_column(
...
+    col_id: i64,
+) -> RustStrWithView {
+    match reader.as_mut().ffi_read_early_column(batch_id, col_id) {
+        Ok(data) => build_from_string(data),
+        Err(err) => {
+            error!(
+                "ffi_read_early_column failed, batch_id={}, col_id={}: {}",
+                batch_id, col_id, err
+            );
+            RustStrWithView::default()
```
JaySon-Huang, line 304: `ffi_read_early_column` and `ffi_read_late_column` return a default empty `RustStrWithView` on error after logging. Unlike `ffi_read_early_block` / `ffi_materialize_selected`, which signal failure with `u64::MAX`, the C++ caller cannot distinguish I/O/protocol failure from a legitimate empty payload.

TiFlash then deserializes `col_data.buff` without checking length, which can produce silent wrong results or crashes instead of a query-visible error.

**Fix direction:** propagate failure explicitly (e.g. sentinel return convention, out-parameter error code, or a dedicated error FFI), and align with the existing late-materialization batch APIs.

Source: rule/JAYSONHUANG-RC-008 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Storages/StorageDisaggregatedColumnar.cpp
```diff
@@ -1439,113 +1770,349 @@ Block RNColumnarInputStream::readImpl()
...
-        if (!ensureReader())
+        if (column.column_id == MutSup::extra_table_id_col_id
+            || early_column_ids.find(column.column_id) == early_column_ids.end())
+            continue;
+        (void)w.elapsedSecondsFromLastTime();
+        auto col_data
+            = late_materialization_interfaces->fn_read_early_column(reader.value(), batch_id, column.column_id);
+        duration_read_sec += w.elapsedSecondsFromLastTime();
+        SCOPE_EXIT({ RustGcHelper::instance().gcRustPtr(col_data.inner.ptr, col_data.inner.type); });
+        ReadBufferFromMemory buf(col_data.buff.data, static_cast<size_t>(col_data.buff.len));
```
JaySon-Huang, line 1879: After `fn_read_early_column` returns, this path deserializes immediately without validating `col_data.buff.len` (same pattern for `fn_read_late_column` below). If the Rust FFI returns an empty buffer on failure, TiFlash may treat it as valid column data for `rows`/`selected_rows`.

Please validate the buffer before `deserializeBinaryBulkWithMultipleStreams`, and throw `Exception` on zero-length or size-mismatched payloads. Match the error-handling behavior of the legacy `readLegacyBlock` path.

Source: rule/JAYSONHUANG-RC-008 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

yongman:
Wait for https://github.com/pingcap/tiflash/pull/11062

dbms/src/Storages/StorageDisaggregatedColumnar.cpp
```diff
@@ -932,6 +1007,148 @@ const String & RNColumnarReadTask::getExecutorID() const
...
+            LOG_DEBUG(
+                shared_reader_context->log,
+                "Columnar late materialization filter is ineligible: reason=unsupported_filter_column_type, "
+                "column_id={}, timezone_cast={}, executor_id={}, table_id={}",
+                column.id,
+                needs_timezone_cast,
+                getExecutorID(),
+                getLogicalTableID());
+            return false;
+        }
```
JaySon-Huang, line 1131: There's also an issue where StorageColumnar late materialization cannot be performed on columns with timezone conversion. We should probably create a follow up issue for this and resolve it in another PR.

dbms/src/Storages/StorageDisaggregatedColumnar.cpp
```diff
@@ -1439,113 +1788,350 @@ Block RNColumnarInputStream::readImpl()
     return readImpl(filter_ignored, false);
 }
 
-Block RNColumnarInputStream::readImpl([[maybe_unused]] FilterPtr & res_filter, [[maybe_unused]] bool return_filter)
+Block RNColumnarInputStream::readLateMaterializedBlock()
```
JaySon-Huang, line 1829: ## Suggestion: add Columnar LM counters to `ColumnarScanContext`

We should expose late-materialization effectiveness (and, optionally, why LM was not used) through `ColumnarScanContext`.

### What to measure

In `readLateMaterializedBlock`, each batch already has:

- `rows` from `fn_read_early_block` — MVCC-visible early rows (filter input)
- `selected_rows` from `fn_materialize_selected` — rows that need late-column materialization (and become output)

The TiFlash-side row-level win is approximately:

```text
late_skipped_rows = Σ (early_rows − selected_rows)
```

Disable / eligibility reasons are separate from effectiveness. Prefer recording the **final** enable/disable outcome for the scan (setting, filter eligibility, late/early ratio, Hub support, density probe), not only `isLateMaterializationFilterEligible`.

### Recommended fields (minimal)

**Effectiveness:**

| Field | Meaning |
|-------|---------|
| `lm_early_rows` | Sum of early-batch rows (LM filter input) |
| `lm_selected_rows` | Sum of `materialize_selected` rows (late materialization / output) |

`lm_skipped_rows` can be derived as `lm_early_rows - lm_selected_rows`, or stored explicitly for parity with DM.

**Process:**

| Field | Meaning |
|-------|---------|
| `lm_batches` | Number of LM batches that completed the early→select path (including empty selections) |
| `lm_empty_batches` | Batches with `selected_rows == 0` |
| `lm_probe_disabled` | Whether density probe disabled LM for this scan (`0`/`1`) |

**Eligibility:**

| Field | Meaning |
|-------|---------|
| `lm_enabled` | Whether LM was actually used (`0`/`1`) |
| `lm_disable_reason_code` | Small enum for the final disable reason |

Avoid floating-point skip ratios in the context (derive from the counters). Avoid free-form reason strings for multi-stream merge.

### Instrumentation points

- `readLateMaterializedBlock`: accumulate `lm_early_rows` / `lm_selected_rows` / batch counters; set `lm_probe_disabled` on probe fallback.
- `initializeLateMaterialization`: set `lm_enabled` / `lm_disable_reason_code` once (CAS / once-per-task).

Lookup path already exists via `dag_context->columnar_scan_context_map[executor_id]`.

### Rollout

1. add atomics to C++ `ColumnarScanContext`, wire merge/`toJson` (EXPLAIN ANALYZE / TiFlash logs).
2. extend `tipb.ColumnarScanContext` and TiDB display for the numeric counters; reason codes can stay TiFlash-local initially.

contrib/tiflash-columnar-hub/hub-runtime/ffi/src/RaftStoreProxyFFI/ProxyFFI.h
```diff
@@ -225,6 +225,8 @@ struct ColumnarScanStats {
   uint64_t rough_check_unknown_packs;
   uint64_t remote_segments;
   uint64_t total_segments;
+  uint64_t lm_late_packs_loaded;
```
JaySon-Huang, line 228: `ColumnarScanStats` gains `lm_late_packs_loaded` / `lm_late_rows_gathered` (+16 bytes) while `fn_columnar_scan_stats` returns this POD by value. `RaftStoreProxyFFI/@version` and `interfaces.rs::RAFT_STORE_PROXY_VERSION` are still `5493270813306750334` and were not regenerated via gen-proxy-ffi after the header change.

Same-commit lockstep builds can still agree on layout, but mixed old/new hub↔TiFlash binaries will not trip the version gate and can truncate or read garbage on the by-value return.

**Fix direction:** regenerate `@version` + the Rust `RAFT_STORE_PROXY_VERSION` constant with gen-proxy-ffi (or equivalent), and ship both sides together.

Source: rule/JAYSONHUANG-RC-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Coprocessor/ColumnarScanContext.h
```diff
@@ -51,6 +67,17 @@ class ColumnarScanContext
     std::atomic<UInt64> remote_segments{0};
     std::atomic<UInt64> total_segments{0};
 
+    std::atomic<UInt64> lm_late_packs_loaded{0};
```
JaySon-Huang, line 70: LM counters are accumulated locally (`merge(ColumnarScanStats)`, `toJson`, disable-reason mask), but `serialize()` / `deserialize()` / `merge(const tipb::ColumnarScanContext&)` still omit every LM field. MPP `ExecutionSummary` therefore drops these metrics across tasks/nodes, so TiDB execution-plan / summary observability cannot see LM effectiveness.

**Follow-up OK:** this does not need to block the current PR. Prefer a separate follow-up PR that (1) lands tipb `ColumnarScanContext` optional fields, (2) wires `serialize`/`merge(tipb)`, and (3) exposes the metrics in TiDB execution-plan observability as an independent enhancement commit.

Source: rule/JAYSONHUANG-RC-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

yongman: Yes, proto and related stats update in tidb side will be committed in follow-up PR. Issue has been created to track. https://github.com/pingcap/tiflash/issues/11071

contrib/tiflash-columnar-hub/hub-runtime/ffi/src/RaftStoreProxyFFI/ProxyFFI.h
yongman: Added `make gen_proxy_ffi` support.

JaySon-Huang (approved):

JinheLin (approved):

yudongusa (approved):

PR #11081, opened 2026-09-08
schedule: add keyspace cpu limiter

### LINKED ISSUE #11082: Lack of self resource protection for multiple keyspace
## Enhancement

If the resource group are not set properly or not enabled, the server cpu resource can be exhusted by one keyspace workload and other keyspace requests being starved.

yongman:

Issue Number: close #11082 

### Problem Summary:

In a multi-keyspace environment, a single keyspace can monopolize pipeline CPU and scheduler slots, including CPU consumed by columnar IO tasks. This can delay requests from other keyspaces and reduce overall tenant isolation.

- Added a shared `KeyspaceCpuLimiter` for both pipeline CPU and IO task pools. A keyspace must acquire the shared limiter before its task is scheduled, so CPU and columnar IO tasks are accounted for together.
- Added two independent per-keyspace limits:
  - `pipeline_keyspace_pool_limit_ratio`: caps concurrent active scheduler tasks.
  - `pipeline_keyspace_cpu_limit_ratio`: caps long-term CPU-time consumption with a per-keyspace token bucket.
- CPU usage is measured with `CLOCK_THREAD_CPUTIME_ID`. Blocking time does not consume tokens, while actual CPU work performed in both CPU and IO pool threads does.
- When the CPU token bucket is exhausted, the task yields and is retried after tokens are refilled. When the active-task limit is reached, new tasks from that keyspace remain queued while tasks from other keyspaces can continue.
- The two limits can be enabled independently. Setting either ratio to `0` disables only that mechanism; setting both to `0` disables the limiter entirely.
- Both settings are validated in `[0.0, 1.0]`, are applied during server startup, and default to `0.0` to preserve existing behavior unless explicitly enabled.
- Added Prometheus metrics for active tasks, configured task limit, remaining CPU tokens, CPU quota, current throttling state, charged CPU seconds, and throttled admissions split by CPU-quota versus active-task-limit reasons.
- Added tests for CPU-only limiting and for pool-only limiting shared across CPU and IO pools.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```
image
image

# TiFlash Keyspace CPU Limiter Comparison Test Report

## 1. Test Overview

- Test date: 2026-09-09
- TiDB endpoint: `127.0.0.1:4000`
- User: `root`; password: empty
- Database: `widecol`
- Table: `widecol.widecol_test_wide`, approximately `152,000,000` rows
- TiFlash service: `tiflash-9000`
- Metrics endpoint: `http://127.0.0.1:8234/metrics`
- Configuration file: `/data/deploy/tiflash-9000/conf/tiflash.toml`
- Configuration before and after the test: `cpu=0.8`, `pool=0`

The test set contains the original nine combinations of `0`, `0.5`, and `1.0`, plus two additional combinations, `0/0.8` and `0.8/0`. All 11 configurations are merged into the result table below.

For each configuration, the test changed both parameters, ran `systemctl restart tiflash-9000`, waited for the service and metrics endpoint to become ready, waited an additional 5 seconds, and then ran the following query through `mycli`:

```sql
SET SESSION tidb_opt_enable_late_materialization=OFF;
SELECT * FROM widecol_test_wide WHERE w_name='gTQB3O';
```

Each query returned one row. All 11 SQL executions and service restarts succeeded. Query duration includes query execution and result transfer time.

## 2. Combined Results

The limiter counters below are cumulative values observed after the query, following the TiFlash restart for that row. No matching limiter series was present in the pre-query snapshot for any round, so the post-query values represent the observed process-local counters for that round. `cpu_s` is `tiflash_pipeline_keyspace_cpu_limiter_cpu_seconds`; the two rejection columns are the `reason` values of `tiflash_pipeline_keyspace_cpu_limiter_throttled`.

| CPU ratio | Pool ratio | Restart/service | Readiness poll (s) | Query time (s) | cpu_s | active_tasks rejects | cpu_quota rejects | max_active_tasks | quota/s | Interpretation |
|---:|---:|:---:|---:|---:|---:|---:|---:|---:|---:|:---|
| 0 | 0 | success/active | 4 | 141.479 | not exposed | not exposed | not exposed | not exposed | not exposed | Both limiters disabled |
| 0 | 0.5 | success/active | 3 | 191.190 | 0.000 | 402,433,486 | 0 | 16 | 0 | Pool limiter only |
| 0 | 0.8 | success/active | 2 | 156.121 | 0.000 | 247,469,894 | 0 | 25 | 0 | Pool limiter only |
| 0 | 1.0 | success/active | 3 | 144.007 | 0.000 | 0 | 0 | 32 | 0 | Pool limiter only |
| 0.5 | 0 | success/active | 4 | 227.676 | 3,183.872 | 0 | 3,533,724 | 0 | 16 | CPU limiter only |
| 0.5 | 0.5 | success/active | 3 | 194.882 | 2,559.912 | 37,752,087 | 218,259 | 16 | 16 | Both limiters enabled |
| 0.5 | 1.0 | success/active | 4 | 235.480 | 3,266.812 | 0 | 3,786,222 | 32 | 16 | Both limiters enabled |
| 0.8 | 0 | success/active | 4 | 142.399 | 3,066.596 | 0 | 78,458 | 0 | 25.6 | CPU limiter only |
| 1.0 | 0 | success/active | 3 | 141.527 | 3,069.927 | 0 | 883 | 0 | 32 | CPU limiter only |
| 1.0 | 0.5 | success/active | 4 | 193.845 | 2,538.236 | 38,933,261 | 0 | 16 | 32 | Both limiters enabled |
| 1.0 | 1.0 | success/active | 3 | 143.627 | 3,181.051 | 0 | 0 | 32 | 32 | Both limiters enabled |

At query completion, the `active_tasks` and `throttled` gauge values were 0 in every round. The observed `cpu_tokens_seconds` values were approximately `0`, `1.6`, `2.56`, or `3.2`, according to the CPU ratio.

## 3. Findings

1. With both ratios set to `0`, the limiter metric series was not exposed, consistent with both limiters being disabled.
2. `pipeline_keyspace_cpu_limit_ratio` controls `cpu_quota_seconds_per_second`: `0.5` maps to `16`, `0.8` maps to `25.6`, and `1.0` maps to `32`.
3. `pipeline_keyspace_pool_limit_ratio` controls `max_active_tasks`: `0.5` maps to `16`, `0.8` maps to `25`, and `1.0` maps to `32`. A ratio of `0` leaves the pool limiter disabled.
4. When both ratios are non-zero, both CPU quota and active-task admission statistics are observable.
5. Query time across the 11 rounds ranged from `141.479s` to `235.480s`. The environment had other concurrent `mycli` query workloads, so the counters and timings cannot be attributed exclusively to this test query.
6. Each configuration was tested once in a fixed order. These results are suitable for behavior and configuration comparison; performance conclusions should use an isolated environment, repeated runs, and randomized order.

## Summary by CodeRabbit

- **New Features**
  - Added optional per-keyspace CPU and active-task limits for pipeline scheduling.
  - Added settings to configure CPU-share and pool-share limits, validated from 0 to 1.
  - CPU and I/O workloads now share keyspace limits, while work from other keyspaces can continue independently.
  - Added Prometheus metrics for per-keyspace activity, CPU usage, quotas, and throttling.

- **Bug Fixes**
  - Improved cleanup of keyspace reservations when tasks are finalized, canceled, drained, or cannot be assigned.

- **Tests**
  - Added coverage for shared limits, quota throttling, disabled limits, and cross-pool scheduling.


JaySon-Huang (commented):
Second Opinion: 1 high and 3 medium findings posted as inline comments.

dbms/src/Flash/Pipeline/Schedule/TaskQueues/IOPriorityQueue.cpp
```diff
@@ -58,32 +61,73 @@ bool IOPriorityQueue::take(TaskPtr & task)
...
-        if (popTask(first_queue, task))
+        if (tryTakeTaskWithoutLock(first_queue, task))
             return true;
-        if (popTask(next_queue, task))
+        if (tryTakeTaskWithoutLock(next_queue, task))
             return true;
-        cv.wait(lock);
+        if (limiter_enabled)
+        {
+            const auto previous_change_id = keyspace_cpu_limiter->getChangeId();
```
JaySon-Huang: High: `getChangeId()` is taken after the scan, while `updateStatistics()` → `release()` does not hold the queue mutex. A release that frees a slot between a failed `tryAcquire` and this snapshot is recorded as the current generation, so `waitForChange` waits for a later event.

With only `pipeline_keyspace_pool_limit_ratio` enabled, `getRefillWaitDuration()` is `milliseconds::max()`, so the default IO queue (and MLFQ) can hang until an unrelated submit/release/finish. `ResourceControlQueue` is only bounded by LAC's 1s timeout.

Snapshot `change_id` before scanning, and split empty-queue wait from token-refill polling. `finish()` must still be able to wake limiter waiters.

Source: rule/XUHUAIYU-PR-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -92,22 +96,35 @@ template <typename Impl>
 void TaskThreadPool<Impl>::handleTask(TaskPtr & task)
 {
     assert(task);
+    const auto * const task_ptr = task.get();
+    SCOPE_EXIT({
+        if (keyspace_cpu_limiter)
+            keyspace_cpu_limiter->release(task_ptr);
```
JaySon-Huang, line 103: Medium: this `SCOPE_EXIT` keeps a raw `Task*` for exception-path release, but the success path already releases in `updateStatistics()`, and `FINISH_STATUS` calls `task.reset()` before `SCOPE_EXIT` runs. The later map lookup does not dereference, so it is usually a no-op; if the allocator reuses the same address for a newly bound task, this can drop someone else's reservation and over-admit.

Track a `released` flag (or null the pointer after the real release) so `SCOPE_EXIT` only runs when `handleTask` still owns the slot.

Source: rule/XUHUAIYU-PR-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Pipeline/Schedule/TaskQueues/KeyspaceCpuLimiter.h
```diff
@@ -0,0 +1,269 @@
...
+/// bucket is charged with thread CPU time, so blocking IO does not consume it
+/// while CPU spent by columnar IO tasks does. Work outside these scheduler
+/// pools is outside of its scope.
+class KeyspaceCpuLimiter
+{
+public:
+    explicit KeyspaceCpuLimiter(size_t max_active_tasks_, UInt64 cpu_quota_per_second_ns_ = 0)
+        : max_active_tasks(max_active_tasks_)
+        , cpu_quota_per_second_ns(cpu_quota_per_second_ns_)
+        , cpu_quota_burst_ns(cpu_quota_per_second_ns_ == 0 ? 0 : std::max<UInt64>(1, cpu_quota_per_second_ns_ / 10))
```
JaySon-Huang: Medium: burst is `max(1, quota/10)` and tokens start at that burst, so a keyspace with 1s of quota can run only ~100ms before throttling, then refills are still capped at 10% of a second. Combined with `take()` polling `waitForChange` every 1ms even when the queue is empty, enabling `pipeline_keyspace_cpu_limit_ratio` will cause frequent yields and idle-thread wakeups on the global limiter mutex.

Is the 10% burst intentional? If not, use ~1s of quota as bucket capacity, and wait on the queue cv when there is no throttled work.

Source: rule/JAYSONHUANG-RC-005 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Pipeline/Schedule/TaskQueues/tests/gtest_io_priority.cpp
```diff
@@ -226,4 +226,43 @@ try
...
+    SCOPE_EXIT({ context2.decActiveRefCount(); });
+
+    auto limiter = std::make_shared<KeyspaceCpuLimiter>(1);
+    IOPriorityQueue queue(limiter);
+    queue.submit(std::make_unique<MockIOTask>(context1, true));
+    queue.submit(std::make_unique<MockIOTask>(context1, true));
+    queue.submit(std::make_unique<MockIOTask>(context2, true));
+
+    TaskPtr task;
+    ASSERT_TRUE(queue.take(task));
```
JaySon-Huang, line 251: Medium: this test never blocks in `waitForChange` — it takes sequentially on one thread. `KeyspaceCpuLimiterTest` is also single-threaded acquire/consume, and the scheduler sharing test can hide a rare hang behind a 5s timeout.

Please add coverage for (1) a concurrent waiter plus the last `release` of a pool-limited keyspace, (2) `finish()` / destructor joining while workers are in limiter wait, (3) reservation release if `execute()` throws. These are the acceptance tests for the wait-protocol hang above.

Source: rule/JAYSONHUANG-RC-006 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Pipeline/Schedule/TaskQueues/KeyspaceCpuLimiter.h
```diff
@@ -0,0 +1,271 @@
...
+        quota.metrics->cpu_seconds_total->Increment(static_cast<double>(cpu_time_ns) / 1'000'000'000.0);
+        updateMetricsWithoutLock(owner_iter->second, quota);
+        return quota.tokens_ns <= 0;
+    }
+
+    /// Token refill is time based, so a queue with only throttled work must
+    /// periodically retry even when no task is submitted or completed.
+    std::chrono::milliseconds getRefillWaitDuration() const
+    {
+        return cpu_quota_per_second_ns == 0 ? std::chrono::milliseconds::max() : std::chrono::milliseconds(1);
```
JaySon-Huang, line 119: High: pool-only mode (`pipeline_keyspace_cpu_limit_ratio=0`, `pipeline_keyspace_pool_limit_ratio>0`) returns `std::chrono::milliseconds::max()` here and passes it straight into `cv.wait_for` (`waitForChange` below).

`wait_for` is `wait_until(now + rel_time)`. Converting `milliseconds::max()` to `steady_clock` nanoseconds overflows int64, so the deadline is in the past (on gcc 11 / libstdc++ the wrapped delta is `-1ms`). `wait_for` therefore returns immediately while the predicate is still false, and `take()`'s outer loop rescans, fails `tryAcquire` on every pending task, and bumps `tiflash_pipeline_keyspace_cpu_limiter_throttled{reason="active_tasks"}` — one waiter can pin a core and hammer the limiter mutex.

Call sites:
- `IOPriorityQueue.cpp` (default IO queue) — affected
- `MultiLevelFeedbackQueue.cpp` (`queue_type=MLFQ`) — affected
- `ResourceControlQueue.cpp` — not affected, because `std::min(wait_dura, …)` caps at LAC's 1s

This matches the PR numbers: cpu=0 / pool=0.5 and 0.8 show ~10^8 active_tasks rejects (tight loop, not 1ms poll); cpu=0 / pool=1.0 is 0 because that config never throttles into this wait.

Fix: when there is no CPU-quota refill, call the untimed `waitForChange(previous_change_id)` (`cv.wait` + predicate). Pool slots are only returned by `release()`, so a time-driven retry is not needed. Do not pass `duration::max()` to `wait_for`.

Source: rule/WINDTALKER-PERF-008 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (commented):
Second Opinion: one additional high finding on pool-only wait_for overflow.

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -92,23 +96,38 @@ template <typename Impl>
...
 {
     assert(task);
+    const auto * const task_ptr = task.get();
+    bool reservation_released = false;
+    SCOPE_EXIT({
+        if (keyspace_cpu_limiter && !reservation_released)
+            keyspace_cpu_limiter->release(task_ptr);
+    });
     TaskTimer timer{task->profile_info};
+    timer.startCPUTime();
```
JaySon-Huang, line 106: Medium: with both ratios at the default 0, `createKeyspaceCpuLimiter()` still returns a non-null `shared_ptr`, so this hot path always pays `clock_gettime(CLOCK_THREAD_CPUTIME_ID)`.

`if (keyspace_cpu_limiter)` / `ptr && consumeCPUTime(..., updateCPUExecutingTime())` does not skip the argument: the pointer is never null, so `updateCPUExecutingTime()` runs before `consumeCPUTime` can early-return on `cpu_quota_per_second_ns == 0`. That clock is not on the vDSO fast path (needs per-thread cputime in the kernel), so it is a real syscall.

Count per `handleTask`: `startCPUTime` once, one more call before the loop, then one per `execute()` round. A single-round task is three syscalls, not zero. `setThreadCPUTimeNs` below is only a store (`getThreadCPUTimeNs` has no readers today).

This is new cost on the default scheduler path, which the PR describes as preserving existing behavior.

Gate `startCPUTime` / `consumeCPUTime` on CPU-quota being enabled (`cpu_quota_per_second_ns != 0`), not on `isEnabled()` — pool-only mode does not need thread CPU time either.

Source: rule/JAYSONHUANG-RC-005 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

dbms/src/Flash/Pipeline/Schedule/TaskQueues/IOPriorityQueue.cpp
```diff
@@ -95,6 +151,14 @@ void IOPriorityQueue::updateStatistics(const TaskPtr &, ExecTaskStatus exec_task
         break;
     default:; // ignore not io status.
     }
+
+    if (keyspace_cpu_limiter)
+    {
+        // A cancelled task can be returned from cancel_task_queue without ever
+        // acquiring a reservation; release() is owner-aware and therefore a no-op.
+        keyspace_cpu_limiter->release(task.get());
+        notifyWaiters();
```
JaySon-Huang, line 160: Medium: `updateStatistics` used to notify nobody. The guard here is pointer non-null, not `isEnabled()`, and Server always passes a non-null limiter. Every task completion now calls `notifyWaiters()` → `cv.notify_all()` even when both ratios are 0.

When the limiter is off, `release()` / `KeyspaceCpuLimiter::notifyAll()` early-return, but `cv.notify_all()` still runs. Idle `take()` waiters wake, see an empty queue, and sleep again. On FINISH that wakeup is pure waste; on yield+resubmit it duplicates the notify that `submit()` already does. Same pattern in `MultiLevelFeedbackQueue::updateStatistics` and `ResourceControlQueue::updateStatistics`.

Please only add this wakeup when `isEnabled()` is true. Do **not** skip `cv.notify_all()` inside the shared `notifyWaiters()` used by `submit` / `cancel` / `finish` — those still need to wake waiters with the limiter off. (Separately, `submit` also changed from `notify_one` to `notify_all`; worth a look on the default path.)

Source: rule/WINDTALKER-PERF-008 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (commented):
Second Opinion: two medium findings on default-off hot-path overhead.

JaySon-Huang (approved):

dbms/src/Flash/Pipeline/Schedule/TaskQueues/MultiLevelFeedbackQueue.cpp
```diff
@@ -156,57 +159,118 @@ void MultiLevelFeedbackQueue<TimeGetter>::submit(std::vector<TaskPtr> & tasks)
...
+
+    for (auto it = unit_queue.task_queue.begin(); it != unit_queue.task_queue.end(); ++it)
+    {
+        const auto keyspace_id = (*it)->getKeyspaceID();
+        if (!keyspace_cpu_limiter->tryAcquire(keyspace_id))
+            continue;
+
+        bool owner_bound = false;
+        SCOPE_EXIT({
+            if (!owner_bound)
```
windtalker, line 271: The `owner_bound` seems always be true?

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
```diff
@@ -92,23 +96,40 @@ template <typename Impl>
...
         auto total_time_spent = timer.updateExecutingTime();
+        const bool cpu_quota_exhausted
+            = cpu_quota_enabled && keyspace_cpu_limiter->consumeCPUTime(task.get(), timer.updateCPUExecutingTime());
         // The executing task should yield if it takes more than `YIELD_MAX_TIME_SPENT_NS`.
-        if (!Impl::isTargetStatus(status_after_exec) || total_time_spent >= YIELD_MAX_TIME_SPENT_NS)
+        if (!Impl::isTargetStatus(status_after_exec) || total_time_spent >= YIELD_MAX_TIME_SPENT_NS
+            || cpu_quota_exhausted)
             break;
     }
+    task->profile_info.addThreadCPUTimeNs(timer.cpu_executing_time - cpu_time_before_exec);
```
windtalker, line 130: Looks like if `cpu_quota_enabled` is false, then the threadCPUTime is meaningless?

dbms/src/Flash/Pipeline/Schedule/TaskQueues/MultiLevelFeedbackQueue.cpp
yongman: If there any exception throw due to task move or queue deletion the owner_bound keeps false and make sure the keyspace will be released.

dbms/src/Flash/Pipeline/Schedule/ThreadPool/TaskThreadPool.cpp
yongman: If cpu_quota_enabled == false, the `startCPUTime()` will not be called and threadCPUTime keeps unchanged.

windtalker (approved):
lgtm

yudongusa (approved):

PR #11089, opened 2026-09-14
fix: stop hash join probes on pipeline early termination

### LINKED ISSUE #11088: pipeline: global LIMIT can hang right outer join by skipping finishOneProbe
## Problem

In the pipeline execution model, a global LIMIT downstream of a hash join may finish one probe pipeline before that probe source reaches EOF. For joins that must scan the build-side hash map after all probes complete, the early-finished probe never calls `Join::finishOneProbe()`.

Other probe streams can reach EOF, call `finishOneProbe()`, and then wait forever for `wait_probe_finished_future`, because `active_probe_threads` never reaches zero.

This can leave the root MPP task and its exchange sender waiting indefinitely. The production case was observed on v8.5.4.

## Affected Join Modes

The reproducer covers `RightOuter`. The same lifecycle requirement is used by the join modes where `needScanHashMapAfterProbe()` is true, including full outer, right semi, and right anti joins.

## Deterministic Reproducer

A minimal pipeline gtest was added locally:

`JoinExecutorTestRunner.PipelineRightOuterJoinLimitMaySkipProbeFinish`

It builds a two-stream probe side for a `RIGHT OUTER JOIN ... LIMIT 1`:

- Probe stream A has two one-row blocks and produces the row that reaches the global limit.
- Probe stream B immediately reaches EOF.
- A test-only failpoint pauses B immediately after it calls `finishOneProbe()`, allowing A to be stopped by LIMIT before it reaches EOF.
- After releasing the failpoint, B transitions to `WAIT_PROBE_FINISH` and cannot make progress because A never decremented the active-probe count.

The test explicitly calls `query_executor->cancel()` for cleanup. Without cancellation it does not complete.

A 60-second verification run showed:

```
11:19:48.442  PROBE -> WAIT_PROBE_FINISH
11:20:48.448  cancellation starts cleanup
wait_for_notify_time_ns = 60,005,557,000
```

The query had no natural progress for the full 60-second post-failpoint observation window.

## Expected Behavior

Early downstream completion caused by global LIMIT must not leave a join waiting for a probe stream that will never receive source EOF. The query should complete naturally without requiring query cancellation.

## Possible Direction

Make pipeline early-stop propagation join-aware: when a downstream sink finishes and an unfinished probe will no longer be driven to EOF, ensure the join records that probe as finished or otherwise resolves the scan-after-probe wait. The fix must correctly wake the `wait_probe_finished_future`, not only condition-variable waiters.

## Regression Coverage

Once fixed, the reproducer should be inverted to assert natural query completion under the same forced scheduling order.

windtalker:

Issue Number: close #11088

Problem Summary:

A pipeline task can stop after a global LIMIT is satisfied while sibling V1 HashJoin probe workers are waiting for every probe input to reach EOF before scanning unmatched build rows or restoring spilled partitions. The early termination was local to one probe worker, leaving peers waiting indefinitely.

```commit-message
fix: stop hash join probes on pipeline early termination
```

Add a small shared `stopProbePhase()` path to V1 HashJoin. A probe operator stopped by pipeline suffix publishes the stop, wakes workers waiting for the normal probe barrier, and closes an in-progress unmatched-build scan. Peer probe operators check the shared stop state at their operator boundaries and finish without starting scan or restore work.

Normal input EOF keeps the existing `finishOneProbe()`/`finalizeProbe()` barrier and remains responsible for normal spill finalization.

Add a pipeline regression test that reproduces a right outer join whose LIMIT causes one probe worker to stop before observing input EOF.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
Fix a potential MPP query hang when a global LIMIT terminates a right or full outer hash join probe early.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed hash join probe completion for right outer joins with row limits.
  - Prevented unfinished probe streams from being treated as complete.
  - Improved synchronization and cleanup when probe processing is interrupted.
  - Ensured post-probe processing only proceeds after all probe streams finish normally.

- **Tests**
  - Added coverage for right outer joins with limits and asynchronous probe completion.
  - Added validation for probe-phase coordination and interrupted execution scenarios.


dbms/src/Interpreters/Join.h
```diff
@@ -379,8 +395,11 @@ class Join
 
     mutable std::condition_variable probe_cv;
     size_t probe_concurrency;
-    size_t active_probe_threads;
-    std::atomic_bool probe_finished{false};
+    // Streams that have not finished probe input normally.
+    size_t pending_probe_streams;
```
gengliqi, line 399: It seems that there is no need to change this name?

gengliqi (approved):
Rest LGTM

xzhangxian1008 (approved):
LGTM

PR #11104, opened 2026-09-18
executor: restrict skip-probe optimization to local probe sources

### LINKED ISSUE #11092: MPP query fails when a consumer finishes early: in-flight writes hit an already-closed tunnel and abort the query

Any MPP query in which an exchange consumer task legitimately finishes before
its upstream producer tasks have finished sending, while the producers still
have packets in flight toward it.

Known triggers:

- A hash join task whose build side is empty skips reading its probe input
  and immediately closes its probe ExchangeReceiver, while the probe-side
  producer tasks are still writing (the skip optimization from #11001 makes
  this window wide and reliable, which is how the bug surfaced).
- The same class of problem was previously observed with LIMIT-style early
  exits (#7177). #7969 removed the gather-level error cascade, but the
  producer-side write failure and the task-level error-packet cascade remain.

The failure itself is a timing race: it requires a producer write to land
after the consumer has already closed. On an unmodified build it occurs with
low probability; it can be reproduced deterministically with a small
delay-injection patch that stretches the producer write window.

The query succeeds. A consumer legitimately finishing early — because of
LIMIT, an empty build side, a filter that eliminates all rows, etc. — is a
fact that upstream producers must tolerate, not an error. In-flight packets
that arrive after the consumer has finished should be discarded silently,
and the producer tasks should finish normally.

The query fails with err msg like:

`ERROR 1105 (HY000): write to tunnel tunnel13+21 which is already closed,
tunnel13+21: unexpectedWriteDone called`

The correct query result is lost. Root cause:

1. A consumer task legitimately finishes early and closes its
   ExchangeReceiver / MPPTunnel while upstream producers are still sending.
2. The MPP exchange layer has no graceful-close handshake: in the async gRPC
   path the close reason (clean vs. erroneous) is not transported, so the
   producer cannot distinguish a clean consumer finish from a real failure,
   and its in-flight write fails with "tunnel is already closed" /
   "unexpectedWriteDone called".
3. The failed producer task reports error packets, which cascade to sibling
   tasks and abort the whole MPP query, even though every task's own work
   was correct.

Note this weakness is not introduced by #11001; it has existed in the
exchange layer since the beginning. #11001 merely added a new early-finisher
(the empty-build skip) that hits the window reliably.

master (exchange-layer weakness predates #11001; reliably triggered by it)

ChangRui-Ryan:

Issue Number: close #11092 

## Problem Summary
The skip-probe optimization introduced in #11001 lets Inner/Semi/RightSemi joins with an empty build side skip reading the probe input entirely. However, when the probe input comes from an`ExchangeReceiver` (remote MPP tasks or local self-exchange), skipping the probe terminates the receiver while remote senders are still transmitting data. The senders then fail with "tunnel already closed" errors, which cascades into task cancellation and complicated query states. Notably, two existing tests from #11001 (`EmptyBuildRightSemiJoinSkipsProbe` /`BuildWithOnlyNullKeysRightSemiJoinSkipsProbe` ) had already constructed exactly this scenario (probe = exchange receiver + empty build); the mock environment simply never exposed the failure.

The skip-probe optimization is now restricted to local probe inputs:

- Added`PhysicalPlanNode::subtreeContainsExchangeReceiver()` , which recursively checks during plan building whether a subtree contains an`ExchangeReceiver` (or`MockExchangeReceiver` ) node.
- `PhysicalJoin` (both the pipeline path and the legacy BlockInputStream path) and`PhysicalJoinV2::buildPipeline` call`join->setLocalProbeSource(!probe()->subtreeContainsExchangeReceiver())` while building the probe pipeline.
- `Join::shouldSkipProbe()` /`HashJoin::shouldSkipProbe()` now require`local_probe_source` to be true; it defaults to`false` so any future execution path that forgets to set it stays on the safe side (optimization silently disabled instead of failing remote senders).
- Local probe sources (table scans, incl. cop mode) keep the optimization; probe inputs from an exchange receiver fall back to the pre-#11001 behavior: the probe side is fully consumed, receivers drain all data, and remote senders finish cleanly.
Judging on the plan tree instead of the runtime source op makes the check independent of executor shapes (ConcatSourceOp wrapping, SharedQueue, multi-builder concurrency) and unifies both execution paths with a single predicate.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Side effects

Documentation

- [ ] Affects user behaviors
- [ ] Contains syntax changes

### Release note

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed joins with empty build inputs to continue consuming rows from remote exchange sources, preventing upstream data from being left unprocessed.
  * Restricted the empty-build probe optimization to local probe inputs for more reliable distributed query execution.
  * Improved handling of nested exchange receivers during query planning and execution.

* **Tests**
  * Added coverage for nested exchange receivers and joins involving remote probe sources, including empty and null-only build inputs.


ChangRui-Ryan:
@coderabbitai Thanks for the review. Responses to the findings:

**1. `setLocalProbeSource` parameter rename (`Join.h:262`, `HashJoin.h:83`) — accepted.**
Renamed `is_local` to `isLocal` in 74a11a6975.

**2. Member rename `local_probe_source` → `localProbeSource` — not applied.**
Keeping the member in snake_case to stay consistent with the adjacent `build_side_empty` and other members in `Join.h` / `HashJoin.h` (e.g. `total_input_build_rows`, `active_build_worker`, `has_build_data_in_memory`). Mixing a camelCase member into this region would create local inconsistency.

**3. Propagate `local_probe_source` in `Join::createRestoreJoin` — intentionally not propagated.**
Restore joins are part of the spill workflow, and #11001 deliberately excludes spilled joins from the skip-probe fast path (`build_side_empty` requires `!isSpilled()`, see `EmptyHashTableAfterSpillStillReadsProbe` in `gtest_spill_join.cpp`). Restore probes also consume spill-restored probe partitions (`probe_partition_blocks` / `hasProbeSideMarkedSpillData`), which skipping would bypass. Since `local_probe_source` defaults to `false`, the restore join stays on the safe side by design; the potential win is limited to the rare combination of a local probe, a spilled build, and a restore partition that ends up empty.

windtalker (approved):
lgtm

